PR #4, opened 2020-06-07
Adding a loading state on 'regio' page, and added some accessibility tweaks

JeroenReumkens:
Hi there! 

I really like to work that is being done on the dashboard. I took some time today to improve a few things I noticed, which because of the short timeline might not have made it into the first version yet. Let me know if you would like to merge these changes or have some feedback. I'd love to help, also with any other more pressing issues. 

This PR has been split into #6 , #7 , #8 and #9 and will probably be closed. 

~~**The changes I made:**~~

~~1. Added a small simple skeleton loading state for the graphs on the region page. That way the container doesn't jump around when the data is loaded.
2. Added screen reader only labels describing what the BarCharts show, for visually impaired users. (The content might need to be checked though).
3. Extended the graph titles on the region page with a (only for screen readers visible) text repeating the region this data is for.
4. Added some focus management which after selecting a region moves the focus to the heading of the first graph, so a user immediately can read the related data. 
5. Added a screen reader only button at the end of the content bringing the user back to the region selector. This way a visually impaired user can switch easier between regions multiple times.~~

PR #8, opened 2020-06-10
feat: Add screenreader labels to the barchart (a11y)

JeroenReumkens:
Hi there!

I initially created a PR #4 that included mulitple changes. In this PR I split the a11y labels for barcharts into a separate PR.

There are two things still needing to be done: 
- [x] Add Jest and write a unit test for the variable replacement util
- [x] Verify that the content I have added is okay ( @VWSCoronaDashboard )

**The changes made:**
1. I added screenreader labels to all bar charts trying to explain what a user sees in the chart. It read outs the value and also (if applicable) the signaalwaarde.
2. Repeated the region in the chart titles on the sidebars of the region page. Eg: Ziekenhuisopnames per dag (in `regio`).
3. Needed to add a small variable replacement utility to replace the regio name inside the translation. 

**Why?**
1. Visually impaired users might have a hard time deciphering the data in the graphs. We can help them with understanding the data better by reading out loud the data. 
2. After the user selected a region (s)he will continue navigating through the website, if it would come across a title like 'ziekenhuisopnames per dag' it would have to guess this is for the region (s)he just selected.

_Example of the chart description:_
image
(notice the paragraph with the chart title in the source code on the right, which is the default title, and the span with the explanation right underneath)

_Example of the region page titles with region added_
image


JeroenReumkens:
@VWSCoronaDashboard I have added the unit test and thus also a very minimal setup for Jest. We can always tweak this in a different PR if needed, I think.

src/components/barScale/index.tsx
```diff
@@ -98,74 +102,83 @@ const BarScale: FunctionComponent<BarscaleProps> = ({
...
-      <div className="scale">
-        <div className="minValue">{min}</div>
-        <div className="maxValue">{max}</div>
+    <>
+      <ScreenReaderOnly>
+        {replaceVariablesInText(screenReaderText, {
+          value,
+          kritiekeWaarde,
+        })}
+      </ScreenReaderOnly>
```
VWSCoronaDashboard, line 111: I am wondering if it's better to move the title and descriptive text to the actual SVG element as per https://www.w3.org/TR/2000/NOTE-SVG-access-20000807/#Equivalent section 2.1.

src/data/textNationaal.json
```diff
@@ -5,6 +5,7 @@
   "ic_opnames_per_dag": {
     "title": "Intensive care-opnames per dag",
```
VWSCoronaDashboard, line 6: We should add this to the SVG element as `title`

src/data/textNationaal.json
```diff
@@ -5,6 +5,7 @@
...
     "text": "Aantal opnames per dag, gemiddeld over de afgelopen 3 dagen.",
+    "screen_reader_graph_content": "{{value}} intensive care-opnames per dag. Signaalwaarde: {{kritiekeWaarde}} opnames per dag.",
```
VWSCoronaDashboard, line 8: I think this is perfect as-is as description for the chart.

src/components/graphHeader/index.tsx
```diff
@@ -1,21 +1,30 @@
...
     <div className={styles.graphHeader}>
       {Icon && (
         <div className={styles.graphHeaderIcon}>
           <Icon />
         </div>
       )}
-      <h3>{title}</h3>
+      <h3>
+        {title}
+        {regio && <ScreenReaderOnly>in {regio}</ScreenReaderOnly>}
```
VWSCoronaDashboard, line 26: Nice addition

src/pages/regio/index.tsx
```diff
@@ -195,6 +201,10 @@ const Regio: FunctionComponentWithLayout<HomeLayoutProps> = () => {
                     state[selectedRegio.code].infected_people_delta_normalized
                       .value
                   }
+                  screenReaderText={
+                    siteText.regionaal_positief_geteste_personen
+                      .screen_reader_graph_content
```
VWSCoronaDashboard, line 206: Let's double-check if this description says "per 100.000 inwoners" for clarity

src/utils/__tests__/replaceVariablesInText.spec.js
```diff
@@ -0,0 +1,52 @@
+import replaceVariablesInText from '../replaceVariablesInText';
```
VWSCoronaDashboard, line 1: This is good for now, but I suspect it's a short term solution. If, and when, internationalization is becoming a concern I propose we move to something like https://formatjs.io/docs/icu-syntax and adhere to standards.

src/utils/replaceVariablesInText.js
```diff
@@ -0,0 +1,28 @@
...
+ */
+const replaceVariablesInText = (translation, variables) => {
+  if (!translation) return '';
+
+  return translation.replace(curlyBracketRegex, (_string, variableName) => {
+    if (!variables) return '';
+
+    return variables[variableName.trim()] ?? '';
+  });
+};
```
VWSCoronaDashboard, line 26: Same comment as above: https://formatjs.io/docs/icu-syntax

Not needed right now, but in time I think we should migrate to a mature solution if we add more language features.

src/components/barScale/index.tsx
JeroenReumkens: I was wondering about that too. But when Googling for the way screen readers handle these labels I found quite some contradicting answers. Looks like there also is quite some difference between all the different screen readers. That's why I opted to add it as a text before the graph, and give the graph an `aria-hidden` label. That way a visually impaired user doesn't even know there is a graph but just thinks it is text data (which is fine I think?).

src/data/textNationaal.json
JeroenReumkens: It is always okay to do it, absolutely. But screenreaders will ignore it, as mentioned in my previous comment. So it can not hurt us, but I also don't think there is really a benefit in having this.
Maybe just add it since it is very quick to add? 😄

src/utils/__tests__/replaceVariablesInText.spec.js
JeroenReumkens: Totally agree!

src/components/barScale/index.tsx
VWSCoronaDashboard: Yeah, I also saw that SVG titles and descriptions are not always exposed to screen readers. I'm leaning towards your solution.

src/data/textNationaal.json
VWSCoronaDashboard: It's definitely not a blocker, assuming we go with the current direction, but it doesn't hurt. Let's do it.

PR #9, opened 2020-06-10
chore: Shift focus to updated contents after changing a region (a11y)

JeroenReumkens:
Hi there!

I initially created a PR #4 that included mulitple changes. In this PR I split the focus management on the region page into a separate PR.

**TODO:**
- [x] Wrap the 'back to region selection' button into the `ScreenReaderOnly` component after #8  is merged. 

**The changes made:**
* Added some focus management that moves the focus to the first heading after the region is selected
* Added a button after the region content (only visible to screen readers) to move the focus back to the region selection. 

**Why?**
* By moving the focus to the first header the user doesn't need to search for the content that it has just triggered. Instead (s)he immediately knows there is content updated on the page and what it is about (Since after #8 the title will also read out the region again)
* Instead of needing to navigate back 'up' in the page someone user their keyboard can immediately focus the input again via this button.


src/components/graphHeader/index.js
```diff
@@ -1,15 +1,20 @@
...
           <Icon />
         </div>
       )}
-      <h3>{title}</h3>
+      {/*
+      An optional ref and the tabIndex makes it focussable via JS,
+      this is used on the region page to focus the first graph header
+      after the region is changed.
+      */}
+      <h3 ref={headingRef} tabIndex={headingRef ? 0 : undefined}>{title}</h3>
```
JeroenReumkens, line 17: @VWSCoronaDashboard In #4  you asked: 

> What is the default behavior without these changes? The focus stays on the last clicked anchor element right? Why is that bad?

I tried already to explain it a bit in the issue description. 
The main reason this can be improved is because the user changes the input field, but doesn't know what or even if something has changed. By moving them towards the first changed (and relevant) content part they know what has happened and where they can find the new content. After #8 that first title we move to will also read out the region, for an extra confirmation.

src/pages/regio/index.js
```diff
@@ -42,6 +42,31 @@ function Regio() {
...
+
+  /**
+   * Focuses the first heading in the content column,
+   * used after region changes to move focus for visually impaired
+   * users so they know what has changed.
+   */
+  const focusFirstHeading = () => {
+    try {
+      if (contentRef.current) contentRef.current.focus();
+    } catch {}
```
JeroenReumkens, line 67: @VWSCoronaDashboard In #4  you asked: 

> Why are you adding a try/catch without a catch?

This try catch is especially relevant in line 55 (since we query something that might be null). There is not catch in that case because it is merely a progressive enhancement and we can't do anything else if it fails. 
For the headingRef I added the try catch as well, just for additional security (I don't want to break cause a runtime error). But if there is a `contentRef.current` it should actually not fail. Do you want me to remove it?

src/components/graphHeader/index.js
VWSCoronaDashboard: That reasoning makes sense.

src/pages/regio/index.js
```diff
@@ -42,6 +42,31 @@ function Regio() {
...
+
+  /**
+   * Focuses region select element. Triggered by a screen reader
+   * only button at the end of the content.
+   */
+  const focusRegioSelect = () => {
+    if (!selectRegioWrapperRef.current) return;
+    try {
+      selectRegioWrapperRef.current.querySelector('input').focus();
+    } catch {}
```
VWSCoronaDashboard, line 56: Try/catch here makes sense. Is there anyway we can make this safer?

VWSCoronaDashboard, line 56: How about something like this?
Not sure if the `.length` is a correct check in this instance, but as an example...

```suggestion
const focusPoint = selectRegioWrapperRef.current.querySelector('input');
if (focusPoint.length > 0) {
  focusPoint.focus()
}
```

src/pages/regio/index.js
VWSCoronaDashboard: I think you can safely remove the second try/catch because I don't see any way this would fail.

VWSCoronaDashboard:
Do you have any links to leading resources into best practices around focus management in SPA's? I have my doubts about making the title focusable, something about it feels not right.

I'd love to be proven wrong on this.

src/pages/regio/index.js
JeroenReumkens: We could replace it with;
```
const input = selectRegioWrapperRef.current.querySelector('input')
if (input) input.focus()
```

Might raise less questions I think.

JeroenReumkens: Oh yes, similar to what I wrote in the previous comment! I think this is less confusing way of doing it. You want me to change it?

VWSCoronaDashboard: Yes please, I'd rather have more strict checks before calling a function then relying on a silent error.

JeroenReumkens:
> Do you have any links to leading resources into best practices around focus management in SPA's? I have my doubts about making the title focusable, something about it feels not right.
> 
> I'd love to be proven wrong on this.

The book Inclusive Components bij a11y guru Heydon Pickering is really nice. In the book there is also a part very similar tot this article: https://inclusive-components.design/a-todo-list/ (search for 'focus management' on that page). 

Next to that you for example have Reach Router which is an accessible React router (by the same devs as React Router). They also focus the first focusable element after a route change.

I did just change the tabindex from 0 to -1, which it should have been from the start. By making it -1 a user can not focus it manually (because that wouldn't have any value), only JS can focus it. On top of that I removed the outline styles for any elements having tabindex equal to -1, which will also make it nicer for non visually impaired users. Let me know what you think!

PR #10, opened 2020-06-18
build: Add ESLint, Prettier and TypeScript Support

VWSCoronaDashboard:
Do not merge until all issues are resolved when running `yarn test` and `yarn build`

Known issues:
- [x] SVG imports are not working. @svgr/webpack throws errors, and babel-inline-svg-loader throws `Cannot find module 'assets/arts.svg'`
- [x] The getLayout property in the layout component is not recognized in pages.
- [x] `Icon` needs typing in `GraphHeader`


VWSCoronaDashboard:
@JeroenReumkens Have you battled with TypeScript and SVG's before? I've tried replacing `@svgr/webpack` with `babel-plugin-inline-react-svg` but no dice.

JeroenReumkens:
@VWSCoronaDashboard The babel plugin should work. Have you added this in the index.d.ts definition file? 

`declare module ‘*.svg’`

VWSCoronaDashboard:
> @VWSCoronaDashboard The babel plugin should work. Have you added this in the index.d.ts definition file?
> 
> `declare module ‘*.svg’`

I got the following, but still nothing. Now I'm Wondering if this is a problem with absolute imports or with SVG definitions.


```
declare module '\*.svg' {
  import React = require('react');
  export const ReactComponent: React.FC<React.SVGProps<SVGSVGElement>>;
  const value: any;
  export default value;
}
```

JeroenReumkens:
> > @VWSCoronaDashboard The babel plugin should work. Have you added this in the index.d.ts definition file?
> > `declare module ‘*.svg’`
> 
> I got the following, but still nothing. Now I'm Wondering if this is a problem with absolute imports or with SVG definitions.
> 

In the past I sometimes also chose to just immediately create a react component for the icons instead of the svg file. Makes life a lot easier. 

Or you could use the SVGR CLI to create them automatically: 
```
    "svgr": "npx @svgr/cli ./public/icons --template ./public/icons/svgrtemplate.js --ext tsx --out-dir ./src/components/Icon/icons"
```

Personally I'd opt for immediately create a tsx file for an icon instead of the svg.

VWSCoronaDashboard:
> In the past I sometimes also chose to just immediately create a react component for the icons instead of the svg file. Makes life a lot easier.

Yeah, but SVGO is so good to have automated :). I found out that the trick was to put `index.d.ts` with the custom type in the actual `assets` folder.

JeroenReumkens:
> > In the past I sometimes also chose to just immediately create a react component for the icons instead of the svg file. Makes life a lot easier.
> 
> Yeah, but SVGO is so good to have automated :). I found out that the trick was to put `index.d.ts` with the custom type in the actual `assets` folder.

Interesting... Good to hear that you were able to fix it though! 😄

PR #37, opened 2020-06-30
feat: Add a new tile with sewage measurements

VWSCoronaDashboard:
"Aantal COVID-19 RNA deeltjes gemeten per mL in rioolwater"


tomudding:
You may consider this nitpicking, but I believe there should be a `-` between _RNA_ and _deeltjes_ as it is a "samenstelling". The result is _RNA-deeltjes_.

PR #45, opened 2020-07-06
feat: search by municipality

VWSCoronaDashboard:
# Summary

This PR includes the following:

- A new SelectMunicipality dropdown that allows users to select a safety region by municipality.
- A heading on the Regionale Cijfers page, showing the currently selected safety region.
- The addition of the safety region name in the headers of cards on the Regionale Cijfers page.
- Replaces the `onClick` handlers in the safety region selection map to be [SVG anchors](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/a), linking to the appropriate safety region page. 

# Detailed design

## Regionale Cijfers page

- The page now exports a function `getStaticProps`, which is ran on build time. It `require()`'s the needed static/unchanging data for the page and transforms it as needed. The data is then passed as props to the page component.

## `SelectMunicipality`

- `SelectMunicipality` is a combobox implemented using [`downshift`](https://github.com/downshift-js/downshift/tree/master/src/hooks/useCombobox#actions)'s `useCombobox` hook. The prop getters of `useCombobox` include aria and other accessibility features by default where needed, following the [Combobox with Listbox Popup example](https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html).
- The dropdown menu shows a segmented list with municipalities, divided into safety regions. The list is filtered while typing in the input field, based on a lowercase match.
- A reset button was added that will clear the input field and subsequently the page.
- `aria-live` status messages that are read out while a user is typing in the input field or makes a selection have been translated to Dutch.

### Additional changes

- This PR now fixes a bug in IE11 where the dropdown menu would not be reset to a height of `0` after closing it, by setting `overflow-y: hidden;` on the dropdown menu element.
- This PR also fixes a bug in IE11 where, after selecting a municipality from the list, the input value would immediately be cleared again. To do this, the functionality that cleared the input when re-opening the dropdown menu after selecting a municipality earlier, was changed to instead select the input value after either clicking the input value or opening the dropdown again.

## `SVGMap`

- The `onClick` handlers have been removed from the `path` elements in the safety region map SVG. They have been replaced with `a` elements, which bring link states by default. Said link states are now used to manage the appearance of the paths.

# Unresolved questions

- How to reset the value of the input field of `SelectMunicipality` when an external circumstance like selecting a safety region in the map changes the selected safety region, _without_ resetting the page itself back to an unselected state.

# To do

- [x] Replace the &times; symbol in the `SelectMunicipality` reset button with an SVG icon and make sure it is centered.


next.config.js
```diff
@@ -31,7 +31,9 @@ const nextConfig = {
       if (cacheGroups.framework) {
         cacheGroups.preact = Object.assign({}, cacheGroups.framework, { test });
         // if you want to merge the 2 small commons+framework chunks:
-        // cacheGroups.commons.name = 'framework';
+        cacheGroups.commons.name = 'framework';
+      } else {
+        cacheGroups.preact = { name: 'commons', chunks: 'all', test };
```
VWSCoronaDashboard, line 36: This was updated following the [using-preact](https://github.com/vercel/next.js/tree/canary/examples/using-preact) example by Vercel.

PR #64, opened 2020-07-14
fix: Correct name of Zuid-Limburg safety region (#59)

Raqbit:
As suggested in #59, I corrected the name of the Veiligheidsregio Zuid-Limburg. (It is currently displayed as `Limburg-Zuid`.)


VWSCoronaDashboard:
Thank you for picking this up.

PR #71, opened 2020-07-18
chore: Reduce maximum body size

Opvolger:
Om DoS attacks lastiger te maken, max 1 MB en geen 80 MB.


VWSCoronaDashboard:
Seems sensible. Thanks!

Issue #96, opened 2020-08-04
kritiekeWaarde is defined for infected people in nursing homes, but not present in the component


boisei0:
The signal value for people infected in nursing homes is defined in the locale file. However, it does not show on the dashboard as it hasn't been added to the component. Awaiting the roadmap mentioned in #90, it should be added to the component if only for the time being, as after all the value has been defined in the locale already.

### Governance
- [x] I've read the contributing document https://github.com/minvws/.github/blob/master/CONTRIBUTING.md
- [x] I've read and understand the Code of Conduct https://github.com/minvws/.github/blob/master/CODE_OF_CONDUCT.md
- [x] I understand that any contributions or suggestions I made may make it into the actual code. I've read the License https://github.com/minvws/nl-covid19-notification-app-coordination/blob/master/LICENSE.txt and the contributor license agreement https://github.com/minvws/nl-covid19-notification-app-coordination/blob/master/CLA.md

Issue #90, opened 2020-08-03
Location of critical value of bar scales in code has risk of being overridden in translations


boisei0:
The critical value, in the code referenced as both “signaalwaarde” and “kritiekeWaarde”, is sourced in the locale. The default value is present in the Dutch locale. If left empty in translations, the Dutch value is used to override the translation. However, if it erroneously isn’t left empty and the value differs from the original, the interpretation strongly changes. 

Should values that act as constants and don’t change upon translating be present in a locale in the first place, rather than be in file with constants? It would also improve the transparency if constants like these are in an easy place to find, rather than having to study the code in order to find them. Furthermore, storing numbers as numerical value rather than as string removes the need for `Number(...)` calls in the code and reduces errors caused by accidental typos slipping into the text values. 

At the moment of writing, the critical value for IC admittances is located at line 102 in the Dutch locale (value 10):
https://github.com/minvws/nl-covid19-data-dashboard/blob/43b7be4c0415607a0f74b78073afe04c5f27ed5a/src/locale/nl.json#L102
and at line 98 in the English locale (value left blank):
https://github.com/minvws/nl-covid19-data-dashboard/blob/43b7be4c0415607a0f74b78073afe04c5f27ed5a/src/locale/en.json#L98

### Governance
- [x] I've read the contributing document https://github.com/minvws/.github/blob/master/CONTRIBUTING.md
- [X] I've read and understand the Code of Conduct https://github.com/minvws/.github/blob/master/CODE_OF_CONDUCT.md
- [x] I understand that any contributions or suggestions I made may make it into the actual code. I've read the License https://github.com/minvws/nl-covid19-notification-app-coordination/blob/master/LICENSE.txt and the contributor license agreement https://github.com/minvws/nl-covid19-notification-app-coordination/blob/master/CLA.md

VWSCoronaDashboard:
@boisei0 

Agreed that these numeric values should ideally not live in a locale file!

If you look at an [older version of the app](https://github.com/minvws/nl-covid19-data-dashboard/blob/0.1.1/src/data/textNationaal.json#L159-L185), you can see that at some point not only the text but also all other _configuration_ ended up in a single json files. In hindsight you could say that was premature abstraction of logic.

Not only the _kritieke waarde_ (which has been renamed to _signaal waarde_) was stored there, but also constants like the min and max value and gradients of specific charts.

Long story short, we're planning to remove these constant from configuration files anyway and place them directly in the JSX. 

[...]

PR #97, opened 2020-08-04
Adds signal value to number of infected people in nursing homes component

boisei0:
As described in #96, this PR adds the signal value for number of infected people in nursing homes following the definition in the locale (see #90). This has been implemented based on other components that already have a signal value described in the component.


VWSCoronaDashboard:
Hi @boisei0, number of infected people in nursing homes doesn’t have a “signaalwaarde”. It has now been removed from the JSON.

PR #126, opened 2020-08-13
Fix mistakes in methodology section on cumulative data

J535D165:
- "Het bestand wordt dagelijks om 10:00 online gezet". Data is extracted from OSIRIS at 10 AM. It comes available around 3 PM (it was 4 PM, but that seems to be changed), isn't it? 
- "registreert het aantal ziekenhuisopnames in de afgelopen 24 uur." isn't the content of the file. A better description is: "registreert het aantal ziekenhuisopnames gemeld bij het RIVM in de afgelopen 24 uur". Technically, this is a derivative of the content of the file, so I replaced it by the definition used by RIVM: https://data.rivm.nl/geonetwork/srv/dut/catalog.search#/metadata/1c0fcd57-1102-4620-9cfa-441e93ea5604?tab=relations


VWSCoronaDashboard:
This suggestion is a bit outdated with the texts currently on the dashboard and therefore we're closing this PR. We've relayed the feedback to the editorial team. Thanks for the suggestion!

PR #135, opened 2020-08-17
Add “Rioolwatermonitoring” to `/regio` page

VWSCoronaDashboard:


src/components/barChart/index.tsx
```diff
@@ -3,15 +3,19 @@ import HighCharts from 'highcharts';
...
-  const total = data.reduce((mem, part) => {
-    return (mem || 0) + (part || 0);
+  const total = data.reduce((mem: number, part): number => {
+    if (typeof part !== 'number') {
+      return mem;
+    }
+    return mem + ((part as number) || 0);
   }, 0);
 
   const options = useMemo(
```
VWSCoronaDashboard, line 21: This should be `useMemo<HighCharts.Options>` to get the right typings

src/components/lineChart/multiDateLine.tsx
```diff
@@ -0,0 +1,208 @@
...
+      min: 0,
+      allowDecimals: false,
+      lineColor: '#C4C4C4',
+      gridLineColor: '#C4C4C4',
+      title: {
+        text: null,
+      },
+      labels: {
+        formatter: function (): string {
+          // @ts-ignore
```
VWSCoronaDashboard, line 137: Remove this ignore

src/components/lineChart/multiDateLine.tsx
```diff
@@ -0,0 +1,208 @@
...
+};
+
+const filterValues = (values: Value[], days: number): Value[] => {
+  const minimumDate = Math.floor(
+    new Date().getTime() / 1000 - days * 24 * 60 * 60
+  );
+  return values
+    .filter((value: Value) => value.date >= minimumDate)
+    .map((value: Value): Value => ({ ...value }));
+};
```
VWSCoronaDashboard, line 165: We need a consistent a reused function for filtering values. Please create this util and align it with the other charts.

src/components/lineChart/multiDateLine.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+const MultiDateLineChart: React.FC<MultiDateLineChartProps> = ({
```
VWSCoronaDashboard, line 167: Let's make this component specifically for "rioolwatermonitoring" until we need another multi-line chart. Then, we can pragmatically think about how we want to make such a component reusable instead. 

- Rename it to `RegionalSewerWaterLineChart`.
- In `getOptions`, both `values` and `secondaryValues` should always be valid so they don't have to be conditional. Secondly, they can have more intention revealing names now that we make it specific for one use case. Such as 'averageValues' and `listOfAllValues` (as it's an array of arrays)

src/components/tiles/regio/SewerWater.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+import { Regionaal } from 'types/data';
+import { SafetyRegion } from 'pages/regio';
+import BarChart from 'components/barChart';
+
+interface IProps {
+  data: Regionaal;
+  selectedRegio: SafetyRegion | undefined;
+  contentRef: React.RefObject<HTMLHeadingElement>;
+}
```
VWSCoronaDashboard, line 21: Has more props than the component uses. Check if they are needed.

src/components/tiles/regio/SewerWater.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+export const SewerWater: React.FC<IProps> = ({ data }) => {
+  const text: typeof siteText.regionaal_rioolwater_metingen =
+    siteText.regionaal_rioolwater_metingen;
```
VWSCoronaDashboard, line 25: Please also make a `const data` and assign the data typings to it.  Example: https://github.com/minvws/nl-covid19-data-dashboard/blob/develop/src/components/tiles/ReproductionIndex.tsx#L27

src/types/data.ts
```diff
@@ -158,3 +160,31 @@ export interface RioolwaterMetingenLastValue {
...
+}
+
+export interface SewerValue extends Value {
+  average: number;
+  date_measurement_unix: number;
+  week_unix: number;
+  rna_per_ml: number;
+  rwzi_awzi_code: string;
+  rwzi_awzi_name: string;
+}
```
VWSCoronaDashboard, line 190: You can run the data through https://app.quicktype.io/ until we have automated this. Then it’s aligned with how we did the other data interfaces.

PR #149, opened 2020-08-19
Update routing & add new persisent layout for `/landelijk`

VWSCoronaDashboard3:
## Summary

- Update routing
  - Redirect `/` to `/landelijk`
  - Create nested routes for metrics on `/landelijk` (`/landelijk/reproductiegetal`)
- Create `NationalLayout`
- Update types and the `getLayout` from `components/layout`
- Add `withChildren` type.
- Add basic layout styles.

## Motivation

First step in moving towards the new vision.

## Detailed design

- To keep the PR small, only a few links and pages are added as well as some basic styling. A follow-up PR would implement other links and styling.
- `NationalLayout` composes the site layout with the new layout specifically for `/landelijk`. This pattern is clean and straightforward. It doesn’t use abstractions, when a new layout is needed for `/regio`, we simply create a new composition.
- The `withChildren` type is added for reasons described in [this article](https://fettblog.eu/typescript-react-why-i-dont-use-react-fc/).
  - Additional reason: it makes modules more straightforward to digest as you can see all the exports on top.
  - One exception where this can’t be used is when a property is added to the component itself (`Component.getLayout = getLayout()`). 

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

- What do you think about `withChildren` and not using `React.FC`?


src/components/layout/NationalLayout.tsx
```diff
@@ -0,0 +1,83 @@
...
+ * - /landelijk/[metric] -> shows aside and content (children)
+ *
+ * More info on persistent layouts:
+ * https://adamwathan.me/2019/10/17/persistent-layout-patterns-in-nextjs/
+ */
+function NationalLayout(props: WithChildren) {
+  const { children } = props;
+  const router = useRouter();
+  const isLargeScreen = useMediaQuery('(min-width: 1000px)', true);
+  const showAside = isLargeScreen || router.route === '/landelijk';
```
VWSCoronaDashboard1, line 45: nit: do we stick the route names in some sort of struct as well? Will the routes be translated as well, for example?

src/pages/landelijk/positief-geteste-mensen.tsx
```diff
@@ -0,0 +1,10 @@
+import { FCWithLayout } from 'components/layout';
+import { getNationalLayout } from 'components/layout/NationalLayout';
+
+const PostivelyTestedPeople: FCWithLayout = () => {
+  return <p>Content van positief geteste mensen</p>;
```
VWSCoronaDashboard4, line 5: Maybe a good idea to add the localize keys from the start, to avoid having to go through it all in bulk later?

src/scss/layout.scss
```diff
@@ -60,3 +62,46 @@
...
+  ul {
+    list-style: none;
+    margin: 0;
+    padding: 0;
+
+    li {
+      padding: 1em 0;
+    }
+
+    li:not(:last-of-type) {
```
VWSCoronaDashboard4, line 103: ```li + li { border-top: ...}``` Is a bit more readable and tailored to how browsers render it.

VWSCoronaDashboard1 (changes requested):

src/pages/landelijk/positief-geteste-mensen.tsx
VWSCoronaDashboard3: This is mock text so it does not need to go into the locale. Next PR will update the aside menu, and the PR after that will fill in the content per page.

src/scss/layout.scss
VWSCoronaDashboard3: That is not the same selector. I need a border on all list items _except_ the last one. I feel the current selector shows that.

src/components/layout/NationalLayout.tsx
VWSCoronaDashboard3: Translated routes is something you can easily do with server-side rendering but this site is static. Which means the only option is duplicating all pages. We decided we will not doing that for the initial translated site. We might investigate later how to add that.

VWSCoronaDashboard1: makes sense ;)

VWSCoronaDashboard1 (approved):

src/scss/layout.scss
VWSCoronaDashboard1: doesn't that selector also do that? li + li means every li that has a sibling no?

VWSCoronaDashboard4 (approved):

PR #151, opened 2020-08-19
Feature/chart time improvements

VWSCoronaDashboard4:
## Summary

Fixes a few issues with the chart time control and the control of the values inside it from LineChart and AreaChart.

## Motivation

Some charts display data tied to dates way in the past, when eg month or week is selected. This is unified to filter to a date X days in the past, based on the chosen option.

Also changed is the possibility to provide which time slots are visible in the chart time control. This is now used only to remove the Week value for some tile that do not make sense to use it (eg measurement once every week or less frequent).

Then there is the Month. That's in a month? 30 days, 31, 29? The label is adjusted to "Last 5 weeks".

## Detailed design

In ChartTimeControl there is now a type TimeframeOption, used to pass it's value around in a secure manner. The component uses the default all/5weeks/week to render the controls dynamically.

LineChart and Area chart used to slice data from arrays to then render it. Now there is a check on the actual date value. For this I have written a generic util in the ChartTimeControl that can be used in LineChart and AreaChart to filter the correct (typed) data. The util file also contains methods to retrieve the amount of days for the provided TimeframeOption.

TimeframeOptions can be passed to LineChart, AreaChart and ChartTimeControl as an optional parameter.

## Drawbacks

Tradeoffs are that the charts can look messy and inconsistent in date values; and that some charts will only display 1 or 2 values.

## Alternatives

The change of "last 5 weeks" could be tweaked, or a definition could be provided elsewhere.

## Unresolved questions

-

### DIFF
--- a/src/components/areaChart/index.tsx
+++ b/src/components/areaChart/index.tsx
@@ -182,7 +182,7 @@ const AreaChart: React.FC<AreaChartProps> = (props) => {
     });
   }, [data]);
 
-  const [timeframe, setTimeframe] = useState<TimeframeOption>('month');
+  const [timeframe, setTimeframe] = useState<TimeframeOption>('5weeks');
 
   const chartOptions = useMemo(() => {
     const getOptionsThunk = (rangeData: TRange[], lineData: TLine[]) =>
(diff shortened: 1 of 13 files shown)


VWSCoronaDashboard1 (approved):

VWSCoronaDashboard3 (approved):
Much better!

PR #157, opened 2020-08-20
Feature/missing translation strings

VWSCoronaDashboard4:
## Summary

Add missing translation strings to locale files and use them in the components / pages.

## Motivation

For proper English UI :-)

## Detailed design

For components that are re-used a lot I created a "common" part that contains the strings per component. They are passed the common one now, but could be adjusted to pass a specific translation for that place.

## Unresolved questions

I could not test the end result of the municipality_in_safety_region_selected properly, not sure if the (A11y) string is only generated in a certain mode by the dropdown combobox component.

### DIFF
--- a/src/components/metadata/index.tsx
+++ b/src/components/metadata/index.tsx
@@ -1,16 +1,17 @@
 import styles from './metadata.module.scss';
-import { TranslationStrings } from 'types/common';
+import siteText from 'locale';
 
 interface IProps {
   dataSource: {
     href: string;
     text: string;
   };
-  text: TranslationStrings;
 }
 
 const Metadata: React.FC<IProps> = (props) => {
-  const { dataSource, text } = props;
+  const { dataSource } = props;
+
+  const text: typeof siteText.common.metadata = siteText.common.metadata;
 
   return (
     <div className={styles.metadataContainer}>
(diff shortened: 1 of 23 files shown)


src/components/metadata/index.tsx
```diff
@@ -1,20 +1,22 @@
...
   dataSource: {
     href: string;
     text: string;
   };
+  text: TranslationStrings;
 }
 
 const Metadata: React.FC<IProps> = (props) => {
-  const { dataSource } = props;
+  const { dataSource, text } = props;
```
VWSCoronaDashboard3, line 13: If you simply import the locale here you prevent adding an extra prop everywhere.

src/types/common.ts
```diff
@@ -0,0 +1 @@
+export type TranslationStrings = Record<string, string>;
```
VWSCoronaDashboard3, line 1: In the `next` branch I have created `types/index.ts` for common types. Can you create that file as well and move this type there? Then we can use short imports (`import { } from 'types'`) for common types.

VWSCoronaDashboard3 (commented):
I didn’t know we forgot so many strings!

VWSCoronaDashboard3 (approved):

PR #160, opened 2020-08-20
Correct date format for EN short dates

VWSCoronaDashboard4:
## Summary & Motivation

The English date format was in the wrong order.

## Detailed design

Based on the locale I set a string for the date format, that can then be used by filling in the template for day and month.

## Drawbacks / Alternatives

We use both imports from date-fns and the webplatform's Intl.DateTimeFormat. Perhaps there is a cleaner way to determine the proper format for the locale.

## Unresolved questions

n/a

### DIFF
--- a/src/utils/formatDate.ts
+++ b/src/utils/formatDate.ts
@@ -2,6 +2,7 @@ import { isToday, isYesterday } from 'date-fns';
 
 import siteText from 'locale';
 import getLocale from 'utils/getLocale';
+import replaceVariablesInText from './replaceVariablesInText';
 
 const locale = getLocale();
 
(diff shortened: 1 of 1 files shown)


VWSCoronaDashboard1 (approved):

VWSCoronaDashboard3 (commented):
I’m not a fan of this. This might be better:

```js
new Intl.DateTimeFormat('en-GB', {dateStyle: 'long'}).formatToParts(date);
```

yields:
```
[
  {
    "type": "day",
    "value": "20"
  },
  {
    "type": "literal",
    "value": " "
  },
  {
    "type": "month",
    "value": "August"
  },
  {
    "type": "literal",
    "value": " "
  },
  {
    "type": "year",
    "value": "2020"
  }
]
```

And we could build the date up without using two `Intl`'s (`Day` and `Month`)

Alternatively:

```js
const dateLong = new Intl.DateTimeFormat('en-GB', {dateStyle: 'long'}).format(date)
const [first, second] = dateLong.split(' ')
// 10 augustus
// August 10

```

PR #162, opened 2020-08-20
Implementation of the municipality map using highcharts

VWSCoronaDashboard1:
## Summary

Rudimentary implementation of the municipality map using highcharts

## Motivation

Feature request

## Detailed design

The map is constructed using two series, one for the country outline and the other for the municipal outlines.
Both are loaded from the static-json directory in the public folder (not sure if this is the correct way of going about it, let me know).
The actual metric data is loaded from json/municipality-data.json, this is a data file that should contain an array of municipal objects in this shape:

```
{
  "Date_of_report": "2020-03-13 10:00:00",
  "Municipality_code": "GM0003",
  "Municipality_name": "Appingedam",
  "Province": "Groningen",
  "Total_reported": 0,
  "Hospital_admission": 0,
  "Deceased": 0
}
``` 

the useMunicipalityData hook loads this file and extracts only those objects with the newest Date_of_report (which should be the current day, but if not then it'll just take the latest, that really depends on the delivered data). The hook then also creates a 'value' property on each item whose value is filled by the given metricName property (Total_reported, Hospital_admission or Deceased).

These three datasources are then fed into the highchart config where the geojson data is cross-referenced with the metric data using the gemcode -> Municipality_code link.

The Chart component itself has three props:
selected?: { id: string }; // this pre-selects the indicated municipality on the map
setSelection: (item: { id: string }) => void; // this method is called after the user click on the map
metric: TMunicipalityMetricName; // this indicates which metric is supposed to be rendered as the chloropleth (Total_reported, Hospital_admission or Deceased)

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

For now I;ve only just added the component to the region page. Its not hooked up to any of the KPIs or whatever. Its also not using a correct gradient for the cholorpleth yet. But functionaly this is pretty much what was expected I believe.


src/components/mapChart/index.tsx
```diff
@@ -1,135 +1,161 @@
...
+        maxColor: '#ff0000',
+      },
+      legend: {
+        enabled: false,
+      },
+      tooltip: {
+        backgroundColor: '#FFF',
+        borderColor: '#01689B',
+        borderRadius: 0,
+        formatter: function (this: { point: TMunicipalityPoint }) {
```
VWSCoronaDashboard4, line 82: Add the return type to the function

src/components/mapChart/index.tsx
```diff
@@ -1,135 +1,161 @@
...
+      metric,
+    ]
+  );
+
+  const ref = useRef<any>();
+
+  useEffect(() => {
+    const chart = ref?.current?.chart;
+    if (selected && chart) {
+      if (!chart) {
```
VWSCoronaDashboard4, line 146: chart is already truthy due to the check by the previous if statement

src/components/mapChart/index.tsx
```diff
@@ -1,135 +1,161 @@
...
+        return;
+      }
+      const point = chart?.series[0].points.find(
+        (p: any) => p.gemcode === selected.id
```
VWSCoronaDashboard4, line 150: any could be replaced by actual type

src/pages/regio/index.tsx
```diff
@@ -165,15 +163,20 @@ const Regio: FCWithLayout<RegioProps> = (props) => {
 
             <div className={styles['map-container']}>
               <div className={styles['safety-region-header']}>
-                <p>{text.your_safety_region}</p>
```
VWSCoronaDashboard4, line 168: Develop was not properly merged here, the translations should be used

src/utils/useMunicipalityData.ts
```diff
@@ -0,0 +1,39 @@
...
+      return [];
+    }
+
+    const max = Math.min(
+      ...data.map((item) => new Date(item.Date_of_report).getTime())
+    );
+
+    const filteredData = data
+      .filter(
+        (item) =>
```
VWSCoronaDashboard4, line 31: add type to the item

src/utils/useMunicipalityData.ts
```diff
@@ -0,0 +1,39 @@
...
+): MunicipalityData[] {
+  const { data } = useSWR<MunicipalityData[]>('/json/municipality-data.json');
+
+  return useMemo(() => {
+    if (!data) {
+      return [];
+    }
+
+    const max = Math.min(
+      ...data.map((item) => new Date(item.Date_of_report).getTime())
```
VWSCoronaDashboard4, line 26: add type to the item

src/components/mapChart/index.tsx
VWSCoronaDashboard1: I'm not sure I follow. IMHO, chart could be undefined when either ref or current is undefined, or when current IS defined but the chart isn't fully initialized yet.

src/pages/regio/index.tsx
VWSCoronaDashboard1: I've created this branch based on 'next', so I haven't merged anything from develop.

src/components/mapChart/index.tsx
VWSCoronaDashboard1: Actually, typing this strictly will get messy. The Highcharts type definitions don't provide a generic Series type that allows for the point property to be typed. So, currently its just typed to the 'Series.Point' type. I've typed the chart variable to Highcharts.Chart which expects the chart.series.point property to be just that. So typing p in the map will trigger a compile error.
I don't really see a neat way of fixing this properly, so if you have a suggestion, I'm all ears.

src/components/mapChart/index.tsx
```diff
@@ -1,135 +1,161 @@
...
+    FeatureCollection<MultiPolygon, MunicipalityProperties>
+  >('/static-json/municipalities-outline.geojson');
+
+  const municipalityData = useMunicipalityData(metric);
+  const [min, max] = useExtent(
+    municipalityData,
+    (item: MunicipalityData) => item[metric]
+  );
+
+  const mapOptions = useMemo(
```
VWSCoronaDashboard3, line 49: In order to get the correct typings for the options you need `useMemo<Highcharts.Options>`

src/utils/useExtent.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * This hook returns the minimum and maximum values for the given array.
+ * Optionally a predicate can be passed in to extract the number values from the given array.
+ *
+ * @param collection the given array
+ * @param predicate the optional predicate
+ */
+export default function useExtent(
+  collection: any[],
+  predicate?: (item: any) => number
```
VWSCoronaDashboard3, line 12: This might be a bit more correct.
```suggestion
  predicate?: (item: Record<string, unknown>) => number
```

VWSCoronaDashboard3: Or if arrays are supported as well.

```ts
Record<string, unknown> | Array<unknown>
```

VWSCoronaDashboard3 (commented):
Looks good! I like that functionality has been spit up into hooks.

src/components/mapChart/index.tsx
```diff
@@ -79,14 +79,22 @@ const MunicipalityMap: React.FC<IProps> = ({
...
-        formatter: function (this: { point: TMunicipalityPoint }) {
-          const { Province, Municipality_name } = this.point;
+        formatter: function (
+          this: TooltipFormatterContextObject
+        ):
+          | false
+          | string
+          | Array<string | null | undefined>
+          | null
+          | undefined {
```
VWSCoronaDashboard3, line 89: Interesting 😅

src/components/mapChart/index.tsx
VWSCoronaDashboard1: Already done

VWSCoronaDashboard3 (approved):

src/components/mapChart/index.tsx
VWSCoronaDashboard1: This is what's in the official Highcharts typings ;)

VWSCoronaDashboard4: if you'd supply the subset that could be returned by your function, it would be plenty? :P

VWSCoronaDashboard4 (approved):

src/components/mapChart/index.tsx
VWSCoronaDashboard1: fair enough, changed it

PR #163, opened 2020-08-21
Lang=EN: Add language switcher

VWSCoronaDashboard4:
## Summary & Motivation

For the lang=en launch:
* Adds the language switcher to the header.
* Adjusts the areachart for reproduction index to exclude the week data.

## Detailed design

The route subpath is added to a base url, to maintain the current page that the user is on (eg when to receive a link from someone else)

Links use the proper lang / hreflang attributes to indicate language changes

The language strings are added to the locale files; yet they are not intended to be translated. It's merely for copy control. (The english text "Website in English" eg should be on the NL site when hovering the EN link.)

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

No specific mobile design was provided; perhaps a more elaborate implementation is wanted there (to occupy less screen estate)

### DIFF
--- a/src/components/layout/index.tsx
+++ b/src/components/layout/index.tsx
@@ -73,23 +73,23 @@ const Layout: FunctionComponentWithLayout<LayoutProps> = (props) => {
         <MaxWidth>
           <div className={styles.languageSwitcher}>
             <a
-              href={`${text.language_switcher.nl_base_url}${router.asPath}`}
+              href={`https://coronadashboard.rijksoverheid.nl${router.asPath}`}
               lang="nl"
               hrefLang="nl"
               className={locale === 'nl' ? styles.languageActive : undefined}
-              title={text.language_switcher.nl_title}
+              title="Website in het Nederlands"
             >
-              {text.language_switcher.nl}
+              NL
             </a>
             |
             <a
-              href={`${text.language_switcher.en_base_url}${router.asPath}`}
+              href={`https://coronadashboard.government.nl${router.asPath}`}
               lang="en-GB"
               hrefLang="en-GB"
               className={locale === 'en-GB' ? styles.languageActive : undefined}
-              title={text.language_switcher.en_title}
+              title="Website in English"
             >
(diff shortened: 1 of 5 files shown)


src/locale/en.json
```diff
@@ -525,5 +525,13 @@
...
     "default_title": "Dashboard Coronavirus COVID-19 | Rijksoverheid.nl",
     "default_url": "https://coronadashboard.rijksoverheid.nl"
+  },
+  "language_switcher": {
+    "en": "EN",
+    "en_title": "Website in English",
+    "en_base_url": "https://coronadashboard.government.nl",
+    "nl": "NL",
+    "nl_title": "Website in het Nederlands",
+    "nl_base_url": "https://coronadashboard.rijksoverheid.nl"
```
VWSCoronaDashboard3, line 535: Does this need to be in the locale? Seems like we can hardcode it as it doesn’t need translations and `NL | EN` and the url will remain the same.

VWSCoronaDashboard4: I've put it in the localization files merely to allow for changes to the copy. I can move it out of that is not needed.

src/components/layout/index.tsx
```diff
@@ -69,6 +71,27 @@ const Layout: FunctionComponentWithLayout<LayoutProps> = (props) => {
         </div>
 
         <MaxWidth>
+          <div className={styles.languageSwitcher}>
+            <a
+              href={`https://coronadashboard.rijksoverheid.nl${router.asPath}`}
+              lang="nl"
+              hrefLang="nl"
+              className={locale === 'nl' ? styles.languageActive : undefined}
+              title="Website in het Nederlands"
```
VWSCoronaDashboard1, line 80: nit: does this need to be a translation key as well? (I'm guessing not, because these texts should be the same in both languages)

VWSCoronaDashboard3 (approved):

VWSCoronaDashboard1 (approved):

PR #164, opened 2020-08-21
Fix rel noopener noreferrer

VWSCoronaDashboard4:
## Summary & Motivation

Fix external links opened in a new tab by providing the security measure `rel="noopener noreferrer"`

## Detailed design

Rewrite the way the options for linkify are handled to always have the rel as default in there.

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

n/a

### DIFF
--- a/src/components/replaceLinks/index.tsx
+++ b/src/components/replaceLinks/index.tsx
@@ -5,7 +5,12 @@ import Linkify, { LinkifyProps } from 'linkifyjs/react';
  */
 
 const ReplaceLinks: React.FC<IProps> = (props) => {
-  const { children, tagName, options } = props;
+  const { children, tagName } = props;
+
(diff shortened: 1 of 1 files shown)


VWSCoronaDashboard3 (commented):
This does not solve the links on for instance the `/verantwoording` page, where links come from markdown to HTML transform. The best solution would be to:

- Replace `snarkdown` and `dompurify` with `unfied` + `remark` + `remark-external-links` + `remark-rehype` + `rehype-sanitize`.
- Replace `ReplaceLinks` with updated `MDToHTMLString`.

VWSCoronaDashboard3:
This PR will be superseded by different PR which will implement steps as mentioned in the previous comment.

PR #176, opened 2020-08-25
Municipality maps on national pages

VWSCoronaDashboard1:
## Summary

- Put the municipality national overview maps on the appropriate pages and show the specified data there as a chloropleth
- Implement 'zooming in' on the safety region of the selected municipality

## Motivation

This is part of the reactors towards vNext of the site

## Detailed design

A DualComponent was added with a leftCol and rightCol prop to display content in two columns on desktop and underneath each other on mobile.

This was used on the national pages positief-geteste-mensen, verpleeghuis-sterfte and ziekenhuis-opnames where the different chloropleths needed to be shown.

The 'zoom in' on the safety region of the selected municipality is implemented in a data-driven fashion. Map features can be rendered only when relevant.
This is done by setting the 'allAreas' option to false. That way only features that can be joined with the provided data will be displayed. (configured by the 'joinBy' option, this configures the geojson property that needs to be matched with a data property).
When a municipality code is selected the useMunicipalityData hook simply retrieves all the other municipality codes that belong to the same safety region and filters the data based on those. This way automatically only that part of the map gets rendered.
The only extra conditional that needed to be added is the rendering of the country outline, this is now only done when NO selected municipality exists.

Furthermore, the typings have been further narrowed in the MunicipalityChart component.

Here's a screenshot of one of the national pages:
image

And one with a zoomed in region:
image

## Drawbacks

n/a

## Alternatives

Maybe my DualComponent isn't the desired solution. Please give alternatives, if any.

## Unresolved questions

The map isn't fully implemented in the regional parts. This is to keep this PR as small as possible.


src/components/dualColumn/index.tsx
```diff
@@ -0,0 +1,18 @@
+import { ReactNode } from 'react';
+
+import styles from './dualColumn.module.scss';
+
+export interface IProps {
+  leftCol: ReactNode;
+  rightCol: ReactNode;
+}
+
+export default function DualColumn(props: IProps) {
```
VWSCoronaDashboard3, line 10: It’s probably best to remove this component for now. Grid support is questionable in IE and with the new layout coming in we had to start clean anyway. It’s okay if everything is single column for now.

src/components/mapChart/index.tsx
```diff
@@ -40,12 +55,53 @@ const MunicipalityMap: React.FC<IProps> = ({
...
+        point: {
+          events: {
+            // @ts-ignore
+            click: function (this: TMunicipalityPoint) {
+              if (setSelection) {
+                this.select(this.selected, false);
+                setSelection({ id: this.Municipality_code });
+              }
+            },
+          },
```
VWSCoronaDashboard3, line 80: Click to zoom will not be a feature. But it might be implemented to check if you can make a zoomed in version?

src/utils/useMunicipalityData.ts
```diff
@@ -1,46 +1,87 @@
...
+  const regionCode = municipalCodeToRegionCode[code];
+
+  return Object.entries<[string, string]>(municipalCodeToRegionCode).reduce<
+    string[]
+  >((aggr, [mun, region]) => {
+    if (region === regionCode) {
+      aggr.push(mun);
+    }
+    return aggr;
+  }, []);
```
VWSCoronaDashboard3, line 34: nit: slightly hard to follow reduce with weird formatting

src/utils/useMunicipalityData.ts
```diff
@@ -1,46 +1,87 @@
...
+    );
+  }, [municipalCode]);
+
   return useMemo(() => {
     if (!data) {
       return [];
     }
 
-    const max: number = Math.min(
+    const maxDate: number = Math.min(
```
VWSCoronaDashboard3, line 64: This says `maxDate` but `.min` is used? Is that correct?

src/utils/useMunicipalityData.ts
```diff
@@ -1,46 +1,87 @@
...
           Boolean(item.Municipality_code) &&
-          new Date(item.Date_of_report).getTime() === max
-      )
-      .map(
-        (item: MunicipalityData): MunicipalityData => ({
-          ...item,
-          value: item[metricName],
-        })
-      );
+          new Date(item.Date_of_report).getTime() === maxDate;
```
VWSCoronaDashboard3, line 76: `.includes` is not supported in IE and we don’t have a polyfill for it. Can you try something else and split this up to make it a bit more readable, if possible.

VWSCoronaDashboard3 (commented):
👍

src/components/dualColumn/index.tsx
VWSCoronaDashboard1: I'm basing this off the design I got from Martijn:
image

src/components/mapChart/index.tsx
VWSCoronaDashboard1: This click handler doesn't perform the zoom, it just selects the feature in the map (i.e. changes its appearance) and reports the selection to the outside using the injected setSelection method.
Currently, this isn't used anywhere yet, so clicking on the map effectively does nothing. But I also thought this wasn't necessarily a huge load of code that was distracting.
Potentially, we could use this click through to the region page from the national page.

src/utils/useMunicipalityData.ts
VWSCoronaDashboard1: I agree, but this is what prettier makes of it. So if I change it, prettier will just put it back :(

src/utils/useMunicipalityData.ts
VWSCoronaDashboard1: Ouch, very well spotted... This is indeed not what is supposed to be happening. The newest date is supposed to be extracted from the data.

src/components/dualColumn/index.tsx
VWSCoronaDashboard1: Instead of grid I suppose I can just use floats instead then?

src/components/mapChart/index.tsx
VWSCoronaDashboard3: I would prefer to only merge code that is required at this point in time. When we need a click handler later, we can add it. But we might not and then we have this debt laying around.

src/utils/useMunicipalityData.ts
VWSCoronaDashboard3: Some inspiration for alternatives: https://twitter.com/jaffathecake/status/1213408744362692608, if you’re up for it. But it‘s a nit 🤷‍♂️

src/components/mapChart/index.tsx
VWSCoronaDashboard1: fair enough

src/utils/useMunicipalityData.ts
VWSCoronaDashboard1: I replaced the reduce with a filter and map call, should be a lot clearer now

src/components/dualColumn/index.tsx
VWSCoronaDashboard3: I’ll adjust when the new designs comes in. We can leave it as is for now.

src/utils/useMunicipalityData.ts
VWSCoronaDashboard1: Removed .includes in favor of indexOf() > -1 and created two named functions for the two filter flavors, should be more readable now.

src/components/dualColumn/index.tsx
VWSCoronaDashboard1: I removed the DualColumn component for now, awaiting the final designs

VWSCoronaDashboard3 (approved):

PR #194, opened 2020-08-26
Refactor React.FC to regular functions

VWSCoronaDashboard5:
## Summary

This PR refactors all places where the `React.FC` type is used throughout the codebase. All these components are turned into regular functions. When the component accepts `children`, the type `WithChildren` is used, see [this article](https://fettblog.eu/typescript-react-why-i-dont-use-react-fc/) for reference.

## Motivation

- Code style preferences
- `React.FC` type will be deprecated

## Detailed design

Example of refactor:
```
// before refactor
const Component: React.FC<IProps> = (props) => { // JSX }

// after refactor
function Component(props: IProps) { // JSX }
```

Furthermore this PR removes files no longer used throughout the codebase and places `export default` above the components for better readability.

## Unresolved Questions

What do we do with `FCWithLayout`, which uses:
```
export type FCWithLayout<Props = void> = React.FC<Props> & {
  getLayout: (page: React.ReactNode) => React.ReactNode;
};
```


src/types/index.ts
```diff
@@ -1,4 +1,7 @@
 // https://fettblog.eu/typescript-react-why-i-dont-use-react-fc/
-export type WithChildren<T = Record<string, unknown>> = T & {
+// Slightly altered version of the WithChildren type found in source
+// Added the {} in order for it to allow multiple children, simply wrapping it in a Fragment didn't work
```
VWSCoronaDashboard3, line 3: This comment is redundant. But a comment on how to use this type might help!

src/types/index.ts
```diff
@@ -1,4 +1,7 @@
...
+// eslint-disable-next-line
```
VWSCoronaDashboard3, line 4: You can you make this more specific by adding the rule name after `eslint-disable-next-line`. That way not every rule is ignored.

VWSCoronaDashboard3 (commented):
LGTM. To small things.

VWSCoronaDashboard3 (approved):

VWSCoronaDashboard4 (approved):

PR #196, opened 2020-08-28
Add JSON schema validation

VWSCoronaDashboard1:
## Summary

- Create JSON schema's for the different flavors of JSON data
- Add node scripts to execute validation for all the file in the public/json files
- Generate typescript interfaces from the JSON schema's

## Motivation

Data validation makes us feel safer.

## Detailed design

A new directory called 'schema' was added, it contains the JSON schema's for the various JSON types.
It also contains a directory called 'validator' which contains the scripts that perform the validations and code generation.

Three yarn scripts were added:

- validate-json This performs validation for the entire public/json folder
- validate-single This allows a user to validate a specific json file
- generate-typescript This generates the typescript interfaces and saves it to the file types/data.d.ts 

## Drawbacks

n/a

## Alternatives

Please let me know

## Unresolved questions

The JSON files are now all dropped in public/json. It might be an idea to create subdirectories per schema there. So public/json/national, public/json/regional, etc
This will make the validation logic a little more generic and scalable.


src/types/municipality.d.ts
```diff
@@ -0,0 +1,13 @@
+export interface MunicipalityMetrics {
+  Total_reported: number;
```
VWSCoronaDashboard, line 2: Why are these keys written with capital letters?

src/types/data.d.ts
```diff
@@ -0,0 +1,378 @@
...
+  max_rna_per_ml: number;
+  min_rna_per_ml: number;
+}
+
+/* tslint:disable */
+/**
+ * This file was automatically generated by json-schema-to-typescript.
+ * DO NOT MODIFY IT BY HAND. Instead, modify the source JSONSchema file,
+ * and run json-schema-to-typescript to regenerate this file.
+ */
```
VWSCoronaDashboard, line 84: Why is this message repeated 2 times? Are you piping the results in a single file?

src/pages/regio/index.tsx
```diff
@@ -158,7 +158,7 @@ const Regio: FCWithLayout<RegioProps> = (props) => {
     <>
       <MaxWidth>
         <LastUpdated
-          lastUpdated={data?.last_generated * 1000}
+          lastUpdated={+data?.last_generated * 1000}
```
VWSCoronaDashboard, line 161: Why are you doing this? Is it type coercion or is the value negative in some cases?

VWSCoronaDashboard (commented):
Overall very happy with this PR. Seems like a good investment in stability!

I've left some comments which are no dealbreakers.

Could you add some documentation around the validators themselves?

src/types/data.d.ts
VWSCoronaDashboard1: yes

src/types/municipality.d.ts
VWSCoronaDashboard1: This is temporary, will be removed once I have implemented the final municipal data. I was working with the wrong data files before. (this interface was in the repo already)

src/pages/regio/index.tsx
VWSCoronaDashboard1: The property is a string, this is a way of coercing it to number

schema/validator/generate-typescript.js
```diff
@@ -0,0 +1,48 @@
...
+
+// The directory where the resulting data.d.ts file will be saved
+const outputPath = path.join(__dirname, '..', '..', 'src', 'types');
+
+const schemaNames = getSchemaNames();
+
+const promises = schemaNames.map(generateTypeScriptFromSchema);
+
+Promise.all(promises).then((result) => {
+  saveFile(result.join('\n'));
```
VWSCoronaDashboard4, line 16: since you `console.info` the generation of the individual schemas, perhaps also `console.info` after writing to the file?

VWSCoronaDashboard4 (approved):

schema/municipal/municipal.json
```diff
@@ -0,0 +1,33 @@
...
+    "last_generated": {
+      "type": "string"
+    },
+    "proto_name": {
+      "type": "string",
+      "pattern": "^GM[0-9]+$"
+    },
+    "name": {
+      "type": "string",
+      "pattern": "^GM[0-9]+$"
```
VWSCoronaDashboard3, line 17: Nice to see you can add regex patterns, which makes this as strict as it can be.

VWSCoronaDashboard3 (approved):
Really impressed! The code, the script to compare all data, but also a CLI script to manually compare a file.

schema/municipal/municipal.json
VWSCoronaDashboard1: Oh, there is a LOT more that the schema is capable of, so I think in the future we can add even more rules to make things even more strict.

PR #197, opened 2020-08-28
perf: Get data on build-time to make a fully static build

VWSCoronaDashboard:
## Summary

This PR replaces most client-side fetches with static builds through getStaticProps.
## Motivation

This will increase (perceived) performance and reduce layout shifting. It also allows us to remove SWR from the packages as data will not have to be fetched client-side or invalidated.

## Detailed design

We use [getStaticProps](https://nextjs.org/docs/basic-features/data-fetching#getstaticprops-static-generation). We are dependent on the pipeline to include the jsons on build-time to do this.

1. First, I added getStaticProps to all pages. If we used SWR on a page, it now uses getStaticProps.
2. I have abstracted away all repetitive code from getStaticProps in a file. Take a look at `src/static-props/nl-data.ts`. This should make it easier to maintain the application. It would be even better if you could somehow re-use a getStaticProps export somewhere else by importing and re-exporting, but that sounds tricky.
3. I needed to put the pageProps (which we get from getStaticProps) in the persistent layout component. I've done this by passing pageProps as an argument in `_app.js`. It wasn't difficult, but maybe the TypeScript parts can be done better.
4. I've changed the routing of the region pages. Instead of using query parameters the app now uses dynamic URLs. In order to write out static HTML files Next.js needs to know the unique paths. A query parameter is not something you can convert into a filename (e.g. you can't write to `region?regio=VR01.html`). There's some duplicate code between the index and the slug page, but I think we can clean that up together with design by just designing a better empty state.
5. Because the data is available on build-time, we can remove a bunch of undefined states in TypeScript. I'm not sure if I caught them all, but this PR should make things more strict already.

## Drawbacks

**Not very DRY**
Re-using getStaticProps across different pages needs to be done by duplicating the logic for all pages. It's annoying, but there's no other way around it. I mitigated this by moving duplicate logic in helper functions. Until `_app.js` supports getStaticProps the duplication is here to stay I'm afraid.

**No de-duping of data**
image
This is a big drawback. We use the same data for all national pages, but Next.js is not smart enough to de-dupe this. This means for every route change you're going to download around 18KB of data for no reason.

I suspect we can make this smaller by not returning the full NL.json file for every page. The page content can use a specific key, reducing the filesize by a huge margin. However, the sidebar needs the values. We can try to filter that out and reduce the duplication across pages. If we can reduce this to <5KB or something I think it should be OK.

## Unresolved questions

1. I'm using `fs` and `path` because the official docs show that. I see we use `require` in other parts of the codebase to load JSON files in getStaticProps. Are there obvious downsides or upsides to either approach? Using require sounds simpler to me. I don't mind rewriting that logic to require if there's no downsides.

2. Do we want to keep SWR? We could pass the staticProps data as [initialData into SWR](https://swr.vercel.app/docs/with-nextjs#pre-rendering). As a result, the page loads will be fast because of the static builds and caching, but SWR will invalidate the data in the background. If, somehow, the data gets updated while you're looking at an older build it should replace the contents automatically.

> The page is still pre-rendered. That means it's SEO friendly, can be cached and accessed very fast. But after hydration, it’s also fully powered by SWR in the client side. Which means the data can be dynamic and update itself over time and user interactions.

3. Am I correct in assuming that data will never be undefined again when using getStaticProps? How does that work after hydration? We need to verify that I didn't break any loading states.


src/pages/landelijk/besmettelijke-mensen.tsx
```diff
@@ -49,9 +52,7 @@ export function InfectiousPeopleBarScale(props: {
   );
 }
 
-const InfectiousPeople: FCWithLayout = () => {
-  const { data } = useSWR(`/json/NL.json`);
-
+const InfectiousPeople: FCWithLayout = ({ data }) => {
```
VWSCoronaDashboard1, line 55:
```
interface IProps {
  data: National;
}

const InfectiousPeople: FCWithLayout<IProps> = ({ data }) => {
```

src/pages/landelijk/besmettelijke-mensen.tsx
```diff
@@ -123,4 +124,18 @@ const InfectiousPeople: FCWithLayout = () => {
 
 InfectiousPeople.getLayout = getNationalLayout();
 
+// This function gets called at build time on server-side.
+// It won't be called on client-side.
+export const getStaticProps: GetStaticProps = async () => {
```
VWSCoronaDashboard1, line 129: `export const getStaticProps: GetStaticProps<IProps> = async () => {`

src/pages/landelijk/besmettelijke-mensen.tsx
```diff
@@ -123,4 +124,18 @@ const InfectiousPeople: FCWithLayout = () => {
...
+  const jsonDirectory = path.join(process.cwd(), 'public/json');
```
VWSCoronaDashboard1, line 130: path.join() takes any number of arguments. So this would work too:

`path.join(process.cwd(), 'public', 'json', 'NL.json');`

It would also ensure cross-platform path delimiters

VWSCoronaDashboard1 (changes requested):

VWSCoronaDashboard1 (approved):

src/components/layout/NationalLayout.tsx
```diff
@@ -28,19 +27,33 @@ import CoronaVirus from 'assets/coronavirus.svg';
...
+      <NationalLayout pageProps={pageProps}>{page}</NationalLayout>
     );
   };
 }
 
+interface PageProps {
+  pageProps: {
+    data: National;
+  };
+}
```
VWSCoronaDashboard, line 55: This type should be required for all pages that use <NationalLayout>. Did I implement that correctly?

src/components/layout/index.tsx
```diff
@@ -22,7 +22,7 @@ export interface LayoutProps {
 }
 
 export type FCWithLayout<Props = void> = React.FC<Props> & {
-  getLayout: (page: React.ReactNode) => React.ReactNode;
+  getLayout: (page: React.ReactNode, pageProps: any) => React.ReactNode;
```
VWSCoronaDashboard, line 25: In theory this can be any, but on NationalLayout it should always be this one:

```ts
interface PageProps {
  pageProps: {
    data: National;
  };
}
```

not sure how do type that out.

src/components/tiles/regio/IntakeHospital.tsx
```diff
@@ -16,8 +16,8 @@ import { Regionaal, RegionaalValue } from 'types/data';
 import siteText from 'locale';
 
 interface IProps {
-  data: Regionaal;
-  selectedRegio: SafetyRegion | undefined;
+  data?: Regionaal;
```
VWSCoronaDashboard, line 19: The region page wrongly assumed that data can not be undefined. If data is undefined, it shows a loading state.

src/pages/_app.tsx
```diff
@@ -50,7 +51,7 @@ function MyApp(props: IProps): React.ReactElement {
         fetcher,
       }}
     >
-      {getLayout(<Component {...pageProps} />)}
+      {getLayout(<Component {...pageProps} />, pageProps)}
```
VWSCoronaDashboard, line 54: This is the magic to get the data from getStaticProps in persistent layouts.

src/pages/landelijk/besmettelijke-mensen.tsx
```diff
@@ -125,4 +125,10 @@ const InfectiousPeople: FCWithLayout = () => {
 
 InfectiousPeople.getLayout = getNationalLayout();
 
+// This function gets called at build time on server-side.
+// It won't be called on client-side.
+export const getStaticProps: GetStaticProps<NationalLayoutProps> = async () => {
+  return getNlData();
```
VWSCoronaDashboard, line 131: I'm amazed that this works, but it works! This keeps it semi-dry.

src/pages/regio/[region].tsx
```diff
@@ -0,0 +1,233 @@
+import fs from 'fs';
```
Comment (VWSCoronaDashboard) [inline src/pages/regio/[region].tsx:1]: This whole file is basically a duplicate of `src/pages/regio/index.tsx` before I reduced it (the index file) to a simpler version that's an empty state with a search bar.

src/pages/regio/[region].tsx
```diff
@@ -0,0 +1,233 @@
...
+   * users so they know what has changed.
+   */
+  const focusFirstHeading = () => {
+    const el = contentRef.current as any;
+    if (el) el.focus();
+  };
+
+  const setSelectedRegio = (safetyRegionCode: SafetyRegion['code']): void => {
+    router.replace('/regio/[region]', `/regio/${safetyRegionCode}`);
+  };
```
Comment (VWSCoronaDashboard) [inline src/pages/regio/[region].tsx:155]: This has changed to support dynamic routes instead of slugs. We might want to add redirects to the app for people who bookmarked the pages with query parameters?

src/pages/regio/[region].tsx
```diff
@@ -0,0 +1,233 @@
...
+
+  useEffect(focusFirstHeading, [data]);
```
Comment (VWSCoronaDashboard) [inline src/pages/regio/[region].tsx:157]: This still works, even when it's as short as it is right now. It basically fires on page load.

src/pages/regio/[region].tsx
Comment (VWSCoronaDashboard) [reply src/pages/regio/[region].tsx:155]: I still need to solve a bug where resetting the searchbar (which returns undefined) triggers a replace() but we should ignore that value.

src/components/layout/NationalLayout.tsx
VWSCoronaDashboard1: If it needs to be used by other pages, then definitely export the interface otherwise it can't be found.

src/pages/_app.tsx
VWSCoronaDashboard1, line 54: Just for completeness, I'd type the Component prop to FCWithLayout in the IProps interface on this page

src/static-props/nl-data.ts
```diff
@@ -0,0 +1,19 @@
+import fs from 'fs';
+import path from 'path';
+
+import { NationalLayoutProps } from 'components/layout/NationalLayout';
+
+interface IProps {
```
VWSCoronaDashboard1, line 6:
```suggestion
export interface IProps {
```

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard3 (changes requested):
Overall looking good! But I have some concerns.

---

> I've changed the routing of the region pages

Please refrain from doing this. This is outside of the scope of this ticket and we are currently completely rebuilding /regio.

> No de-duping of data

This is indeed a big drawback. I’m sure webpack is smart enough to de-dupe imports. Are we sure this is the right approach over `import`?

> 

> I'm using fs and path because the official docs show that. I see we use require in other parts of the codebase to load JSON files in getStaticProps. Are there obvious downsides or upsides to either approach? Using require sounds simpler to me. I don't mind rewriting that logic to require if there's no downsides.

AFAIK, there are no downsides to `require`. Both fix directory slashes for different operating systems

> Do we want to keep SWR?

I don’t think we need it anymore.

VWSCoronaDashboard:
> Please refrain from doing this. This is outside of the scope of this ticket and we are currently completely rebuilding /regio.

This is unfortunately unavoidable if we want static pages. We can choose to wait with this improvement until things are more stable in the next release before committing to such a change?

> This is indeed a big drawback. I’m sure webpack is smart enough to de-dupe imports. Are we sure this is the right approach over import?

I have my doubts too :)

PR #198, opened 2020-08-28
Add new municipalities level (via toggle) to /regio

VWSCoronaDashboard4:
## Summary & Motivation

Adds data for municipalities to the region page.

## Detailed design

* A new map is added, it plots data based on the municipalities' geodata.
* A new mapping between municipalities / gemcode / vrcode is added to aid the select box.
* The region page is adjusted to toggle between Municipality / Safety region
* Extra (mostly duplicate, structure wise) tiles added
* Sewer tile for municipalities is adjusted to properly support zero, one or more RWZI locations.
* The select municipality component is adjusted to support two structures


src/components/mapChart/municipality.tsx
```diff
@@ -0,0 +1,176 @@
...
+          {...({ ref } as any)}
+          highcharts={Highcharts}
+          containerProps={{ style: { width: '631px', height: '100%' } }}
+          constructorType={'mapChart'}
+          options={mapOptions}
+        />
+      </div>
+    </div>
+  ) : (
+    <Fragment></Fragment>
```
VWSCoronaDashboard1, line 172: You can just return null here, no need for an empty Fragment

src/components/tiles/municipality/IntakeHospital.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+import siteText from 'locale';
+
+interface IProps {
+  data: RegionaalMunicipality;
+  selectedRegio: MunicipalityMapping | undefined;
+  contentRef: React.RefObject<HTMLHeadingElement>;
+}
+
+export const IntakeHospitalMunicipality: React.FC<IProps> = (props) => {
```
VWSCoronaDashboard1, line 25: I think we've been moving away from using the React.FC interface in favor of just a function with a typed param.
I.e. function IntakeHospitalMunicipality(props: IProps)

src/components/selectMunicipality/index.tsx
```diff
@@ -200,6 +218,7 @@ const SelectMunicipality: React.FC<SelectMunicipalityProps> = (props): any => {
         })}
       >
         {isOpen &&
+          regionType === 'safetyRegion' &&
```
VWSCoronaDashboard1, line 221: I see a lot of conditional rendering in this component. Would it be a cleaner solution to perhaps split this up into two smaller components?

src/components/tiles/municipality/IntakeHospital.tsx
```diff
@@ -0,0 +1,129 @@
...
+          headingRef={contentRef}
+          regio={selectedRegio?.name}
+        />
+
+        <p>{siteText.regionaal_municipality_ziekenhuisopnames_per_dag.text}</p>
+
+        {!selectedRegio && <RegioDataLoading />}
+
+        {selectedRegio && (
+          <>
```
VWSCoronaDashboard1, line 45: This seems to have quite a bit of conditional rendering as well. Perhaps also a candidate to split up into smaller pieces?

src/components/tiles/municipality/PositivelyTestedPeople.tsx
```diff
@@ -0,0 +1,149 @@
...
+          }
+          regio={selectedRegio?.name}
+        />
+
+        <p>{siteText.regionaal_municipality_positief_geteste_personen.text}</p>
+
+        {!selectedRegio && <RegioDataLoading />}
+
+        {selectedRegio && (
+          <>
```
VWSCoronaDashboard1, line 45: And another one that can perhaps be split up into smaller components.

src/components/tiles/municipality/SewerWater.tsx
```diff
@@ -0,0 +1,247 @@
...
+        ) || [];
+
+      averageValues = data?.sewer_measurements?.values || [];
+      allValues =
+        data?.results_per_sewer_installation_per_municipality?.values || [];
+      averageValueKey = 'average';
+      averageDateKey = 'week_unix';
+      averageLabelText = text.graph_average_label_text;
+    }
+  }
```
VWSCoronaDashboard1, line 105: This code could probably be refactored into a hook, and perhaps could use some useMemo() since there's quite a bit of data massaging going on

src/pages/index.tsx
```diff
@@ -73,7 +73,11 @@ const Home: FunctionComponentWithLayout = () => {
       </Head>
 
       <MaxWidth>
-        <LastUpdated lastUpdated={data?.last_generated * 1000} />
+        <LastUpdated
+          lastUpdated={
+            data?.last_generated ? parseInt(data.last_generated, 10) * 1000 : 0
+          }
```
VWSCoronaDashboard1, line 79: Wasn't the parseInt logic added to the LastUpdated component, therefore data?.last_generated can just be set as the prop value?

VWSCoronaDashboard1 (changes requested):

src/components/tiles/municipality/IntakeHospital.tsx
VWSCoronaDashboard4: The intake hospital file is structurally equal to the one for safety regions. Optimizing it would be better suited for a separate task/PR.

src/components/tiles/municipality/PositivelyTestedPeople.tsx
VWSCoronaDashboard4: Same here; The positively tested people file is structurally equal to the one for safety regions. Optimizing it would be better suited for a separate task/PR.

VWSCoronaDashboard3 (changes requested):

PR #213, opened 2020-09-01
Feature: conditional redirect on '/landelijk'

VWSCoronaDashboard5:
## Summary

Rather controversial to trigger a redirect based on viewport-size, but it is the only way to handle the current design where the mobile view shows a list of all `Medische Factoren` and desktop automatically sets the first tile `/landelijk/positief-geteste-mensen` to be active. 

## Motivation

Have the first tile selected by default on desktop views.

## Detailed design

Allow for the first tile to be active on a visit on `/landelijk` where `isLargeScreen` is true on `NationalLayout`


VWSCoronaDashboard3 (commented):
Can you test if moving this logic up into the React tree makes it more snappy? Can you also try `useLayoutEffect`?

VWSCoronaDashboard:
@VWSCoronaDashboard3 Maybe unrelated to this PR, but I saw some redirects in the next.config.js. Are we sure that works with a fully static site? I suspect that only works in Node.js environments (e.g. when running next start).

VWSCoronaDashboard3:
> @VWSCoronaDashboard3 Maybe unrelated to this PR, but I saw some redirects in the next.config.js. Are we sure that works with a fully static site? I suspect that only works in Node.js environments (e.g. when running next start).

`next.config.js` is a static file, not aware of the view port. We need a redirect on desktop screens for `/landelijk`, but not for mobile.

VWSCoronaDashboard3 (approved):

PR #223, opened 2020-09-02
feature: "terug naar alle cijfers" (back button) on mobile

VWSCoronaDashboard5:
## Summary

This PR adds a link that allows mobile users to go back to the `/landelijk` overview after selecting a number view.

## Motivation

This enables users to navigate back to the overview.

## Detailed design

I started on a `useRoutes` file, this could be used in the future to write routing logic. For now it holds a list of all the current pages. I needed these pages in order for the render logic (see the `showBackButton` boolean based on `includes(router.pathname)`. The showBackButton is showed when:
1. the current route is one of the routes of the `/landelijk` pages
2. the viewport is mobile/tablet (under the breakpoint of `1000px`).

Furthermore this PR adds a `box-shadow-bottom` variable that allows for only setting the box shadow on the bottom side of the element. I also reused the `Arrow` used in the `Collapse` component and rotated it.

## Alternatives

I started working on this PR by adding the button in the `NationalLayout` compositions, but this would conflict with the paddings and margins applied to the compositions, I also think it makes more sense adding the anchor the `nav`, which is in the `layout/index` file.

## Unresolved questions

n/a 

### Governance

- [x] Documentation is added


src/components/layout/index.tsx
```diff
@@ -159,6 +165,14 @@ function Layout(props: WithChildren<LayoutProps>) {
               </li>
             </ul>
           </MaxWidth>
+          {showBackButton && (
+            <Link href="/landelijk" passHref>
```
VWSCoronaDashboard3, line 169: What is `passHref`?

src/components/layout/index.tsx
```diff
@@ -49,6 +52,9 @@ function Layout(props: WithChildren<LayoutProps>) {
 
   const locale = getLocale();
   const showSmallLogo = useMediaQuery('(max-width: 480px)', true);
+  const showBackButton =
+    useMediaQuery('(max-width: 1000px)', false) &&
+    Object.values(PAGES.landelijk).includes(router.pathname);
```
VWSCoronaDashboard3, line 57: I don’t think this is required? You might in the wrong layout file. It should probably be `NationalLayout`

VWSCoronaDashboard3 (changes requested):

src/components/layout/layout.module.scss
```diff
@@ -72,6 +72,25 @@ $language-switcher-offset: 55px;
...
 
+  > a {
+    background: white;
+    display: block;
+    width: 100%;
+    padding: 1em;
+    box-shadow: $box-shadow-bottom;
+    text-decoration: none;
+
+    &:hover {
```
VWSCoronaDashboard4, line 83: also add &:focus

src/scss/variables.scss
```diff
@@ -34,3 +34,5 @@ $default-margin: 1rem;
 
 $box-shadow: 0 -1px 1px 0 lightgrey, 0 1px 1px 0 lightgrey,
   0 2px 2px 0 lightgrey, 0 4px 4px 0 lightgrey, 0 8px 8px 0 lightgrey;
+
+$box-shadow-bottom: 0 4px 2px -2px lightgrey;
```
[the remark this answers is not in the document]
VWSCoronaDashboard5: i see a merge conflict in `scss/variables.scss` which is probably caused by this, will look into it now

src/components/layout/index.tsx
VWSCoronaDashboard5: at first I thought that `passHref` was needed if you use the `Link` when wrapping it around an `a`, but [looking into it](https://nextjs.org/docs/api-reference/next/link#if-the-child-is-a-custom-component-that-wraps-an-a-tag) now I see that it is only needed when you use libraries such as `stlyed systems` and want to style the `a` in a `const` (and than you can't set the `href` to the component). so in this case it is not needed, I'll remove it 👍

src/components/layout/index.tsx
VWSCoronaDashboard5: See `Alternatives` in my PR Description

VWSCoronaDashboard5: > I started working on this PR by adding the button in the NationalLayout compositions, but this would conflict with the paddings and margins applied to the compositions, I also think it makes more sense adding the anchor the nav, which is in the layout/index file.

VWSCoronaDashboard5: oops, by `NationalLayout compositions` I'm referring to the `ContentHeader` component

src/components/layout/layout.module.scss
VWSCoronaDashboard4: Fix could've combined it like: `&:hover, &:focus { ... }`

VWSCoronaDashboard4 (approved):

src/scss/variables.scss
```diff
@@ -32,5 +32,7 @@ $default-radius: 5px;
 $default-padding: 1rem;
 $default-margin: 1rem;
 
-$box-shadow: 0 -1px 1px 0 #e5e5e5, 0 1px 1px 0 #e5e5e5,
-  0 2px 2px 0 #e5e5e5, 0 4px 4px 0 #e5e5e5, 0 6px 6px 0 #e5e5e5;
+$box-shadow: 0 -1px 1px 0 #e5e5e5, 0 1px 1px 0 #e5e5e5, 0 2px 2px 0 #e5e5e5,
+  0 4px 4px 0 #e5e5e5, 0 6px 6px 0 #e5e5e5;
+
+$box-shadow-bottom: 0 4px 2px -2px lightgrey;
```
VWSCoronaDashboard3, line 38: This should be the same color as the other box-shadow. But you might be able to use the one above because the top and sides shadow will be cut off if the element is full-width and against another element on top.

src/scss/layout.scss
```diff
@@ -77,6 +77,30 @@
 .national-layout,
 .safety-region-layout {
   display: flex;
+  flex-direction: column;
+
+  > a {
```
VWSCoronaDashboard3, line 82: Can you create a class specially for this button? This will trip things up if we add another link.

VWSCoronaDashboard3 (commented):
Almost good to go!

VWSCoronaDashboard3 (approved):

PR #227, opened 2020-09-02
Add new combobox and dynamic routes for /veiligheidsregio

VWSCoronaDashboard3:
## Summary

- Add new combobox component (with solid types!)
- Add `useThrottle` 
- Implement dynamic routes for /veiligheidsregio
- Demo dynamic data fetching in one page

## Motivation

New vision

## Detailed design
- `downshift` is hard to use and has a big bundle size footprint so a new combobox has been created with `@reach/combobox`.

It accepts a generic type which extends the `{name: string}` type. This means you can pass an array of objects with strictly typed keys and values, but it has to at least contain a `name: string`. This also makes the `handleSelect` function more strict.

```js
<Combobox<TSafetyRegion>
  handleSelect={handleSafeRegionSelect}
  options={safetyRegions}
/>
```
A similar typed approach has been taken in `useThrottle`.

- Pages in `veiligheidsregio` have now moved to `veiligheidsregio/[code]`
- New `index` file in `veiligheidsregio`

## Unresolved questions

This means we need to keep `useSRW`. Unlike stated in #197 .


VWSCoronaDashboard1:
Really nice, the new Combobox is super clean. Cool!

VWSCoronaDashboard1 (approved):

src/utils/useThrottle.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export default useThrottle;
+
+// This hook uses a generic type to match the input with the output.
+//
+// Example:
+// ```
+// const throttledTerm = useThrottle<string>(term, 100);
+// ```
+function useThrottle<Value>(value: Value, limit: number): Value {
```
VWSCoronaDashboard4, line 11: Isn't the more appropriate term for this behavior a `debounce` instead of `throttle`?

src/components/comboBox/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+};
+
+type TProps<Option extends TOption> = {
+  options: Option[];
+  handleSelect: (option: Option) => void;
+};
+
+export default ComboBox;
+
+function ComboBox<Option extends TOption>(props: TProps<Option>) {
```
VWSCoronaDashboard4, line 41: looks neat!

src/components/comboBox/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const { options, handleSelect } = props;
+
+  const inputRef = useRef<HTMLInputElement>();
+  const [term, setTerm] = useState<string>('');
+  const results = useSearchedOptions<Option>(term, options);
+
+  function handleChange(event: React.ChangeEvent<HTMLInputElement>) {
```
VWSCoronaDashboard4, line 48:
```suggestion
  function handleChange(event: React.ChangeEvent<HTMLInputElement>): void {
```

src/components/comboBox/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+    setTerm(event.target.value);
+  }
+
+  function onSelect(name: string) {
```
VWSCoronaDashboard4, line 52:
```suggestion
  function onSelect(name: string): void {
```

src/utils/useThrottle.ts
VWSCoronaDashboard3: They’re not the same thing: https://css-tricks.com/the-difference-between-throttling-and-debouncing/

src/components/comboBox/index.tsx
VWSCoronaDashboard3: Event handlers are inherently void, I think. Plus return types mostly add value when you actually return something, somewhere 🤷‍♂️

PR #240, opened 2020-09-04
feat: Allow combobox to search for alternative spellings

VWSCoronaDashboard:
## Summary

Currently, the combobox values have to be equal to municipality names in the GeoJSON files.
E.g. we can't match `'s Gravenhage`, the correct spelling of the municipality, to `s-gravenhage`, the format used in the GeoJSON.

image

In this PR I show how we can not only correct the display of the municipalities, but I also show how we can search for extra terms to match the correct municipality. E.g. you can now search for `Den Haag` as well.

image

## Motivation
The search function will be easier to use by visitors.

## Detailed design

I've added support for an optional `displayName` key. For example, `option.displayName = "'s Gravenhage"` Whenever a search option provides this as a value, we use it in the UI. If it can't be found, we still use `option.name` as we did before. 

I've added a new array value as optional key to items, called `searchTerms`, which we use to match alternative spellings of search items.

E.g.
```json
  {
    "name": "s-Gravenhage",
    "displayName": "'s Gravenhage",
    "safetyRegion": "VR15",
    "gemcode": "GM0518",
    "searchTerms": ["Den Haag"]
  }
```

Both `displayName` and `searchTerms` are passed as extra keys to matchSorter, so they are taken into account while searching.

```ts
    return matchSorter(options, throttledTerm, {
      keys: [(item: Option) => item.name, 'searchTerms', 'displayName'],
    });
```

## Drawbacks

I can't see any drawbacks. Most of the heavy lifting is done by matchSorter. All we need to do is add alternative spellings to some regions and municipalities in the codebase.

Because handleSelect is a callback function (Inversion of control) it means how we show the options in the list has no impact on existing selection or redirection logic.

## Unresolved questions

~I'm not happy with the property name `searchTerms`, should we rename that to `alternativeSpellings` or similar?~ Solved by renaming to variations.

### DIFF
--- a/src/components/layout/MunicipalityLayout.tsx
+++ b/src/components/layout/MunicipalityLayout.tsx
@@ -29,6 +29,7 @@ type TMuncipality = {
   displayName?: string;
   safetyRegion: string;
   gemcode: string;
+  searchTerms: string[];
 };
 
 export function getMunicipalityLayout() {
--- a/src/components/layout/MunicipalityLayout.tsx
+++ b/src/components/layout/MunicipalityLayout.tsx
@@ -29,7 +29,7 @@ type TMuncipality = {
   displayName?: string;
   safetyRegion: string;
   gemcode: string;
-  searchTerms: string[];
+  searchTerms?: string[];
 };
 
 export function getMunicipalityLayout() {
--- a/src/components/layout/MunicipalityLayout.tsx
+++ b/src/components/layout/MunicipalityLayout.tsx
@@ -24,14 +24,7 @@ import municipalities from 'data/gemeente_veiligheidsregio.json';
 
 export default MunicipalityLayout;
 
-type TMuncipality = {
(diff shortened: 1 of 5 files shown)


src/components/layout/MunicipalityLayout.tsx
```diff
@@ -19,8 +20,18 @@ import { WithChildren } from 'types';
...
 export default MunicipalityLayout;
 
+type TMuncipality = {
+  name: string;
+  displayName?: string;
+  safetyRegion: string;
+  gemcode: string;
+  searchTerms?: string[];
+};
+
```
VWSCoronaDashboard1, line 34:
```suggestion
export type TMunicipality = typeof municipalities;
```
This will automatically construct a type based on the JSON shape.

VWSCoronaDashboard1 (approved):

PR #246, opened 2020-09-07
Implement static props and paths instead of useSWR

VWSCoronaDashboard3:
## Summary

- Pass `pageProps` in `_app.tsx` to `getLayout`.
- Use passed props in layout components instead of `useSWR`
- Create `src/static-props/nl-data.ts` and implement on all `/landelijk` pages.
- Create `src/static-props/municipality-data.ts` and implement on all `/gemeente` pages.
- Create `src/static-props/safefyregion-data.ts` and implement on all `/veiligheidsregio` pages.

## Motivation

- Pre-build pages instead of fetching data on run-time.
- Only build valid pages and redirect to 404 for invalid codes.
- Remove the need for `useSWR`, resulting in a smaller bundle size.

## Detailed design

n\a

## Drawbacks

Not sure

## Alternatives

- Run time fetches
- `import` statements

## Unresolved questions

1. We should figure out a way to get rid of the remaining uses of `useSWR`:

Screenshot 2020-09-07 at 16 42 05

2. We should try to inject the locale via `getStaticProps` otherwise we will never be able to only ship the needed locale (currently `en.json` and `nl.json` are always in the bundle). This is because the build step cannot determine whether only one import is needed because we export them conditionally.

### DIFF
--- a/src/components/layout/MunicipalityLayout.tsx
+++ b/src/components/layout/MunicipalityLayout.tsx
@@ -1,7 +1,5 @@
 import Link from 'next/link';
 import Head from 'next/head';
-import { useState } from 'react';
-import useSWR from 'swr';
 import { useRouter } from 'next/router';
 
 import TitleWithIcon from 'components/titleWithIcon';
@@ -21,17 +19,21 @@ import siteText from 'locale';
 import { WithChildren } from 'types';
 
 import useMediaQuery from 'utils/useMediaQuery';
-import { Municipal } from 'types/data';
 
 import municipalities from 'data/gemeente_veiligheidsregio.json';
+import { IMunicipalityData } from 'static-props/municipality-data';
 
 export default MunicipalityLayout;
 
 export type TMunicipality = typeof municipalities;
+
 export function getMunicipalityLayout() {
-  return function (page: React.ReactNode): React.ReactNode {
+  return function (
+    page: React.ReactNode,
+    pageProps: IMunicipalityData
+  ): React.ReactNode {
     return getSiteLayout(siteText.gemeente_metadata)(
-      <MunicipalityLayout>{page}</MunicipalityLayout>
+      <MunicipalityLayout {...pageProps}>{page}</MunicipalityLayout>
     );
   };
 }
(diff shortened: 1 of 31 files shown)


VWSCoronaDashboard:
> We should try to inject the locale via getStaticProps otherwise we will never be able to only ship the needed locale (currently en.json and nl.json are always in the bundle). This is because the build step cannot determine whether only one import is needed because we export them conditionally.

You can read this from `process.env.NEXT_PUBLIC_LOCALE` which should be available on build-time and run-time.

src/pages/veiligheidsregio/index.tsx
```diff
@@ -1,10 +1,18 @@
...
 
-const SafetyRegion: FCWithLayout = () => {
+const SafetyRegion: FCWithLayout<ISafetyRegionData> = () => {
   return null;
 };
 
 SafetyRegion.getLayout = getSafetyRegionLayout();
 
+export const getStaticProps = getSafetyRegionData();
+export const getStaticPaths = getSafetyRegionPaths();
```
VWSCoronaDashboard, line 16: 🙌 Impressed that this is possible!

VWSCoronaDashboard (approved):
Hyped for this!

src/components/layout/MunicipalityLayout.tsx
```diff
@@ -21,17 +19,21 @@ import siteText from 'locale';
...
 
 import useMediaQuery from 'utils/useMediaQuery';
-import { Municipal } from 'types/data';
 
 import municipalities from 'data/gemeente_veiligheidsregio.json';
+import { IMunicipalityData } from 'static-props/municipality-data';
 
 export default MunicipalityLayout;
 
 export type TMunicipality = typeof municipalities;
```
VWSCoronaDashboard1, line 28: I know this was already part of the repo, but now that I look at it again, this is probably nicer:
```suggestion
export type TMunicipalities = typeof municipalities;
export type TMunicipality = TMunicipalities[number];
```

src/components/layout/MunicipalityLayout.tsx
```diff
@@ -76,8 +75,6 @@ function MunicipalityLayout(props: WithChildren) {
   }
 
   function handleMunicipalitySelect(region: TMunicipality[number]) {
```
VWSCoronaDashboard1, line 77: If the aforementioned change is made, then this can simply become:
```suggestion
  function handleMunicipalitySelect(region: TMunicipality) {
```

VWSCoronaDashboard4 (approved):

VWSCoronaDashboard1 (approved):

PR #256, opened 2020-09-09
Feature/vx

VWSCoronaDashboard1:
## Summary

Total conversion of the chloropleth maps. Highcharts was kicked out in favor of @vx which is a more low-level charting lib specifically aimed at React. It's basically D3 with a bunch of React helpers which gives much more control over how a chart is being rendered.

## Motivation

Highcharts became an impediment as we were slowly fighting the configuration details instead of actually making progress. When it became apparent we had to make design concessions because of another Highchart shortcoming, we chose to ditch Highcharts altogether and look for a new library. 

## Detailed design

Currently, only the chloropleths are ported and they both live in the components/vx folder. The main entry points being MunicipalityMap and SafetyRegionMap. Both of which wraps a MunicipalityChloropleth or SafetyRegionChloropleth component in a div that controls its sizing.
The municipal map will zoom into the corresponding safety region when a selected municipality code is passed in. This is done by filtering the given map features on just the municipalities that belong to the same safety region and feeding this into the fitSize property of the Mercator component. (hooray for data-driven components)
The SafetyRegionMap component is largely the same, it just loads different map data and chloropleth data.

## Drawbacks

It takes time to do the conversion of highcharts to vx. But this is an investment that will pay itself back in ease of development in the long run.

## Alternatives

Please let us know ;)

## Unresolved questions

Currently, the useMunicipalityData and useRegionData still depend on SWR to load its data. What is the preferred way of pulling in the data here? Just a regular import?

### Governance

- [ ] Documentation is added


next.config.js
```diff
@@ -1,11 +1,19 @@
...
+  'react-use-measure',
+]);
+
 const withBundleAnalyzer = require('@next/bundle-analyzer')({
   enabled: process.env.ANALYZE === 'true',
 });
 
 const nextConfig = {
   webpack(config, { dev }) {
+    config.optimization.minimizer = [];
```
VWSCoronaDashboard, line 16: We should not forget to revert this

next.config.js
```diff
@@ -1,11 +1,19 @@
 const withPlugins = require('next-compose-plugins');
 const withOptimizedImages = require('next-optimized-images');
+const withTM = require('next-transpile-modules')([
+  '@juggle/resize-observer',
+  '@vx/tooltip',
+  '@vx/event',
+  'react-use-measure',
```
VWSCoronaDashboard, line 7: I want to disable these 1-by-1 when the feature is complete, so we only transpile the absolute minimum of packages ourselves.

package.json
```diff
@@ -25,7 +29,9 @@
     "remark-external-links": "^7.0.0",
     "remark-parse": "^8.0.3",
     "remark-rehype": "^7.0.0",
+    "resize-observer-polyfill": "^1.5.1",
```
VWSCoronaDashboard, line 32: We can remove this one because we use @juggle/resize-observer?

src/components/vx/use-chart-dimensions.tsx
```diff
@@ -0,0 +1,68 @@
+import { ResizeObserver } from '@juggle/resize-observer';
```
VWSCoronaDashboard, line 1: I'll put in some documentation here to say that I took this logic from https://wattenberger.com/blog/react-and-d3

src/utils/useMunicipalityData.ts
```diff
@@ -30,28 +21,30 @@ function createMunicipalCodeFilter<T extends { gmcode: string }>(
...
+          ...item,
+          value: (item as any)[metricName],
+        };
+        return aggr;
+      },
+      {}
     );
-  }, [metricItems, metricName, municipalCodes]);
+
+    return filteredData;
```
VWSCoronaDashboard, line 48: If we extract this logic to another helper file, we can re-use it in getStaticProps to retrieve municipality data for a specific page. Then we don't need SWR anymore for municipality data.

The important bit is that we only return the data it needs to build the page, to keep the sizes down. If that proves to be too difficult, we might be better of lazy-loading this data client-side like we do right now with SWR so we only download `/json/municipalities.json` once.

We will need to measure the performance of both options before we make a decision. I wouldn't even want to make it a part of this PR yet, to keep things easier for ourselves.

VWSCoronaDashboard (commented):
Left some generic comments around the codebase, no explicit approval or rejection meant.

src/components/vx/MunicipalityMap.tsx
```diff
@@ -0,0 +1,28 @@
...
+};
+
+export default function MunicipalityMap(props: IMunicipalityMapProps) {
+  const [ref, dms] = useChartDimensions(chartSettings);
+
+  const { width, height } = dms;
+
+  return (
+    <div ref={ref} className={styles.chloroplethContainer}>
+      <MunicipalityChloropleth width={width} height={height} {...props} />
```
VWSCoronaDashboard, line 25: We made a small mistake here. We should be using boundedWidth and boundedHeight for the chart, because that's the width and height with the margins subtracted.

next.config.js
VWSCoronaDashboard1: Good point, I'll do that right away

package.json
VWSCoronaDashboard1: I'll try and see if anything explodes or not

src/components/vx/MunicipalityMap.tsx
VWSCoronaDashboard: This fix is pushed.

VWSCoronaDashboard4:
Reviewed the code and it looks good and reasonable. Some of the specific data logic / type stuff are harder to review unless your work with it or take more time to get the mental picture.

If I would comment stuff it would only be minor type stuff (return types, getting rid of some any types, simplifying some types); tidying up imports (nitpick) but I guess that should not hold us back for merging this.

VWSCoronaDashboard3 (approved):
Impressive! Can’t wait to see how it looks and feels in the browser.

PR #262, opened 2020-09-09
chore: visual tweaks to charts

VWSCoronaDashboard5:
## Summary

- adds an area fill to the linecharts with subtle fill-color
- restyles the barchart tile to have a two-column layout
- repositions the controls for the linechart to be centered under the chart
- sets the max value of the line-chart to always show the signaalwaarde value (when available)

## Motivation

design tweaks

### DIFF
--- a/src/scss/layout.scss
+++ b/src/scss/layout.scss
@@ -68,9 +68,9 @@
   margin-top: 2rem;
   box-shadow: $box-shadow;
 
-  h3,
-  p {
+  .article-text {
     max-width: 30rem;
+    // color: blue;
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/scss/layout.scss
```diff
@@ -67,6 +67,11 @@
   margin: 0;
   margin-top: 2rem;
   box-shadow: $box-shadow;
+
+  h3,
```
VWSCoronaDashboard4, line 71: It's handier to put a maxwidth on a parent element; this is a pretty generic selector in the end that might have some unintended side effects.

VWSCoronaDashboard1 (approved):

PR #270, opened 2020-09-10
Visual tweaks

VWSCoronaDashboard3:
## Summary

- On mobile, make the article cards full-width
- On mobile, fix styling of radio group component
- Disable “Rioolwatermeting” when no data is available & disabled styles
- Change size of link arrows in metric links
- Add search icon to `comboBox`
- Create utils to get the correct gemeente/regio name based on a code.
- Update chloropleth styles


src/scss/layout.scss
```diff
@@ -76,10 +81,11 @@
 
 .safety-region-aside,
 .municipality-aside {
-  min-height: 30em;
+  min-height: 35em;
 
   [data-reach-combobox] {
-    padding: 1em;
+    margin: 2em 1em;
```
VWSCoronaDashboard4, line 87: I've deliberately put a padding on this one in the IE fixes, since IE renders it oddly with just margin on smaller screens. Perhaps good to double check?

src/scss/layout.scss
```diff
@@ -246,6 +251,21 @@ a.metric-link {
   }
 }
 
+.metric-not-available {
+  padding: 1rem;
+  color: $text-color-grey;
+
+  svg {
+    fill: currentColor;
```
VWSCoronaDashboard4, line 259: Or just `fill: $text-color-grey`

src/utils/getLocalTitleForCode.ts
```diff
@@ -0,0 +1,26 @@
+import replaceVariablesInText from './replaceVariablesInText';
+
+import regios from 'data';
+import municipalities from 'data/gemeente_veiligheidsregio.json';
+
+export { getLocalTitleForRegion, getLocalTitleForMuncipality };
+
+function getLocalTitleForRegion(title: string, code: string): string {
```
VWSCoronaDashboard4, line 8: The name of the safety region / municipality is now passed as the string `name` in props on all safety region and municipality pages, you could pass that instead of the code?

VWSCoronaDashboard4 (commented):
Looking good!

Added some comments :-)

src/scss/layout.scss
VWSCoronaDashboard3: I checked in IE, it looks pixel perfect. Quite surprised, honestly.

src/utils/getLocalTitleForCode.ts
VWSCoronaDashboard3: I can’t use the `name`because it’s just the code. The idea is that eventually the backend will send the actual name and then we can get rid of this util. Until then, passing `code` is intended.

Screenshot 2020-09-10 at 12 33 49

VWSCoronaDashboard1 (approved):

PR #272, opened 2020-09-10
chore: charts visual tweaks (fix for linechart max value)

VWSCoronaDashboard5:


src/scss/layout.scss
```diff
@@ -67,6 +67,11 @@
   margin: 0;
   margin-top: 2rem;
   box-shadow: $box-shadow;
+
+  .article-text {
+    max-width: 30rem;
+    // color: blue;
```
VWSCoronaDashboard4, line 73: Remove debug code

src/components/lineChart/index.tsx
```diff
@@ -76,7 +76,7 @@ function getOptions(
     },
     yAxis: {
       min: 0,
-      max: values.length > 0 ? null : 1,
+      max: values.length > 0 ? null : Math.max(signaalwaarde as number, 1),
```
VWSCoronaDashboard4, line 79: signaalwaarde can be undefined, I guess that's also why you get a type error.

I think it works well if you just use `signaalwaarde || 1`

Issue #307, opened 2020-09-15
Sample data not included, application doesn't load without JSONs in place


marten:
### Describe the bug, issue or concern

I wanted to investigate reports of incorrect application of gradients on the barScale component. Unfortunately, I can't seem to get the frontend to load without putting any data in place in `public/json`, but the README doesn't specify where to get this data from.

Would it be possible to include a set of sample data with the repository for development? Or otherwise describe how to get a copy of the current data from the production dashboard (ideally a yarn command).

### To Reproduce

1. Clone a fresh copy of the repository
2. Run `yarn` as described in the README
3. Run `yarn dev` as described in the README
4. Open http://localhost:3000 and see error `Error: ENOENT: no such file or directory, open '/Users/marten/src/nl-covid19-data-dashboard/public/json/NL.json'`

### Expected behavior

Dashboard opens with static sample data. It is acceptable if there is a manual step to put the sample data in place but then that needs to be described in the README.

### Screenshots

![image](https://user-images.githubusercontent.com/277/93191030-58ed3980-f744-11ea-83f9-4fa7ebaaa1da.png)

**Desktop (please complete the following information):**
 - OS: macOS Catalina
 - Browser Firefox Developer Edition
 - Version 81

### Additional context

n/a

### Governance
- [x] I've read the contributing document https://github.com/minvws/.github/blob/master/CONTRIBUTING.md
- [x] I've read and understand the Code of Conduct https://github.com/minvws/.github/blob/master/CODE_OF_CONDUCT.md
- [x] I understand that any contributions or suggestions I made may make it into the actual code. I've read the License https://github.com/minvws/nl-covid19-notification-app-coordination/blob/master/LICENSE.txt and the contributor license agreement https://github.com/minvws/nl-covid19-notification-app-coordination/blob/master/CLA.md

jwz104:
I was just creating the same issue! The readme specifies something about that the data needs to be put into `/public/json`, but I'm not able to get the right json data. The calculation can be found in the [nl-covid19-data-backend-processing](https://github.com/minvws/nl-covid19-data-backend-processing) repo. But I'm not able to generate the actual data.

marten:
I've had some luck downloading the files from the production site btw, but you'll need to pull files per municipality and region.

```
mkdir public/json
curl -o public/json/NL.json https://coronadashboard.rijksoverheid.nl/json/NL.json
[...]

PR #333, opened 2020-09-16
feat: Load data from public website if files are missing in local file system

VWSCoronaDashboard:
## Summary

We want to make it easier for people to work with the latest data sets.
https://github.com/minvws/nl-covid19-data-dashboard/issues/307

Todos

- [ ] Load RANGES.json with getStaticProps
- [ ] Load MUNICIPALITIES.json with getStaticProps
- [ ] Remove uswSWR from the application

## Motivation

It's difficult for new people to contribute to this project, e.g. make a PR to solve an issue. By introducing a fallback to production data we make it easier for people to contribute to this codebase.

By placing JSON files in `/public/json/` you can override the fallback mechanism, allowing maintainers to work on new data structures.

## Detailed design

Some if/else statements like this should cover most of it

```tsx
export default function getNlData(): () => IProps {
  return async function () {
    let nlData: National;

    if (fs.existsSync(filePath)) {
      nlData = JSON.parse(fileContents);
    } else {
      const res = await fetch(
        'https://coronadashboard.rijksoverheid.nl/json/NL.json'
      );
      nlData = await res.json();
    }

    const lastGenerated = nlData.last_generated;

    return {
      props: {
        data: nlData,
        lastGenerated,
      },
    };
  };
}
```

## Drawbacks

All these switch statements will introduce maintenance complexity.

## Alternatives

- Publishing a zip file with latest data and serve it on the website (feasible, easy)
- Manually downloading all files from the server with a yarn download script (good way to get banned for a DoS attack)

## Unresolved questions
1.  Do we want to introduce an environment variable to enable this behavior or do we enable this on development by default?

~2. SWR fetches are blocked when we make a request to coronadashboard.rijksoverheid.nl because of the NO-CORS policy. Is there a smart way to prevent that?~

Let's just remove useSWR and go 🚀 fully static 🚀.


VWSCoronaDashboard:
Closed because publishing a zip file will be much easier and introduce less maintenance.

PR #338, opened 2020-09-16
Add GGD measurements

VWSCoronaDashboard4:
## Summary & Motivation

Adds GGD measurements to the UI.

## Detailed design

Most things are similar to the other tiles.

Is there any feedback on the way I render the KPI sentence around the percentage?

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added


src/pages/landelijk/positief-geteste-mensen.tsx
```diff
@@ -199,6 +202,62 @@ const PostivelyTestedPeople: FCWithLayout<INationalData> = (props) => {
...
+                  dangerouslySetInnerHTML={{
+                    __html: replaceVariablesInText(
+                      text.percentage_ggd_positief_getest_getest_week_uitleg,
+                      {
+                        percentage: `<span class="text-light-blue inline-kpi">${formatDecimal(
+                          percentageDataGGD?.percentage_infected_ggd
+                        )}%</span>`,
+                        totaal: `<span class="text-dark-blue inline-kpi">${formatDecimal(
+                          percentageDataGGD?.total_tested_ggd
+                        )}</span>`,
```
VWSCoronaDashboard1, line 250: Indeed this looks very verbose in order to just replace a few values. I wouldn't know of a cleaner way to do this though, but this is something we might need to discuss and see if we can come up with a nicer solution. For now, it works, so let's leave it.

VWSCoronaDashboard1 (approved):

src/locale/nl.json
```diff
@@ -137,7 +137,19 @@
...
+    "percentage_ggd_bron": {
+      "href": "https://coronadashboard.rijksoverheid.nl/#ggd",
+      "text": "GGD"
+    },
+    "percentage_ggd_toelichting": "Dit getal zegt iets over het totaal aantal personen dat via de GGD getest is, en hoeveel daarvan een positieve uitslag hadden. Let op dat de meeste testen via de GGD lopen, maar dit niet voor alle testen geldt. Daarom wijkt het absolute aantal positief geteste personen af van het aantal positief geteste personen door de GGD.",
+    "percentage_ggd_totaal_getest_week_titel": "Aantal geteste personen via GGD",
+    "percentage_ggd_totaal_getest_week_uitleg": "Aantal personen die in één week via de GGD getest zijn.",
+    "percentage_ggd_positief_getest_week_titel": "Positief getest personen via GGD",
+    "percentage_ggd_positief_getest_week_uitleg": "Aantal en percentage positief geteste personen die in één week via de GGD getest zijn.",
+    "percentage_ggd_positief_getest_getest_week_uitleg": "dit is {{percentage}} van {{totaal}} geteste personen"
```
VWSCoronaDashboard, line 152: nit: Might be easier to put this all in one object, to keep them contained/namespaced.

src/pages/landelijk/positief-geteste-mensen.tsx
```diff
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useState, Fragment } from 'react';
```
VWSCoronaDashboard, line 1: No need for Fragment, you can use `<>`

VWSCoronaDashboard4: No, that's not possible for the way it is used.

VWSCoronaDashboard1 (approved):

src/pages/landelijk/positief-geteste-mensen.tsx
```diff
@@ -25,10 +25,14 @@ import positiveTestedPeopleTooltip from 'components/chloropleth/tooltips/municip
 import positiveTestedPeopleTooltipRegion from 'components/chloropleth/tooltips/region/positiveTestedPeopleTooltip';
 import MunicipalityLegenda from 'components/chloropleth/legenda/MunicipalityLegenda';
 import SafetyRegionLegenda from 'components/chloropleth/legenda/SafetyRegionLegenda';
+import replaceKpisInText from 'utils/replaceKpisInText';
 
 const text: typeof siteText.positief_geteste_personen =
   siteText.positief_geteste_personen;
 
+const percentageGgdText: typeof siteText.positief_geteste_personen_ggd =
```
VWSCoronaDashboard8, line 33: This looks weird. Using typeof on a value to extract a type, and then assigning the same value to the type. Typescript should infer the same time if you leave out the type definition @VWSCoronaDashboard4 

Looks like it happened before in this file, so maybe that inspired you...

src/pages/landelijk/positief-geteste-mensen.tsx
```diff
@@ -199,6 +205,70 @@ const PostivelyTestedPeople: FCWithLayout<INationalData> = (props) => {
           </>
         )}
       </article>
+
+      {percentageDataGGD && (
+        <>
+          <ContentHeader
+            category={'\u00A0'}
+            title={percentageGgdText.titel}
```
VWSCoronaDashboard8, line 213: There's a lot of `percentageGgdText.` type of prefixes in this file. I would destructure it first.

src/utils/replaceKpisInText.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * - If a specific variable is NOT given, it will replace it with an empty string.
+ * - If no translation is given, an empty string will be returned.
+ *
+ * @param translation - Translation string with curly brackets for variables.
+ * @param kpis - An object with keys representing kpi object containing a value and class name available for replacement.
+ */
+
+const replaceKpisInText = (
+  translation?: string | undefined | null,
```
VWSCoronaDashboard8, line 26: Why undefined *and* null? Also undefined is already implied by the `?`

src/utils/replaceKpisInText.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Where everything between curly brackets is accepted as a placeholder
+ * for a variable. The second argument is an object with keys representing
+ * the string between curly brackets. In this case it would be:
+ * { type: 'curly' }
+ *
+ * - If a specific variable is NOT given, it will replace it with an empty string.
+ * - If no translation is given, an empty string will be returned.
+ *
+ * @param translation - Translation string with curly brackets for variables.
+ * @param kpis - An object with keys representing kpi object containing a value and class name available for replacement.
```
VWSCoronaDashboard8, line 22: These king of annotations are a little redundant. Typescript will tell you the shape of the kpis. So now you have to remember to update the comments whenever the code changes.

src/utils/replaceKpisInText.ts
VWSCoronaDashboard8: In what situation would you want to call "replace in text" without passing an actual text? Similarly why would you call this to replace kpis in a string if you do not have any kpis? In other words, are you sure these function arguments need to be optional?

src/utils/replaceKpisInText.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+const replaceKpisInText = (
+  translation?: string | undefined | null,
+  kpis?: Kpi[]
+): string => {
+  const variables: { [key: string]: string | undefined } = {};
```
VWSCoronaDashboard8, line 29: I don't see any code below that would make the result undefined. Looks like all iterations will produce a string.

VWSCoronaDashboard8 (changes requested):

src/pages/landelijk/positief-geteste-mensen.tsx
VWSCoronaDashboard1: This is legacy code, the typing is no longer necessary since the entire locale struct is typed in locale/index.ts

src/pages/landelijk/positief-geteste-mensen.tsx
VWSCoronaDashboard4: I've adjusted this variable name, but destructing it first causes other troubles (and it's not consistent with our use of `text.` in other pages)

VWSCoronaDashboard:
In this situation I would prefer a different message

image

would be better as:

**"Geen van de GGD testen was positief"**

VWSCoronaDashboard (commented):
Please add tests for your new utility method.

Besides that, I think we should start thinking about different i18n and l10n solutions to make maintenance of this project a bit easier. I know bundle size is a concern, but still...

src/pages/landelijk/positief-geteste-mensen.tsx
```diff
@@ -173,6 +196,69 @@ const PostivelyTestedPeople: FCWithLayout<INationalData> = (props) => {
...
+                          className: 'text-light-blue',
+                        },
+                        {
+                          name: 'denominator',
+                          value: formatNumber(ggdData?.total_tested_ggd),
+                          className: 'text-dark-blue',
+                        },
+                      ]
+                    ),
+                  }}
```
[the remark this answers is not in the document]
VWSCoronaDashboard4: Not sure if any of these packages would solve / improve any of the attached code. What specific part do you want to see to be simplified?

src/pages/landelijk/positief-geteste-mensen.tsx
```diff
@@ -173,6 +196,69 @@ const PostivelyTestedPeople: FCWithLayout<INationalData> = (props) => {
...
         )}
       </article>
+
+      {ggdData && (
+        <>
+          <ContentHeader
+            category={'\u00A0'}
+            title={ggdText.titel}
+            id="ggd"
+            Icon={Fragment}
```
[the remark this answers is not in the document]
VWSCoronaDashboard4: Now that I look at this again... I'll work on simplifying this! It's working around the problem that without the icon it messes up the spacing of the layout, which is a silly excuse.

VWSCoronaDashboard4:
@VWSCoronaDashboard The -% would be a data issue, I've added an extra safeguard to prevent this from getting displayed.

src/pages/landelijk/positief-geteste-mensen.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard: It would make pluralization easier, and it would remove the need for our custom replacement methods.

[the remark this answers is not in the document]
VWSCoronaDashboard: But we can take that off the table and put it on the tech debt backlog!

src/components/layout/Content.tsx
```diff
@@ -35,4 +42,7 @@ interface IContentHeaderProps {
       text: string;
     };
   };
+  category?: string;
+  Icon?: React.ComponentType;
```
VWSCoronaDashboard, line 46: 👍

VWSCoronaDashboard (approved):

PR #345, opened 2020-09-17
perf: Increase tooltip performance

VWSCoronaDashboard:
## Summary

This PR ensures tooltips don't cause wasted re-renders anymore. This makes the chart feel much more responsive on IE11.
## Motivation

The tooltips on the Choropleths cause very expensive re-renders of the whole component. We want to see if we can tell React to only update the tooltips without expensive reconciliation procedures.

## Detailed design

1. Don't trust `@vx` because we don't know what happens under the hood
2. Don't use `setState()` because that triggers a complete re-render
3. Use `zustand` to bypass the normal React state system 

## TODO

- [x] Check in IE11
- [x] Test performance (70ms renders to 0.3ms 🎉 )
- [x] Fix the layout/styling


VWSCoronaDashboard1:
This is great stuff

PR #348, opened 2020-09-17
Improve Line Chart Layout and Signaalwaarde

VWSCoronaDashboard8:
## Summary

* Add "Signaalwaarde" label to the plotline
* Encapsulate chart title and description in chart component
* Align chart time controls with the title on wide screens
* Adapt all existing line charts to use the new title and description

### Governance

- [ ] Documentation is added


src/components/chartTimeControls/chartTimeControls.module.scss
```diff
@@ -1,15 +1,18 @@
 @import 'scss/variables.scss';
 
-.chart-radio-group {
+.root {
+  /*
+    I think these are not right here. The parent component should be responsible
+    for the positioning.
+  */
```
VWSCoronaDashboard1, line 7: Comments like these tend to become noise IMHO and will just get lost. Either turn it into a TODO, or better, add a ticket in Trello, that way we can be fairly certain that action will be taken.

VWSCoronaDashboard8: With "turn into TODO" you mean just prefixing it with `@TODO` in the comment? That's what I'm used to doing in code. It's merely an indication for anyone coming back to change / update this component in the future that it might be something worth fixing. I'm happy to prefix it with @TODO.

I wouldn't put this in Trello because we cloud easily flood the board with small tasks like and loose focus.

PR #354, opened 2020-09-18
Fix: Escalation map mobile functionality

VWSCoronaDashboard1:
## Summary

- The escalation map is now actually visible on mobile
- Clicking on the map on mobile will show the tooltip, clicking on the tooltip will trigger the route change
- Tooltip stays within viewport now
- Updated escalation colors to final versions

## Motivation

Feature request

## Detailed design

- The tooltip positioning all happens in the useEffect hook within the tooltip component.
- The escalation tool tip render has been augmented with an onclick handler that performs the route change

## Drawbacks

The switch between the click functionality on the map is now done based on screen size detection. Preferable this becomes some sort of feature detection where we can determine whether the current screen has touch capabilities or not.

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added

### DIFF
--- a/src/components/chloropleth/Chloropleth.tsx
+++ b/src/components/chloropleth/Chloropleth.tsx
@@ -10,9 +10,15 @@ import styles from './chloropleth.module.scss';
 import { localPoint } from '@vx/event';
 
 import Tooltip from './tooltips/tooltip';
+import useMediaQuery from 'utils/useMediaQuery';
 
 const tooltipStore = create((set) => ({
   tooltip: null,
+  updateTooltip: (tooltip: any) => {
+    set({
+      tooltip,
+    });
+  },
   showTooltip: (hoveredElement: any) => {
     return set({
       tooltip: {
@@ -100,6 +106,7 @@ export default function Chloropleth<T>(props: TProps<T>) {
 
   const clipPathId = useRef(`_${Math.random().toString(36).substring(2, 15)}`);
   const timout = useRef<any>(-1);
+  const isLargeScreen = useMediaQuery('(min-width: 1000px)');
 
   const {
     width = 0,
@@ -125,7 +132,7 @@ export default function Chloropleth<T>(props: TProps<T>) {
         className={styles.svgMap}
         onMouseOver={svgMouseOver(timout, showTooltip)}
         onMouseOut={svgMouseOut(timout, hideTooltip)}
-        onClick={svgClick(onPathClick)}
+        onClick={svgClick(onPathClick, showTooltip, isLargeScreen)}
       >
         <clipPath id={clipPathId.current}>
           <rect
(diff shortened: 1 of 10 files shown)


src/components/chloropleth/Chloropleth.tsx
```diff
@@ -10,9 +10,15 @@ import styles from './chloropleth.module.scss';
 import { localPoint } from '@vx/event';
 
 import Tooltip from './tooltips/tooltip';
+import useMediaQuery from 'utils/useMediaQuery';
 
 const tooltipStore = create((set) => ({
   tooltip: null,
+  updateTooltip: (tooltip: any) => {
```
VWSCoronaDashboard, line 17: We can/should improve upon the :any type

src/components/chloropleth/Chloropleth.tsx
```diff
@@ -125,7 +132,7 @@ export default function Chloropleth<T>(props: TProps<T>) {
         className={styles.svgMap}
         onMouseOver={svgMouseOver(timout, showTooltip)}
         onMouseOut={svgMouseOut(timout, hideTooltip)}
-        onClick={svgClick(onPathClick)}
+        onClick={svgClick(onPathClick, showTooltip, isLargeScreen)}
```
VWSCoronaDashboard, line 135: Let's tackle and verify this logic in another ticket.

src/components/chloropleth/SafetyRegionChloropleth.tsx
```diff
@@ -78,15 +78,15 @@ const escalationThresholds: RegionalThresholds = {
...
-      color: '#FACB3E',
+      color: '#D95790',
       threshold: 2,
     },
     {
-      color: '#F5512D',
+      color: '#A11050',
       threshold: 3,
     },
   ],
```
VWSCoronaDashboard, line 92: Fun fact, this should be a categorical (aka ordinal) scale. Not a treshold scale.
https://observablehq.com/@d3/d3-scaleordinal


We know for a fact that a number will always be 1, 2, or 3. And those numbers correspond to a fixed color.

src/pages/veiligheidsregio/index.module.scss
```diff
@@ -29,6 +30,7 @@
     padding-right: 0.5em;
     margin: 0;
     font-size: 1.2em;
+    cursor: pointer;
```
VWSCoronaDashboard, line 33: Remove this line

src/pages/veiligheidsregio/index.module.scss
```diff
@@ -46,6 +48,7 @@
 
   .escalationInfo {
```
VWSCoronaDashboard, line 49: Since this is nested under .escalationTooltip everything will have a tooltip. No need to repeat cursor: pointer.

src/pages/veiligheidsregio/index.module.scss
```diff
@@ -46,6 +48,7 @@
...
     padding: 1em;
+    cursor: pointer;
```
VWSCoronaDashboard, line 51: Remove this too

src/pages/veiligheidsregio/index.tsx
```diff
@@ -59,6 +56,7 @@ const escalationTooltipContent = (context: any): ReactNode => {
 // lots of unnecessary null checks on those pages.
 const SafetyRegion: FCWithLayout<any> = () => {
   const router = useRouter();
+  const isLargeScreen = useMediaQuery('(min-width: 1000px)');
```
VWSCoronaDashboard, line 59: This 1000 value comes back quite a few times. Can we make it a single source of truth somewhere?

VWSCoronaDashboard (commented):
Most important is the feedback on the scale, the tresholdScale is the "wrong" scale for the escalation map. "Wrong" as in it will work as desired, but the way it's configured is counter-intutive.

src/scss/layout.scss
```diff
@@ -195,6 +194,12 @@
       display: none;
     }
   }
+
+  .small-screen-aside-tendency {
```
VWSCoronaDashboard4, line 198: it's not handy to use this class in styling other than the showing / hiding of the asidde/content. (For a different PR: There could be a lit of comment added here about that.)

Maybe just use `.safety-region-aside` inside this mediaquery?

src/components/chloropleth/Chloropleth.tsx
VWSCoronaDashboard1: Yes, let's find out who wrote that code ;)

VWSCoronaDashboard1: Fixed now, the entire tooltip system has been typed

src/scss/layout.scss
VWSCoronaDashboard1: Please add this as a Trello ticket, otherwise, it'll get lost

PR #361, opened 2020-09-18
fix: Apply markdown parsing to toelichting key on safety region page

VWSCoronaDashboard:
## Summary

We need to parse markdown formatted text with the `MDToHTMLString` function in getStaticProps. There is no markdown formatter in the React codebase itself, because we want to keep that out of the bundle.

## Detailed design

Make sure you look at these lines. I removed the getLastGenerated method and replaced it with this more specific one.

```tsx
  const text = require('../../locale/index').default;

  const serializedContent = MDToHTMLString(
    text.veiligheidsregio_index.selecteer_toelichting
  );

  text.veiligheidsregio_index.selecteer_toelichting = serializedContent;

  return { props: { text, lastGenerated } };
}
```

## Unresolved questions

I might've skipped on some types. But... that's a concern for the future.


src/pages/veiligheidsregio/index.tsx
```diff
@@ -90,12 +102,35 @@ const SafetyRegion: FCWithLayout<any> = () => {
             tooltipContent={escalationTooltip(router)}
           />
         </div>
+        {!isLargeScreen && <EscalationMapLegenda text={text} />}
```
VWSCoronaDashboard1, line 105: Remove this, the legenda isn't conditonally rendered anymore

src/pages/veiligheidsregio/index.tsx
```diff
@@ -76,10 +85,13 @@ const SafetyRegion: FCWithLayout<any> = () => {
...
-            {text.veiligheidsregio_index.selecteer_toelichting}
-          </p>
-          <EscalationMapLegenda />
+          <div
+            className="text-max-width"
+            dangerouslySetInnerHTML={{
+              __html: text.veiligheidsregio_index.selecteer_toelichting,
+            }}
+          />
+          {isLargeScreen && <EscalationMapLegenda text={text} />}
```
VWSCoronaDashboard1, line 94:
```suggestion
          <EscalationMapLegenda text={text} />
```
Don't render legend conditionally. I think this is a merge error probably.

VWSCoronaDashboard1 (changes requested):

src/pages/veiligheidsregio/index.tsx
VWSCoronaDashboard: it is

PR #391, opened 2020-09-21
add week-unix-values to tooltip in linecharts that use weekly data

VWSCoronaDashboard5:
## Summary

The data used in the line-charts of Rioolwater and Huisartsen is the average value over time (one week), the fact that this value is based on a week-basis should be made more clear. This ticket changes the tooltips shown for those graphs to show the `week_start_unix` and the `week_end_unix` in the tooltip for the concerned value.


src/components/lineChart/nationalPractitionerLineChart.tsx
```diff
@@ -0,0 +1,186 @@
...
+} from '~/components/chartTimeControls';
+import { getFilteredValues } from '~/components/chartTimeControls/chartTimeControlUtils';
+
+import styles from './lineChart.module.scss';
+import { formatNumber } from '~/utils/formatNumber';
+import { formatDate } from '~/utils/formatDate';
+
+interface Value {
+  date: number;
+  value: number | undefined | null;
```
VWSCoronaDashboard8, line 17: Are both undefined and null needed?

src/components/lineChart/nationalPractitionerLineChart.tsx
```diff
@@ -0,0 +1,186 @@
...
+};
+
+type NationalPractitionerLineChartProps = {
+  values: Value[];
+  text: string;
+  timeframeOptions?: TimeframeOption[];
+};
+
+function getOptions(values: Value[]): Highcharts.Options {
+  const multipleAverageValues = values.length > 1;
```
VWSCoronaDashboard8, line 33: Prefix booleans to make clear it's a boolean. In this case `hasMultipleValues` or something. Otherwise the name is very vague and it even wrongly implies that it holds multiple values.

src/components/lineChart/nationalPractitionerLineChart.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+  const weeklyMeasurements: Week[] = values.map((value) => value.week);
+
+  const series: SeriesAreaOptions[] = [
+    {
+      type: 'area',
+      data: values.map((value) => [value.date, value.value]),
```
VWSCoronaDashboard8, line 40: Nitpick; Personally I like using `x` in these type of short inline functions. It saves space and in this case avoids value.value. So `values.map(x => [x.date, x.value])`

src/components/lineChart/nationalPractitionerLineChart.tsx
```diff
@@ -0,0 +1,186 @@
...
+      fillColor: 'rgba(51, 145, 204, 0.2)',
+      allowPointSelect: false,
+      marker: {
+        symbol: 'circle',
+        enabled: !multipleAverageValues,
+      },
+      events: {
+        legendItemClick: function () {
+          return false;
+        },
```
VWSCoronaDashboard8, line 53:
```suggestion
        legendItemClick: () => false,
```

src/components/lineChart/nationalPractitionerLineChart.tsx
```diff
@@ -0,0 +1,186 @@
...
+        align: 'right',
+        // types say `rotation` needs to be a number,
+        // but that doesn’t work.
+        rotation: '0' as any,
+        formatter: function (): string {
+          if (this.isFirst || this.isLast) {
+            return formatDate(this.value * 1000, 'axis');
+          }
+          return '';
+        },
```
VWSCoronaDashboard8, line 113: I don't think a return type definition should be required here. Also an arrow function would be nice for inline. But I guess you got this code from somewhere else.

 ```suggestion
        formatter: () =>
          this.isFirst || this.isLast
            ? formatDate(this.value * 1000, 'axis')
            : '',
   
```

File: src/components/lineChart/nationalPractitionerLineChart.tsx [hunk 6]
```diff
@@ -0,0 +1,186 @@
...
+      },
+    },
+    tooltip: {
+      backgroundColor: '#FFF',
+      borderColor: '#01689B',
+      borderRadius: 0,
+      formatter: function (): string {
```
VWSCoronaDashboard8, line 120: Similar as previous comment

src/components/lineChart/nationalPractitionerLineChart.tsx
```diff
@@ -0,0 +1,186 @@
...
+      gridLineColor: '#C4C4C4',
+      title: {
+        text: null,
+      },
+      labels: {
+        formatter: function (): string {
+          // @ts-ignore
+          return formatNumber(this.value);
+        },
+      },
```
VWSCoronaDashboard8, line 142:
```suggestion
        formatter: () => formatNumber(this.value),
```

src/components/lineChart/nationalPractitionerLineChart.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+export function NationalPractitionerLineChart({
+  values,
+  text,
+  timeframeOptions,
+}: NationalPractitionerLineChartProps) {
+  const [timeframe, setTimeframe] = useState<TimeframeOption>('5weeks');
+
+  const chartOptions = useMemo(() => {
+    const filteredValues = getFilteredValues<Value>(
```
VWSCoronaDashboard8, line 161: I don't think `<Value>` is required if the type is inferred from the `values` argument

src/components/lineChart/regionalSewerWaterLineChart.tsx
```diff
@@ -109,7 +121,11 @@ function getOptions(
         if (this.series.name !== text.average_label_text) {
           return false;
         }
-        return `${formatDate(this.x * 1000)}: ${formatNumber(this.y)}`;
+        const { start, end } = weeklyMeasurements[this.point.index];
```
VWSCoronaDashboard8, line 124: Nitpick; I think it would be good to check the return value before using the values.

VWSCoronaDashboard4, line 124: I'd prefer a method over a calculated array. Eg have `function getWeekStartEndByIndx(index: number): Week[] { ... }`

src/utils/assert.ts
```diff
@@ -0,0 +1,8 @@
+export default function assert(
```
VWSCoronaDashboard8, line 1: This should probably be a named export now.

src/components/lineChart/nationalPractitionerLineChart.tsx
VWSCoronaDashboard8: Oops this won't work like that because you don't get the same "this" in the fat arrow context. So keep it a normal function declaration, but you can still do this:

```
formatter: function () {
         return this.isFirst || this.isLast
           ? formatDate(this.value * 1000, 'axis')
           : '' },
```

VWSCoronaDashboard5: I applied a refactor in the other places where this syntax was used

src/components/lineChart/lineChartWithWeekTooltip.tsx
```diff
@@ -0,0 +1,179 @@
...
+    tooltip: {
+      backgroundColor: '#FFF',
+      borderColor: '#01689B',
+      borderRadius: 0,
+      formatter: function () {
+        const { start, end } = getWeekStartEndByIndex(
+          values.map((x) => x.week),
+          this.point.index
+        );
+        assert(typeof start === 'number' && typeof end === 'number');
```
VWSCoronaDashboard5, line 117: @VWSCoronaDashboard8 is this what you meant by checking the return value before using the values through an assert function?

src/components/lineChart/lineChartWithWeekTooltip.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+import styles from './lineChart.module.scss';
+import { formatNumber } from '~/utils/formatNumber';
+import { formatDate } from '~/utils/formatDate';
+import { getWeekStartEndByIndex, Week } from '~/utils/getWeekStartEndByIndex';
+import { assert } from '~/utils/assert';
+
+interface Value {
+  date: number;
+  value: number | undefined;
```
VWSCoronaDashboard8, line 19:
```suggestion
  value?: number;
```

src/components/lineChart/lineChartWithWeekTooltip.tsx
VWSCoronaDashboard8: I meant something like this:

```
const weekStartEnd = getWeekStartEndByIndex(
          values.map((x) => x.week),
          this.point.index
        );

assert(weekStartEnd, `Missing week start/end for index ${this.point.index}`)
```

... and then you can safely destructure.

But I am not familiar with the data so I'm not sure it out-of-range index is something we should worry about. I generally like this kind of "defensive programming" because it catches problems early and gives better error messages. But there's a balance of course, because you don't want to litter the code with all sorts of unnecessary assert statements.

src/utils/getWeekStartEndByIndex.ts
```diff
@@ -0,0 +1,8 @@
+export type Week = {
+  start: number;
+  end: number;
+};
+
+export function getWeekStartEndByIndex(weekSet: Week[], index: number): Week {
```
VWSCoronaDashboard8, line 6: I think this is a silly utility. All it does is get an item from an index. Why not just access the items directly?

You could make it generic at least so it accepts any type of array, but then I still don't see the point. Unless you want to build the assert in here, then it at least adds some value if this pattern is used often.

src/components/lineChart/lineChartWithWeekTooltip.tsx
```diff
@@ -0,0 +1,179 @@
...
+  week: Week;
+}
+
+type LineChartProps = {
+  values: Value[];
+  title: string;
```
VWSCoronaDashboard8, line 25: The other line chart has a description next to a title. Did you leave it out on purpose here?

src/components/lineChart/lineChartWithWeekTooltip.tsx
```diff
@@ -0,0 +1,179 @@
...
+      borderColor: '#01689B',
+      borderRadius: 0,
+      formatter: function () {
+        const { start, end } = getWeekStartEndByIndex(
+          values.map((x) => x.week),
+          this.point.index
+        );
+        assert(typeof start === 'number' && typeof end === 'number');
+
+        return `<strong>${formatDate(start * 1000, 'short')} - ${formatDate(
```
VWSCoronaDashboard8, line 119: I see this formatDate(x * 1000, ...) quite often. Would be a good candidate for a utility function. Maybe check all uses of formatDate, because I would expect the conversion of unix epoch to seconds to be part of the format function if all we use is epoch.

Might be better handled in a different PR if it is all over the codebase.

src/components/lineChart/lineChartWithWeekTooltip.tsx
```diff
@@ -0,0 +1,179 @@
...
+      minRange: 0.1,
+      allowDecimals: false,
+      lineColor: '#C4C4C4',
+      gridLineColor: '#C4C4C4',
+      title: {
+        text: null,
+      },
+      labels: {
+        formatter: function (): string {
+          // @ts-ignore
```
VWSCoronaDashboard8, line 136: This can be simplified like the other instance.

src/components/lineChart/lineChartWithWeekTooltip.tsx
```diff
@@ -0,0 +1,179 @@
...
+    title: {
+      text: undefined,
+    },
+    series,
+  };
+
+  return options;
+}
+
+export function LineChart({ values, title, timeframeOptions }: LineChartProps) {
```
VWSCoronaDashboard8, line 150: I think it's best to export a more specific name `LineChartWithWeekTooltip`, because it will clash with the other LineChart if we start using barrel files. Also, I don't see an advantage in having them both share the same name.

src/components/lineChart/lineChartWithWeekTooltip.tsx
```diff
@@ -0,0 +1,179 @@
...
+      (value: Value) => value.date * 1000
+    );
+    return getOptions(filteredValues);
+  }, [values, timeframe]);
+
+  return (
+    <section className={styles.root}>
+      <header className={styles.header}>
+        <div className={styles.titleAndDescription}>
+          {title && <h3>{title}</h3>}
```
VWSCoronaDashboard8, line 166: Same question here about the description. It was already optional in de other LineChart, so why is it not the same here?

src/components/lineChart/regionalSewerWaterLineChart.tsx
```diff
@@ -4,12 +4,15 @@ import HighchartsReact from 'highcharts-react-official';
...
 import { formatNumber } from '~/utils/formatNumber';
 import { formatDate } from '~/utils/formatDate';
+import { getWeekStartEndByIndex, Week } from '~/utils/getWeekStartEndByIndex';
 
 type TranslationStrings = Record<string, string>;
 
 interface Value {
   date: number;
-  value: number | undefined | null;
+  value: number | undefined;
```
VWSCoronaDashboard8, line 13:
```suggestion
  value?: number;
```

src/components/lineChart/regionalSewerWaterLineChart.tsx
```diff
@@ -21,24 +24,26 @@ function getOptions(
   averageValues: Value[],
   text: TranslationStrings
 ): Highcharts.Options {
-  const multipleAverageValues = averageValues.length > 1;
+  const hasMultipleValues = averageValues.length > 1;
+  const weekSet: Week[] = averageValues.map((value) => ({
```
VWSCoronaDashboard8, line 28: Nitpick; I name lists usually just using the plural. So `const weeks: Week[] =` If it's plural you kind of know that it's a list already. If its a map instead I'll make it explicit in the name.

A "set" is also a little confusing because the Javascript `Set()` would be a special array-like object with only unique values and different API from Array.

src/components/lineChart/regionalSewerWaterLineChart.tsx
```diff
@@ -109,7 +113,14 @@ function getOptions(
         if (this.series.name !== text.average_label_text) {
           return false;
         }
-        return `${formatDate(this.x * 1000)}: ${formatNumber(this.y)}`;
+        const { start, end } = getWeekStartEndByIndex(
+          weekSet,
+          this.point.index
+        );
+        return `<strong>${formatDate(start * 1000, 'short')} - ${formatDate(
+          end * 1000,
```
VWSCoronaDashboard8, line 121: This pattern again :)

src/utils/assert.ts
```diff
@@ -0,0 +1,5 @@
+export function assert(condition: any, msg?: string): asserts condition {
```
VWSCoronaDashboard8, line 1: I don't think `msg` should be optional. If you do an assertion you should specify a clear error message.

VWSCoronaDashboard8 (changes requested):
Don't let the amount of comments discourage you. I like picking out details 😉

src/components/lineChart/regionalSewerWaterLineChart.tsx
VWSCoronaDashboard8: @VWSCoronaDashboard4 I only see a point in creating a specific method if the method contains something like an assertion. Otherwise you might as well use something like Lodash `get()`, no? Or am I missing something?

src/components/lineChart/lineChartWithWeekTooltip.tsx
VWSCoronaDashboard8: I would advise you to move the assertion to the `getWeekStartEndByIndex` utility, and than it's not so silly anymore 😉

src/components/lineChart/lineChartWithWeekTooltip.tsx
VWSCoronaDashboard5: I could add it as an extra prop, but these charts have no descriptions according to the design (also no descriptions there at prod)

src/components/lineChart/lineChartWithWeekTooltip.tsx
VWSCoronaDashboard5: good point. I will make this refactor a separate PR!

PR #395, opened 2020-09-22
Fix chloropleth map resizing

VWSCoronaDashboard8:
## Summary

Fix and re-implement the useChartDimensions hook.

## Motivation

The map resizing wasn't working because the resize listener was not firing in the useChartDimensions hook. This problem has been solved by replacing the observer with a hook based approach, and in addition, the code could be greatly simplified.

### DIFF
--- a/src/components/chloropleth/hooks/useChartDimensions.tsx
+++ b/src/components/chloropleth/hooks/useChartDimensions.tsx
@@ -53,7 +53,7 @@ const useChartDimensions = (
    * The ref will be initialized from the outside by mutating the return value of this
    * hook. This is a bit funky.
    *
-   * @TODO pass ref as prop
+   * @REF https://trello.com/c/i25FG3jk/548-usechartdemensions-pass-ref-as-prop
    */
   const ref = useRef<HTMLElement | null>(null);
 
--- a/src/components/chloropleth/hooks/useChartDimensions.tsx
+++ b/src/components/chloropleth/hooks/useChartDimensions.tsx
@@ -1,12 +1,12 @@
-import { ResizeObserver } from '@juggle/resize-observer';
-import { useRef, useState, useEffect, MutableRefObject } from 'react';
+import { useRef, MutableRefObject } from 'react';
+import useResizeObserver from 'use-resize-observer';
 
 /**
- * This custom hook has been taken from the wonderful blog of
- * Amelia Wattenberger. https://wattenberger.com/blog/react-and-d3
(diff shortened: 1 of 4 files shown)


src/components/chloropleth/hooks/useChartDimensions.tsx
```diff
@@ -60,54 +47,33 @@ export type TCombinedChartDimensions = TChartDimensions & {
...
 
 const useChartDimensions = (
-  passedSettings?: TChartDimensions
+  chartDimensions: TChartDimensions = {}
 ): [MutableRefObject<HTMLElement | null>, TCombinedChartDimensions] => {
+  /**
+   * The ref will be initialized from the outside by mutating the return value of this
+   * hook. This is a bit funky.
+   *
+   * @TODO pass ref as prop
```
VWSCoronaDashboard, line 56: @Todo left here

src/components/chloropleth/hooks/useChartDimensions.tsx
```diff
@@ -60,54 +47,33 @@ export type TCombinedChartDimensions = TChartDimensions & {
...
 const useChartDimensions = (
-  passedSettings?: TChartDimensions
+  chartDimensions: TChartDimensions = {}
 ): [MutableRefObject<HTMLElement | null>, TCombinedChartDimensions] => {
+  /**
+   * The ref will be initialized from the outside by mutating the return value of this
+   * hook. This is a bit funky.
+   *
+   * @REF https://trello.com/c/i25FG3jk/548-usechartdemensions-pass-ref-as-prop
+   */
```
VWSCoronaDashboard8, line 57: I removed the @TODO and added a reference @VWSCoronaDashboard

VWSCoronaDashboard1, line 57: I'm not sure if returning a ref in a hook is 'funky'. I see it used like that in quite a few instances. If the ref is passed as a prop I'd say its dependency injection. When creating the ref inside the hook and returning it, it could be considered a type of callback and more of an inversion of control.

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard4 (approved):

src/components/chloropleth/hooks/useChartDimensions.tsx
VWSCoronaDashboard8: @VWSCoronaDashboard1 It reminded me more of returning a mutable state. But if this pattern is used more often maybe it's ok. I'll revisit it later.

PR #407, opened 2020-09-22
Barrel files

VWSCoronaDashboard8:
## Summary

A draft PR to verify the working of barrel files and their effect on bundle size / tree shaking. Webpack needs to know which files have side effects. For example, modules that inject a polyfill would be considered as having a side effect.

https://webpack.js.org/configuration/optimization/#optimizationsideeffects

In this PR a (partial) barrel file is created for ~/components and it is used for imports in the 404 page. So the size of the compiled 404 page is what you should look at here:

Before:

```
Page                                                           Size     First Load JS
┌ ● /                                                          2.57 kB         214 kB
├   └ css/b8f09c2cd1b1ac1dcb75.css                             2.82 kB
├   /_app                                                      0 B            63.4 kB
├ ○ /404                                                       452 B           101 kB
├ ● /500                                                       441 B           101 kB          
```

Without config:

```
Page                                                           Size     First Load JS
┌ ● /                                                          2.57 kB         214 kB
├   └ css/b8f09c2cd1b1ac1dcb75.css                             2.82 kB
├   /_app                                                      0 B            63.4 kB
├ ○ /404                                                       32.8 kB         248 kB
├   └ css/bef9b58462241312ebcb.css                             1.79 kB
├ ● /500                                                       441 B           101 kB
```

With config:

```
Page                                                           Size     First Load JS
┌ ● /                                                          2.57 kB         214 kB
├   └ css/a2852150e50fc414760f.css                             2.81 kB
├   /_app                                                      0 B            63.4 kB
├ ○ /404                                                       452 B           101 kB
├ ● /500                                                       441 B           101 kB
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,5 +106,8 @@
     "validate-json": "node schema/validator/validate-json.js",
     "validate-single": "node schema/validator/validate-single.js",
     "generate-typescript": "node schema/validator/generate-typescript.js"
-  }
+  },
+  "sideEffects": [
+    "./src/some-side-effectful-file.js"
+  ]
 }
--- a/src/components/index.ts
+++ b/src/components/index.ts
@@ -1,0 +1,15 @@
+export * from './areaChart';
+export * from './barChart';
+export * from './barScale';
+export * from './chartRegionControls';
+export * from './charts';
+export * from './chartTimeControls';
+// export * from './chloropleth';
+export * from './comboBox';
+// export * from './gemeente';
+// export * from './landelijk';
+export * from './layout';
+export * from './legenda';
+export * from './lineChart';
+
+export * from './maxWidth';
--- a/src/components/chloropleth/Chloropleth.tsx
+++ b/src/components/chloropleth/Chloropleth.tsx
@@ -164,14 +164,23 @@ export function Chloropleth<T>(props: TProps<T>) {
           transform={`translate(${marginLeft},${marginTop})`}
           clipPath={`url(#${clipPathId.current})`}
         >
-          <Mercator data={featureCollection.features} fitSize={sizeToFit}>
+          <Mercator
+            data={featureCollection.features}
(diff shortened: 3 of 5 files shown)


VWSCoronaDashboard4:
If we are testing bundle sizes, I'd be more interested in a more content loaded page. Do you have a comparison for that as well?

src/components/index.ts
```diff
@@ -0,0 +1,15 @@
+export * from './areaChart';
+export * from './barChart';
+export * from './barScale';
```
VWSCoronaDashboard8, line 3: I know these charts would normally not be included in the barrel file because we use dynamic exports from ./charts but this is just to add some weight to the size of the barrel.

VWSCoronaDashboard8:
@VWSCoronaDashboard4 I demonstrated what happens if tree shaking is not working. It clearly explodes the bundle size. With the right config there was no overhead. So I'm not sure what else you'd like me to demonstrate. How would it be any different for a page that has more components? But I'm happy to move this PR forward and start creating more barrel files and then doing another check later.

package.json
```diff
@@ -106,5 +106,8 @@
     "validate-json": "node schema/validator/validate-json.js",
     "validate-single": "node schema/validator/validate-single.js",
     "generate-typescript": "node schema/validator/generate-typescript.js"
-  }
+  },
+  "sideEffects": [
+    "./src/some-side-effectful-file.js"
+  ]
```
VWSCoronaDashboard, line 112: So we would need to indicate all files with side-effects? What is a side-effect in this context?

VWSCoronaDashboard:
Can you explain the configuration in more detail? How will it impact maintenance?

VWSCoronaDashboard1:
> Can you explain the configuration in more detail? How will it impact maintenance?

I concur, do the advantages of barrels really outweigh the maintenance?

PR #414, opened 2020-09-23
Define chart y-range based on values and signaalwaarde

VWSCoronaDashboard8:
Calculate the chart y-max value based on incoming values + the signaalwaarde with 10 margin. Whichever is highest defines the maximum.


src/components/lineChart/index.tsx
```diff
@@ -203,3 +206,18 @@ export default function LineChart({
...
   );
 }
+
+function calculateYMax(values: Value[], signaalwaarde = -Infinity) {
+  /**
+   * From all the defined values, extract the highest number so we know how to
+   * scale the y-axis
+   */
+  const maxValue = values
+    .filter(isDefined)
```
VWSCoronaDashboard4, line 216: Do we need an external package for this?

src/components/lineChart/index.tsx
```diff
@@ -26,7 +27,9 @@ interface LineChartProps {
   timeframeOptions?: TimeframeOption[];
 }
 
-function getChartOptions(values: Value[], signaalwaarde?: number | undefined) {
+function getChartOptions(values: Value[], signaalwaarde?: number) {
```
VWSCoronaDashboard4, line 30: :100:

src/components/lineChart/index.tsx
VWSCoronaDashboard8: It's very handy, because typescript doesn't yet infer types correctly after a filter declaration. This will be part of the language but until than this package is handy yes. I don't think it gives any real overhead in terms of bytes, but I'll check

VWSCoronaDashboard8: We could copy the code of that single function to our own utils.

VWSCoronaDashboard8: Also see [package readme](https://www.npmjs.com/package/ts-is-present)

VWSCoronaDashboard4: Ah, for type stuff it makes more sense!

VWSCoronaDashboard1 (approved):

PR #417, opened 2020-09-23
refactor dateFormat

VWSCoronaDashboard5:
## Summary

The `unix_time` values we receive resemble the amount of seconds since Epoch. Since JavaScript interprets epoch times through milliseconds, all the values passed through these functions had to be multiplied by `1000`. Everywhere through the codebase the `formatDate` function was written with `formatDate(value * 1000, style)`. We figured it would make sense to do this multiplication in the function, as well as refactoring the places where the formatter was used.

### Governance

- [x] Documentation is added


src/components/areaChart/index.tsx
```diff
@@ -186,7 +186,7 @@ export default function AreaChart(props: AreaChartProps) {
   const rangeData: TRange[] = useMemo(() => {
     return data
       .sort((a, b) => a.date - b.date)
-      .map((d) => [new Date(d.date * 1000), d.min, d.max]);
+      .map((d) => [new Date(d.date), d.min, d.max]);
```
VWSCoronaDashboard8, line 189: This will create the wrong date. You still need * 1000 here

src/components/layout/index.tsx
```diff
@@ -68,7 +68,7 @@ function Layout(props: WithChildren<LayoutProps & ILastGeneratedData>) {
   const dateTime = new Date(Number(lastGenerated) * 1000).toISOString();
 
   const dateOfInsertion = lastGenerated
-    ? formatDate(Number(lastGenerated) * 1000, 'long')
+    ? formatDate(Number(lastGenerated), 'long')
```
VWSCoronaDashboard8, line 71: Nitpick; lastGenerated is cast to number multiple times in this file it seems. I'd store it in a const `lastGeneratedEpoch` or something

src/utils/formatDate.ts
```diff
@@ -41,19 +36,26 @@ const Day = new Intl.DateTimeFormat(locale, {
 });
 
 export function formatDate(
-  value: number | Date,
+  epochOrDate: number,
```
VWSCoronaDashboard8, line 39: this can now be `epoch` or `unix_time` as it is called in other parts of the code it seems

src/utils/formatDate.ts
```diff
@@ -41,19 +36,26 @@ const Day = new Intl.DateTimeFormat(locale, {
...
   style?: 'long' | 'medium' | 'short' | 'relative' | 'iso' | 'axis'
 ): string {
-  if (style === 'iso') return new Date(value).toISOString(); // '2020-07-23T10:01:16.000Z'
-  if (style === 'long') return Long.format(value); // '23 juli 2020 om 12:01'
-  if (style === 'medium') return Medium.format(value); // '23 juli 2020'
+  /**
+   * JavaScript uses milliseconds since EPOCH, therefore the value
+   * formatted by the format() function needs to be multiplied by 1000
+   * to format to an accurate dateTime
+   */
```
VWSCoronaDashboard8, line 46: If you still want to use Date you could do:

```
const epoch = typeof epochOrDate === 'number' ? epochOrDate : epochOrDate.getMilliseconds()
```
and then from there work with epoch. All the functions below expect ms so you'd need to convert date first, and also make that clear in the name, otherwise it becomes confusing.

src/components/areaChart/index.tsx
VWSCoronaDashboard8: I think this is a bug in the original code. `new Date(d.date * 1000)` was wrong because date is already unix time, so multiplying it with 1000 is not correct. I think the chart only worked because the same mistake is made for the y and x axis.

src/utils/formatDate.ts
VWSCoronaDashboard5: 😅

src/components/areaChart/index.tsx
VWSCoronaDashboard8: Correction (for the record) this is not true. As we concluded that all date timestamps in the data are supplied in seconds not milliseconds, so the * 1000 conversion is necessary.

VWSCoronaDashboard1:
I know it can be annoying, but please do take some time to write proper PR comments. This PR has quite a bit of logic in it so a clear and concise filled-out PR template helps the reviewers to determine what it is exactly they ought to be reviewing.

PR #420, opened 2020-09-23
fix lastEdited in footer

VWSCoronaDashboard5:
## Summary

There's an inconsistency on how the time on which the dashboard was last edited is interpreted across different browsers, environments and languages (locales).
- it seems that there is a timezone malfunction (being two hours ahead on production when new data is generated)
- when you view the time on page visit, the right time is shown, but after browsing across several pages this time changes (hydration issue perhaps)
- the formatted string shows an English time formatting `September 16, 2020 at 2:00 PM` on the Dutch translation (locale)
- the formatted string leaves out the entire formatted timestring on the English pages

## Motivation

Why are we doing this? What use cases does it support? What is the expected
outcome?

Please focus on explaining the motivation so that if this PR is not accepted,
the motivation could be used to develop alternative solutions. In other words,
enumerate the constraints you are trying to solve without coupling them too
closely to the solution you have in mind.

## Detailed design

This is the bulk of the PR. Explain the design in enough detail for somebody
familiar with Opus to understand, and for somebody familiar with the
front-end to implement. This should get into specifics and corner-cases,
and include examples of how the feature is used. Any new terminology should be
defined here.

## Drawbacks

- implementation cost, both in term of code size and complexity
- maintenance cost, how complex is the solution and how prone to change is it?
- integration of this feature with other existing and planned features

There are tradeoffs to choosing any path. Attempt to identify them here.

## Alternatives

What other designs have been considered? What is the impact of not doing this?

## Unresolved questions

### Governance

- [ ] Documentation is added

### DIFF
--- a/src/components/layout/index.tsx
+++ b/src/components/layout/index.tsx
@@ -65,8 +65,7 @@ function Layout(props: WithChildren<LayoutProps & ILastGeneratedData>) {
   const locale = getLocale();
   const showSmallLogo = useMediaQuery('(max-width: 480px)', true);
 
-  const dateTime = new Date(Number(lastGenerated) * 1000).toISOString();
-
+  const dateTime = formatDate(Number(lastGenerated), 'iso');
   const dateOfInsertion = lastGenerated
     ? formatDate(Number(lastGenerated), 'long')
     : undefined;
--- a/src/components/layout/index.tsx
+++ b/src/components/layout/index.tsx
@@ -255,7 +255,7 @@ function Layout(props: WithChildren<LayoutProps & ILastGeneratedData>) {
                     __html: replaceVariablesInText(
                       text.laatst_bijgewerkt.message,
                       {
-                        dateOfInsertion: `<time dateTime=${dateTime}>${dateOfInsertion}</time>`,
+                        dateOfInsertion: `<time datetime=${dateTime}>${dateOfInsertion}</time>`,
                       }
                     ),
                   }}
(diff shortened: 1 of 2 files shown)


VWSCoronaDashboard4:
I've pushed some small improvements.

I think the 2 hour difference should be fixed by the timezine that I've added.

The EN time on NL pages is probably due to the fact that Node on the server does not understand the Int.DateTimeFormat for Dutch.

src/components/layout/index.tsx
```diff
@@ -65,11 +65,14 @@ function Layout(props: WithChildren<LayoutProps & ILastGeneratedData>) {
   const locale = getLocale();
   const showSmallLogo = useMediaQuery('(max-width: 480px)', true);
 
-  const dateTime = new Date(Number(lastGenerated) * 1000).toISOString();
+  const { dateTime, dateOfInsertion } = useMemo(() => {
```
0x80, line 68: I would be careful with useMemo. I general I would not use it is not clear it will have a real impact on performance because it can create overhead and introduce bugs too. This looks like it might be premature optimization. 

Also see https://kentcdodds.com/blog/usememo-and-usecallback

src/components/layout/index.tsx
```diff
@@ -65,8 +65,7 @@ function Layout(props: WithChildren<LayoutProps & ILastGeneratedData>) {
   const locale = getLocale();
   const showSmallLogo = useMediaQuery('(max-width: 480px)', true);
 
-  const dateTime = new Date(Number(lastGenerated) * 1000).toISOString();
-
+  const dateTime = formatDate(Number(lastGenerated), 'iso');
```
VWSCoronaDashboard, line 68: Do we really want ISO format in the footer?

VWSCoronaDashboard4: it's used as `datetime` attribute, not to display the text string

PR #422, opened 2020-09-24
Improve chloropleth tiles part 1

VWSCoronaDashboard8:
## Summary

https://trello.com/c/n0leX1zC/552-choropleth-tegel-verbetering

minus the legend redesign

## Motivation

I have not used the 40/60 layout as it seems the chart didn't look right using it. The chloropleth didn't take up the extra space, because somehow it doesn't scale well to the maximum of its container.

I have used a grid layout for both the Chloropleth composition with legend and text. I also applied a grid layout to the controls. This will affect all other controls in a way that:

 - Makes buttons evenly sized
 - Doesn't grow the buttons to full width in the container anymore. This makes it more in line with how controls were already rendered for the line chart.

Screenshot 2020-09-24 at 12 00 07
Screenshot 2020-09-24 at 12 09 57

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,5 +110,9 @@
     "validate-json": "node schema/validator/validate-json.js",
     "validate-single": "node schema/validator/validate-single.js",
     "generate-typescript": "node schema/validator/generate-typescript.js"
-  }
+  },
+  "browserslist": [
+    "defaults",
+    "not ie < 11"
+  ]
 }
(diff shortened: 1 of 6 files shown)


VWSCoronaDashboard1:
Unfortunately, this is how it renders on desktop in IE11

image

VWSCoronaDashboard1 (approved):

PR #424, opened 2020-09-24
Feature/menu improvements

VWSCoronaDashboard4:
## Summary

Adjust menu to be visible immediately on mobile on homepage.

## Motivation

Better UX for mobile users.

## Detailed design

This is a concatenation of the following changes:

1. Non-JS version always has the menu and the content visible, the menu at the top.
2. A JS class is added to the HTML to hide the menu or content at will.
3. A hook `useMenuState` is added to handle menu state and opening and closing.

## Drawbacks

n/a

## Alternatives

* Menu could also be defaulted underneath the content on non-homepage pages.

## Unresolved questions

* Are there more adjustments needed to the non-JS version?
* I'd suggest we add a third skiplink for accessibility to quickly skip over the menu directly to the actual content, now the "content" goes to the menu and "navigation" goes to the main navigation.

### Governance

- [ ] Documentation is added


src/utils/useMenuState.ts
```diff
@@ -0,0 +1,58 @@
...
+    }
+    setMenu(true);
+  };
+
+  const handleMenuClick = (
+    event: React.MouseEvent<HTMLAnchorElement, MouseEvent> | undefined
+  ) => {
+    if (event?.currentTarget) {
+      event.currentTarget.blur();
+    }
```
VWSCoronaDashboard1, line 34:
```suggestion
    event?.currentTarget?.blur();
```

src/utils/useMenuState.ts
```diff
@@ -0,0 +1,58 @@
...
+ */
+export function useMenuState(defaultOpen = false): MenuStateReturn {
+  const [isMenuOpen, setMenu] = useState<boolean>(defaultOpen);
+
+  const openMenu = (
+    event: React.MouseEvent<HTMLAnchorElement, MouseEvent> | undefined
+  ): void => {
+    if (event) {
+      event.preventDefault();
+    }
```
VWSCoronaDashboard1, line 25:
```suggestion
    event?.preventDefault();
```

src/utils/useMenuState.ts
```diff
@@ -0,0 +1,58 @@
...
+   * */
+  useEffect(() => {
+    const handleCloseMenu = () => {
+      setMenu(false);
+    };
+    Router.events.on('beforeHistoryChange', handleCloseMenu);
+    return () => {
+      Router.events.off('beforeHistoryChange', handleCloseMenu);
+    };
+  });
```
VWSCoronaDashboard1, line 51:
```suggestion
  useEffect(() => {
    const handleCloseMenu = () => {
      setMenu(false);
    };
    Router.events.on('beforeHistoryChange', handleCloseMenu);
    return () => {
      Router.events.off('beforeHistoryChange', handleCloseMenu);
    };
  }, []);
```
I'm guessing this hook is supposed to only run once, so therefore it would need an empty dependency array.

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard1:
Apart from the small nitpicks I added I don't see anything wrong with this. The skiplink addition might indeed be a good idea with the idea of accessibility in mind.

src/components/layout/MunicipalityLayout.tsx
```diff
@@ -24,6 +24,7 @@ import GetestIcon from '~/assets/test.svg';
 import Ziekenhuis from '~/assets/ziekenhuis.svg';
 import RioolwaterMonitoring from '~/assets/rioolwater-monitoring.svg';
 import Arrow from '~/assets/arrow.svg';
+import { useMenuState } from '~/utils/useMenuState';
```
VWSCoronaDashboard8, line 27: Nitpick; I wouldn't classify this as a utility if you don't use it many times over in different places. And the fact that it's a hook contributes even more to this feeling.

If the logic is only used here I would keep it in this folder or the layout / parent folder.

src/components/layout/NationalLayout.tsx
```diff
@@ -33,6 +33,7 @@ import siteText from '~/locale/index';
 import { WithChildren } from '~/types/index';
 
 import { INationalData } from '~/static-props/nl-data';
+import { useMenuState } from '~/utils/useMenuState';
```
VWSCoronaDashboard8, line 36: So I guess keep it the layout folder then ;)

src/components/layout/sidebar.tsx
```diff
@@ -0,0 +1,11 @@
+import { WithChildren } from '~/types/index';
+
+interface SidebarProps {
+  openedTendency: string;
```
VWSCoronaDashboard8, line 4: Sure you want to call this "tendency"? It is very indeterministic by nature. I would suggest `openedMode` or something.

And I think you can be strict on the types of string expected here i.e. `"aside" | "content"`

src/utils/useMenuState.ts
```diff
@@ -0,0 +1,58 @@
...
+  openMenu: (event?: React.MouseEvent<HTMLAnchorElement, MouseEvent>) => void;
+  handleMenuClick: (
+    event?: React.MouseEvent<HTMLAnchorElement, MouseEvent>
+  ) => void;
+}
+
+/**
+ * Keep track of the menu state and handle the opening and closing.
+ * Returns controls to the state.
+ * @param defaultOpen
```
VWSCoronaDashboard8, line 15: I think these kind of doctype annotations are redundant in Typescript

src/utils/useMenuState.ts
```diff
@@ -0,0 +1,58 @@
+import { useState, useEffect } from 'react';
+import { Router } from 'next/router';
+
+interface MenuStateReturn {
```
VWSCoronaDashboard8, line 4: I think this can simply be named `MenuState` as it is irrelevant that it's being returned somewhere. That's what your IDE will tell you :)

src/utils/useMenuState.ts
```diff
@@ -0,0 +1,58 @@
...
+  ) => void;
+}
+
+/**
+ * Keep track of the menu state and handle the opening and closing.
+ * Returns controls to the state.
+ * @param defaultOpen
+ */
+export function useMenuState(defaultOpen = false): MenuStateReturn {
+  const [isMenuOpen, setMenu] = useState<boolean>(defaultOpen);
```
VWSCoronaDashboard8, line 18: Nitpick;

```suggestion
  const [isMenuOpen, setMenuOpen] = useState<boolean>(defaultOpen);
```

You're not really "setting a menu" here.

src/utils/useMenuState.ts
VWSCoronaDashboard8: I find `event && event.preventDefault();` a little more explicit for some reason... But both are fine with me 👍

src/utils/useMenuState.ts
VWSCoronaDashboard8, line 51: It's a little bit code duplication but this is also an option IMO

```suggestion
  useEffect(() => {
    Router.events.on('beforeHistoryChange', () => setMenu(false));
    return () => {
      Router.events.off('beforeHistoryChange',  () => setMenu(false));
    };
  }, []);
```

VWSCoronaDashboard8 (changes requested):
👍 Nice one regardless of the comments :)

src/utils/useMenuState.ts
VWSCoronaDashboard1: It is however a requirement right now that we try and document the hooks. The comment doesn't need to go into the implementation, but should just give a general description about what the hook does.

src/utils/useMenuState.ts
VWSCoronaDashboard4: Not sure if these will do the same. For event handlers, especially event-un-handlers arrow functions do not work as arguments (only references to a named function / variable containing a function.)

src/components/layout/NationalLayout.tsx
```diff
@@ -33,7 +33,7 @@ import siteText from '~/locale/index';
 import { WithChildren } from '~/types/index';
 
 import { INationalData } from '~/static-props/nl-data';
-import { useMenuState } from '~/utils/useMenuState';
+import { useMenuState } from '~/components/layout/useMenuState';
```
VWSCoronaDashboard8, line 36: Nipick; When you use feature local files, a relative import is cleaner and also clearly shows you that the dependency is local to the file.

```suggestion
import { useMenuState } from './useMenuState';
```

src/components/layout/SafetyRegionLayout.tsx
```diff
@@ -22,7 +22,7 @@ import GetestIcon from '~/assets/test.svg';
 import Ziekenhuis from '~/assets/ziekenhuis.svg';
 import RioolwaterMonitoring from '~/assets/rioolwater-monitoring.svg';
 import Arrow from '~/assets/arrow.svg';
-import { useMenuState } from '~/utils/useMenuState';
+import { useMenuState } from '~/components/layout/useMenuState';
```
VWSCoronaDashboard8, line 25:
```suggestion
import { useMenuState } from './useMenuState';
```

VWSCoronaDashboard8 (approved):

src/components/layout/SafetyRegionLayout.tsx
VWSCoronaDashboard4: So much for VScode auto fix paths ... :-)

PR #425, opened 2020-09-24
Update and standardize schemas

VWSCoronaDashboard8:
## Summary

New schemas were required for building some upcoming regional charts. I have attempted to standardize and simplify the way we define schemas so that we can use this as a basis for standardizing the other existing schemas down the line.

- All data belonging to a single subject is grouped in a single schema.
- Adds reusable definitions for time series and other types.
- Adds new regional schemas for intensive-cares, hospitals, and nursing homes.

Note that the "definitions" at the bottom of each file are now just copy-pasted. We might find a way to put these in a separate file so we can keep it DRY, but for now, it's already valuable if things can be copied I think.

### DIFF
--- a/schema/regional/hospitals.json
+++ b/schema/regional/hospitals.json
@@ -5,9 +5,12 @@
   "type": "object",
   "properties": {
     "intake_average": { "$ref": "#/definitions/time_series_value" },
+    "beds_occupied": {
+      "type": "number"
+    },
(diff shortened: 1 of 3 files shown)


VWSCoronaDashboard5 (approved):

VWSCoronaDashboard8:
This proposal was nog technically possible.

PR #430, opened 2020-09-25
chore: Use typeof window instead of process.env for browser checking

RobbyUitbeijerse:
## Summary

Implemented a utility for checking whenever code is running client side, to replace the usage of process.browser 

## Motivation

To not use the process.browser which is already removed from the next.js core and marked legacy / deprecated (https://github.com/vercel/next.js/pull/7651)


VWSCoronaDashboard (approved):

VWSCoronaDashboard:
Thanks @RobbyUitbeijerse , this looks good!

PR #432, opened 2020-09-25
chore: remove unused dependencies

RobbyUitbeijerse:
Hi there!

I was aiming to reduce the Docker image size by using the node:12-alpine alternative of the node:12 image, it's a much more lightweight alternative to the default Node docker image and maintained by the same team https://hub.docker.com/_/node

Nowadays `yarn` is included, so basically the existing Dockerfile and tools should work without any problem, but the total image size will be reduced by quite a bit.

After the image was adjusted, I noticed that `next-optimized-images` was installed & included but not actually used (found out during testing the alpine image compatibility). During removing the plug-in I noticed some other unnecessary / outdated things which I also removed.. 

## A complete overview of the changes:

- Adjusted node:12 to node:12-alpine for a much smaller Docker image size (please test this well, I wasn't actually able to complete the build process due to missing json files)
- Removed next-optimized-images as the plug-in doesn't seem to be used at all (also the related optimization packages)
- Removed 'react-use-measure' from the transpile process as that dependency doesn't seem to be in the project (anymore)
- ~~Removed the .babelrc file as Next.js provides the exact same setting(s) as default already: https://nextjs.org/docs/advanced-features/customizing-babel-config~~
- Removed unfetch as it wasn't used (probably because Next.js provides a fetch polyfill already included in the core)
- Enabled React strict mode for proper React warnings https://nextjs.org/docs/api-reference/next.config.js/react-strict-mode

Please let me know if there are any questions.

**[EDIT]** the .babelrc is apparently needed for Jest to function properly, so I've put that one back.


VWSCoronaDashboard:
I don't think `node:12-alpine` is going to help much, because the production image uses `bitnami/nginx:latest`. But I'll try it out locally!

RobbyUitbeijerse:
@VWSCoronaDashboard I think you are actually correct! I'm running into trouble actually building the project due to missing a lot of JSON files in public/json, so wasn't able to actually complete the build process and check the outcome.

I also figured that you need bash in order to run the commitHash handler in next.config.js, meaning that even if it would reduce the image size, you would still need to run `apk add bash` in order to make it available in the alpine container. Not sure if it's worth the trouble.

removing next-optimized-images and it's packages already reduces the node_modules by quite a bit. For now: I've adjusted the PR, kept it at just cleaning a couple of extra dependencies;

- linkifyjs is now removed, it wasn't used
- removed @types/geojson from the dependencies as it was present in the devDepedencies already
- Version locked @vx/geo and @vx/scale to prevent any incompatibility issues due to installing latest
- specified the version for d3-scale to prevent future issues (not sure if it will ever get a new 2.* release, but better safe than sorry)

VWSCoronaDashboard:
About the JSON files, we're looking into ways to automate that. For now the simplest solution seems to be publishing a zip file which we can download with a yarn command. 

If you want to run and test locally, you can use this zip file for now. Put the contents in `./public/json/`

[json-27-09-2020.zip](https://github.com/minvws/nl-covid19-data-dashboard/files/5291544/json-27-09-2020.zip)

VWSCoronaDashboard (approved):

PR #453, opened 2020-09-28
Match area chart to line chart implementation

VWSCoronaDashboard8:
Line charts were recently updated with improved layout, better y-axis scaling and signaalwaarde labeling with both text and value. The area chart was not adapted.

This PR matches area-chart with the current line-chart implementation.


src/components/areaChart/areaChart.module.scss
```diff
@@ -0,0 +1,36 @@
+@import '~/scss/variables.scss';
+
+$breakpoint: 1200px;
+
+.root {
```
VWSCoronaDashboard4, line 5: empty property block

src/components/areaChart/areaChart.module.scss
```diff
@@ -0,0 +1,36 @@
+@import '~/scss/variables.scss';
+
+$breakpoint: 1200px;
```
VWSCoronaDashboard4, line 3: not sure about this variable name, it's very generic

src/components/areaChart/index.tsx
```diff
@@ -21,6 +21,8 @@ type TRange = [Date, number | null, number | null];
 type TLine = [Date, number | null];
 
 interface AreaChartProps {
+  title: string;
+  description?: string;
```
VWSCoronaDashboard4, line 25: optional properties should be at the bottom

src/components/areaChart/index.tsx
```diff
@@ -110,7 +153,9 @@ function getOptions(props: IGetOptions): Highcharts.Options {
       xDateFormat: '%d %b %y',
       formatter(): string {
         const rangePoint = rangeData.find((el) => el[0].getTime() === this.x);
-        // @ts-ignore
+
+        if (!rangePoint) return '';
```
VWSCoronaDashboard4, line 157: you can also return false; then it will not display an empty tooltip

VWSCoronaDashboard4 (commented):
Some nitpicks; some things I comment on already occur on other places in the areachart, maybe it just needs a little bit of brushing up in a separate PR.

src/components/areaChart/index.tsx
VWSCoronaDashboard1: Is there a lint rule for this perhaps? Would be nice if these kinds of things get triggered.

VWSCoronaDashboard1 (approved):

src/components/areaChart/areaChart.module.scss
VWSCoronaDashboard8: It's also very local. I think that's the beauty of css modules, that you can keep the naming super simple.

VWSCoronaDashboard8:
@VWSCoronaDashboard1 A lint rule for what exactly?

src/components/areaChart/index.tsx
VWSCoronaDashboard8: That's a pretty subjective claim I think. I don't agree with it. I think it helps if you try to logically group things together and title and description in my mind belong together.

It's different for function arguments. There I agree that you would place optional arguments at the end, so that you can call the function leaving those out. But I don't see a point in interface definitions.

src/components/areaChart/index.tsx
VWSCoronaDashboard8: 👍 The function definition claimed it wanted to return a string, so I've removed that also.

src/components/areaChart/areaChart.module.scss
VWSCoronaDashboard: Agree, with module files this does not matter at all.

src/components/areaChart/index.tsx
VWSCoronaDashboard: I think putting optional at the end mostly makes sense for arguments. Other types, I don't see a reason to be dogmatic.

src/components/areaChart/index.tsx
```diff
@@ -230,13 +256,42 @@ export default function AreaChart(props: AreaChartProps) {
...
+     */
+    .filter(([_date, a, b]) => a !== null && b !== null)
+    .flatMap(([_date, a, b]) => [a, b] as [number, number]);
+
+  /**
+   * Adding an absolute value to the yMax like in LineChart doesn't seem to
+   * work well for AreaChart given the values it is rendered with. So for
+   * now we use a (relative) 20% increase.
+   */
+  return Math.max(signaalwaarde * 1.2, ...flatValues);
```
VWSCoronaDashboard, line 296: While this solves your problem, I don't think this is the right solution. Adding padding/margin to your graph is a concern for the presentational layer. 

`calculateYMax` now is an impure function which does not return what its name implies. 

I think it's easy to fix though:

```
  const PADDING_INCREASE = 1.2;
  const yMax = calculateYMax(rangeData, signaalwaarde) * PADDING_INCREASE;
```

src/components/lineChart/index.tsx
```diff
@@ -119,6 +119,7 @@ function getChartOptions(values: Value[], signaalwaarde?: number) {
             {
               value: signaalwaarde,
               color: 'transparent',
+              zIndex: 10,
```
VWSCoronaDashboard, line 122: This is a duplicate of 104, so should this be a constant at the top of the file?

VWSCoronaDashboard (changes requested):
Please don't put the padding multiplier in your max function. Besides that I'm good with this.

src/components/areaChart/index.tsx
VWSCoronaDashboard8: Impure as in has side-effects? 

The result is a value that is used to set the ymax of the chart, so I don't really see how this function name is so misleading. Also have a look at lineChart where I did the exact same and no one made any comments there.

I feel like we're really nitpicking here, especially considering the current state of the codebase 🤷‍♂️

VWSCoronaDashboard8: @VWSCoronaDashboard Your proposed solution doesn't produce the same result. The function currently only adds to the max if the signaalwaarde + extra margin is a higher value than the maximum value in the time series.

VWSCoronaDashboard8: I can make it like this `const yMax = calculateYMax(rangeData, signaalwaarde * PADDING_INCREASE)`. That could be the compromise.

VWSCoronaDashboard4 (approved):

PR #457, opened 2020-09-28
Fix searchbar not functioning in IE11

VWSCoronaDashboard1:
## Summary

The searchbar on the municipal and region pages wasn't working due to a crash under the hood in some other piece of code.

## Motivation

Bug report

## Detailed design

Added the polyfills intl-getcanonicallocales and intl-datetimeformat (this caused the crash in IE11).
I only added the timezone data for Europe/Amsterdam since adding the full set would bloat our code with an additional 280k, so this might seem a bit hacky, but there's no clean way of getting this config in. (as far as I could find)

## Drawbacks

Polyfills are never fun, but there's no way around it.

## Alternatives

Suggestions are welcome

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added


VWSCoronaDashboard4 (approved):

src/utils/formatDate.ts
```diff
@@ -1,3 +1,23 @@
+import '@formatjs/intl-getcanonicallocales/polyfill';
+import '@formatjs/intl-datetimeformat/polyfill';
+import '@formatjs/intl-datetimeformat/locale-data/en';
+import '@formatjs/intl-datetimeformat/locale-data/nl';
+
+if ('DateTimeFormat' in Intl && (Intl.DateTimeFormat as any).__addTZData) {
```
VWSCoronaDashboard8, line 6: In code that is impossible to understand it would be nice to have some comments

VWSCoronaDashboard8 (approved):

src/utils/formatDate.ts
VWSCoronaDashboard1: done

PR #465, opened 2020-09-29
Remove grid layout from gemeente selection map

VWSCoronaDashboard8:
Municipality selection map as rendered in the same 50/50 column layout as other Chloroplets, but that is not preferable. This PR removes the grid layout so that the map takes up the full width.

Screenshot 2020-09-29 at 10 10 33


VWSCoronaDashboard (approved):

VWSCoronaDashboard4:
Shouldn't the borders be in blue?

VWSCoronaDashboard1 (approved):

PR #490, opened 2020-09-29
feat: new tooltip design

VWSCoronaDashboard5:
## Summary

This PR adds a new tooltip design for the choropleth maps on pages. The updated design shows the name and value of the node in the map and allows for using the tooltip as a link (using an `onClick`) to the node's page (i.e. `/gemeente/GM0627/positief-geteste-mensen` on a node in a municipality choropleth).

The UX of this feature is rather implicit still, but it uses the same implementation as used in the Escalation (Risico) map, which is when the visitor is on mobile, the tooltip is showed on a tap (mobile `:hover`) and then allows the user to use the tooltip as a link to navigate.

### Governance

- [x] Documentation is added


src/components/chloropleth/tooltips/municipal/positiveTestedPeopleTooltip.tsx
```diff
@@ -1,18 +1,36 @@
+import { NextRouter } from 'next/router';
 import { ReactNode } from 'react';
 import { MunicipalityProperties } from '../../shared';
-import styles from '~/components/chloropleth/chloropleth.module.scss';
+import styles from '../tooltip.module.scss';
 
-export function positiveTestedPeopleMunicipalTooltip(
+export const positiveTestedPeopleMunicipalTooltip = (router: NextRouter) => (
```
VWSCoronaDashboard1, line 6: since this has now become a factory method as well, I suggest you rename it to 'createPositiveTestedPeopleMunicipalTooltip'

src/components/chloropleth/tooltips/municipal/positiveTestedPeopleTooltip.tsx
```diff
@@ -1,18 +1,36 @@
...
   context: MunicipalityProperties & { value: number }
-): ReactNode {
+): ReactNode => {
+  const onSelectRegion = (event: any) => {
+    event.stopPropagation();
+    router.push(
+      '/gemeente/[code]/positief-geteste-mensen',
+      `/gemeente/${context.gemcode}/positief-geteste-mensen`
+    );
```
VWSCoronaDashboard1, line 14: Instead of duplicating this code, you can also import the createSelectMunicipalHandler factory here:
`const handler = createSelectMunicipalHandler(router);
handler(context);
`

src/components/chloropleth/tooltips/region/positiveTestedPeopleTooltip.tsx
```diff
@@ -1,18 +1,36 @@
+import { NextRouter } from 'next/router';
 import { ReactNode } from 'react';
 import { SafetyRegionProperties } from '../../shared';
-import styles from '~/components/chloropleth/chloropleth.module.scss';
+import styles from '../tooltip.module.scss';
 
-export function positiveTestedPeopleRegionalTooltip(
+export const positiveTestedPeopleRegionalTooltip = (router: NextRouter) => (
```
VWSCoronaDashboard1, line 6: same naming suggestion as above

src/components/chloropleth/tooltips/region/positiveTestedPeopleTooltip.tsx
```diff
@@ -1,18 +1,36 @@
...
   context: SafetyRegionProperties & { value: number }
-): ReactNode {
+): ReactNode => {
+  const onSelectRegion = (event: any) => {
+    event.stopPropagation();
+    router.push(
+      '/veiligheidsregio/[code]/positief-geteste-mensen',
+      `/veiligheidsregio/${context.vrcode}/positief-geteste-mensen`
+    );
+  };
```
VWSCoronaDashboard1, line 15: Same duplication suggestion as above

VWSCoronaDashboard1 (changes requested):

src/components/chloropleth/tooltips/municipal/positiveTestedPeopleTooltip.tsx
```diff
@@ -1,18 +1,35 @@
...
-import styles from '~/components/chloropleth/chloropleth.module.scss';
+import { createSelectMunicipalHandler } from '../../../chloropleth/selectHandlers/createSelectMunicipalHandler';
+import styles from '../tooltip.module.scss';
+
+export const createPositiveTestedPeopleMunicipalTooltip = (
+  router: NextRouter
+) => (context: MunicipalityProperties & { value: number }): ReactNode => {
+  const onSelectRegion = (event: any) => {
+    event.stopPropagation();
+    const handler = createSelectMunicipalHandler(router);
```
VWSCoronaDashboard1, line 12: You can actually move the creation of this handler outside of the onSelectRegion instance, that way it won't have to be created every time onSelectRegion is invoked

src/components/chloropleth/tooltips/region/positiveTestedPeopleTooltip.tsx
```diff
@@ -1,18 +1,35 @@
...
-import styles from '~/components/chloropleth/chloropleth.module.scss';
+import { createSelectRegionHandler } from '../../../chloropleth/selectHandlers/createSelectRegionHandler';
+import styles from '../tooltip.module.scss';
+
+export const createPositiveTestedPeopleRegionalTooltip = (
+  router: NextRouter
+) => (context: SafetyRegionProperties & { value: number }): ReactNode => {
+  const onSelectRegion = (event: any) => {
+    event.stopPropagation();
+    const handler = createSelectRegionHandler(router);
```
VWSCoronaDashboard1, line 12: same as the comment above

VWSCoronaDashboard1 (changes requested):

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard4 (approved):

src/components/chloropleth/tooltips/region/positiveTestedPeopleTooltip.tsx
```diff
@@ -1,18 +1,35 @@
...
-  context: SafetyRegionProperties & { value: number }
-): ReactNode {
   return (
     context && (
-      <div className={styles.defaultTooltip}>
-        <strong>{context.vrname}</strong>
-        <br />
-        {context.value !== undefined && `${context.value} / 100.000`}
-        {!context.value === undefined && '-'}
+      <div className={styles.escalationTooltip} onClick={onSelectRegion}>
```
VWSCoronaDashboard8, line 19: The use of CSS modules here goes against the idea of modules to separate components. It is maybe not the fault of this PR because the CSS file already existed, but you are importing from a module that seems to be shared amongst multiple different tooltip components. So it becomes a sort of "global" CSS again, only now global to all tooltips. This is not a good separation of concerns IMO. 

If a tooltip is so different that it can not be implemented by extending an existing tooltip component via props, it should be a component by itself with its own css module.

That way you can also simplify class names in generic ones like root, header, info...

VWSCoronaDashboard8 (changes requested):

src/components/chloropleth/tooltips/region/positiveTestedPeopleTooltip.tsx
VWSCoronaDashboard8: I will approve the PR since this might be something better fixed when refactoring to styled components.

VWSCoronaDashboard8 (approved):

src/components/chloropleth/tooltips/region/positiveTestedPeopleTooltip.tsx
VWSCoronaDashboard1: Let's chalk this up to technical debt and add a ticket to the tech debt column?

PR #500, opened 2020-09-30
Add createDate factory method to turn unix timestamps into dates and fix areachart bug

VWSCoronaDashboard1:
## Summary

- Add createDate factory method to turn unix timestamps into dates
- Use createDate everywhere necessary to create Date objects
- Refactor formatDate into two methods: formatDateFromSeconds and formatDateFromMilliseconds
- Fix date bug in areachart where dates where milliseconds where multiplied too often

## Motivation

Bug report

## Detailed design

In the areachart milliseconds were incorrectly multiplied by 1000 and that value was then given to the formatDate method, which multiplied it y a thousand again, leading to all sorts of illogical results.
To avoid any direct new Date() calls a factory method was added that takes a unix timestamp (in seconds) and creates a JS Date from that.
The formatDate method was split into a formatDateFromSeconds  and formatDateFromMilliseconds method so that the areachart can use the formatDateFromMilliseconds in its tooltip and axis formatters.

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added

### DIFF
--- a/src/components/areaChart/index.tsx
+++ b/src/components/areaChart/index.tsx
@@ -42,6 +42,83 @@ type IGetOptions = Omit<AreaChartProps, 'data' | 'title' | 'description'> & {
   lineData: TLine[];
 };
 
+export default function AreaChart(props: AreaChartProps) {
+  const {
+    rangeLegendLabel,
+    lineLegendLabel,
+    data,
+    signaalwaarde,
+    timeframeOptions,
+    title,
+    description,
+  } = props;
+
+  const rangeData: TRange[] = useMemo(() => {
+    return data
+      .sort((a, b) => a.date - b.date)
+      .map((d) => [createDate(d.date), d.min, d.max]);
+  }, [data]);
+
+  const lineData: TLine[] = useMemo(() => {
+    return data.map((value) => {
+      return [createDate(value.date), value.avg];
+    });
+  }, [data]);
+
+  const [timeframe, setTimeframe] = useState<TimeframeOption>('5weeks');
+
+  const chartOptions = useMemo(() => {
+    const getOptionsThunk = (rangeData: TRange[], lineData: TLine[]) =>
+      getChartOptions({
+        rangeData,
+        lineData,
+        signaalwaarde,
+        rangeLegendLabel,
+        lineLegendLabel,
+      });
+
+    const filteredRange = getFilteredValues<TRange>(
+      rangeData,
+      timeframe,
+      (value: TRange) => value[0].getTime()
+    );
+
+    const filteredLine = getFilteredValues<TLine>(
+      lineData,
+      timeframe,
+      (value: TLine) => value[0].getTime()
+    );
+
+    return getOptionsThunk(filteredRange, filteredLine);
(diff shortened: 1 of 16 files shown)


src/utils/createDate.ts
```diff
@@ -0,0 +1,10 @@
+const milliseconds = 1000;
+
+/**
+ * JavaScript uses milliseconds since EPOCH, therefore the value
+ * used to create the new Date instance needs to be multiplied by 1000
+ * to create an accurate dateTime
+ */
+export function createDate(seconds: number): Date {
+  return new Date(seconds * milliseconds);
```
VWSCoronaDashboard8, line 9:
```suggestion
  return new Date(seconds * 1000);
```

I think it's simple enough. Or name the variable better because "milliseconds" is not the same as 1000. You could name it `oneSecondInMilliseconds` but I think it's a little overkill

src/utils/formatDate.ts
```diff
@@ -73,18 +73,25 @@ export function formatDate(
    * formatted by the format() function needs to be multiplied by 1000
    * to format to an accurate dateTime
    */
-  const value = seconds * 1000;
+  const milliseconds = seconds * 1000;
```
VWSCoronaDashboard8, line 76: See, here it is actually milliseconds ;)

src/utils/formatDate.ts
```diff
@@ -73,18 +73,25 @@ export function formatDate(
...
 
-  if (style === 'iso') return new Date(value).toISOString(); // '2020-07-23T10:01:16.000Z'
-  if (style === 'long') return Long.format(value); // '23 juli 2020 om 12:01'
-  if (style === 'medium') return Medium.format(value); // '23 juli 2020'
+  return formatDateFromMilliseconds(milliseconds, style);
+}
+
+export function formatDateFromMilliseconds(
```
VWSCoronaDashboard8, line 81: I don't see a real need for this function. The duplicate code makes me prefer the alternative; call the already available `formatDateFromSeconds` with the correct value.

VWSCoronaDashboard8 (approved):
Just some nitpicks.

src/components/areaChart/index.tsx
```diff
@@ -248,13 +249,13 @@ export default function AreaChart(props: AreaChartProps) {
     const filteredRange = getFilteredValues<TRange>(
       rangeData,
       timeframe,
-      (value: TRange) => value[0].getTime() * 1000
+      (value: TRange) => value[0].getTime()
```
VWSCoronaDashboard4, line 252: Why keep the `getTime` here? we could pass the date object and adjust the `getMinimumUnixForTimeframe` ?

VWSCoronaDashboard4 (approved):
Agree with the other comments. One opinion / question added :-)

src/components/areaChart/index.tsx
VWSCoronaDashboard1: I just fixed the bug as far as I could fathom it ;)

PR #503, opened 2020-09-30
Sprint 7 Schemas

VWSCoronaDashboard8:
This PR holds all changes and additions to schemas that we need for upcoming features. It also attempts to standardize and clean up certain things moving forward.

- regional intensive care admissions
- regional intensive care beds occupied
- regional hospital (kliniek) admissions
- regional hospital (kliniek) beds occupied
- regional nursing home data
- regional GGD data
- national nursing home data  (*), adding the percentage of locations infected.
- difference value for huisarts vermoedens
- difference value for positively tested people
- use of __index.json files for compound schema definitions entry point    
- README documentation on various design decisions and naming conventions

(*) I've kept the national nursing home data the same as regional nursing home data, so it replaces the national indicators we had before, in an effort to standardize.


schema/README.MD
```diff
@@ -0,0 +1,83 @@
+# JSON input data validation
```
VWSCoronaDashboard4, line 1: :100: thanks for extending this

VWSCoronaDashboard4 (approved):
Looking good!

VWSCoronaDashboard1 (changes requested):

VWSCoronaDashboard1:
One other thing: by changing the convention for the schema entry point you also broke the generate-typescript script. Probably an easy fix, but should be added to this PR.

VWSCoronaDashboard1 (approved):

PR #521, opened 2020-10-02
Validator and code generator port to typescript

VWSCoronaDashboard1:
## Summary

As the title suggests

## Motivation

Tech debt

## Detailed design

The validation and code generation code was moved from the schema directory into the src/tools directory.
The javascript sources were ported to typescript and the yarn commands updated accordingly.
Added chalk for some console coloring and meow for pretty usage banners.

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added


src/tools/tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
```
VWSCoronaDashboard8, line 1: Was there a need to create a separate tsconfig? I suspect you could have re-used the existing one in the project root.

src/tools/code-generation/generate-typescript.ts
```diff
@@ -21,33 +21,37 @@ Promise.all(promises).then((result) => {
...
+      generateOptions
+    ).then((ts) => {
       console.info(`Generated typescript for schema '${schemaName}'`);
       return ts;
     });
   });
 }
 
-function saveFile(ts) {
+function saveFile(typescriptSources: string) {
```
VWSCoronaDashboard8, line 52: Nitpick; maybe call it typeDefinitions or something.

src/tools/validator/getSchemaNames.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+export function getSchemaNames(): string[] {
+  const schemaDirectoryPath = schemaDirectory;
+  const contents = fs.readdirSync(schemaDirectoryPath);
+  return contents.filter((item) => {
+    const isDir = fs
+      .lstatSync(path.join(schemaDirectoryPath, item))
+      .isDirectory();
+    return isDir;
+  });
```
VWSCoronaDashboard8, line 21:
```suggestion
  return contents.filter((item) => fs
      .lstatSync(path.join(schemaDirectoryPath, item))
      .isDirectory());
```

src/tools/validator/schemaValidator.ts
```diff
@@ -1,35 +1,36 @@
-const Ajv = require('ajv');
-const fs = require('fs');
-const path = require('path');
+import Ajv, { ValidateFunction } from 'ajv';
```
VWSCoronaDashboard8, line 1: What a shitty name for a library. How are we supposed to know what ajv does? 😅

src/tools/validator/schemaValidator.ts
```diff
@@ -1,35 +1,36 @@
...
     this.basePath = path.dirname(schemaPath);
-    this.schemaPath = schemaPath;
   }
 
   /**
-   * @returns {Promise} A Promise object that will resolve to a validator function.
+   * @returns A Promise object that will resolve to a validator function.
    */
-  init() {
+  init(): Promise<ValidateFunction> {
```
VWSCoronaDashboard8, line 33: Is the return type annotation necessary here? What does it think you will return?

src/tools/validator/schemaValidator.ts
```diff
@@ -39,11 +40,9 @@ class SchemaValidator {
     const validator = new Ajv({
       loadSchema: loadSchema.bind(null, this.basePath),
       $data: true,
-    }); // options can be passed, e.g. {allErrors: true}
-    return validator.compileAsync(schema).then(function (validate) {
-      return validate;
     });
+    return validator.compileAsync(schema).then((validate) => {
+      return validate;
+    }) as Promise<ValidateFunction>;
```
VWSCoronaDashboard8, line 46: I suspect this already tell the function everything

src/tools/validator/schemaValidator.ts
```diff
@@ -1,35 +1,36 @@
...
 }
 
 /**
  * This class creates an instance with the given schema path.
  * The instance returns a validator function that is able to validate a JSON object
  * according to the given schema.
  */
-class SchemaValidator {
-  constructor(schemaPath) {
+export class SchemaValidator {
```
VWSCoronaDashboard8, line 23: The fact that this is class is a bit confusing. The class reads and does things on init() / construction, but doesn't actually have any members that are being called after instantiation it seems. It could have simply been a function that gets passed the correct data and performs its task. It would make the code simpler and easier to understand.

src/tools/validator/validate-json.ts
```diff
@@ -17,20 +18,22 @@ const schemaToJsonLookup = {
...
-const results = Object.keys(schemaToJsonLookup).reduce((aggr, schemaName) => {
+const validationResults = Object.keys(schemaToJsonLookup).reduce<
+  PromiseLike<boolean>[]
+>((validationAggregate, schemaName) => {
   const promises = validate(schemaName, schemaToJsonLookup[schemaName]);
-  aggr.push(...promises);
-  return aggr;
+  validationAggregate.push(...promises);
+  return validationAggregate;
 }, []);
```
VWSCoronaDashboard8, line 28: The reduce makes it way more complicated than needed. And I personally like to prefix promised results with "promised". I suggest doing this:

```
const promisedValidationResults = Object.keys(schemaToJsonLookup).map(schemaName => validate(schemaName, schemaToJsonLookup[schemaName]))
```

src/tools/validator/validate-single.ts
```diff
@@ -1,15 +1,15 @@
...
+import { getSchemaNames, schemaDirectory } from './getSchemaNames';
+import { jsonBasePath } from './jsonBasePath';
 
 const validSchemaNames = getSchemaNames();
 
-var myArgs = process.argv.slice(2);
+const cliArgs = process.argv.slice(2);
 
-if (myArgs.length !== 2) {
+if (cliArgs.length !== 2) {
```
VWSCoronaDashboard8, line 12: There are some nice libraries to help with this type of thing, like [meow](https://www.npmjs.com/package/meow)

VWSCoronaDashboard8 (changes requested):

src/tools/tsconfig.json
VWSCoronaDashboard1: well, you suspect wrong ;)

src/tools/code-generation/generate-typescript.ts
VWSCoronaDashboard1: fair enough

src/tools/validator/schemaValidator.ts
VWSCoronaDashboard1: it might not be, but I don't see this as a carnal sin in this instance, it doesn't clutter the sources or anything. But I can remove it.

src/tools/validator/schemaValidator.ts
VWSCoronaDashboard1: Indeed. The extra cast is necessary because the Ajv definitions are wrong. The definitions claim it returns PromiseLike, but instead we get an actual Promise. PromiseLike doesn't contain a definition for the catch() method, which I use downstream so I had to hack the cast in here.

src/tools/validator/schemaValidator.ts
VWSCoronaDashboard1: Its an acronym for 'Another JSON Validator', which is arguably the least inventive name possible. Its the most widely used one though, so we'll have to make due.

src/tools/validator/schemaValidator.ts
VWSCoronaDashboard1: Init was a poorly named class method, reality is that the init was invoked multiple times so it needed a rename. I've refactored the class into an async factory method called createValidateFunction

src/tools/validator/validate-single.ts
VWSCoronaDashboard1: I've added meow because it does some nice boilerplate stuff. Unfortunately, it only performs flag validation and not parameter validation (or at least not according to the docs) so it didn't really clean up a lot IMHO.

As a general comment to the above comments: I took the time to beautify stuff because today is my day off. These tools were written in a hurry and are also not really subjected to an elaborate future of modification, so I won't spend much more time fixing nitpicks on it after this ;)

src/tools/validator/getSchemaNames.ts
```diff
@@ -11,9 +11,8 @@ export const schemaDirectory = path.join(__dirname, '..', '..', '..', 'schema');
  * @returns {array} The list of schema names
  */
 export function getSchemaNames(): string[] {
-  const schemaDirectoryPath = schemaDirectory;
-  const contents = fs.readdirSync(schemaDirectoryPath);
-  return contents.filter((item) => fs
-      .lstatSync(path.join(schemaDirectoryPath, item))
-      .isDirectory());
+  const contents = fs.readdirSync(schemaDirectory);
```
VWSCoronaDashboard8, line 14: nitpick; not really a need to store the intermediate results.

src/tools/code-generation/generate-typescript.ts
```diff
@@ -0,0 +1,58 @@
...
+const promises = schemaNames.map(generateTypeScriptFromSchema);
+
+Promise.all(promises).then((result) => {
+  saveDefinitionsFile(result.join('\n'));
+});
+
+/**
+ * Loads the given schema by name and generates typescript interfaces from it.
+ *
+ * @param schemaName the given schema name
```
VWSCoronaDashboard4, line 24: I'd remove these @param and @returns, and add the Typescript return type the TS way

src/tools/validator/validate-json.ts
```diff
@@ -49,8 +49,8 @@ Promise.all(validationResults)
  * @param fileNames An array of json file names
  * @returns An array of promises that will resolve either to true or false dependent on the validation result
  */
-function validate(schemaName: string, fileNames: string[]) {
-  const validatorInstance = new SchemaValidator(
+async function validate(schemaName: string, fileNames: string[]) {
+  const validateFunction = await createValidateFunction(
```
VWSCoronaDashboard8, line 53: Nitpick; If you only ever call the validator once in a script pass, there was not really a need for a factory function I think. You could have just passed the data together with the other arguments in a single function call.

src/tools/validator/createValidateFunction.ts
```diff
@@ -0,0 +1,38 @@
+import Ajv, { ValidateFunction } from 'ajv';
+import fs from 'fs';
+import path from 'path';
+
+/**
+ * Loads the given uri and parses its conts to JSON
+ * @param {string} basePath The given base directory name
+ * @param {string} uri The given filename
+ */
+function loadSchema(basePath: string, uri: string): Promise<any> {
```
VWSCoronaDashboard4, line 10: Same for the @params and @returns in this file

src/tools/validator/validate-json.ts
```diff
@@ -61,21 +61,16 @@ function validate(schemaName: string, fileNames: string[]) {
...
-          throw new Error(`${fileName} is invalid`);
-        }
-        console.log(`${fileName} is valid`);
-        return true;
-      })
-      .catch(() => {
-        return false;
-      });
+    const valid = validateFunction(data);
+    if (!valid) {
```
VWSCoronaDashboard8, line 65:
```suggestion
    const isValid = validateFunction(data);
    if (!isValid) {
```

or 

```suggestion
    if (!validateFunction(data)) {
```

VWSCoronaDashboard8 (approved):
👍

VWSCoronaDashboard4 (changes requested):
Looking good :100: Some methods miss a typescript return where they would be handy. The docblocks do not add much value and are sometimes not aligned with the params/returns types.

PR #525, opened 2020-10-05
Fixes skip link scroll behaviour in IE11

VWSCoronaDashboard4:
## Summary

Fixes a bug in IE11 that skip links (to the GGD data in this case) cause an odd cut off of the page, resulting a fairly un-usuable page.

Fixed it by making sure the skip link target has an invisible part of it rendered at the left edge of the screen.

### DIFF
--- a/src/components/layout/layout.module.scss
+++ b/src/components/layout/layout.module.scss
@@ -245,6 +245,15 @@ $language-switcher-offset: 55px;
   margin-left: 4rem;
 }
 
+/*
+  Ensure skip link anchors to have a (non visible) start at the left side of the screen
+  This fixes odd skip-link behaviour in IE11
+*/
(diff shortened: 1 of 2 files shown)


src/components/layout/layout.module.scss
```diff
@@ -245,6 +245,15 @@ $language-switcher-offset: 55px;
...
 }
 
+/*
+  Ensure skip link anchors to have a (non visible) start at the left side of the screen
+  This fixes odd skip-link behaviour in IE11
+*/
+.contentHeader[id] {
+  margin-left: -100vw;
+  padding-left: 100vw;
+}
```
VWSCoronaDashboard1, line 255: IE hacks FTW! haha

VWSCoronaDashboard1 (approved):

PR #529, opened 2020-10-05
Add municipalities total_positive_tested_people to required properties list

VWSCoronaDashboard5:
## Summary

This PR adds the properties required to the schemas in order to show the absolute number of positive tested people to the tooltips in the chloropleths of regional and municipality. This data is not yet in place, so in order to complete this feature request, the schema must first be updated.

### Governance

- [ ] Documentation is added
- [] Test cases are added or updated


VWSCoronaDashboard4:
Can you run `yarn generate-typescript` and commit the generated type definitions?

src/types/data.d.ts
```diff
@@ -89,6 +89,7 @@ export interface PositiveTestedPeople {
   date_of_report_unix: number;
   gmcode: string;
   positive_tested_people: number;
+  total_positive_tested_people?: number;
```
VWSCoronaDashboard8, line 92: Why is this value optional? I suspect it should be added to the required field in the schema.

src/types/data.d.ts
```diff
@@ -543,6 +544,7 @@ export interface RegionPositiveTestedPeople {
   date_of_report_unix: number;
   vrcode: string;
   positive_tested_people: number;
+  total_positive_tested_people?: number;
```
VWSCoronaDashboard8, line 547: Same here.

schema/municipalities/positive_tested_people.json
```diff
@@ -20,6 +20,9 @@
     "positive_tested_people": {
       "type": "number"
     },
+    "total_positive_tested_people": {
+      "type": "number"
```
VWSCoronaDashboard8, line 24: number is a floating point. Unless this is an average I think it should be integer

VWSCoronaDashboard8 (changes requested):

src/types/data.d.ts
VWSCoronaDashboard5: my assumption is that the generated `data.d.ts` file made this value optional because the `"total_positive_tested_people"` was not included in the `required` array in `schema/x/positive_tested_people.json`. The reason why I left it out for now is because if this PR gets merged before we receive the data from the backend, the schema's will cause errors (for missing values).

VWSCoronaDashboard8: Yes that's what I meant with the required field. I don't think it will generate errors, only for the validate script but that is not being run anywhere automatically. 

I think it's odd to code on types that are not correct, because you are likely to forget to change them afterwards. I have been working also with JSONs today that did not match the schema yet. It was no issue for me. @VWSCoronaDashboard1 Can you confirm?

VWSCoronaDashboard1: The schema validations are only run automatically in the production build pipe, so yes, we don't have to make stuff optional anymore to prevent the develop build from breaking

VWSCoronaDashboard8 (approved):
Change the PR title before merging because it is not very informative or correct. I would suggest something like this:

"Add municipalities total_positive_tested_people to required properties list"

PR #537, opened 2020-10-06
Feature/map aspect ratio

VWSCoronaDashboard4:
## Summary

Enlarge maps to full width of parent container (especially noticeable on smaller/medium screens), ensuring aspect ratio.

## Detailed design

* `useChartDimensions` now accepts an aspect ratio param;
* aspect ratio of 1.2 is passed for our current safety region and municipality maps
* fixed and semi-fixed widths using `isLargeScreen` are removed
* the large map on the `/gemeentes` endpoint has styling for the container to ensure the map does not render too big

### DIFF
--- a/src/components/chloropleth/hooks/useChartDimensions.tsx
+++ b/src/components/chloropleth/hooks/useChartDimensions.tsx
@@ -47,7 +47,6 @@ export type TCombinedChartDimensions = TChartDimensions & {
 };
 
 export const useChartDimensions = (
-  chartDimensions: TChartDimensions = {},
   aspectRatio?: number
 ): [MutableRefObject<HTMLElement | null>, TCombinedChartDimensions] => {
   /**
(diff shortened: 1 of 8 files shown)


src/components/chloropleth/hooks/useChartDimensions.tsx
```diff
@@ -47,7 +47,8 @@ export type TCombinedChartDimensions = TChartDimensions & {
 };
 
 export const useChartDimensions = (
-  chartDimensions: TChartDimensions = {}
+  chartDimensions: TChartDimensions = {},
+  aspectRatio?: number
```
VWSCoronaDashboard1, line 51: won't it be a tad neater to switch the arguments? That way you won't have to pass in undefined for the chartDimensions.
In fact, the chartDimensions param might just be removed since I don't think its actually used anywhere.

VWSCoronaDashboard1 (approved):

src/components/chloropleth/hooks/useChartDimensions.tsx
VWSCoronaDashboard4: Good one, I'll nuke it!

VWSCoronaDashboard5 (approved):

PR #549, opened 2020-10-07
Add workaround for content header metadata sources

VWSCoronaDashboard8:
For national IC and ziekenhuis opnames pages we need a double content header metadata mention for sources. The current abstractions do not allow this. Since this data is likely removed from the headers in the next sprint, the problem is now solved via copy-paste-hack. No need to spend any real effort to make this right.

### DIFF
--- a/src/components/contentHeaderMetadataHack/index.tsx
+++ b/src/components/contentHeaderMetadataHack/index.tsx
@@ -1,0 +1,61 @@
+import { TitleWithIcon } from '~/components/titleWithIcon';
+import styles from './layout.module.scss';
+import { MetadataHack } from './metadataHack';
+
+/**
+ * An alteration from layout/ContentHeader in order to render two sources in metadata.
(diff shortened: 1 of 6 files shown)


VWSCoronaDashboard1 (approved):

VWSCoronaDashboard4:
What if we would adjust the original dataSource to only accept arrays of objects? Then, for all other occurrences of the content header we can just add the required square brackets to make it work?

But I agree, the current setup of the ContentHeader component forces is in dark, twisted corners.

VWSCoronaDashboard8:
@VWSCoronaDashboard4 Passing an array was my initial idea of solving this issue, but it is hard to get it right with the way it is designed. It would be easy of every source had it's own icon and title but things are combined with & on one line. And since we're likely taking it out in the next sprint and the abstraction is currently a mess, I didn't think it was worth the effort.

VWSCoronaDashboard4 (approved):

PR #550, opened 2020-10-07
add check for content in verantwoording MD

VWSCoronaDashboard5:
## Summary

An empty string in the `verantwoording` object in `nl.json` still returns a `<div>` filled with html elements in the map function on `pages/verantwoording.tsx`. This PR adds a ternary to check if the content (title and text) is longer than 0 characters and otherwise returns `null`.

### Governance

- [x] Documentation is added


src/pages/verantwoording.tsx
```diff
@@ -68,16 +66,19 @@ const Verantwoording: FCWithLayout<{ text: any; lastGenerated: string }> = (
...
-                  <h3>{item.cijfer}</h3>
-                  <div
-                    dangerouslySetInnerHTML={{
-                      __html: item.verantwoording,
-                    }}
-                  />
-                </Fragment>
-              ))}
+              {text.verantwoording.cijfers.map((item: ICijfer) => {
+                return item.cijfer.length > 0 &&
```
VWSCoronaDashboard4, line 70: This can be simplified:
* the `return` is not needed if you keep you wrap it with round brackets (instead of creating a block with the `{ }`
* `> 0` is redundant, because `0` is falsy
* a ternary operator is less readable, elsewhere we use `&&` instead of the `?`, so you can remove the ` : null`

VWSCoronaDashboard4 (approved):

src/pages/verantwoording.tsx
```diff
@@ -68,16 +66,20 @@ const Verantwoording: FCWithLayout<{ text: any; lastGenerated: string }> = (
...
-                      __html: item.verantwoording,
-                    }}
-                  />
-                </Fragment>
-              ))}
+              {text.verantwoording.cijfers.map(
+                (item: ICijfer) =>
+                  item.verantwoording &&
+                  item.cijfer && (
+                    <Fragment key={`item-${item.cijfer}`}>
```
VWSCoronaDashboard8, line 73:
```suggestion
                    <Fragment key={item.cijfer}>
```
Not your code, since it was already there, but this would have the same effect

VWSCoronaDashboard8 (approved):

src/pages/verantwoording.tsx
```diff
@@ -26,9 +26,7 @@ interface StaticProps {
 
 export async function getStaticProps(): Promise<StaticProps> {
   const text = require('../locale/index').default;
-  const serializedContent = text.verantwoording.cijfers.map(function (
-    item: ICijfer
-  ) {
+  const serializedContent = text.verantwoording.cijfers.map((item: ICijfer) => {
     return { ...item, verantwoording: MDToHTMLString(item.verantwoording) };
   });
```
VWSCoronaDashboard8, line 32: nitpick; You don't need the return statement with a one-line lambda function.

```suggestion
  const serializedContent = text.verantwoording.cijfers.map((item: ICijfer) => ({ ...item, verantwoording: MDToHTMLString(item.verantwoording)});
```

VWSCoronaDashboard8: Also the type annotation `ICijfer` should be inferred by the compiler already

VWSCoronaDashboard1 (approved):

src/pages/verantwoording.tsx
```diff
@@ -26,9 +26,12 @@ interface StaticProps {
 
 export async function getStaticProps(): Promise<StaticProps> {
   const text = require('../locale/index').default;
-  const serializedContent = text.verantwoording.cijfers.map((item: ICijfer) => {
-    return { ...item, verantwoording: MDToHTMLString(item.verantwoording) };
-  });
+  const serializedContent = text.verantwoording.cijfers.map(
+    (item: ICijfer) => ({
```
VWSCoronaDashboard8, line 30: No need for ICijfer

PR #552, opened 2020-10-07
Show percentage of infected locations on country and region pages

VWSCoronaDashboard1:
## Summary

Show percentage of infected locations on country and region pages.
Add choropleth to the country page as well.

## Motivation

Feature request

## Detailed design

Mostly just simple renderings of values form the json data.
Added new threshold data for the nursing_home.infected_locations_total metric.

## Drawbacks

When the Regions data file was still 'flat', it was rather easy to associate the threshold data with the metric. It was just a Record that looked up positive_tested_people, hospital_admissions or escalation_levels and returned a data struct that described the thresholds and its colors.

Now that nursing home data was added, a new level was introduced: nursing_home.infected_locations_total, etc

So, resolving the threshold data for the given metric and now metric sub key gets rather convoluted, I would appreciate some suggestions on how to clean this up a little.
Of course, it's entirely possible to just inject the threshold data in the choropleth component, but that way there's no way of making sure that the choropleth color data really is meant for the actual data being rendered in the map.

Maybe I'm overthinking it ;)

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added


src/components/chloropleth/legenda/hooks/useSafetyRegionLegendaData.ts
```diff
@@ -1,10 +1,15 @@
...
 
 import { useLegendaItems } from './useLegendaItems';
-import { thresholds } from '~/components/chloropleth/SafetyRegionChloropleth';
+import { regionThresholds } from '~/components/chloropleth/regionThresholds';
 
-export function useSafetyRegionLegendaData(metric: TRegionMetricName) {
-  const ths = thresholds[metric];
+export function useSafetyRegionLegendaData(
+  metric: TRegionMetricName,
+  metricPropertyName?: TRegionsNursingHomeMetricName
```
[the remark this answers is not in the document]
VWSCoronaDashboard4: either have everything in an object like this, though I'd not be much in favor of adding lodash on top of it

Having an array of these, and then a straighforward method provided to filter it down to one to get to the right value would work too. It's confusing and tricky to read things like `(regionThresholds[metric] as any)[metricPropertyName]`, it feels like a way to circumvent typescript from doing its thing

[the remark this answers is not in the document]
VWSCoronaDashboard8: I'm not sure I follow the point about `as any`. I agree that this should be avoided at all cost, but grabbing some data out of an object with lodash.get() or something similar you can still type the response Threshold | undefined and then do runtime assertions to verify that the response is not undefined.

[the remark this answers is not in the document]
VWSCoronaDashboard1: I've refactored those bits into a separate function that doesn't have all the fiddly any casts, so I think its more readable than what it was.

package.json
```diff
@@ -16,6 +16,7 @@
     "highcharts": "^8.2.0",
     "highcharts-react-official": "^3.0.0",
     "jsdom": "^16.4.0",
+    "lodash.get": "^4.4.2",
```
VWSCoronaDashboard8, line 19: I don't think it is necessary to be this specific with lodash dependencies. You can just use lodash or (if you want to be safe) lodash-es. And then use `import { get } from "lodash"`. Tree shaking should make sure it doesn't bloat the bundles.

src/components/chloropleth/legenda/hooks/useSafetyRegionLegendaData.ts
```diff
@@ -1,15 +1,34 @@
...
+import {
+  ChoroplethThresholds,
+  TRegionMetricName,
+  TRegionsNursingHomeMetricName,
+} from '../../shared';
 
 import { useLegendaItems } from './useLegendaItems';
 import { regionThresholds } from '~/components/chloropleth/regionThresholds';
 
+import get from 'lodash.get';
```
VWSCoronaDashboard8, line 10: Lodash has more useful functions, like omit and pick and chain. I use it quite regularly, and it seems to be already a dependency of one of our dependencies. If I do `import {get} from "lodash"` the dependency and its types are already found by the compiler in our node_modules folder.

src/components/chloropleth/legenda/hooks/useSafetyRegionLegendaData.ts
```diff
@@ -1,15 +1,34 @@
...
+
+export function getSelectedThreshold(
+  metricName?: TRegionMetricName,
+  metricPropertyName?: string
+) {
+  if (!metricName) {
+    return undefined;
```
VWSCoronaDashboard8, line 17: nitpick; just `return` is already returning undefined. Maybe you like this style for clarity, but I am not used to doing this.

src/components/chloropleth/legenda/hooks/useSafetyRegionLegendaData.ts
```diff
@@ -1,15 +1,34 @@
...
+} from '../../shared';
 
 import { useLegendaItems } from './useLegendaItems';
 import { regionThresholds } from '~/components/chloropleth/regionThresholds';
 
+import get from 'lodash.get';
+
+export function getSelectedThreshold(
+  metricName?: TRegionMetricName,
+  metricPropertyName?: string
```
VWSCoronaDashboard8, line 14: You might as well combine these two parameters into something like `metricPropertyPath` I find the combination name and propertyName a bit confusing to start with, because both refer to a property in the json structure.

You will lose the TRegionMetricName information but I doubt it's very valuable anyway. See the comment below.

src/components/chloropleth/legenda/hooks/useSafetyRegionLegendaData.ts
```diff
@@ -1,15 +1,34 @@
...
+  if (!metricName) {
+    return undefined;
+  }
+
+  const thresholdInfo = metricPropertyName
+    ? get(regionThresholds, `${metricName}.${metricPropertyName}`) ??
+      regionThresholds[metricName]
+    : regionThresholds[metricName];
+
+  return thresholdInfo as ChoroplethThresholds;
```
VWSCoronaDashboard8, line 25:
```suggestion
  const thresholdInfo = get(regionThresholds, metricProperyPath) 
  assert(thresholdInfo, `Failed to locate threshold info at metric property path ${metricPropertyPath}`)
  
  return thresholdInfo as ChoroplethThresholds;
```

VWSCoronaDashboard8 (approved):
Approved with some suggestions

src/components/chloropleth/legenda/hooks/useSafetyRegionLegendaData.ts
VWSCoronaDashboard1: Yea, its just that it doesn't work that way ;) Its entirely possible that a valid metricProperty name is passed, but no specific thresholds exist for that subkey. In that case regionThresholds[metricName] needs to be the fallback

VWSCoronaDashboard8: Ah bummer. Well, maybe you can add a comment there about the fallback mechanism then.

src/components/chloropleth/legenda/hooks/useSafetyRegionLegendaData.ts
VWSCoronaDashboard1: I can rename it to metricValueName. The split is logical since the Nursinghome metric has several different values.
If I send the name as a period delimited path it'll mean I have to stitch it together conditionally at every place where this function is called. Since the second parameter is optional.

package.json
VWSCoronaDashboard1: Weird, I only saw the lodash typings in the package.json, so I figured the actual lib wasn't added yet

VWSCoronaDashboard1: Fixed

PR #553, opened 2020-10-08
Sort time-series data

VWSCoronaDashboard8:
In this PR all time-series data gets sorted in the page build/bundle stage. 

First, all properties are identified that contain a { values: [] } object. Then for each of those time-series, the used timestamp is detected (there are currently 3). If no known timestamp is found, the code will throw and the application build will break. This way we can be sure that all data is sorted.

There was some additional runtime sorting happening in the code, which I have removed since it should be redundant.

### DIFF
--- a/src/static-props/data-sorting.ts
+++ b/src/static-props/data-sorting.ts
@@ -1,0 +1,161 @@
+import { Municipal, National, Regionaal } from '~/types/data.d';
+
+export function sortNationalTimeSeriesInDataInPlace(data: National) {
+  const timeSeriesPropertyNames = getNationalTimeSeriesPropertyNames(data);
+
+  for (const propertyName of timeSeriesPropertyNames) {
+    const timeSeries = data[propertyName] as TimeSeriesData<Timestamped>;
+    (data[propertyName] as TimeSeriesData<
+      Timestamped
+    >).values = sortTimeSeriesValues(timeSeries.values);
+  }
+}
+
+export function sortRegionalTimeSeriesInDataInPlace(data: Regionaal) {
+  const timeSeriesPropertyNames = getRegionalTimeSeriesPropertyNames(data);
+
+  for (const propertyName of timeSeriesPropertyNames) {
+    /**
+     * There is one property in the dataset that contains timeseries nested
+     * inside values, so we need to process that separately.
+     */
+    if (propertyName === 'results_per_sewer_installation_per_region') {
+      const nestedSeries = data[propertyName] as RegionalSewerTimeSeriesData<
+        Timestamped
+      >;
+
+      nestedSeries.values = nestedSeries.values.map((x) => {
+        x.values = sortTimeSeriesValues(x.values);
+        return x;
+      });
+
+      // Skip the remainder of this loop
+      continue;
+    }
+
+    const timeSeries = data[propertyName] as TimeSeriesData<Timestamped>;
+
+    (data[propertyName] as TimeSeriesData<
+      Timestamped
+    >).values = sortTimeSeriesValues(timeSeries.values);
+  }
+}
+
+export function sortMunicipalTimeSeriesInDataInPlace(data: Municipal) {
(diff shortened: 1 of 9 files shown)


src/static-props/nl-data.ts
```diff
@@ -43,6 +43,8 @@ export default function getNlData(): () => IProps {
 
     const lastGenerated = data.last_generated;
 
+    sortNationalTimeSeriesInDataInPlace(data);
```
VWSCoronaDashboard4, line 46: Good improvement to have this calculated in the static props!

src/static-props/data-sorting.ts
```diff
@@ -0,0 +1,179 @@
+import { Municipal, National, Regionaal } from '~/types/data.d';
+
+export function sortNationalTimeSeriesInDataInPlace(data: National) {
```
VWSCoronaDashboard4, line 3: Overall this file is good to follow. There are some concerns that might better be split into several files. At the bottom there are some generic utilities, that could be moved into its own helper/util file. For national, regional, municipal: considered splitting them into their own file, eg `data-sorting-municipal.ts`

src/static-props/data-sorting.ts
```diff
@@ -0,0 +1,179 @@
...
+interface TimeSeriesData<T> {
+  values: T[];
+}
+
+interface RegionalSewerTimeSeriesData<T> {
+  values: TimeSeriesData<T>[];
+}
+function isTimeSeries(
+  value: unknown | TimeSeriesData<Timestamped>
+): value is TimeSeriesData<Timestamped> {
```
VWSCoronaDashboard4, line 154: TIL about specifying return types like this

VWSCoronaDashboard4 (commented):
Good improvement!

src/static-props/data-sorting.ts
VWSCoronaDashboard8: These functions are only used in this file, so I wouldn't classify them as utilities. For me, utilities are very non-specific by nature and should be used in multiple locations, otherwise, you're just dislocating them from their specific context.

Splitting things to different files doesn't feel like an improvement to me.

src/static-props/data-sorting.ts
VWSCoronaDashboard8: What is TIL and what are you suggesting? Return types like this is how you define type guards.

VWSCoronaDashboard4 (approved):

src/static-props/data-sorting.ts
VWSCoronaDashboard4: This was meant as compliment :-) TIL = today I learned, as I didn't know this specific TS feature.

src/static-props/municipality-data.ts
```diff
@@ -54,7 +55,9 @@ export function getMunicipalityData() {
 
     const filePath = path.join(process.cwd(), 'public', 'json', `${code}.json`);
     const fileContents = fs.readFileSync(filePath, 'utf8');
-    const data = JSON.parse(fileContents);
+    const data = JSON.parse(fileContents) as Municipal;
```
VWSCoronaDashboard1, line 58: nitpick: Why doesn't the Typescript global definitions STILL not have a generic argument for JSON.parse?
JSON.parse<Municipal> just reads so much nicer ;)

src/static-props/data-sorting.ts
VWSCoronaDashboard1: TIL === Today I Learned

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard1:
Only general comment I have is that perhaps the data sorting methods could be unit tested?

VWSCoronaDashboard8:
@VWSCoronaDashboard1 I agree, but given the current time pressure, I didn't want to do this now. Also, there are only two things that can really go wrong; missing a time series property (and it's all or nothing because they use the same function to detect them) and a mismatch in timestamp definition. That last one throws and breaks the build, so there is not so much to test really.

PR #556, opened 2020-10-08
Fix padding for region control

VWSCoronaDashboard4:
## Summary

Fixes paddings on region select

The removed grid properties did not have any effect, elsewhere the display is set to flex.

### DIFF
--- a/src/components/chartRegionControls/chartregioncontrols.module.scss
+++ b/src/components/chartRegionControls/chartregioncontrols.module.scss
@@ -5,7 +5,12 @@
   margin: 0 0 1rem;
   max-width: 100%;
   overflow: hidden;
-  display: grid;
-  grid-template-columns: 1fr 1fr;
-  grid-template-rows: auto;
+
+  /* Extra specificity to overrule radio-group rules */
+  &.select-region-group {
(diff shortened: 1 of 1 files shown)


src/components/chartRegionControls/chartregioncontrols.module.scss
```diff
@@ -5,7 +5,12 @@
   margin: 0 0 1rem;
   max-width: 100%;
   overflow: hidden;
-  display: grid;
-  grid-template-columns: 1fr 1fr;
-  grid-template-rows: auto;
```
VWSCoronaDashboard8, line 10: By fixing the padding you also altered the design. So this PR / title is a bit misleading. Why did you remove the grid layout? It made the buttons equal size.

VWSCoronaDashboard4: The constrained width in this case was giving troubles. Text in the labels ran too far to the borders: hence why the paddings were wrong.

After discussing with design, the same layout solution as the chart time controls was wanted.

VWSCoronaDashboard8 (approved):

PR #557, opened 2020-10-08
implement data warning component

VWSCoronaDashboard5:
## Summary

From tuesday on, the dashboard will use a different data-set for the hospital admissions, this data might have a slight delay compared to the data we get from RIVM. This PR adds a warning message to update the user on this data-delay. It needs to be implemented on the following pages:
- [x] Landelijk Ziekenhuis-opnames 
- [x] Regionaal Ziekenhuis-opnames 
- [x] Gemeentelijk Ziekenhuis-opnames 

Furthermore this PR moves the chloropleth map to the bottom of the page, changin the order of the components in the layout.

### Governance

- [x] Documentation is added


src/components/dataWarning/dataWarning.module.scss
```diff
@@ -0,0 +1,14 @@
+@import '~/scss/variables.scss';
+
+.dataWarning {
+  display: grid;
+  grid-template-columns: 25px 1fr;
+  grid-gap: 12px;
+  margin-bottom: 24px;
+
+  .dataWarningText {
```
VWSCoronaDashboard8, line 9: Since it's an isolated module you can keep the CSS names basic. Like .root and .text. Also no need to nest the text I suspect.

src/components/dataWarning/dataWarning.module.scss
```diff
@@ -0,0 +1,14 @@
+@import '~/scss/variables.scss';
+
+.dataWarning {
+  display: grid;
+  grid-template-columns: 25px 1fr;
+  grid-gap: 12px;
+  margin-bottom: 24px;
```
VWSCoronaDashboard8, line 7: Ideally, a component knows nothing about its placement in the context so it becomes really reusable. The margin-bottom on the container would be set by the parent component which handles the placement of its children. 

By setting the margin here, you are making assumptions on how and where the component is used. This is not a good "separation of concerns". This mistake has been made everywhere in the codebase, so I don't mind if you don't solve this now. I think we should tackle this when we start to work on layout components.

VWSCoronaDashboard8 (approved):

src/components/dataWarning/dataWarning.module.scss
```diff
@@ -0,0 +1,14 @@
+@import '~/scss/variables.scss';
+
+.dataWarning {
+  display: grid;
+  grid-template-columns: 25px 1fr;
```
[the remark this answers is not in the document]
VWSCoronaDashboard4: I'd agree that grids are overkill in this case. Flexbox is used elsewhere for similar situations.

src/components/dataWarning/dataWarning.module.scss
VWSCoronaDashboard5: I completely understand your point and I think it is very good feedback. Let's discuss a refactor later, for now I will keep the margin-bottom in the component styles, since applying margins from the layout would take a while for now.

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard4 (approved):

VWSCoronaDashboard1 (approved):

src/scss/layout.scss
```diff
@@ -474,13 +487,18 @@ a.metric-link {
   @media (min-width: $chloropleth-breakpoint) {
     display: grid;
     grid-template-columns: 1fr 1fr;
-    grid-template-rows: auto 1fr auto;
+    grid-template-rows: auto auto 1fr auto;
     grid-template-areas:
+      'w w'
```
VWSCoronaDashboard5, line 492: this `w` is supposed to resemble the grid-area for the "warning", let me know if a more elegant solution is required

PR #559, opened 2020-10-09
fixes IE11 alignment of metadata footers

VWSCoronaDashboard4:
## Summary

Fixes the alignment of metadata footers in IE11.

## Detailed design

* The flexbox use in `article-top` goes funky in IE11 because it does not understand the shorthand. Specifying it in full works.
* The choropleth did not specify where the metadata should be placed. Extended the grid and put it in place.

### DIFF
--- a/src/scss/layout.scss
+++ b/src/scss/layout.scss
@@ -96,7 +96,7 @@
   is pushed to the bottom.
   */
   .article-top {
-    flex: 1;
+    flex: 1 1 auto;
   }
 
   .article-footer {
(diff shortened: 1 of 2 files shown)


src/scss/layout.scss
```diff
@@ -474,10 +474,11 @@ a.metric-link {
   @media (min-width: $chloropleth-breakpoint) {
     display: grid;
     grid-template-columns: 1fr 1fr;
-    grid-template-rows: auto 1fr;
+    grid-template-rows: auto 1fr auto;
```
VWSCoronaDashboard8, line 477: I'm surprised this made a difference. If set to "auto 1fr" I would expect it already uses auto again for the 3rd row.

VWSCoronaDashboard8 (approved):

PR #580, opened 2020-10-12
Fix region input blur on select

VWSCoronaDashboard8:
Fixes a problem where blur event did not work on region search input. This is especially annoying on mobile, because it keeps the keyboard view open.

In addition, I have cleaned up some other things that came on my path.

### DIFF
--- a/src/utils/useThrottle.ts
+++ b/src/utils/useThrottle.ts
@@ -1,19 +1,19 @@
 import { useState, useEffect, useRef } from 'react';
 
-export function useThrottle<T>(value: T, intervalMs: number): T {
+export function useThrottle<T>(value: T, delayMs: number): T {
   const [throttledValue, setThrottledValue] = useState<T>(value);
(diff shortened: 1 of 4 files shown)


VWSCoronaDashboard1 (approved):

src/utils/useThrottle.ts
```diff
@@ -6,20 +6,20 @@ import { useState, useEffect, useRef } from 'react';
 // ```
 // const throttledTerm = useThrottle<string>(term, 100);
 // ```
-export function useThrottle<Value>(value: Value, limit: number): Value {
-  const [throttledValue, setThrottledValue] = useState<Value>(value);
+export function useThrottle<T>(value: T, intervalMs: number): T {
```
VWSCoronaDashboard4, line 9: Consider renaming `limit`/`intervalMs` to `delay` instead.

VWSCoronaDashboard4 (approved):

PR #584, opened 2020-10-13
Bugfix/percentage nursing home choropleth

VWSCoronaDashboard4:
## Summary

Adjusts choropleth on besmette verpleeghuizen to use percentage data instead of total.

### DIFF
--- a/src/components/chloropleth/regionThresholds.ts
+++ b/src/components/chloropleth/regionThresholds.ts
@@ -109,7 +109,7 @@ const nursingHomeThresholds: ChoroplethThresholds = {
   ],
 };
 
-const nursingHomeInfectedThresholds: ChoroplethThresholds = {
+const nursingHomeInfectedLocationsPercentageThresholds: ChoroplethThresholds = {
   thresholds: [
     {
       color: '#c0e8fc',
@@ -143,7 +143,7 @@ export const regionThresholds: Record<
   hospital_admissions: hospitalAdmissionsThresholds,
   escalation_levels: escalationThresholds,
   nursing_home: {
-    infected_locations_percentage: nursingHomeInfectedThresholds,
+    infected_locations_percentage: nursingHomeInfectedLocationsPercentageThresholds,
     newly_infected_people: nursingHomeThresholds,
   } as Record<Partial<TRegionsNursingHomeMetricName>, ChoroplethThresholds>,
 };
--- a/src/components/chloropleth/regionThresholds.ts
+++ b/src/components/chloropleth/regionThresholds.ts
@@ -109,6 +109,31 @@ const nursingHomeThresholds: ChoroplethThresholds = {
   ],
 };
 
+const nursingHomeInfectedThresholds: ChoroplethThresholds = {
+  thresholds: [
+    {
+      color: '#c0e8fc',
+      threshold: 0,
+    },
+    {
+      color: '#87cbf8',
+      threshold: 10,
+    },
+    {
+      color: '#5dafe4',
+      threshold: 20,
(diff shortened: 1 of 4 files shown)


src/components/chloropleth/tooltips/region/createInfectedLocationsRegionalTooltip.tsx
```diff
@@ -21,7 +21,9 @@ export const createInfectedLocationsRegionalTooltip = (router: NextRouter) => (
       <TooltipContent title={context.vrname} onSelect={onSelect}>
         <strong>
           {`${
-            context.value !== undefined ? context.value : '-'
+            context.infected_locations_total !== undefined
+              ? context.infected_locations_total
+              : '-'
```
VWSCoronaDashboard5, line 26: couldn't this be written as `context?.infected_locations_total ? context.infected_locations_total : '-'`

src/components/chloropleth/regionThresholds.ts
```diff
@@ -118,7 +143,7 @@ export const regionThresholds: Record<
   hospital_admissions: hospitalAdmissionsThresholds,
   escalation_levels: escalationThresholds,
   nursing_home: {
-    infected_locations_total: nursingHomeThresholds,
+    infected_locations_percentage: nursingHomeInfectedThresholds,
```
VWSCoronaDashboard8, line 146: There are more types of "infected" data in nursing homes, no?  Maybe call it specifically `nursingHomeInfectedLocationsPercentageThresholds`. It's a long name. Ideally, these thresholds would be defined grouped with the chart where they are used because then you don't have this problem, but that's for another time.

src/components/chloropleth/tooltips/region/createInfectedLocationsRegionalTooltip.tsx
VWSCoronaDashboard8, line 26: This ternary thing with `-` fallback is what formatNumber does I think, so it would make sense to just use that.

VWSCoronaDashboard8 (changes requested):

src/components/chloropleth/tooltips/region/createInfectedLocationsRegionalTooltip.tsx
VWSCoronaDashboard8: You can write it `context?.infected_locations_total ?? '-'` even :)

VWSCoronaDashboard4: `?.` checks if context is undefined, not if infected_locations_total is undefined. But I'll fix it with the suggested numberformatter way :-)

src/components/chloropleth/tooltips/region/createInfectedLocationsRegionalTooltip.tsx
```diff
@@ -20,9 +20,9 @@ export const createInfectedLocationsRegionalTooltip = (router: NextRouter) => (
     context && (
       <TooltipContent title={context.vrname} onSelect={onSelect}>
         <strong>
-          {`${
-            context.value !== undefined ? context.value : '-'
-          } (${formatNumber(context.infected_locations_percentage)}%)`}
+          {`${formatNumber(
+            context.infected_locations_percentage
+          )}% (${formatNumber(context.infected_locations_total)})`}
```
VWSCoronaDashboard8, line 25: Are you sure the percentage should not be rendered _after_ the total like in other places?

VWSCoronaDashboard8 (approved):

PR #602, opened 2020-10-13
Adds descriptions for hospital linecharts

VWSCoronaDashboard4:
## Summary

Adds lokalize keys for hospital linecharts on all 3 levels.


src/locale/en.json
```diff
@@ -111,6 +111,7 @@
     "map_titel": "Distribution of hospital admissions in the Netherlands",
     "map_toelichting": "These figures show the daily hospital admissions in the Netherlands.",
     "linechart_titel": "Progression in hospital admissions over time",
+    "linechart_description": "",
```
VWSCoronaDashboard4, line 114: The linechart component handles these empty ones well; they will/can be filled in with lokalize from now.

PR #631, opened 2020-10-14
remove blue bars and implement metric components (sidebar and pages)

VWSCoronaDashboard5:
## Summary

This PR replaces the `BarScale` components used in both the sidebar and on individual pages (`National`, `Regional` and `Municipal`) and replaces them with a new `Metric` component. In the sidebar factory methods for each `Metric` component are used, similar to the previous setup where factory methods were used for each `BarScale`. Each factory method makes sure the right data and text is passed on to the `MetricKPI` component. In the sidebar each factory method passes the `last_value` of each corresponding data entry to the component.

The `Metric` component (called `MetricKPI`) is a simple component that takes a `value`, where passing a custom `formatter` is optional. 

Furthermore, the individual pages where the blue `BarScale` component was formerly used, a new two-column layout is implemented, rendering the `last_value` object in the `column-item`

## Unresolved questions

I made new files for each factory method of the `Metric` components, do we still want to keep the `BarScale` factory methods for each level? (`National`, `Regional` and `Municipal`)

### Governance

- [x] Documentation is added


src/components/layout/MunicipalityLayout.tsx
```diff
@@ -224,9 +222,8 @@ function MunicipalityLayout(props: WithChildren<IMunicipalityData>) {
                           }
                         />
                         <span>
-                          <SewerWaterBarScale
-                            data={getSewerWaterBarScaleData(data)}
-                            showAxis={true}
+                          <SewerWaterMetric
+                            data={getSewerWaterBarScaleData(data) ?? undefined}
```
VWSCoronaDashboard8, line 226: What is `?? undefined ` for? If the getData function returns data or undefined you can just pass it like that.

src/components/metricKPI/index.tsx
```diff
@@ -0,0 +1,22 @@
+import styles from './styles.module.scss';
+
+type IProps = {
+  label?: string;
```
VWSCoronaDashboard8, line 4: I think title is a better name, looking at how you used it in the other components

src/components/metricKPI/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+  value: number | null | undefined;
+  format: any | false;
```
VWSCoronaDashboard8, line 6: Why false and not just define it as an optional property?

VWSCoronaDashboard8 (commented):
There's a lot of ? and undefined checks on data that I think should just be there, but maybe it's a result of wrong types generated on json schemes that didn't set properties as required.

I suspect a lot of them can be cleared out though.

src/components/metricKPI/index.tsx
VWSCoronaDashboard8: Also the function signature for "format" should be clear, instead of using any. for example `(number | string) => string`

src/pages/veiligheidsregio/[code]/rioolwater.tsx
```diff
@@ -66,7 +66,11 @@ const SewerWater: FCWithLayout<ISafetyRegionData> = (props) => {
         <div className="column-item column-item-extra-margin">
           <h3>{text.barscale_titel}</h3>
 
-          <SewerWaterBarScale data={barScaleData} showAxis={true} />
+          <h3>
+            <span className="text-blue kpi" data-cy="infected_daily_total">
+              {formatNumber(barScaleData?.value)}
```
Comment (VWSCoronaDashboard5) [inline src/pages/veiligheidsregio/[code]/rioolwater.tsx:71]: passing it its data through the memoized `barScaleData` (from `getSewerWaterBarScaleData`) does not make sense, i'll change this right away.

Comment (VWSCoronaDashboard5) [reply src/pages/veiligheidsregio/[code]/rioolwater.tsx:71]: however, i think it is better to change this in a separate PR to avoid mixing concerns

src/components/common/nursing-home-infected-locations-metric.tsx
```diff
@@ -0,0 +1,35 @@
...
+  const description = data?.date_of_insertion_unix
+    ? replaceVariablesInText(text.dateOfReport, {
+        dateOfReport: formatDateFromSeconds(
+          data.date_of_insertion_unix,
+          'relative'
+        ),
+      })
+    : undefined;
+
+  if (!data) return null;
```
VWSCoronaDashboard8, line 25: nitpick; it would make more sense to do this check first, before you start using the data. That applies to all metric components below as well.

Then I think you should be able to assume that data.date_of_insertion_unix is always available, if the type definitions are correct, so you don't need the ternary `? :`

src/components/metricKPI/index.tsx
```diff
@@ -0,0 +1,22 @@
+import styles from './styles.module.scss';
+
+type IProps = {
+  title?: string;
+  value: number | undefined;
```
VWSCoronaDashboard8, line 5:
```suggestion
  value?: number;
```

src/pages/landelijk/verpleeghuis-positief-geteste-personen.tsx
```diff
@@ -48,10 +48,11 @@ const NursingHomeInfectedPeople: FCWithLayout<INationalData> = (props) => {
         <div className="column-item column-item-extra-margin">
           <h3>{text.barscale_titel}</h3>
 
-          <NursingHomeInfectedPeopleBarScale
-            value={data.last_value.newly_infected_people}
-            showAxis={true}
-          />
+          <h3>
```
VWSCoronaDashboard8, line 51: I don't think semantically this should be a h3 element. There is already one above it.

src/pages/landelijk/verpleeghuis-sterfte.tsx
```diff
@@ -36,10 +36,11 @@ const NursingHomeDeaths: FCWithLayout<INationalData> = (props) => {
         <div className="column-item column-item-extra-margin">
           <h3>{text.barscale_titel}</h3>
 
-          <NursingHomeDeathsBarScale
-            value={data.last_value.deceased_daily}
-            showAxis={true}
-          />
+          <h3>
```
VWSCoronaDashboard8, line 39: Same comment about h3

src/pages/veiligheidsregio/[code]/positief-geteste-mensen.tsx
```diff
@@ -83,19 +83,17 @@ const PostivelyTestedPeople: FCWithLayout<ISafetyRegionData> = (props) => {
...
-                  Math.round(
-                    resultsPerRegion.last_value
-                      .total_reported_increase_per_region
-                  )
-                )}
-              </span>
-            </h3>
-          )}
+          <h3>{text.kpi_titel}</h3>
+          <h3>
```
Comment (VWSCoronaDashboard8) [inline src/pages/veiligheidsregio/[code]/positief-geteste-mensen.tsx:87]: etc

src/pages/veiligheidsregio/[code]/rioolwater.tsx
Comment (VWSCoronaDashboard8) [reply src/pages/veiligheidsregio/[code]/rioolwater.tsx:71]: h3 again

src/pages/veiligheidsregio/[code]/verpleeghuis-besmette-locaties.tsx
```diff
@@ -62,35 +61,22 @@ const NursingHomeInfectedLocations: FCWithLayout<ISafetyRegionData> = (
...
-            rangeKey="total_new_reported_locations"
-            gradient={[
-              {
-                color: '#3391CC',
-                value: 0,
-              },
-            ]}
-            showAxis={true}
-          />
+          <h3>
```
Comment (VWSCoronaDashboard8) [inline src/pages/veiligheidsregio/[code]/verpleeghuis-besmette-locaties.tsx:64]: I think you get my point ;)

VWSCoronaDashboard8 (changes requested):

src/components/metricKPI/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+  format?: (value?: number) => string;
+  description?: string;
+};
+
+export function MetricKPI(props: IProps) {
+  const { value, format, title, description } = props;
+
+  return (
+    <div className={styles.root}>
+      <p className={styles.title}>{title}</p>
```
VWSCoronaDashboard, line 15: Semantically this might need to be a smaller heading.

src/components/metricKPI/styles.module.scss
```diff
@@ -0,0 +1,28 @@
...
+  margin: 0;
+}
+
+.value {
+  font-size: 1.25em;
+  font-weight: bold;
+}
+
+.description {
+  margin-left: 1em !important;
```
VWSCoronaDashboard, line 26: This is why you should not do `.wrapper > p` and always target classnames. It prevents specificity issues.

PR #646, opened 2020-10-16
Add styled system and various component abstractions

VWSCoronaDashboard8:


src/components-styled/two-kpi-section.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export function TwoKpiSection({ children }: TwoKpiSectionProps) {
+  const childrenCount = React.Children.count(children);
+
+  assert(
+    childrenCount < 3,
+    `TwoKpiSection can only have 1 or 2 children, but received ${childrenCount}`
+  );
+
+  const childrenArray = React.Children.toArray(children);
```
VWSCoronaDashboard, line 17: The array is a bit clunky, but it works. We could also think about using prop slots, but I don't really have a strong opinion on this. Especially since it's implementation details and not part of a public API.

src/components-styled/two-kpi-section.tsx
```diff
@@ -0,0 +1,28 @@
...
+  children: React.ReactNode;
+}
+
+export function TwoKpiSection({ children }: TwoKpiSectionProps) {
+  const childrenCount = React.Children.count(children);
+
+  assert(
+    childrenCount < 3,
+    `TwoKpiSection can only have 1 or 2 children, but received ${childrenCount}`
+  );
```
VWSCoronaDashboard, line 15: Nice!
Would TS not be able to guard against this too? Or is this because children is too opaque for TS?

src/style/theme.ts
```diff
@@ -0,0 +1,35 @@
...
+const breakpoints = ['40em', '52em', '64em', '80em'];
+
+const mediaQueries = {
+  sm: `@media screen and (min-width: ${breakpoints[0]})`,
+  md: `@media screen and (min-width: ${breakpoints[1]})`,
+  lg: `@media screen and (min-width: ${breakpoints[2]})`,
+  xl: `@media screen and (min-width: ${breakpoints[3]})`,
+};
+
+const space = [0, 4, 8, 16, 32, 64, 128, 256, 512];
```
VWSCoronaDashboard, line 12: These are pixels right? 👍

VWSCoronaDashboard8: Yes, but I was copy-pasted. I was planning to convert them to em still. If someone changes the browser font-size I would expect the whole layout to move with it.

VWSCoronaDashboard: I think pixels or rems are actually preferred, so nothing wrong with this!

src/style/theme.ts
```diff
@@ -0,0 +1,101 @@
...
+ * 1200 = 75em
+ * 1000 ~ 64em
+ * 960 = 60em
+ * 992 = 62em
+ * 768 = 48em
+ * 700 ~ 44em
+ * 420 ~ 26em
+ */
+
+interface Breakpoints extends Array<string> {
```
VWSCoronaDashboard4, line 48: Maybe we can bring these down to 3 even, once we get the hang of it.

PR #655, opened 2020-10-18
feat: Optional subtitle to center icon with multiple headings

nickspaargaren:
TitleWithIcon


VWSCoronaDashboard (approved):
Seems like this improves both API and alignment. Thanks Nick!

PR #657, opened 2020-10-19
Fix SSR size of barchart

VWSCoronaDashboard1:
## Summary

The SSR size of the infected people barchart tile caused the barchart to initially render with a wrong size that displayed faulty axis values. This PR fixes that.

## Motivation

Bug report

## Detailed design

Having the flex shrink and grow options set to true caused the SSR problem. A new class was added called column-item-fixed that defines a column with a fixed width of 50%. This fixes the SSR issue.

## Drawbacks

n/a

## Alternatives

Let's hear it ;)

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added

### DIFF
--- a/src/scss/layout.scss
+++ b/src/scss/layout.scss
@@ -432,6 +432,12 @@ a.metric-link {
       margin: 0;
     }
 
+    .column-item-fixed {
+      flex: 0 0 50%;
+      display: flex;
+      flex-direction: column;
+    }
+
     .column-item {
       flex: 1 1 50%;
       display: flex;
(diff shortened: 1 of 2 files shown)


src/scss/layout.scss
```diff
@@ -432,6 +432,12 @@ a.metric-link {
       margin: 0;
     }
 
+    .column-item-fixed {
```
VWSCoronaDashboard8, line 435: Please add a comment describing the problem that caused you to find this solution. When we create layout components or in general for future reference we need to be aware of exactly why this was a fix and for what.

VWSCoronaDashboard8 (approved):

src/scss/layout.scss
VWSCoronaDashboard8: Not sure if the comment is best placed in the css or the component, or maybe some combination.

PR #658, opened 2020-10-19
Collapsable sections for /over and /verantwoording

VWSCoronaDashboard4:
## Summary

Long sections in `/over` and `/verantwoording` are now collapsed by default, and can be opened to read the text.

## Detailed design

After a trial run with HTML5's `<details>`, I found out it will not support animations to the degree we want to use :-(

I've used `@reach/disclosure` to get accessibility needs out of the box. For our animations, I needed to undo the `display: none`, causing links to become tab-able. I've added a method that enables/disables these links on open/close.


public/images/chevron-down.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="18" height="11" viewBox="0 0 18 11" fill="none" xmlns="http://www.w3.org/2000/svg">
```
VWSCoronaDashboard4, line 1: I've copied this one from @VWSCoronaDashboard1 's PR.

public/images/chevron-down.svg
```diff
@@ -0,0 +1,3 @@
...
+<path transform="rotate(90, 9, 9)" d="M1 17L9 9L1 1" stroke="#01689B" stroke-width="2" stroke-linecap="round"/>
```
VWSCoronaDashboard, line 2: if you need it, you can make stroke `currentColor` so you can style the color, hover:color with CSS

VWSCoronaDashboard1 (approved):

src/components/collapsable/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+  }, [open, panelReference]);
+
+  useEffect(() => {
+    setLinkTabability();
+    setContentHeight();
+    checkLocationHash();
+  }, [checkLocationHash, setLinkTabability, setContentHeight]);
+
+  const router = useRouter();
+  router.events?.on('hashChangeComplete', checkLocationHash);
```
VWSCoronaDashboard8, line 70: Does this maybe needs to be unsubscribed when the component unmounts? I guess you could call on() in the useEffect and if there is an `off` method you can call that in its cleanup method.

src/pages/verantwoording.tsx
```diff
@@ -69,19 +75,16 @@ const Verantwoording: FCWithLayout<{ text: any; lastGenerated: string }> = (
...
-                    <Fragment key={item.cijfer}>
-                      <h3>{item.cijfer}</h3>
-                      <div
-                        dangerouslySetInnerHTML={{
-                          __html: item.verantwoording,
-                        }}
-                      />
-                    </Fragment>
-                  )
+              {text.verantwoording.cijfers.map((item: ICijfer) =>
```
VWSCoronaDashboard8, line 78: Nitpick; ICijfer should already be inferred by TS from the cijfers content, so you can leave it out.

Unless the types are not the same in which case the type of cijfers should be fixed I guess.

src/utils/skipLinks.ts
```diff
@@ -0,0 +1,19 @@
...
+  return source.toLowerCase().replace(/[^a-zA-Z0-9]+/g, '-');
+}
+
+/**
+ * Ensures unique IDs for the provided array of objects,
+ * by adding -1 or -2 to duplicate IDs.
+ */
+export function ensureUniqueSkipLinkIds(members: { id: string }[]): void {
+  members.forEach((member) => {
+    const catches = members.filter((mem) => mem.id === member.id);
```
VWSCoronaDashboard8, line 11: nitpick; for simple map / filter statements I usually use x as a value shorthand. In this case it also avoids the naming collision with member so you don't have to use the slightly confusing name `mem`

src/utils/skipLinks.ts
```diff
@@ -0,0 +1,19 @@
+export function getSkipLinkId(source: string): string {
+  return source.toLowerCase().replace(/[^a-zA-Z0-9]+/g, '-');
```
VWSCoronaDashboard8, line 2: There is a module called `slugify` that also replaces special characters to make sure things are URL safe. https://www.npmjs.com/package/@sindresorhus/slugify

nitpick; the replace pattern contains upper case letters, but your calling toLowerCase first :)

VWSCoronaDashboard8 (approved):
Nice one! Very clean 👍

src/components/collapsable/collapsable.module.scss
```diff
@@ -0,0 +1,81 @@
...
+  width: 100%;
+
+  /* our styling */
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0.75em;
+  color: $icon-color;
+  cursor: pointer;
+  font-size: 1.125em;
```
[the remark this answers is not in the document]
VWSCoronaDashboard4: not sure if I understand you correctly, what is wrong with `em`?

src/components/collapsable/index.tsx
VWSCoronaDashboard4: Good catch, I'll add a `.off`

VWSCoronaDashboard8 (approved):

PR #668, opened 2020-10-20
Add total_installation_count key to average_sewer_installation_per_region schema

VWSCoronaDashboard1:
## Summary

The installation count that was used to calculate the averages needs to be shown in the dashboard. We cannot simply use the installation count from the results_per_sewer_installation_per_region struct since apparently some logic is involved. So therefore this count will need to be added to the average_sewer_installation_per_region data.

## Motivation

Feature request

## Detailed design

n/a

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added

### DIFF
--- a/schema/regional/average_sewer_installation_per_region.json
+++ b/schema/regional/average_sewer_installation_per_region.json
@@ -9,6 +9,7 @@
         "week_end_unix",
         "vrcode",
         "average",
+        "total_installation_count",
         "date_of_insertion_unix"
       ],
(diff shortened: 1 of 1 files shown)


VWSCoronaDashboard8 (approved):
`installations_total` would be more in line with the newer property naming conventions (where I've used subject name postfixed with optional info about the data), but since this is part of an older schema I don't mind much.

PR #672, opened 2020-10-20
Fix compile type errors and overall clean up

VWSCoronaDashboard8:
There were some compilation errors. Not sure where they came from. Then I tried to improve the data typing but it snowballed a bit.

### DIFF
--- a/src/components/barScale/index.tsx
+++ b/src/components/barScale/index.tsx
@@ -42,10 +42,6 @@ export function BarScale({
 
   const text = siteText.common.barScale;
 
-  if (value === undefined) {
-    return null;
-  }
-
   const [xMin, xMax] = scale.domain();
 
   const textAlign = scaleThreshold()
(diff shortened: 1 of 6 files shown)


VWSCoronaDashboard4 (approved):

src/components/barScale/index.tsx
```diff
@@ -1,13 +1,11 @@
-import styles from './styles.module.scss';
+import { scaleQuantile, scaleThreshold } from 'd3-scale';
```
VWSCoronaDashboard4, line 1: Maybe we should sort the imports once in a big PR to avoid these less readable diffs.

VWSCoronaDashboard1 (approved):

src/components/barScale/index.tsx
VWSCoronaDashboard8: Yes, that was also a plan I had in mind. We could do one during a weekend if we close all big PRs. VSCode has an extension that can do a "organize imports" on all files in a directory.

PR #689, opened 2020-10-21
Linecharts in visx instead of HighCharts

VWSCoronaDashboard:
## Summary

This PR is meant as a PoC how we can replace the current `HighCharts` linechart components with a more lightweight, custom built solution with `visx`.

## Motivation

HighCharts adds an un-acceptable size to the bundle. This impacts performance heavily. It also doesn't work with static rendering and CSS-in-JS.

We're using `visx` for the maps as well, and besides some small performance issues we're quite happy with the API. 

The end-game here is to remove HighCharts from the bundle completely.

## Detailed design

I took [this example](https://airbnb.io/visx/threshold) from `visx` as starting point and starting hacking away until it resembled our well-known linechart.

## Drawbacks

HighCharts is feature complete and performant (talking about responsive tooltips) out of the box. We'll have to invest more in these topics ourselves, but the benefits will be worth it.

## Unresolved questions

- [ ] How do we make the tooltips more responsive/performant? There's a lot of wasted renders happening.
- [ ] The tooltips positioning is incorrect, should we update the bisector?
- [ ] Are we happy with the API design?


VWSCoronaDashboard7:
How accessible is the visx implementation?

VWSCoronaDashboard:
> How accessible is the visx implementation?

Since you have full control over the SVG rendering, just as accessible or better than HighCharts. But nothing comes for free. We should add it to the list of questions to resolve.

VWSCoronaDashboard:
Closed because there's a newer PR that does the same thing.

PR #693, opened 2020-10-21
Adjust opening of mobile menu to add a new route (location/history)

VWSCoronaDashboard4:
## Summary

Adjust opening of mobile menu to add a new route (location/history) so the back button shows the menu again.

## Detailed design

* Open menu now pushes a new route.
* The route state listens to changes of routes, and shows/hides the menu accordingly.
* Scroll behavior is adjusted to scroll based on menu routes (to the top of the menu) and any other routes (to the top of the page)

Feedback welcome :-)


src/components/layout/useMenuState.ts
```diff
@@ -36,19 +36,22 @@ export function useMenuState(defaultOpen = false): MenuState {
...
-    const handleCloseMenu = () => {
-      setMenuOpen(false);
-    };
-    Router.events.on('beforeHistoryChange', handleCloseMenu);
+    Router.events.on('beforeHistoryChange', setMenuFromRouterState);
+    Router.events.on('routeChangeComplete', setMenuFromRouterState);
     return () => {
-      Router.events.off('beforeHistoryChange', handleCloseMenu);
+      Router.events.off('beforeHistoryChange', setMenuFromRouterState);
+      Router.events.off('routeChangeComplete', setMenuFromRouterState);
```
VWSCoronaDashboard8, line 54: This comment is about the interface of the return type below.

To avoid confusion I like to always work in pairs of onEventX / handleEventX, where the "on" version is used to expose some handler to the outside (so on props), and "handle" is used internally to handle an event. The "handle" function can then call the "on" props function if needed. This makes it clearer when you need to do both local processing _and_ call a props callback (which is quite common).

Also I think you can then simplify the props funtions.

The handleMenuClick becomes onMenuClick, and you might not need to pass the event. Its odd that you define inside this component that it needs to do a blur() and then pass it to the outside. I think you can either:
1. Not pass it to the outside, because you just do that blurring internally and that's enough or
2. Let the caller pass in a handler where the calling context can decide what to do on `onMenuClick` (blur or not or something else), like you would normally do with a callback.
3. A combination of both. Where you define / call a handleMenuClick function locally that does the blurring, and then optionally call a onMenuClick handler if one is passed in.

Maybe this is better explained verbally :)

src/components/layout/useMenuState.ts
```diff
@@ -16,18 +16,18 @@ interface MenuState {
...
-
-    /* Ensure the top of the main navigation (or if not found: the sidebar) is in view */
-    requestAnimationFrame(() => {
-      window.document
-        .querySelector('#main-navigation,aside')
-        ?.scrollIntoView(true);
-    });
+
+    router.push(
+      router.pathname + '?menu',
```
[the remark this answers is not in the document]
VWSCoronaDashboard4: I've added a method that builds up the full URL, including query parameters.

VWSCoronaDashboard1 (approved):

src/components/layout/useMenuState.ts
```diff
@@ -16,45 +13,66 @@ interface MenuState {
 export function useMenuState(defaultOpen = false): MenuState {
   const [isMenuOpen, setMenuOpen] = useState<boolean>(defaultOpen);
 
+  const router = useRouter();
+
+  /*
+   * Retrieve the URL for the current route with menu state,
+   * including query params.
+   */
+  function getFullMenuUrl(): string {
```
VWSCoronaDashboard8, line 22: This function still feels a bit too specific and fragile, because it is tied to the use of specific URL query parameters. But if it works I'm ok with this for now.

src/components/layout/useMenuState.ts
```diff
@@ -16,45 +13,66 @@ interface MenuState {
...
+    let fullMenuUrl = `${router.pathname}?menu`;
+
+    if (router.query.code) {
+      fullMenuUrl = fullMenuUrl.replace('[code]', router.query.code as string);
+    }
+
+    const queryParams = Object.entries(router.query)
+      .filter(([key]) => ['menu', 'code'].indexOf(key) === -1)
```
VWSCoronaDashboard8, line 30: nitpick; I think this is more readable

```suggestion
      .filter(([key]) => ['menu', 'code'].includes(key))
```

src/components/layout/useMenuState.ts
```diff
@@ -16,45 +13,66 @@ interface MenuState {
...
+      .map(([key, value]) => {
+        return [
+          encodeURIComponent(key),
+          encodeURIComponent(value as string),
+        ].join('=');
+      })
+      .join('&');
```
VWSCoronaDashboard8, line 37: It looks like this function will strip any URL query parameter that is not menu or code. This might be problematic even though we might not use any other right now. 

Wouldn't it work if you just started with the `router.pathname` and then build the query string from router.query where you add the "menu" option to it, and then iterate over all elements to build the query string. Then any query parameter will still make it to the new URL. 

I don't see why opening the menu should have the side-effect of stripping all query parameters besides "code"

src/components/layout/useMenuState.ts
```diff
@@ -16,45 +13,66 @@ interface MenuState {
...
+
+    if (queryParams) {
+      fullMenuUrl += `&${queryParams}`;
+    }
+
```
VWSCoronaDashboard8, line 42: Here's a more generic function to build a query string from an object like router.query:

```
function makeQueryString(
  params: { [key: string]: string | number | boolean | undefined } = {},
) {
  const queryString = new URLSearchParams();

  Object.entries(params)
    .filter(([_, value]) => typeof value !== undefined)
    .forEach(([name, value]) => queryString.append(name, String(value)));

  return queryString.toString();
}
```

URL encode is already part of this API I think

VWSCoronaDashboard8 (changes requested):

src/components/layout/useMenuState.ts
VWSCoronaDashboard4: This code filters out `menu` and `code` to no duplicate them in the final URL. The loop merely adds all non menu and non-code query params at the end.

src/components/layout/useMenuState.ts
VWSCoronaDashboard4: I've looked into that. That would require a polyfill for URLSearchParams, while the method you post is making use of it while still looking fairly similar to what is happening in my code.

src/components/layout/useMenuState.ts
```diff
@@ -36,19 +36,22 @@ export function useMenuState(defaultOpen = false): MenuState {
...
-    };
-    Router.events.on('beforeHistoryChange', handleCloseMenu);
+    Router.events.on('beforeHistoryChange', setMenuFromRouterState);
+    Router.events.on('routeChangeComplete', setMenuFromRouterState);
     return () => {
-      Router.events.off('beforeHistoryChange', handleCloseMenu);
+      Router.events.off('beforeHistoryChange', setMenuFromRouterState);
+      Router.events.off('routeChangeComplete', setMenuFromRouterState);
     };
   }, []);
```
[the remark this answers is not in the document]
VWSCoronaDashboard4: It gives a warning. As far as the documentation for `useEffect` goes, providing an empty array will call it on mount and unmount (for that it will call the returned function, that contains the `off`). That is different behavior than with no dependencies in the array: it will then update only when the deps change.

PR #703, opened 2020-10-22
Adjustments to collapsable component and pages using it

VWSCoronaDashboard4:
## Summary

Adjustments based on the QA feedback for the collapsable component.

## Detailed design

* Design: Borders for any screen width, alignment of chevron.
* Moved FAQ section to its own page
* Fix bug that hides content on resize (done using a new `useWindowResizeDebounce`)


src/pages/veelgestelde-vragen.tsx
```diff
@@ -0,0 +1,108 @@
...
+    item: IVraagEnAntwoord
+  ) {
+    return {
+      ...item,
+      id: getSkipLinkId(item.vraag),
+      antwoord: MDToHTMLString(item.antwoord),
+    };
+  });
+
+  ensureUniqueSkipLinkIds(serializedContent);
```
VWSCoronaDashboard8, line 37: So this function mutating its arguments? If possible without much hassle I would much prefer it returns the new value instead of mutating, so it becomes something like:

```suggestion
 const processedContent = ensureUniqueSkipLinkIds(serializedContent);
```

src/pages/veelgestelde-vragen.tsx
```diff
@@ -0,0 +1,108 @@
...
+  text.over_veelgestelde_vragen.vragen = serializedContent;
+
+  const filePath = path.join(process.cwd(), 'public', 'json', 'NL.json');
+  const fileContents = fs.readFileSync(filePath, 'utf8');
+  const lastGenerated = JSON.parse(fileContents).last_generated;
+
+  return { props: { text, lastGenerated } };
+}
+
+const Verantwoording: FCWithLayout<{ text: any; lastGenerated: string }> = (
```
VWSCoronaDashboard8, line 47: Can we type the text here to the normal language definition? Otherwise much of the code below is not type-checked. Use of `any` should really be a last resort.

src/pages/veelgestelde-vragen.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+      <div className={styles.container}>
+        <MaxWidth>
+          <div className={styles.maxwidth}>
+            <h2>{text.over_veelgestelde_vragen.titel}</h2>
+            <p>{text.over_veelgestelde_vragen.paragraaf}</p>
+            <article className={styles.faqList}>
+              {text.over_veelgestelde_vragen.vragen.map(
+                (item: IVraagEnAntwoord) => {
+                  //@TODO, Why does this sometimes return empty strings for the
```
VWSCoronaDashboard8, line 76: By reading this todo I get the impression that the PR is not yet ready? Or was the code already there before?

src/utils/useWindowResizeDebounce.ts
```diff
@@ -0,0 +1,19 @@
+import { useEffect, useRef } from 'react';
+
+export const useWindowResizeDebounce = (callback: any, delay: number): void => {
```
VWSCoronaDashboard8, line 3: I think you should be able to type callback at least to a generic function. `() => void` or something.

src/utils/useWindowResizeDebounce.ts
```diff
@@ -0,0 +1,19 @@
...
+  const timeout = useRef<number | null>(null);
```
VWSCoronaDashboard8, line 4: I don't see a need to use `null` here.

```suggestion
  const timeout = useRef<number>();
```

VWSCoronaDashboard8 (changes requested):

src/pages/veelgestelde-vragen.tsx
VWSCoronaDashboard4: This comment is just moved along. There is no quick fix for this since it has to do with our markdown implementation.

src/pages/veelgestelde-vragen.tsx
VWSCoronaDashboard4: This re-applies this method. I'd be up for improving that method, but as a separate PR.

src/pages/veelgestelde-vragen.tsx
VWSCoronaDashboard4: I've adjusted this, also on the other text heavy pages.

src/utils/useWindowResizeDebounce.ts
VWSCoronaDashboard1: You'll have to initialize the state with a valid number in that case though, otherwise the compiler will complain

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard1:
Only general comment I would make is that you need to start over and implement the entire thing using styled system components  :) :) :)

src/utils/useWindowResizeDebounce.ts
VWSCoronaDashboard8: In that case just use undefined

timeout = useRef<number|undefined>()

VWSCoronaDashboard8 (approved):

PR #706, opened 2020-10-22
Feature/escalation amounts

VWSCoronaDashboard4:
## Summary

Adds amounts of escalation levels to be used by notification text message on homepage.

## Detailed design

The calculation is done in `getStaticProps`.

## Unresolved questions

* Optimizations/alternatives to the calculation would be welcome.
* I needed to do some typescript foo to pass the object to `replaceVariablesInText`, comments/improvements on that would be helpful.


src/pages/index.tsx
```diff
@@ -18,14 +18,34 @@ import { FCWithLayout } from '~/components/layout';
...
+}
+
+interface INationalHomepageData {
+  data: National;
+  text: TALLLanguages;
+  lastGenerated: string;
+  escalationLevelAmounts: EscalationLevelAmounts;
+}
+
+interface EscalationLevelAmounts {
```
VWSCoronaDashboard8, line 39: nit; I prefer "count" over "amounts".

src/pages/index.tsx
```diff
@@ -18,14 +18,34 @@ import { FCWithLayout } from '~/components/layout';
...
+  escalationLevel1: number;
+  escalationLevel2: number;
```
VWSCoronaDashboard8, line 41: If these are properties of the type EscalationLevelAmounts, there is no need to prefix them with `escalation`, since that is already implied by the context.

src/pages/index.tsx
```diff
@@ -121,13 +146,33 @@ const Home: FCWithLayout<INationalData> = (props) => {
...
 
-interface StaticProps {
-  props: {
-    data: National;
-    text: TALLLanguages;
-    lastGenerated: string;
+/**
+ * Calculate the amounts of regions with a certain escalation level
+ */
+const getEcalationAmounts = (
```
VWSCoronaDashboard8, line 152: Typo. Maybe install a spell checker? ;)

src/pages/index.tsx
```diff
@@ -121,13 +146,33 @@ const Home: FCWithLayout<INationalData> = (props) => {
...
+  escalationLevels: EscalationLevels[] | undefined
```
VWSCoronaDashboard8, line 153:
```suggestion
  escalationLevels?: EscalationLevels[]
```
But also, are you sure we need to support undefined?

src/pages/index.tsx
```diff
@@ -121,13 +146,33 @@ const Home: FCWithLayout<INationalData> = (props) => {
...
   };
-}
+
+  if (escalationLevels) {
+    escalationLevels.forEach((region) => {
+      assert(
+        [1, 2, 3, 4, 5].indexOf(region.escalation_level) !== -1,
+        'Escalation level amount not supported. Value needs to be 1-5.'
+      );
+      const key = `escalationLevel${region.escalation_level}` as keyof EscalationLevelAmounts;
```
VWSCoronaDashboard8, line 169: Nitpick; If you use "1", "2", "3" etc as the keys for your EscalationLevelAmounts type, you could use the level value straight away. I think the code would still be clear, but that's up to you.

VWSCoronaDashboard8 (approved):
Nothing serious but some suggestions

src/pages/index.tsx
VWSCoronaDashboard4: The keys are what will be available in the translation files for replacement: `{{escalationLevel1}}` will be replaced for the integer. Truncating that to `{{1}}` does not seem desirable to me.

src/pages/index.tsx
VWSCoronaDashboard4: The value is not required in the data (yet). I'd prefer to keep that as a separate todo/PR.

src/pages/index.tsx
VWSCoronaDashboard4: I'll add a comment about this. Since these will end up in the lokalize strings they need to be a bit more verbose.

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard1 (approved):

PR #708, opened 2020-10-22
Custom JSON validation for vrcodes

VWSCoronaDashboard1:
## Summary

We encountered a data drop where the blocks of data from one region were added to all of the other regions.
In this case the average_sewer_installation_per_region data for region VR25 was added to all the other regions.
To prevent another situation like this a custom keyword was added to the JSON validation that is able to check
whether the VRCode on a subobject is equal to the 'code' property on the root object.

## Motivation

Bug report

## Detailed design

A custom keyword called 'equalsRootProperty' was added in the createValidateFunction.ts file.
equalsRootProperty can be configured with the name of a property that ought to exist on the root document.
For example:
``
"gmcode": {
   "equalsRootProperty": "code"
},
``
This will check if the value of the property 'gmcode ' is equal to the value of the property 'code' on the root object.

Lastly, some simple sanity checks were added to validate-json.ts that checks the correct number of regional and municipal files.

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added

### DIFF
--- a/schema/municipal/__index.json
+++ b/schema/municipal/__index.json
@@ -20,12 +20,10 @@
       "pattern": "^GM[0-9]+$"
     },
     "name": {
-      "type": "string",
-      "pattern": "^GM[0-9]+$"
+      "const": { "$data": "1/proto_name" }
     },
     "code": {
-      "type": "string",
-      "pattern": "^GM[0-9]+$"
+      "const": { "$data": "1/proto_name" }
     },
     "hospital_admissions": {
       "$ref": "hospital_admissions.json"
--- a/src/tools/validator/createValidateFunction.ts
+++ b/src/tools/validator/createValidateFunction.ts
@@ -51,7 +51,7 @@ export function createValidateFunction(schemaPath: string) {
           (equalsRootProperty as any).errors = [
             {
               keyword: 'equalsRootProperty',
-              message: `the property ${_dataPath}'s value '${data}' must be equal to the root property ${schema}'s value '${rootValue}'`,
+              message: `the property '${_dataPath}' value '${data}' must be equal to the root property '${schema}' value '${rootValue}'`,
               params: {
                 keyword: 'equalsRootProperty',
                 value: schema,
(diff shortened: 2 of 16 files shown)


src/tools/validator/createValidateFunction.ts
```diff
@@ -31,6 +31,39 @@ export function createValidateFunction(schemaPath: string) {
...
+      _dataPath?: string,
+      _parentData?: any | any[],
+      _parentDataProperty?: string | number,
+      rootData?: any | any[]
+    ): boolean {
+      if (rootData) {
+        const rootValue = (rootData as any)[schema as string];
+        const validated = data === rootValue;
+        if (!validated) {
+          (equalsRootProperty as any).errors = [
```
VWSCoronaDashboard4, line 51: This is a pretty odd way of communicating the error array. Ar there other ways, or is this schema's go-to way of doing this?

src/tools/validator/validate-json.ts
```diff
@@ -18,6 +18,24 @@ const schemaToJsonLookup: Record<string, string[]> = {
   regions: ['REGIONS.json'],
 };
 
+if (schemaToJsonLookup.regional.length !== 25) {
```
VWSCoronaDashboard4, line 21: :100:

schema/municipal/__index.json
```diff
@@ -20,12 +20,10 @@
       "pattern": "^GM[0-9]+$"
     },
     "name": {
-      "type": "string",
-      "pattern": "^GM[0-9]+$"
```
VWSCoronaDashboard4, line 24: The GM and VR regexes are removed. Shouldn't there be some place still, to verify the format?

VWSCoronaDashboard4 (approved):
Nice, readable improvements!

Some nitpicky questions, but approved already!

src/tools/validator/createValidateFunction.ts
VWSCoronaDashboard1: This all AJV, I swear I didn't make this up myself. It also took quite some googling to find out how it actually worked since the docs are rather hazy...

VWSCoronaDashboard8 (approved):

schema/municipal/__index.json
VWSCoronaDashboard1: They are replaced by a const and $data check that ensures that this property value is equal to the proto_name property. That property still has the pattern match. Too bad this const/$data trick doesn't work for sub schema's, therefore I had to write the custom keyword.

PR #717, opened 2020-10-24
Sewer header meta weekrange

VWSCoronaDashboard8:
## Summary

Add week range metadata to content header for all sewer data pages.

## Motivation
This was solved using a hack because the content header abstraction is not suitable for this. This component will be refactored in the near future so we can remove this hack and also a similar hack that was done to display multiple sources.

I have move the metadata component so that it is a sub-component of the content header. And also moved the content header out of the layout folder. To keep code better grouped together and have the hacks live next to the original.


src/pages/gemeente/[code]/rioolwater.tsx
```diff
@@ -33,6 +34,9 @@ const SewerWater: FCWithLayout<IMunicipalityData> = (props) => {
     };
   }, [data]);
 
+  const sewerAverages = data.sewer_measurements;
+  assert(sewerAverages, 'Missing sewer measurements data');
```
Comment (VWSCoronaDashboard4) [inline src/pages/gemeente/[code]/rioolwater.tsx:38]: This data can be undefined for some municipalities; this check seems too strict in my opinion. Or has this changed in the data structure, recently?

Comment (VWSCoronaDashboard8) [reply src/pages/gemeente/[code]/rioolwater.tsx:38]: OK but that means the typescript definitions are not correct then. I will see if I can change that.

Comment (VWSCoronaDashboard8) [reply src/pages/gemeente/[code]/rioolwater.tsx]: @VWSCoronaDashboard4 The type is correct I think, I have changed the check to be a normal early return. And added a comment for clarity.

VWSCoronaDashboard4 (approved):

PR #724, opened 2020-10-26
Fixes clipping of region control due to overflow

VWSCoronaDashboard4:
## Summary

Outline is now visible again.

As far as I could see on different screen sizes, the overflow did not do anything for the maps.

### DIFF
--- a/src/scss/layout.scss
+++ b/src/scss/layout.scss
@@ -556,7 +556,6 @@ a.metric-link {
 
     margin: 0 0 1rem;
     max-width: 100%;
-    overflow: hidden;
 
     @media (min-width: $chloropleth-breakpoint) {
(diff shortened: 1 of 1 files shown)


VWSCoronaDashboard8 (approved):

src/scss/layout.scss
```diff
@@ -556,7 +556,6 @@ a.metric-link {
 
     margin: 0 0 1rem;
     max-width: 100%;
-    overflow: hidden;
```
VWSCoronaDashboard8, line 559: Seems odd that a.metric-link class is used on the radio buttons.

VWSCoronaDashboard4: It's not, that's how the diff collapsed

PR #755, opened 2020-10-27
align header metadata icons on flex-start

VWSCoronaDashboard5:
## Summary

This PR aligns the icons (clock-icon and source-icon both places in `<span>` elements), to the start of their flex-containers. Since the `svg` elements of the icons are rather small, the icons would be aligned slightly above the text. This is why the `margin-top` is added. Conventionally you would address this with a `classname` to avoid specificity issues, but since the `css` is imported and its styles apply locally, assigning the styles to `span` wouldn't cause any effects on other elements.

### Governance

- [x] Documentation is added


src/components/contentHeader/metadata.module.scss
```diff
@@ -14,5 +14,6 @@
     display: flex;
     align-items: center;
     min-width: 1.8rem;
+    margin-top: 4px;
```
VWSCoronaDashboard, line 17: This is why we need to look into https://seek-oss.github.io/capsize/

VWSCoronaDashboard (approved):

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard8 (changes requested):
There are 2 other contentHeader components which are hacks based on this one (right next to it in the components folder). I suspect you need to apply these changes there too.

VWSCoronaDashboard8 (approved):

PR #757, opened 2020-10-27
Generate language dependent sitemap on build w/ generate-sitemap.js

VWSCoronaDashboard9:
## Summary

Adds a script to the Next.js build that generates a sitemap based on the locale value.

https://coronadashboard.government.nl/sitemap.xml will have `.goverment.nl` urls and 
https://coronadashboard.rijksoverheid.nl/sitemap.xml will have `.rijksoverheid.nl` urls. 

## Motivation

We currently generate around 1600 static pages. The sitemap serves as a roadmap for search engines and will be beneficial for SEO.

## Detailed design

Started out as a TypeScript script: https://github.com/minvws/nl-covid19-data-dashboard/pull/757/commits/3896f84949f626ef13d619b25c0080b5c1da0dd2, but ran into ES6 / CommonJS compatibility issues when I added it to `next.config.js`. The only remaining evidence is the regio data which can't be imported (https://github.com/minvws/nl-covid19-data-dashboard/pull/757/files#diff-9c2bc74157356a94ea111145b0eed6e36512fd21794e2a041ee65ccbcd622179R7). Open to any suggestions!

The sitemap priorities can be updated quite easily, they're currently set as follows:
Homepage, 1
Error pages (404/500), 0
Landelijke, Gemeente and Regio pages, 0.8
All other pages, 0.6

### Governance

- [ ] Documentation is added (N/A)
- [ ] Test cases are added or updated (N/A)


src/tools/sitemap/generate-sitemap.js
```diff
@@ -0,0 +1,115 @@
...
+    });
+  });
+
+  const sitemap = `
+    <?xml version="1.0" encoding="UTF-8"?>
+    <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" 
+      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.sitemaps.org/schemas/sitemap/0.9
+        http://www.sitemaps.org/schemas/sitemap/0.9/sitemap.xsd">
+      <url>
+        <loc>https://coronadashboard.rijksoverheid.nl/</loc>
```
VWSCoronaDashboard4, line 89: The domain variable can also be applied here.

src/tools/sitemap/generate-sitemap.js
```diff
@@ -0,0 +1,115 @@
+/* eslint no-console: 0 */
+const fs = require('fs');
+const globby = require('globby');
+const prettier = require('prettier');
+const gemeenteData = require('../../data/gemeente_veiligheidsregio.json');
+
+// regioData being generated as we can't import an ES export into CommonJS
+const regioData = [...Array(25).keys()].map((n) =>
+  n + 1 < 10 ? `VR0${n + 1}` : `VR${n + 1}`
```
VWSCoronaDashboard4, line 9: Nitpick: you can use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart for this, since you don't have to worry about old browsers/polyfills for this.

src/tools/sitemap/generate-sitemap.js
```diff
@@ -0,0 +1,115 @@
...
+                    <loc>${`https://coronadashboard.${domain}.nl${p.path}`}</loc>
+                    <priority>${p.priority}</priority>
+                </url>
+            `;
+        })
+        .join('')}
+    </urlset>
+  `;
+
+  const formatted = prettier.format(sitemap, {
```
VWSCoronaDashboard4, line 105: Neat!

VWSCoronaDashboard4 (approved):

VWSCoronaDashboard (changes requested):
Maybe the sitemap should not be part of the git history. We're building a different one for each language? I'd say remove it and add it to gitignore?

src/tools/sitemap/generate-sitemap.js
```diff
@@ -0,0 +1,115 @@
...
+  ];
+
+  const pathsWithPriorities = paths.map((path) => {
+    const priority = priorities.find((priority) =>
+      path.includes(priority.path)
+    );
+
+    return {
+      path: path,
+      priority: priority != undefined ? priority.value : 0.6,
```
VWSCoronaDashboard8, line 55: nitpick; !==

Or maybe better
```suggestion
      priority: priority?.value ?? 0.6
```
But now I doubt if ?? is available outside of TS

src/tools/sitemap/generate-sitemap.js
```diff
@@ -0,0 +1,115 @@
...
+    };
+  });
+
+  const allPaths = pathsWithPriorities.filter(
```
VWSCoronaDashboard8, line 59: I find it a little odd that "all paths" are a filtered version of "paths with priorities". Going by the names I would expect the opposite :)

VWSCoronaDashboard8 (approved):
Very nice! 💯

src/tools/sitemap/generate-sitemap.js
VWSCoronaDashboard: TS is not the issue, it's that NodeJS needs to support this, but that's easy to try out. These files and the webpack config file are not transpiled.

VWSCoronaDashboard8: Yes, that's what I mean. The current version of javascript (supported by Node.js) is likely not recognizing this. You'll have to stick with the node version used in the build pipeline I guess.

VWSCoronaDashboard (approved):

VWSCoronaDashboard7 (dismissed):
Since we're introducing a new environmental, we should update the Dockerfile in this PR as well to make sure the sitemap is correctly generated at build time.

src/tools/sitemap/generate-sitemap.js
VWSCoronaDashboard7: > Yes, that's what I mean. The current version of javascript (supported by Node.js) is likely not recognizing this. You'll have to stick with the node version used in the build pipeline I guess.

If this is correct, the node version in use would be the one as defined in the Dockerfile from this repository. So there's a degree of control there.

VWSCoronaDashboard:
> Since we're introducing a new environmental, we should update the Dockerfile in this PR as well to make sure the sitemap is correctly generated at build time.

`NEXT_PUBLIC_LOCALE` has been in the codebase for a while now, but we should double-check if everything works as expected when merged to develop.

PR #758, opened 2020-10-27
replace WithChildren type with ReactNode

VWSCoronaDashboard5:
## Summary

This PR removes the `WithChildren` type from all components that accept `children`, and replaces it for `ReactNode` by `React`. The children are no longer optional. The `WithChildren` file is removed.

### Governance

- [x] Documentation is added


src/components/layout/index.tsx
```diff
@@ -46,7 +45,9 @@ export function getLayout(layoutProps: LayoutProps, lastGenerated: string) {
 
 export default Layout;
 
-function Layout(props: WithChildren<LayoutProps & ILastGeneratedData>) {
+function Layout(
+  props: LayoutProps & ILastGeneratedData & { children: React.ReactNode }
```
VWSCoronaDashboard8, line 49: Here you can integrate things with the already existing layout props

```suggestion
export interface LayoutProps extends ILastGeneratedData {
  url?: string;
  title: string;
  description?: string;
  openGraphImage?: string;
  twitterImage?: string;
  children: React.ReactNode;
}
```

src/components/legenda/index.tsx
```diff
@@ -1,10 +1,14 @@
 import * as React from 'react';
-import { WithChildren } from '~/types/index';
+
+interface IProps {
```
VWSCoronaDashboard8, line 3: Nitpick; I would call it `LegendaProps` I'm not sure if we reached concensus yet over the `I`-prefix, but in most recent code I haven't seen it.

src/components/loadingPlaceholder/index.tsx
```diff
@@ -1,7 +1,11 @@
-import { WithChildren } from '~/types/index';
 import styles from './loadingPlaceholder.module.scss';
 
-export function LoadingPlaceholder({ children }: WithChildren) {
+interface IProps {
```
VWSCoronaDashboard8, line 3: same comment

src/components/maxWidth/index.tsx
```diff
@@ -1,6 +1,10 @@
-import { WithChildren } from '~/types/index';
 import styles from './maxWidth.module.scss';
 
-export function MaxWidth({ children }: WithChildren) {
+interface IProps {
```
VWSCoronaDashboard8, line 3: etc

src/components/scalingSVG/index.tsx
```diff
@@ -1,14 +1,13 @@
 import CSS from 'csstype';
-import { WithChildren } from '~/types/index';
 import styles from './scalingSVG.module.scss';
 
 interface IProps {
   width: number;
   height: number;
-  children: WithChildren;
+  children: React.ReactNode;
```
VWSCoronaDashboard8, line 7: 👍  this was a weird one

VWSCoronaDashboard8 (changes requested):

src/components/legenda/index.tsx
VWSCoronaDashboard4: I usually use the name of the component + `Props`; so `LegendaProps` in this case.

VWSCoronaDashboard1: Also, Legenda needs to be changed to Legend, since the former is a Dutch term. (My mistake, sorry)

PR #759, opened 2020-10-27
remove null types from places where it should return undefined

VWSCoronaDashboard5:
## Summary

This PR removes the `null` types from data type definitions from places where it should return `undefined`. It keeps the places where `null` should be returned for a specific reason, such as `oneInstallation`.

### Governance

- [x] Documentation is added


VWSCoronaDashboard1 (approved):

src/components/gemeente/sewer-water-barscale.tsx
```diff
@@ -6,12 +6,12 @@ import siteText from '~/locale/index';
...
 
 export function SewerWaterBarScale(props: {
-  data: SewerWaterBarScaleData | null;
+  data: SewerWaterBarScaleData;
   showAxis: boolean;
 }) {
   const { data, showAxis } = props;
 
-  if (data === null)
+  if (data === undefined)
```
VWSCoronaDashboard8, line 14: Looks like data is not optional since you removed null. So either this check doesn't make sense or you need to make set the data type to be optional. Whatever is correct from a data perspective.

src/components/gemeente/sewer-water-metric.tsx
```diff
@@ -9,9 +9,7 @@ import siteText from '~/locale/index';
 const text = siteText.common.metricKPI;
 const title = siteText.gemeente_rioolwater_metingen.barscale_titel;
 
-export function SewerWaterMetric(props: {
-  data: SewerWaterBarScaleData | null;
-}) {
+export function SewerWaterMetric(props: { data: SewerWaterBarScaleData }) {
   const { data } = props;
 
   if (!data) return null;
```
VWSCoronaDashboard8, line 15: Same here. This check is now redundant

src/components/veiligheidsregio/sewer-water-metric.tsx
```diff
@@ -9,9 +9,7 @@ import siteText from '~/locale/index';
 const text = siteText.common.metricKPI;
 const title = siteText.veiligheidsregio_rioolwater_metingen.titel_kpi;
 
-export function SewerWaterMetric(props: {
-  data: SewerWaterBarScaleData | null;
-}) {
+export function SewerWaterMetric(props: { data: SewerWaterBarScaleData }) {
   const { data } = props;
 
   if (!data) return null;
```
VWSCoronaDashboard8, line 15: etc

src/components/veiligheidsregio/sewer-water-barscale.tsx
```diff
@@ -3,7 +3,7 @@ import { SewerWaterBarScaleData } from '~/utils/sewer-water/safety-region-sewer-
 import siteText from '~/locale/index';
 
 export function SewerWaterBarScale(props: {
-  data: SewerWaterBarScaleData | null;
+  data: SewerWaterBarScaleData;
   showAxis: boolean;
 }) {
   const { data, showAxis } = props;
```
VWSCoronaDashboard8, line 9: null check below here as well

src/components/gemeente/sewer-water-barscale.tsx
VWSCoronaDashboard8: If this compiles that means data is non-optional probably, so just remove the check then.

VWSCoronaDashboard8 (changes requested):

PR #760, opened 2020-10-27
Update sewer schemas to use corrected values and unify naming of things

VWSCoronaDashboard8:
## Summary

* Update sewer schemas replacing `rna_per_ml` with `rna_normalized`
* Rename sewer schema metric keys and to be consistent
* Rename schema types to be consistent
* Adds regions/sewer.json schema for Choropleth
* Update types based on new schemas
* Update code based on new types
* Clean up a lot of code around sewer data handing involving types, null and undefined checks, and data fallbacks.

### Root key changes

- NL `rioolwater_metingen` => `sewer`
- NL `rioolwater_metingen_per_rwzi` => `sewer_per_installation`
- VR `average_sewer_installation_per_region` => `sewer`
- VR `results_per_sewer_installation_per_region` => `sewer_per_installation`
- GM `sewer_measurements` => `sewer`
- GM `results_per_sewer_installation_per_municipality` => `sewer_per_installation`

### Property changes

- NL sewer_per_installation `rna_per_ml` => `rna_normalized`
- VR sewer_per_installation `rna_per_ml` => `rna_normalized`
- GM sewer_per_installation `rna_per_ml` => `rna_normalized`

### New choropleths

- REGIONS sewer


schema/municipal/sewer_per_installation.json
```diff
@@ -12,7 +12,7 @@
         "rwzi_awzi_code",
         "rwzi_awzi_name",
         "gmcode",
-        "rna_per_ml",
+        "rna_corrected",
```
VWSCoronaDashboard4, line 15: for other places where an adjustment has been made, the term "normalized" is in use:

"In the simplest cases, normalization of ratings means adjusting values measured on different scales to a notionally common scale, often prior to averaging." `</wikipedia>`

VWSCoronaDashboard8: I considered it, but I wasn't sure if normalized maybe also means scaling the values to a prefixed range. I will use that term then 👍

VWSCoronaDashboard4 (approved):

VWSCoronaDashboard1 (approved):

schema/municipal/__index.json
```diff
@@ -2,14 +2,16 @@
   "$schema": "http://json-schema.org/draft-07/schema#",
   "type": "object",
   "title": "municipal",
-  "additionalProperties": false,
+  "additionalProperties": true,
```
VWSCoronaDashboard1, line 5: is this temporary? if not, why do we allow additional properties?

VWSCoronaDashboard1 (approved):

schema/municipal/__index.json
VWSCoronaDashboard8: I figured on the root level it can't hurt. For example then I can merge the PR about behavior, without breaking the build pipeline until we have data for behavior.

It makes the transition to new data a little easier I think, and I can not think of a scenario where this causes issues.

VWSCoronaDashboard1: Aha, ok, so its temporary, no problem then

VWSCoronaDashboard8: Well not necessarily temporary. I don't see a problem enabling this always. Can you think of a scenario where this is problematic? Your app needs to know what all required data is available at root level. Any additional things won't hurt. It's different for properties in nested levels because that would bloat the data.

VWSCoronaDashboard1: It won't break anything true. It might just feel weird that on production there could be data in there that's unused, but since we pull out only the parts that we need it shouldn't be a problem.

src/utils/sewer-water/municipality-sewer-water.util.ts
```diff
@@ -1,15 +1,16 @@
...
 import { XrangePointOptionsObject } from 'highcharts';
 import { replaceVariablesInText } from '~/utils/replaceVariablesInText';
 
 import siteText from '~/locale/index';
+import { assert } from '../assert';
+
+/**
+ * @TODO these helpers for VR and GM should be merged into one using generics.
+ * All of this code seems duplicate now that the type names are unified.
+ */
```
VWSCoronaDashboard4, line 13: This refactor clears up a lot of things! :100:

schema/municipal/__index.json
```diff
@@ -33,11 +33,11 @@
     "positive_tested_people": {
       "$ref": "positive_tested_people.json"
     },
-    "sewer_measurements": {
-      "$ref": "sewer_measurements.json"
+    "sewer": {
```
VWSCoronaDashboard4, line 36: Have you considered naming this eg `sewer_averages` ? Now it could read like `sewer` contains all sewer related data, which that is not the case.

VWSCoronaDashboard4 (approved):

schema/regions/sewer.json
```diff
@@ -0,0 +1,38 @@
...
+    },
+    "week_start_unix": {
+      "type": "integer"
+    },
+    "week_end_unix": {
+      "type": "integer"
+    },
+    "vrcode": {
+      "type": "string",
+      "pattern": "^VR[0-9]+$"
```
VWSCoronaDashboard1, line 26:
```suggestion
      "equalsRootProperty": "code"
```
This checks if the vrcode is the same as the 'code' property in the root file.

src/components/landelijk/sewer-water-barscale.tsx
```diff
@@ -1,17 +1,15 @@
 import { BarScale } from '~/components/barScale';
 import siteText from '~/locale/index';
-import { RioolwaterMetingen } from '~/types/data.d';
+import { NationalSewer } from '~/types/data.d';
```
VWSCoronaDashboard1, line 3: A national sewer, how bad do you reckon would smell that? ;)

VWSCoronaDashboard1 (changes requested):

schema/municipal/__index.json
VWSCoronaDashboard8: Yes, that would have been a good name, but the value property inside sewer is already named "average". This way we could add more values than just averages. For example it also has the installation count, but I guess that is also linked to the average kind of. In any case I don't think the name is wrong or that confusing at the moment.

src/pages/gemeente/[code]/rioolwater.tsx
```diff
@@ -74,7 +74,7 @@ const SewerWater: FCWithLayout<IMunicipalityData> = (props) => {
       />
 
       <TwoKpiSection>
-        {barScaleData.value !== undefined && (
+        {barScaleData && barScaleData.value && (
```
Comment (VWSCoronaDashboard4) [inline src/pages/gemeente/[code]/rioolwater.tsx:77]: What if `barScaleData.value === 0` ?

Comment (VWSCoronaDashboard8) [reply src/pages/gemeente/[code]/rioolwater.tsx:77]: yup good catch :)

VWSCoronaDashboard1 (approved):

PR #767, opened 2020-10-27
Simple schema for locale files with placeholder validation

VWSCoronaDashboard1:
## Summary

This PR adds a very simple schema for the locale files that only checks a few properties in order to ensure they are arrays instead of objects (due to Lokalize sometimes making this change for some strange reason).
A custom validation was also added that loops recursively through all of the string values of the locale files and looks for invalid placeholders. A valid placeholder would be `{{placeholderName}}`, an invalid one would be `{placeHolderName}` or `{placeholderName}}`.

## Motivation

Joery request

## Detailed design

validate-json now uses a slightly more elaborate structure that keeps all of the schema, file and custom validation logic in one place. The validation phase was changed a bit where the optional custom validators are executed.
Currently there's only one custom validator called 'validPlaceholders', but I tried to make the system a little bit generic because there's a good chance we'll have add other checks in the future.

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added


src/tools/validator/custom-validations/valid-placeholders.ts
```diff
@@ -0,0 +1,60 @@
...
+ * the error occurred along with the offending placeholder.
+ *
+ */
+export const validPlaceholders = (
+  input: any,
+  parentName?: string
+): string[] | undefined => {
+  const parentSuffix = parentName ? `${parentName}.` : '';
+
+  const result = Object.entries(input)
```
VWSCoronaDashboard8, line 18: Nitpick; You set input to `any` and then assume you can call Object.entries() on it. Something like `Record<string, unknown>` would be a little more strict. Also, then the flatMap input types would also be inferred correctly.

src/tools/validator/custom-validations/valid-placeholders.ts
```diff
@@ -0,0 +1,60 @@
...
+        prefix !== '{{' ||
+        suffix !== '}}' ||
+        middle.indexOf('{') > -1 ||
+        middle.indexOf('}') > -1
+      ) {
+        return matched;
+      }
+      return undefined;
+    })
+    .filter(Boolean) as string[];
```
VWSCoronaDashboard8, line 59:
```suggestion
    .filter(isDefined);
```

if you use isDefined from ts-is-present, then typescript can infer the correct return type and the code is more robust, because if you made a mistake in the map the compiler will not complain since you explicitly cast the return type.

src/tools/validator/custom-validations/valid-placeholders.ts
```diff
@@ -0,0 +1,60 @@
...
+      if (typeof value === 'object') {
+        return validPlaceholders(value, `${parentSuffix}${propertyName}`);
+      }
+    })
+    .filter(Boolean) as string[];
+
+  return result.length ? result : undefined;
+};
+
+function validatePlaceHolders(text: string): string[] {
```
VWSCoronaDashboard8, line 40: nit; the return type annotation should be redundant. I prefer to omit it.

src/tools/validator/custom-validations/valid-placeholders.ts
```diff
@@ -0,0 +1,60 @@
...
+): string[] | undefined => {
+  const parentSuffix = parentName ? `${parentName}.` : '';
+
+  const result = Object.entries(input)
+    .flatMap(([propertyName, value]: [string, any]) => {
+      if (typeof value === 'string') {
+        const result = validatePlaceHolders(value);
+        if (result.length) {
+          return result.map(
+            (placeholder: string) =>
```
VWSCoronaDashboard8, line 24: `: string` is redundant I think. TS should know the type of results.

src/tools/validator/custom-validations/valid-placeholders.ts
```diff
@@ -0,0 +1,60 @@
...
+          );
+        }
+        return;
+      }
+
+      if (typeof value === 'object') {
+        return validPlaceholders(value, `${parentSuffix}${propertyName}`);
+      }
+    })
+    .filter(Boolean) as string[];
```
VWSCoronaDashboard8, line 35: See comment below about filter and return type

src/tools/validator/validate-json.ts
```diff
@@ -1,36 +1,56 @@
...
-import { jsonBasePath } from './jsonBasePath';
-import { schemaDirectory } from './getSchemaNames';
+import { createValidateFunction } from './create-validate-function';
+import { jsonBasePath, localeBasePath } from './base-paths';
+import { schemaDirectory } from './get-schema-names';
 import chalk from 'chalk';
+import { validPlaceholders } from './custom-validations/valid-placeholders';
 
-const allJsonFiles = fs.readdirSync(jsonBasePath);
+type CustomValidationFunction = (input: any) => string[] | undefined;
```
VWSCoronaDashboard8, line 10:
```suggestion
type CustomValidationFunction = (input: Record<string, unknown>) => string[] | undefined;
```

src/tools/validator/validate-json.ts
```diff
@@ -47,10 +67,14 @@ const validationPromises = Object.keys(schemaToJsonLookup).map<
 // up by CI.
 Promise.all(validationPromises)
   .then((validationResults) => {
-    const flatResult = validationResults.flatMap((bools) => bools);
+    const flatResult = validationResults.flatMap(
+      (booleanResults) => booleanResults
+    );
+
```
VWSCoronaDashboard8, line 73: You're not really mapping anything so I think this will do

```suggestion
    const flatResult = validationResults.flat()
```

src/tools/validator/validate-json.ts
```diff
@@ -60,29 +84,47 @@ Promise.all(validationPromises)
...
-    });
+  return schemaInfo.files.map((fileName) => {
+    const contentAsString = fs.readFileSync(
+      path.join(schemaInfo.basePath, fileName),
+      {
+        encoding: 'utf8',
+      }
+    );
 
     const data = JSON.parse(contentAsString);
```
VWSCoronaDashboard8, line 108: Might be good to put a try / catch around this. Otherwise, the validation will stop / exit at the first file which can't be parsed.

src/tools/validator/validate-json.ts
```diff
@@ -60,29 +84,47 @@ Promise.all(validationPromises)
...
 
-    const isValid = validateFunction(data);
+    let isValid = validateFunction(data);
+    let schemaErrors: any[] = validateFunction.errors ?? [];
```
VWSCoronaDashboard8, line 111: I feel you can be more specific than `any`. An error can not be a whole lot of things or my imagination is just too limited ;)

src/tools/validator/validate-json.ts
```diff
@@ -60,29 +84,47 @@ Promise.all(validationPromises)
...
+    if (schemaInfo.customValidations) {
+      const errors = schemaInfo.customValidations
+        .flatMap((validationFunc) => validationFunc(data))
+        .filter(Boolean);
```
VWSCoronaDashboard8, line 115: Probably helps with typing to use the isDefined I mentioned earlier

src/tools/validator/validate-json.ts
VWSCoronaDashboard1: Aha, I wasn't familiar with flat() haha, thanks

src/tools/validator/custom-validations/valid-placeholders.ts
```diff
@@ -0,0 +1,61 @@
...
+  return result.length ? result : undefined;
+};
+
+function validatePlaceHolders(text: string) {
+  const matches = [...(text.matchAll(/({[^}]+[}]+)/g) as any)];
+
+  return matches
+    .map((matchInfo: string[]) => {
+      const matched = matchInfo[0];
+      const prefix = matched.substr(0, 2);
```
VWSCoronaDashboard4, line 47: If you define group the open/close brackets to be in capturing groups, you can use `matchInfo[1]` to `matchInfo[2]` instead of this substr.

src/tools/validator/custom-validations/valid-placeholders.ts
```diff
@@ -0,0 +1,61 @@
...
+      const suffix = matched.substr(-2);
+      const middle = matched.substr(2, matched.length - 4);
+      if (
+        prefix !== '{{' ||
+        suffix !== '}}' ||
+        middle.indexOf('{') > -1 ||
+        middle.indexOf('}') > -1
```
VWSCoronaDashboard4, line 54: The regex already tests for this part to not include the close bracket. If you extend the regex to also exclude the opening bracket, both of these `indexOf` uses can be removed.

src/tools/validator/custom-validations/valid-placeholders.ts
```diff
@@ -0,0 +1,61 @@
...
+        return validPlaceholders(value, `${parentSuffix}${propertyName}`);
+      }
+    })
+    .filter(isDefined);
+
+  return result.length ? result : undefined;
+};
+
+function validatePlaceHolders(text: string) {
+  const matches = [...(text.matchAll(/({[^}]+[}]+)/g) as any)];
```
VWSCoronaDashboard4, line 42: If you use this regex instead: 

```/({+)([^{}]+)(}+)/g```

You only need to check `matchInfo[1]` and `matchInfo[3]` are of length 2. `matchInfo[2]` is the middlepart and the regex ensures there are no `{` or `}` in there.

(The square brackets in `[}]+` were not needed: it's a character group of only one character.)

File: src/tools/validator/schema-information.ts [hunk 14]
```diff
@@ -0,0 +1,47 @@
...
+
+const localeJsons = fs.readdirSync(localeBasePath);
+const allJsonFiles = fs.readdirSync(jsonBasePath).concat(localeJsons);
+
+// This struct defines which JSON files should be validated with which schema.
+export const schemaInformation: Record<string, SchemaInfo> = {
+  national: { files: ['NL.json'], basePath: jsonBasePath },
+  ranges: { files: ['RANGES.json'], basePath: jsonBasePath },
+  regional: {
+    files: filterFilenames(allJsonFiles, new RegExp('^VR[0-9]+\\.json$')),
```
VWSCoronaDashboard4, line 23: Why use `new RegExp(...)` instead of eg `/^VR[0-9]+\.json$/` ? Double character encodings for both JS strings and regexes make the regexes more complicated to read.

Same comment for the other two regexes below.

src/tools/validator/custom-validations/valid-placeholders.ts
VWSCoronaDashboard1: Actually no, what I'm checking here is a mangled placeholder like this: {{placeho{lderName}}, the regex willcapture this entire string.

src/tools/validator/custom-validations/valid-placeholders.ts
VWSCoronaDashboard1: Hm, again, your regex won't match a mangled placeholder like  `{{placeho{lderName}}`, it will match only this part: `{lderName}}`.
In all honesty, regex will only go so far to check these. I'd have to write a parser to filter out all of the possible mistakes. But I considered this 'good enough' for now.

src/tools/validator/schema-information.ts
VWSCoronaDashboard1: Fixed

src/tools/validator/custom-validations/valid-placeholders.ts
VWSCoronaDashboard1: Actually, the check for '}' isn't necessary, these will not be in the middle part

src/tools/validator/custom-validations/valid-placeholders.ts
VWSCoronaDashboard1: Good catch, fixed!

src/tools/validator/custom-validations/valid-placeholders.ts
VWSCoronaDashboard1: I've changed the check completely. Removed the grouping completely in favor of just matching the entire string and then using this regex /[{]{2}[^{}]+[}]{2}/ to check if the matched string adheres to the correct format.

src/tools/validator/custom-validations/valid-placeholders.ts
VWSCoronaDashboard1: you're right about the 'one character group' though, I've changed that

src/tools/validator/execute-validations.ts
```diff
@@ -0,0 +1,40 @@
...
+ * described in the given schema info.
+ * The validation result along with any possible error messages are returned.
+ *
+ */
+export function executeValidations(
+  validateFunction: ValidateFunction,
+  data: any,
+  schemaInfo: SchemaInfo
+): ValidationResult {
+  let isValid = validateFunction(data) as boolean;
```
VWSCoronaDashboard8, line 22: Why isn't boolean already the function return type? Since the function is for validating something I would expect the outcome to be strongly typed.

VWSCoronaDashboard8 (approved):

src/tools/validator/execute-validations.ts
VWSCoronaDashboard1: ajv defines the result as `boolean | PromiseLike<any>` because validations are optionally async. A generic would have been nicer but unfortunately, I don't have control over their definitions.

PR #774, opened 2020-10-28
Add schemas for behavior for nl, vr and regions

VWSCoronaDashboard8:
## Summary

Adds schemas for implementing charts based on behavior research.

### Root key changes

- NL `behavior` ADD
- VR `behavior` ADD
- REGIONS `behavior` ADD

### Property changes

All three added schemas use the following properties/ids for labeling behavior

- wash_hands
- keep_distance
- work_from_home
- avoid_crowds
- symptoms_stay_home
- symptoms_get_tested
- wear_mask_public_indoors
- wear_mask_public_transport
- sneeze_cough_elbow
- max_visitors_home

To accompany the data which is split between support and compliance, all of these keys have a `_support` or `_compliance` suffix in the schemas.

Then for each of the keys there is another key with an extra suffix to indicate the trend using `_trend` which can be up/down/equal. These trend values are not the same as other calculated differences. If the results over time change from 90% to 92% the trend can still be considered "equal" due to error margins in the data.

I have kept all values type integer because even though they are represented as percentages, they are by nature of "steekproef" not accurate at all.

There are 3 properties that have not been included in the first research so I have left them (and their suffixed properties) out of the required list. These are:

- wear_mask_public_indoors
- wear_mask_public_transport
- sneeze_cough_elbow

### DIFF
--- a/schema/national/__index.json
+++ b/schema/national/__index.json
@@ -23,7 +23,8 @@
     "hospital_beds_occupied",
     "intensive_care_beds_occupied",
     "ggd",
-    "nursing_home"
+    "nursing_home",
+    "behavior"
   ],
   "additionalProperties": false,
   "properties": {
@@ -98,6 +99,9 @@
     },
     "restrictions": {
       "$ref": "restrictions.json"
+    },
+    "behavior": {
+      "$ref": "behavior.json"
     }
   }
 }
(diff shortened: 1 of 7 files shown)


VWSCoronaDashboard1 (approved):

VWSCoronaDashboard4 (approved):

VWSCoronaDashboard5 (approved):

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard8:
Closed in favor of #875

PR #778, opened 2020-10-28
Feature/collapsable styled system

VWSCoronaDashboard4:
## Summary

Use styled-system for the collapsable component.

Also fixes clipping bug on-resize.

## Detailed design

* This enhances the third party components from Reach with styled components, to supply our styles.
* The logic for animating is adjusted. Fully opened panels do not have the max-height any more, and a isAnimating property controls the max-height during the animation. Some other code became unnesscary by using that approach, so overall it clears up some things too.

## Unresolved questions

For the previous version I wrote the hook `useWindowResizeDebounce`, which is not used any more. Shall I remove that?


src/components-styled/collapsable.tsx
```diff
@@ -0,0 +1,185 @@
...
+   * Collapsed content should not be accessible using the tab functionality.
+   */
+  const setLinkTabability = useCallback(
+    (open) => {
+      if (!panelReference) {
+        return;
+      }
+
+      const links = panelReference.querySelectorAll('a');
+      Array.from(links).forEach((link) => {
```
VWSCoronaDashboard1, line 148: Array.from, is that supported by IE11 or polyfilled?
https://caniuse.com/?search=Array.from

VWSCoronaDashboard1 (changes requested):

src/components-styled/collapsable.tsx
VWSCoronaDashboard8: I think TS transpiles to the set target in config, which is I believe es5

VWSCoronaDashboard8 (changes requested):

src/components-styled/collapsable.tsx
VWSCoronaDashboard1: We've had issues with things like this before, so I'm not entirely clear what typescript transpiles and what not. So, I guess we'll just have to test it.

src/components-styled/collapsable.tsx
```diff
@@ -0,0 +1,185 @@
...
+    setLinkTabability(open);
+  }, [checkLocationHash, setLinkTabability, open]);
+
+  useEffect(() => {
+    window.addEventListener('hashchange', checkLocationHash, false);
+    return () => {
+      window.removeEventListener('hashchange', checkLocationHash, false);
+    };
+    /* eslint-disable-next-line */
+  }, []); // should not use dependencies in array: use effect mimics mount / unmount
```
VWSCoronaDashboard10, line 166: This is buggy: your component will now always trigger the `checkLocationHash` which is initially defined using the initial value of `id`. The moment the `id` variable mutates, you will not be checking the location hash against the new value of `id`.

It's perfectly fine to remove/attach the event listener when the callback changes:


```suggestion
  useEffect(() => {
    window.addEventListener('hashchange', checkLocationHash, false);
    return () => {
      window.removeEventListener('hashchange', checkLocationHash, false);
    };
  }, [checkLocationHash]);
```

src/components-styled/collapsable.tsx
```diff
@@ -0,0 +1,185 @@
...
+    <Box as="section" borderTop={'1px solid #dfdfdf'} {...(props as any)}>
+      <Disclosure open={open} onChange={toggle}>
+        <Summary>{summary}</Summary>
+        <Panel
+          ref={setPanelReference}
+          onTransitionEnd={() => setIsAnimating(false)}
+          style={{
+            /* panel max height is only controlled when collapsed, or during animations */
+            maxHeight: !open || isAnimating ? `${panelHeight}px` : undefined,
+          }}
```
VWSCoronaDashboard10, line 178: Do we need to pass a `style` prop? I think this translates to inline styles, doesn't styled-system has something like a `cx` or `css` prop so styles will be applied with classnames?

VWSCoronaDashboard10 (changes requested):
Spotted at least one bug :-)

src/components-styled/collapsable.tsx
```diff
@@ -0,0 +1,185 @@
...
+}
+
+export const Collapsable = (props: CollapsableProps) => {
+  const { summary, children, id } = props;
+
+  const [open, setOpen] = useState(false);
+  const [panelHeight, setPanelHeight] = useState<number>(0);
+  const [isAnimating, setIsAnimating] = useState(false);
+
+  let panelReference: HTMLDivElement | undefined = undefined;
```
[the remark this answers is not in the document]
VWSCoronaDashboard8: Do you need to pass null? I'm trying to get everyone to avoid null as much as possible :)

src/components-styled/collapsable.tsx
VWSCoronaDashboard8: Good catch

src/components-styled/collapsable.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard4: I'll add a comment about why I did this in the code; `panelReference` is used as a dependency for the useCallback hook, and React errors that it is used before it is defined if I remove initial setting to undefined.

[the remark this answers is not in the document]
VWSCoronaDashboard10: > Do you need to pass null? I'm trying to get everyone to avoid null as much as possible :)

I'm afraid we need to pass `null`, TS will complain when going for `undefined`.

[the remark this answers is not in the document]
VWSCoronaDashboard10: I think the `setLinkTabability` can be a regular function/callback instead of having it wrapped in a useCallback hook

src/components-styled/collapsable.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+  /* Store the latest reference to the panel */
+  const setPanelReference = (element: HTMLDivElement) => {
+    if (!element) {
+      return;
+    }
+    panelReference = element;
```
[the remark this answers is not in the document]
VWSCoronaDashboard4: It's what I used in the previous implementation (see deleted collapsable/index); while working on the bug, the reference seemed to get lost from time to time, hence why I converted it to this structure. Since I've also adjusted other things, I'll check if I can use the `useRef` again.

VWSCoronaDashboard4:
@VWSCoronaDashboard10 `NodeList.prototype.forEach()` does not work in IE11, and we have not set up anything for it to be transpiled or polyfilled.

src/components-styled/collapsable.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+  const panelReference = useRef<HTMLDivElement>(null);
+
+  /* Store the latest reference to the panel */
+  // const setPanelReference = (element: HTMLDivElement) => {
+  //   if (!element) {
+  //     return;
+  //   }
+  //   panelReference = element;
+  // };
```
VWSCoronaDashboard8, line 93: Comments can be removed I guess

src/components-styled/collapsable.tsx
VWSCoronaDashboard8: @VWSCoronaDashboard4 Did you forget about this one?

src/components-styled/collapsable.tsx
```diff
@@ -73,40 +74,39 @@ interface CollapsableProps extends BoxProps {
...
-
+export const Collapsable = ({ summary, children, id }: CollapsableProps) => {
   const [open, setOpen] = useState(false);
-  const [panelHeight, setPanelHeight] = useState<number>(0);
+  const [panelHeight, setPanelHeight] = useState(0);
   const [isAnimating, setIsAnimating] = useState(false);
 
-  let panelReference: HTMLDivElement | undefined = undefined;
+  /* Needs to explicitly use undefined to use panelReference as dependancy for useCallback */
+  // let panelReference: HTMLDivElement | undefined = undefined;
```
VWSCoronaDashboard10, line 83: comments can be removed

src/components-styled/collapsable.tsx
```diff
@@ -73,40 +74,39 @@ interface CollapsableProps extends BoxProps {
...
-      return;
-    }
-    panelReference = element;
-  };
+  // const setPanelReference = (element: HTMLDivElement) => {
+  //   if (!element) {
+  //     return;
+  //   }
+  //   panelReference = element;
+  // };
```
VWSCoronaDashboard10, line 93: same here

src/components-styled/collapsable.tsx
```diff
@@ -0,0 +1,184 @@
...
+import styled from 'styled-components';
+import { css } from '@styled-system/css';
+import { colors } from '~/style/theme';
+
+const Summary = styled(DisclosureButton)(
+  css({
+    alignItems: 'flex-start',
+    bg: 'transparent',
+    border: 'none',
+    color: colors.blue,
```
VWSCoronaDashboard8, line 18: You should not have to import colors. Just write "blue" and it should map to the theme, just like paddings etc

src/components-styled/collapsable.tsx
```diff
@@ -0,0 +1,184 @@
...
+    fontSize: 3,
+    fontWeight: 'bold',
+    justifyContent: 'space-between',
+    m: 0,
+    p: 3,
+    textAlign: 'left',
+    width: '100%',
+
+    '&:focus': {
+      outline: `1px dashed ${colors.blue}`,
```
VWSCoronaDashboard8, line 30: Same here

src/components-styled/collapsable.tsx
```diff
@@ -0,0 +1,184 @@
...
+    },
+
+    '&:active[data-state="collapsed"]': {
+      bg: colors.shadow,
```
VWSCoronaDashboard8, line 34: etc

VWSCoronaDashboard8 (changes requested):

src/components-styled/collapsable.tsx
VWSCoronaDashboard10: Are you sure that will work here, wrapped inside a string with multiple configuration?

src/components-styled/collapsable.tsx
VWSCoronaDashboard10: Het mentioned that IE11 doesn't support that method

src/components-styled/collapsable.tsx
VWSCoronaDashboard10: I'd assume that it should work using `outline-color: blue` though

VWSCoronaDashboard4:
@VWSCoronaDashboard10 Array.from works fine in IE11 with our setup of transpiling

VWSCoronaDashboard8:
@VWSCoronaDashboard4 Ah right, I missed your comment about forEach not being supported by IE11 because it was not inline with the subject thread.

VWSCoronaDashboard8 (approved):

src/components-styled/collapsable.tsx
```diff
@@ -165,7 +154,7 @@ export const Collapsable = ({ summary, children, id }: CollapsableProps) => {
   }, [checkLocationHash]);
 
   return (
-    <Box as="section" borderTop={`1px solid ${colors.lightGray}`} id={id}>
+    <Box as="section" borderTop={`1px solid lightGray`} id={id}>
```
VWSCoronaDashboard10: Can you confirm (with inspect element) that `lightGray` maps to `#dfdfdf` instead of actually being `lightGray`?

src/style/theme.ts
```diff
@@ -84,6 +84,7 @@ const colors = {
   icon: '#01689b',
   button: '#01689b',
   shadow: '#e5e5e5',
+  lightGray: '#dfdfdf',
```
VWSCoronaDashboard5, line 87: we're using both 'grey' and 'gray' throughout the codebase. If we are going to continue using several different grey-tones and both spellings ('grey' and 'gray'), we could write it like:

```
const colors = {
blue: '#01689b',
greys: ['#dfdfdf', '#anotherGreyTone']
}

// followed by:
colors.grays = colors.greys
```

this would allow for a greytone scale that's more easy to read and use, and it would accept both spellings.

VWSCoronaDashboard5 (approved):

PR #779, opened 2020-10-28
Add restrictions schema

VWSCoronaDashboard1:
## Summary

Add JSON schemas for the restrictions overview and the restrictions per region.

## Motivation

Feature request

## Detailed design

The restrictions are a list of objects that look like this:
```
{
  "restriction_id": "4_er_op_uit_1",
  "target_region": "nl",
  "escalation_level": 4,
  "category_id": "er_op_uit",
  "restriction_order": 1
}
```
Where _target_region_ can be either **nl** or **vr**, indicating whether the restriction is limited to a region or is of national concern. The escalation level indicates the column the restriction is placed in and the category id indicates the row.
The restriction order indicates which order within the cell this restriction is rendered in.
By selecting all the items with _target_region_ **nl** the national restrictions overview can be assembled.

The region (VR*.json) files now have a new property called 'restrictions' which holds an array of restriction ids that map to the above shown _restriction_id_ field. Using this mapping the region restrictions view can be assembled.

The _restriction_id_ value can be used to map to Lokalize keys and icons.

A new custom validation function was added which will check if the list of restriction identifiers in each region file actually exists in the main RESTRICTIONS.json
A new custom keyword validRestrictionId was added to validate the formatting and values of the restriction_id fields

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added


VWSCoronaDashboard8 (changes requested):
Let's discuss the details in a call

schema/restrictions/__index.json
```diff
@@ -0,0 +1,63 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "title": "restrictions",
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: These are all restrictions, regional and national

src/tools/validator/custom-validations/validate-restriction-ids.ts
```diff
@@ -18,15 +19,23 @@ export const validRestrictionIds = (
...
+  let restrictionData: Restrictions | undefined;
+  try {
+    restrictionData = JSON.parse(
+      fs.readFileSync(sourcePath, { encoding: 'utf8' })
+    );
+    /* eslint-disable-next-line */
+  } catch (e) {}
+
+  if (restrictionData === undefined) {
+    return [`Restrictions file '${sourcePath}' could not be parsed to JSON.`];
```
VWSCoronaDashboard8, line 31: I would expect this to go in the catch block, and then you don't have to use eslint disable. 

I see what you're trying to do here with undefined. Maybe replace it with assert(restrictionData, 'some error') but it should always be there if the catch is not invoked so it's a bit double. The resulting code will be a bit clearer I think.

src/tools/validator/custom-validations/valid-restriction-ds.ts
```diff
@@ -0,0 +1,38 @@
+import fs from 'fs';
+import path from 'path';
+import { isDefined } from 'ts-is-present';
+import { jsonBasePath } from '../base-paths';
+
+export const validRestrictionIds = (
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: I made this change manually, since it also needed to change the import name in another file

src/tools/validator/custom-validations/valid-restriction-ds.ts
```diff
@@ -0,0 +1,38 @@
...
+  input: Record<string, unknown>
+): string[] | undefined => {
+  if (!input.restrictions) {
+    return [`Property 'restrictions' doesn't exist`];
+  }
+  const regionRestrictions = input.restrictions as any;
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: I hadn't generated the typescript by then, now I have so I'm using the generated types now

src/tools/validator/custom-validations/validate-restriction-ids.ts
```diff
@@ -18,15 +19,23 @@ export const validRestrictionIds = (
...
+  }
 
-  const restrictionIds = (restrictionData.values as any[]).map(
+  const restrictionIds = restrictionData.values.map(
     (value: any) => value.identifier
```
VWSCoronaDashboard8:
```suggestion
    (value) => value.identifier
```
With functions like map I often use a shorthand x. It often makes the formatter place things on the same line

```suggestion
    (x) => x.identifier
```

src/tools/validator/custom-validations/validate-restriction-ids.ts
```diff
@@ -18,15 +19,23 @@ export const validRestrictionIds = (
...
   );
 
-  const result = (regionRestrictions.values as any[])
+  const result = regionRestrictions.values
     .map((id: number) => {
```
VWSCoronaDashboard8, line 39: I think this might be a mistake. You are mapping over values but use the value as id. In any case the number type annotation should be redundant/removed

VWSCoronaDashboard8: Ah no regionRestrictions is a lists of ids right, then the code is still correct, only remove the :number annotation

VWSCoronaDashboard1: Indeed, the id has now also changed from a number type to a string, since the identifier will be in the format `<escalation-level>_<category-id>_<unique-number>`

schema/restrictions/__index.json
```diff
@@ -0,0 +1,80 @@
...
+          "type": "string",
+          "enum": [
+            "er_op_uit",
+            "bezoek",
+            "samenkomst",
+            "huwelijk",
+            "verpleeghuis",
+            "horeca",
+            "sport",
+            "reizen_binnenland",
```
VWSCoronaDashboard8, line 65: In my code for behavior I have create all ids in English, but I don't mind really

src/tools/validator/custom-validations/validate-restriction-ids.ts
VWSCoronaDashboard1: Fixed

schema/restrictions/__index.json
```diff
@@ -0,0 +1,80 @@
...
+        "$ref": "#/definitions/value"
+      }
+    }
+  },
+  "definitions": {
+    "value": {
+      "title": "restrictions_value",
+      "additionalProperties": false,
+      "required": [
+        "identifier",
```
VWSCoronaDashboard8, line 35: Id is one of the few names I don't mind as acronym :) I restriction_id would be a bit more clear, since we also have a category_id and restriction_order

VWSCoronaDashboard8 (approved):

schema/restrictions/__index.json
VWSCoronaDashboard1: fair enough

VWSCoronaDashboard4 (approved):

PR #789, opened 2020-10-29
remove onzekerheidsmarge

VWSCoronaDashboard5:
## Summary

This PR removes the 'onzekerheidsmarge' from the chart on reproductiegetal. Previously, this graph showed a blue linechart, surrounded with a grey area-fill to indicate the 'onzekerheidsmarge' (uncertainty margin). This grey fill is removed, leaving only the blue line. The chart is now a `LineChart`.

Furthermore this PR adds a couple changes to the `LineChart` component. It introduces a boolean setting to set the `fill`, which is `true` by default. In the `calculateYMax` function I slightly adjusted the `Math.max()` function to allow for smaller numbers. When the `maxValue` is now below `signaalwaarde`, the `yMax` will be equal to `signaalwaarde * 2` to make sure the `signaalwaarde` line floats in the middle of the `yAxis.

The filter in the `calculateYMax` was changed to check on `isFilled` rather than `isDefined` to filter out `null` values in the `values` array.

### Governance

- [x] Documentation is added


src/components/lineChart/index.tsx
```diff
@@ -11,7 +11,7 @@ import styles from './lineChart.module.scss';
 
 export type Value = {
   date: number;
-  value: number;
+  value: number | null;
```
VWSCoronaDashboard4, line 14: Can we prevent this ` | null` from being added? perhaps modifying the type of whatever is passed to the linechart?

VWSCoronaDashboard8: Yes that would be preferable. But I told @VWSCoronaDashboard5 that this would be the easiest way, because I remembered I removed the `| null` recently. But you're right, the problem should preferably be fixed at the source.

I would suggest to do that in a different PR

src/components/lineChart/index.tsx
```diff
@@ -148,7 +150,12 @@ function getChartOptions(
         showInLegend: false,
         color: '#3391CC',
         // hex to rgb converted, added opacity
-        fillColor: 'rgba(51, 145, 204, 0.2)',
+        /* since this chart has type 'area', a fillColor of `undefined` will return
```
VWSCoronaDashboard8, line 153: nitpick; 

I prefer to use /** for all comments. I find it more readable. If you hit enter it will add new rows and you can reformat with the vscode rewrap extension.

VWSCoronaDashboard8 (approved):

PR #791, opened 2020-10-30
Fixes bug in collapsable that prevents them from closing

VWSCoronaDashboard4:
## Summary

Collapsables linked with an anchor link should open on page load, and then be able to get closed. This fixes a bug in the closing bit.

## Detailed design

Since the `open` state was part of the dependencies for `useEffect`, the `checkLocationHash` got re-evaluated for every toggle. Fixed by shuffling the logic a bit apart. `requestAnimationFrame` fixes odd scrolling behavior on page load when using these links (it would stop mid-page).

### DIFF
--- a/src/components-styled/collapsable.tsx
+++ b/src/components-styled/collapsable.tsx
@@ -86,7 +86,6 @@ export const Collapsable = ({ summary, children, id }: CollapsableProps) => {
       return;
     }
     setOpen(!open);
-    setLinkTabability(!open);
     startAnimation(!open);
   }
 
@@ -141,16 +140,10 @@ export const Collapsable = ({ summary, children, id }: CollapsableProps) => {
     [panelReference]
   );
 
-  useEffect(() => {
-    setLinkTabability(open);
-  }, [setLinkTabability, open]);
+  useEffect(() => setLinkTabability(open), [setLinkTabability, open]);
 
   useEffect(() => {
     requestAnimationFrame(() => checkLocationHash());
-    window.addEventListener('hashchange', checkLocationHash, false);
-    return () => {
-      window.removeEventListener('hashchange', checkLocationHash, false);
-    };
   }, [checkLocationHash]);
 
   return (
--- a/src/components-styled/collapsable.tsx
+++ b/src/components-styled/collapsable.tsx
@@ -142,11 +142,11 @@ export const Collapsable = ({ summary, children, id }: CollapsableProps) => {
   );
 
   useEffect(() => {
-    checkLocationHash();
     setLinkTabability(open);
-  }, [checkLocationHash, setLinkTabability, open]);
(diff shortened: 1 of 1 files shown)


src/components-styled/collapsable.tsx
```diff
@@ -142,11 +142,11 @@ export const Collapsable = ({ summary, children, id }: CollapsableProps) => {
...
-    checkLocationHash();
     setLinkTabability(open);
-  }, [checkLocationHash, setLinkTabability, open]);
+  }, [setLinkTabability, open]);
 
   useEffect(() => {
+    requestAnimationFrame(() => checkLocationHash());
     window.addEventListener('hashchange', checkLocationHash, false);
     return () => {
       window.removeEventListener('hashchange', checkLocationHash, false);
```
VWSCoronaDashboard10, line 152: Will hashchanges even occur during the lifecycle of this component? Isn't it sufficient to grab the value of the hash when the component mounts?

VWSCoronaDashboard4: Not with the current setup, no. I can remove it, and in case it causes troubles in the future, these events can be added again.

src/components-styled/collapsable.tsx
```diff
@@ -141,16 +140,10 @@ export const Collapsable = ({ summary, children, id }: CollapsableProps) => {
...
-  }, [checkLocationHash, setLinkTabability, open]);
+  useEffect(() => setLinkTabability(open), [setLinkTabability, open]);
 
   useEffect(() => {
-    window.addEventListener('hashchange', checkLocationHash, false);
-    return () => {
-      window.removeEventListener('hashchange', checkLocationHash, false);
-    };
+    requestAnimationFrame(() => checkLocationHash());
   }, [checkLocationHash]);
```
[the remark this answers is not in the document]
VWSCoronaDashboard10: And for readability you could move the tabable stuff to a separate hook:

```tsx
function Collapsable() {
  // ... 

  const isTabableRef = useIsTabable<HTMLDivElement>(isOpen)

  // ...
  return <Panel ref={isTabableRef}>...</Panel>
}



// pseudo code, not tested but in theory something like this should work:
function useIsTabable<T>(isEnabled: boolean) {
  const ref = useRef<T>(null)

  React.useEffect(() => {
    if (!ref.current) {
      return;
    }

    const links = ref.current.querySelectorAll('a');

    Array.from(links).forEach((link) => {
      link.setAttribute('tabindex', open ? '0' : '-1');
    });
      
  }, [isEnabled])

  return ref
}
```

// ...

VWSCoronaDashboard4:
@VWSCoronaDashboard10 Let's stick to the bugfix; we can always make everything nicer if we take more time to do it.

VWSCoronaDashboard10 (approved):

PR #792, opened 2020-10-30
Enable use of css-prop

VWSCoronaDashboard10:
This PR allows us to use the `css` prop on styled components and dom elements. The css prop accepts the return value of styled-systems `css`-function. TS will warn when anything else is received, but the app keeps working.

```tsx
const Foo = styled.div``

<Foo css={css({ p: [1, 2, 3] })}>
  <img css={css({ my: 5 })} />
</Foo>

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -3,7 +3,7 @@
   "plugins": [
     "lodash",
     [
-      "styled-components",
+      "babel-plugin-styled-components",
       {
         "ssr": true,
         "displayName": true
--- a/src/types/css.d.ts
+++ b/src/types/css.d.ts
@@ -1,0 +1,12 @@
+import { CssFunctionReturnType } from '@styled-system/css';
+import {} from 'react';
+
+declare module 'react' {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: .babelrc [hunk 1]
```diff
@@ -3,7 +3,7 @@
   "plugins": [
     "lodash",
     [
-      "styled-components",
+      "babel-plugin-styled-components",
```
VWSCoronaDashboard10, line 6: this change is not really necessary, just an improvement for readability.

src/types/css.d.ts
```diff
@@ -0,0 +1,12 @@
+import { CssFunctionReturnType } from '@styled-system/css';
+import {} from 'react';
```
VWSCoronaDashboard8, line 2: What is this for?

VWSCoronaDashboard8 (approved):

src/types/css.d.ts
VWSCoronaDashboard10: good catch, looks like it isn't necessary

PR #793, opened 2020-10-30
Add Lokalise automerge GitHub action

VWSCoronaDashboard1:
## Summary

This adds a Github action that is able to validate, approve, and merge pull requests from Lokalise automatically.

## Motivation

Feature request.

## Detailed design

The action only triggers on pull requests on the develop and master branches. It also only triggers when the json files in the src/locale directory have been changed.

Each step of action checks if the current actor is called VWSCoronaDashboard (the user under which Lokalise runs) and if the PR branch name is prefixed like 'lokalise-*'. With these checks we can be fairly certain that the PR originates from Lokalise.

If these conditions are met, the repository gets pulled and a yarn install is performed.
After that two yarn commands are run: `yarn validate-single locale en.json` and `yarn validate-single locale nl.json`

If these validations come back without errors, the PR is given the label 'lokalise' and is auto-approved.

The final job checks for the existence of the 'lokalise' label in which case the PR is automatically merged and deleted afterwards.

## Drawbacks

This action will also run during normal PRs where the locale files are changed. In that case the action will not DO anything, but it will take a few moments longer to run the checks.

## Alternatives

Please let me know.

## Unresolved questions

Ideally, this action will only run when either the **src/locale/nl.json** file or the **src/locale/en.json** has been changed. Unfortunately, I haven't been able to get this to work in this way. 
Consider this configuration:
```
paths:
  - 'src/locale/*.json'
```
This ensures that the action triggers when the locale files are changed, but unfortunately not exclusively. So, when apart from these files also other files have been changed (like in a regular PR), the action will still trigger. (Just without ending up doing anything).
I would really like this action to trigger exclusively when only the locale files are in the changeset. So if anyone has ideas about this, I am all ears.

~~You'll also see that in each step of the action this if statement is added:
`if: github.actor == 'VWSCoronaDashboard' && startsWith(github.head_ref, 'lokalise-') == true`
Again ideally, I'd like to be able to perform such a check at the start of a job and be able to end the job when the condition is not met. I have not found a clean way of actually doing this. Therefore the check is added to each step. Any suggestions here are welcome too.~~

### Governance

- [ ] Documentation is added

### DIFF
--- a/.github/workflows/lokalise-automerger.yml
+++ b/.github/workflows/lokalise-automerger.yml
@@ -1,4 +1,4 @@
-name: Lokalize PR check
+name: Lokalise PR validation, auto approve and merge
 on:
   pull_request:
     branches:
--- a/.github/workflows/lokalise-automerger.yml
+++ b/.github/workflows/lokalise-automerger.yml
@@ -8,11 +8,12 @@ on:
       - 'src/locale/*.json'
 jobs:
   autocheckandapprove:
+    # More handy functions: https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#startswith
+    if: github.actor == 'VWSCoronaDashboard' && startsWith(github.head_ref, 'lokalise-') == true
+    name: Validate, label and approve
     runs-on: ubuntu-latest
     steps:
       - name: Remove label
-        # More handy functions: https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#startswith
-        if: github.actor == 'VWSCoronaDashboard' && startsWith(github.head_ref, 'lokalise-') == true
         uses: buildsville/add-remove-label@v1
         with:
           token: ${{ github.token }}
@@ -21,29 +22,24 @@ jobs:
 
       - name: Checkout repo
         uses: actions/checkout@v2
-        if: github.actor == 'VWSCoronaDashboard' && startsWith(github.head_ref, 'lokalise-') == true
 
       - name: yarn install
-        if: github.actor == 'VWSCoronaDashboard' && startsWith(github.head_ref, 'lokalise-') == true
         uses: borales/actions-yarn@v2.0.0
         with:
           cmd: install
 
       - name: EN locale validation
-        if: github.actor == 'VWSCoronaDashboard' && startsWith(github.head_ref, 'lokalise-') == true
         uses: borales/actions-yarn@v2.0.0
         with:
           cmd: validate-single locale en.json
 
       - name: NL locale validation
-        if: github.actor == 'VWSCoronaDashboard' && startsWith(github.head_ref, 'lokalise-') == true
         uses: borales/actions-yarn@v2.0.0
         with:
           cmd: validate-single locale nl.json
 
       - name: Add label
         uses: buildsville/add-remove-label@v1
-        if: github.actor == 'VWSCoronaDashboard' && startsWith(github.head_ref, 'lokalise-') == true
         with:
           token: ${{ github.token }}
           label: lokalise
@@ -51,12 +47,12 @@ jobs:
 
       - name: Auto approve
         uses: hmarr/auto-approve-action@v2.0.0
-        if: github.actor == 'VWSCoronaDashboard' && startsWith(github.head_ref, 'lokalise-') == true
         with:
           github-token: ${{ github.token }}
 
   automerge:
     name: Auto merge after successful checks
+    if: github.actor == 'VWSCoronaDashboard' && startsWith(github.head_ref, 'lokalise-') == true
     # By default, jobs run in parallel. To run the jobs sequentially, the keyword "needs" is needed.
     # Auto merge action can be done only when the PR is approved, hence this automerge needs autocheckandapprove as a prerequisite
     needs: autocheckandapprove
@@ -64,7 +60,6 @@ jobs:
     steps:
       - name: Auto merge
         uses: pascalgn/automerge-action@v0.12.0
-        if: github.actor == 'VWSCoronaDashboard' && startsWith(github.head_ref, 'lokalise-') == true
         env:
           GITHUB_TOKEN: ${{ github.token }}
           MERGE_LABELS: lokalise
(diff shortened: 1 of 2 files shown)


.github/workflows/lokalise-automerger.yml
```diff
@@ -0,0 +1,71 @@
...
+on:
+  pull_request:
+    branches:
+      - master
+      - develop
+    paths:
+      - 'src/locale/*.json'
+jobs:
+  autocheckandapprove:
+    runs-on: ubuntu-latest
```
VWSCoronaDashboard10, line 11: about the following: 
>You'll also see that in each step of the action this if statement is added:
if: github.actor == 'VWSCoronaDashboard' && startsWith(github.head_ref, 'lokalise-') == true
Again ideally, I'd like to be able to perform such a check at the start of a job and be able to end the job when the condition is not met. I have not found a clean way of actually doing this. Therefore the check is added to each step. Any suggestions here are welcome too.

Doesn't [`jobs.<job_id>.if`](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif) do the job?

```suggestion
jobs:
  autocheckandapprove:
    if: github.actor == 'VWSCoronaDashboard' && startsWith(github.head_ref, 'lokalise-') == true
    runs-on: ubuntu-latest
```

VWSCoronaDashboard1: hahaha, well, duh, that sounds very likely indeed. I completely missed that one! I am going to test it out a.s.a.p.! Thanks!

.github/workflows/lokalise-automerger.yml
```diff
@@ -0,0 +1,71 @@
...
+          type: add
+
+      - name: Auto approve
+        uses: hmarr/auto-approve-action@v2.0.0
+        if: github.actor == 'VWSCoronaDashboard' && startsWith(github.head_ref, 'lokalise-') == true
+        with:
+          github-token: ${{ github.token }}
+  automerge:
+    name: Auto merge after successful checks
+    # By default, jobs run in parallel. To run the jobs sequentially, they keywords "needs" is needed.
```
VWSCoronaDashboard10, line 59:
```suggestion
    # By default, jobs run in parallel. To run the jobs sequentially, the keyword "needs" is needed.
```

.github/workflows/lokalise-automerger.yml
```diff
@@ -0,0 +1,71 @@
...
+    # Auto merge action can be done only when the PR is approved, hence this automerge needs autocheckandapprove as a prerequisite
+    needs: autocheckandapprove
+    runs-on: ubuntu-latest
+    steps:
+      - name: Auto merge
+        uses: pascalgn/automerge-action@v0.12.0
+        if: github.actor == 'VWSCoronaDashboard' && startsWith(github.head_ref, 'lokalise-') == true
```
VWSCoronaDashboard10, line 66: I'm not sure this `if` is still necessary when the previous job it depends on (`autocheckandapprove`) is canceled due to `jobs.<job_id>.if`

.github/workflows/lokalise-automerger.yml
```diff
@@ -0,0 +1,71 @@
...
+          token: ${{ github.token }}
+          label: lokalise
+          type: add
+
+      - name: Auto approve
+        uses: hmarr/auto-approve-action@v2.0.0
+        if: github.actor == 'VWSCoronaDashboard' && startsWith(github.head_ref, 'lokalise-') == true
+        with:
+          github-token: ${{ github.token }}
+  automerge:
```
VWSCoronaDashboard10, line 57: A new line would be nice over here
```suggestion

  automerge:
```

.github/workflows/lokalise-automerger.yml
```diff
@@ -0,0 +1,71 @@
+name: Lokalise PR validation, auto approve and merge
+on:
+  pull_request:
+    branches:
+      - master
+      - develop
+    paths:
+      - 'src/locale/*.json'
```
VWSCoronaDashboard10, line 8: > Ideally, this action will only run when either the src/locale/nl.json file or the src/locale/en.json has been changed.

In that case, shouldn't the paths be the following?

```suggestion
    paths:
      - 'src/locale/en.json'
      - 'src/locale/nl.json'
```

VWSCoronaDashboard1: No, this will effectively do the same. If BESIDES these two files OTHER files are changed as well, the action will still run. WHat I'm trying to achieve is that this action will run EXCLUSIVELY when ONLY these two files are in the changeset.

VWSCoronaDashboard10: Ah ok, maybe something like https://stackoverflow.com/questions/59288971/retrieving-list-of-modified-files-in-github- might point you in the right direction, but I'm not sure if it's worth the effort

VWSCoronaDashboard1: Yes, with the use of some other actions I'll be able to get it to work, but all of those solutions will require the action to be started up. If it would be possible just by configuring the paths at the top of the action, in that case it won't even be started by Github, which is the speedup I'm looking for.
But yes, it's a nitpick, so currently not worth the bother.

.github/workflows/lokalise-automerger.yml
```diff
@@ -0,0 +1,67 @@
...
+          type: add
+
+      - name: Auto approve
+        uses: hmarr/auto-approve-action@v2.0.0
+        with:
+          github-token: ${{ github.token }}
+
+  automerge:
+    name: Auto merge after successful checks
+    if: github.actor == 'VWSCoronaDashboard' && startsWith(github.head_ref, 'lokalise-') == true
```
VWSCoronaDashboard10, line 55: I still doubt whether this `if` is necessary if the previous step has not been performed, but feel free to keep it like this :-)

VWSCoronaDashboard10 (approved):

.github/workflows/lokalise-automerger.yml
VWSCoronaDashboard1: It's a sanity check, it's true that it's not necessary. But I keep thinking I'm forgetting some kind of angle in which it could be abused. So I'm keeping this in just to keep me sleeping soundly ;)

VWSCoronaDashboard10: Perfect 👍

PR #802, opened 2020-10-30
Cleanup

VWSCoronaDashboard10:
## Summary

In my journey to discover the codebase I've ran into several improvements I tried to pick up, here's a summary of some changes:
- add `compile` script in package.json to be able to detect project-wide compilation issues
- renamed prop `anchorName` to `name` because it looks better compared to `<Anchor anchorName={name} />`
- removed several `any` types
- tried to wrap my head around the types used in the Choropleth. I believe there are some problems with types from different libraries. They have previously been ignored by casting stuff to `any`, but the result is that it's currently not very easy to fix these.
- Fixed several instances of react refs
- import locale data with `import` syntax instead of `require` in order to enable TS typings

### DIFF
--- a/package.json
+++ b/package.json
@@ -125,8 +125,7 @@
     "cy:open": "cypress open",
     "e2e:ci": "yarn build && start-test 3000 cy:ci",
     "cy:ci": "cypress run --headless --browser chrome",
-    "download": "download https://coronadashboard.rijksoverheid.nl/latest-data.zip -e -o public/",
-    "compile": "tsc --noEmit"
+    "download": "download https://coronadashboard.rijksoverheid.nl/latest-data.zip -e -o public/"
   },
   "browserslist": [
     "defaults",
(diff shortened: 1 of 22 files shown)


package.json
```diff
@@ -125,7 +125,8 @@
     "cy:open": "cypress open",
     "e2e:ci": "yarn build && start-test 3000 cy:ci",
     "cy:ci": "cypress run --headless --browser chrome",
-    "download": "download https://coronadashboard.rijksoverheid.nl/latest-data.zip -e -o public/"
+    "download": "download https://coronadashboard.rijksoverheid.nl/latest-data.zip -e -o public/",
+    "compile": "tsc --noEmit"
```
VWSCoronaDashboard1, line 129: There's already this one:
`"type-check": "tsc --pretty --noEmit"`
Granted, that one adds the prettier flag as well, but does largely the same.

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard1:
Overall, great tweaks, especially all of the type narrowing! Awesome!

VWSCoronaDashboard8 (approved):
💯

PR #810, opened 2020-11-02
Fix choropleth size

VWSCoronaDashboard10:
## Summary

The rendering of the choropleth was off in both IE11 as other browsers. We're moving from scss to styled-components, but the latter [won't prefix grid-related css properties for IE11](https://github.com/thysultan/stylis.js/issues/51). One of the possible fixes is to use flexbox instead of grid, which this PR will implement.

Besides the above, the following has been picked up:
- Introduce a `useBreakpoints` hook
- Use the new `ChoroplethTile` component on the app's landing page
- Fix the theme's `mediaQueries` collection, it shouldn't include the `@media` for correct usage with `window.matchMedia`
- Fix jest test-run by providing a ThemeProvider during tests

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -4,8 +4,6 @@ import path from 'path';
 import { useState } from 'react';
 import ExternalLink from '~/assets/external-link.svg';
 import Notification from '~/assets/notification.svg';
-import { ChartRegionControls } from '~/components-styled/chart-region-controls';
-import { ChoroplethLegenda } from '~/components/choropleth/legenda/ChoroplethLegenda';
 import { useSafetyRegionLegendaData } from '~/components/choropleth/legenda/hooks/useSafetyRegionLegendaData';
 import { MunicipalityChoropleth } from '~/components/choropleth/MunicipalityChoropleth';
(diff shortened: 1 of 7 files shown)


VWSCoronaDashboard1 (approved):

src/pages/index.tsx
```diff
@@ -86,65 +89,55 @@ const Home: FCWithLayout<INationalHomepageData> = (props) => {
...
+      <ChoroplethTile
+        title={text.veiligheidsregio_index.selecteer_titel}
+        description={
+          <>
+            <span
+              dangerouslySetInnerHTML={{
+                __html: text.veiligheidsregio_index.selecteer_toelichting,
+              }}
+            />
+            <EscalationMapLegenda text={text} />
```
VWSCoronaDashboard8, line 101: Maybe add a comment that we need a different abstraction for this. Appending the legenda to the description even though there is a legend prop looks like a hack.

VWSCoronaDashboard10:
@VWSCoronaDashboard1 I had to change some testing related code, you might want to take another look at those changes.

VWSCoronaDashboard8 (approved):

PR #831, opened 2020-11-03
Fixes flatMap errors in Edge17

VWSCoronaDashboard4:
## Summary

Fix flatMap errors in Edge17 by including a polyfill.

## Detailed design

Polyfill is now added for each use of flatMap; we could move that / optimize that later as this is prone to future errors.


VWSCoronaDashboard10:
Are you sure this is necessary? When I search inside the `.next/static` directory for `flatMap` there's already a hit in the `.next/static/chunks/polyfill.js` file

src/components/areaChart/index.tsx
```diff
@@ -9,6 +9,9 @@ import { formatNumber } from '~/utils/formatNumber';
 import { getFilteredValues, TimeframeOption } from '~/utils/timeframe';
 import styles from './areaChart.module.scss';
 
+/* Polyfill for flatMap */
+import 'core-js/features/array/flat-map';
```
VWSCoronaDashboard8, line 13: This is a bit risky because you have to remember to import the polyfill. I would opt to add it to _app.tsx.

VWSCoronaDashboard8 (changes requested):

VWSCoronaDashboard10:
IE11 polyfills should already be covered by next.js: https://nextjs.org/docs/basic-features/supported-browsers-features#polyfills

VWSCoronaDashboard8 (approved):

PR #865, opened 2020-11-05
Implement differences in standardized way

VWSCoronaDashboard8:
A proposal to define property differences in a standardized manner. Things to note:

- Every level (NL/VR/GM) gets its own differences file named __difference.json. This file holds differences for any property in the data. 
- Each property uses a standardized definition for either integer or decimal value.
- The file + property names are concatenated so we can have a flat structure. For example the difference for `national/huisarts_verdenkingen.json` => property `incidentie` will live in `national/__difference.json` => property `huisarts_verdenkingen__incidentie`. Note the double underscore to make a distinction between data context and property in the concatenation.

The aim of this setup is to define differences with the least effort for both backend and frontend while also avoiding code duplication.

### DIFF
--- a/schema/national/__difference.json
+++ b/schema/national/__difference.json
@@ -1,0 +1,54 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "national_differences",
+  "type": "object",
+  "properties": {
+    "huisarts_verdenkingen__incidentie": {
+      "$ref": "#/definitions/diff_integer"
+    },
+    "huisarts_verdenkingen__geschat_aantal": {
+      "$ref": "#/definitions/diff_integer"
+    },
+    "infected_people_total__infected_daily_total": {
+      "$ref": "#/definitions/diff_integer"
+    }
+  },
+  "required": ["huisarts_verdenkingen"],
+  "additionalProperties": false,
+  "definitions": {
+    "diff_integer": {
+      "title": "national_diff_integer",
+      "type": "object",
+      "properties": {
+        "old_value": {
+          "type": "integer"
+        },
+        "difference": {
+          "type": "integer"
+        },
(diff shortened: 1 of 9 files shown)


schema/national/__difference.json
```diff
@@ -0,0 +1,54 @@
...
+      "$ref": "#/definitions/diff_integer"
+    },
+    "huisarts_verdenkingen__geschat_aantal": {
+      "$ref": "#/definitions/diff_integer"
+    },
+    "infected_people_total__infected_daily_total": {
+      "$ref": "#/definitions/diff_integer"
+    }
+  },
+  "required": ["huisarts_verdenkingen"],
```
VWSCoronaDashboard1, line 16:
```suggestion
  "required": ["huisarts_verdenkingen__incidentie", "huisarts_verdenkingen__geschat_aantal", "infected_people_total__infected_daily_total"],
```

VWSCoronaDashboard1 (approved):

schema/national/__difference.json
```diff
@@ -0,0 +1,54 @@
...
+          "type": "integer"
+        },
+        "old_date_of_report_unix": {
+          "type": "integer"
+        }
+      },
+      "required": ["old_value", "difference", "old_date_of_report_unix"],
+      "additionalProperties": false
+    },
+    "diff_decimal": {
```
VWSCoronaDashboard4, line 36: this one is not yet needed?

VWSCoronaDashboard4 (approved):

schema/national/__difference.json
VWSCoronaDashboard8: I'm placing the definitions there because they will be used. This was copy-paste and it won't hurt. It takes more time to figure out what the definition should be when you need to start using a decimal diff.

PR #874, opened 2020-11-06
Proposal: loading locale data using Context API

VWSCoronaDashboard9:
## Summary

This PR utilises the React Context API to provide locale data to the pages / components in the application.

Additional notes:

- It looks like one _huge_ PR, in reality 90% of the files is merely a case of replacing:

```
import text from '~/locale/index';
```
with
```
import { useContext } from 'react';
import LocaleContext, { ILocale } from '~/locale/localeContext';
...
const { siteText }: ILocale = useContext(LocaleContext);
```
- Apart from a couple of instances where utils are at play, it can be implemented incrementally (not all at once like this PR)
- Depending on whether or not we upgrade Next or implement a CMS, this might become obsolute

## Motivation

Whilst looking at the website performance, we realised that all locale data is being loaded on every page - regardless of domain. This approach should result in only the data that’s actually being used, to be loaded.

I ran lighthouse (incognito) and analyze locally and there does seem to be some improvements.

**Before**
Screenshot 2020-11-06 at 15 24 01
Screenshot 2020-11-06 at 15 23 48

**After**
Screenshot 2020-11-06 at 15 24 44
Screenshot 2020-11-06 at 15 24 28

"All chunks" before 13MB, after 7MB.

## Detailed design

Essentially, I've created a React Context: https://github.com/minvws/nl-covid19-data-dashboard/pull/874/files#diff-2e6da01c39f5b33ad2a61d72b2b7eea73a090ba12e10bd2a36e9b86e949631fb and the application is wrapped in a ContextProvider (see _app: https://github.com/minvws/nl-covid19-data-dashboard/pull/874/files#diff-88a36bc7a53bfa58c6f451464798d631d785160dfacfdaf9479be7a0b1d7e5e5). The Context Provider is populated with the current text on load. Wherever you want to use `siteText` the locale data can be imported using the `useContext` hook. The implementation can be improved to set data upfront, but that would entail _always_ loading the nl data.

Some additional improvements:
- more consistency between `siteText` and `text`
- setting the data types for locale data doesn't require importing the data first
- locale data is type'd everywhere
- in some cases where we were loading everything for one string (like in DataWarning), text is now passed as props

## Drawbacks

- our strategy might change moving forward
- the performance "gains" aren't guaranteed
- will require lots of QA (not complicated, just far reaching)

### Governance

- [ ] Documentation is added


VWSCoronaDashboard:
This comes _very_ close to how react-intl works out of the box.

PR #883, opened 2020-11-10
Topic/gedrag tabel

VWSCoronaDashboard4:
## Summary

Draft PR to display additon for the gedrag page.


src/domain/behavior/behavior-table-tile.tsx
```diff
@@ -0,0 +1,190 @@
...
+    'compliance'
+  );
+
+  return (
+    <Tile>
+      <h3>{text.basisregels.title}</h3>
+      <Box display="flex" justifyContent="start">
+        <BehaviorTypeControls
+          onChange={onChangeControls}
+        ></BehaviorTypeControls>
```
VWSCoronaDashboard10, line 151:
```suggestion
        <BehaviorTypeControls onChange={onChangeControls} />
```

src/domain/behavior/behavior-table-tile.tsx
```diff
@@ -0,0 +1,190 @@
...
+} {
+  const compliance = formatBehaviorType(behavior, 'compliance');
+  const support = formatBehaviorType(behavior, 'support');
+
+  return sortBehavior(compliance, support);
+}
+
+export function BehaviorTableTile({ text, behavior }: BehaviorTileProps) {
+  const { sortedCompliance, sortedSupport } = formatAndSortBehavior(behavior);
+  const [behaviorType, onChangeControls] = useState<'compliance' | 'support'>(
```
VWSCoronaDashboard10, line 141: Usually we'd call the setter something like `setBehaviorType`, which in this case can be used as a handler:

```tsx
<BehaviorTypeControls onChange={setBehaviorType} />
```

src/domain/behavior/behavior-table-tile.tsx
```diff
@@ -0,0 +1,190 @@
...
+      <h3>{text.basisregels.title}</h3>
+      <Box display="flex" justifyContent="start">
+        <BehaviorTypeControls
+          onChange={onChangeControls}
+        ></BehaviorTypeControls>
+      </Box>
+
+      <p>{text.basisregels.intro[behaviorType]}</p>
+      <MobileScroll>
+        <Table>
```
VWSCoronaDashboard10, line 156: another nitpick thing: since these components almost have no styling and they're used on a single place, it would imho be fine to write the css inline.

eg:

```suggestion
      <div css={css({ overflow: 'auto' })}>
        <table css={css({ width: '100%' })}>
```

It's not really necessary to rewrite it, but it might improve development speed when you're writing tiny bits of css like this.

src/domain/behavior/behavior-table-tile.tsx
```diff
@@ -0,0 +1,190 @@
...
+  'wash_hands',
+  'keep_distance',
+  'work_from_home',
+  'avoid_crowds',
+  'symptoms_stay_home',
+  'symptoms_get_tested',
+  'wear_mask_public_indoors',
+  'wear_mask_public_transport',
+  'sneeze_cough_elbow',
+  'max_visitors_home',
```
VWSCoronaDashboard10, line 31: I think we want to rename this to `max_visitors`, otherwise we won't be able to locate the data which is named `max_visitors_compliance` and `max_visitors_support`.

Another fix would be to discuss this name with backend and have them serve it under `max_visitors_home_compliance` and `max_visitors_home_support`.

VWSCoronaDashboard10 (commented):
Nice work 👍  couple of comments and questions and I think I spotted one bug with the `max_visitors_home`-key.

src/components/layout/NationalLayout.tsx
```diff
@@ -309,6 +309,21 @@ function NationalLayout(props: NationalLayoutProps) {
...
+                <Link href="/landelijk/gedrag">
+                  <a
+                    onClick={blur}
+                    className={getClassName('/landelijk/gedrag')}
+                  >
+                    <TitleWithIcon
+                      Icon={Gedrag}
+                      title={siteText.nl_gedrag.sidebar.titel}
+                    />
+                    <span>@TODO sidebar metric gedrag</span>
```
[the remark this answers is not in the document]
VWSCoronaDashboard4: Let's fix that (and also the icon's appearance) in a small, separate, PR

VWSCoronaDashboard10 (approved):

PR #915, opened 2020-11-12
Topic/gedrag sidebar texts

VWSCoronaDashboard4:
## Summary

* Adjusts the sidebar with correct text and date range
* Fixes Gedrag icon
* Makes common texts (between national and safety region level) common


src/components/layout/SafetyRegionLayout.tsx
```diff
@@ -333,7 +334,9 @@ function SafetyRegionLayout(
                         Icon={Gedrag}
                         title={siteText.nl_gedrag.sidebar.titel}
                       />
-                      <span>@TODO sidebar metric gedrag</span>
+                      <span>
+                        <BehaviorMetric data={data.behavior} />
+                      </span>
```
VWSCoronaDashboard10, line 339: Does it make sense to move this container `span` to the `BehaviorMetric` component itself?

src/locale/en.json
```diff
@@ -174,15 +174,24 @@
     "max_visitors": "Ontvang het maximaal aantal personen thuis"
   },
 
+  "gedrag_common": {
+    "compliance": "Naleving",
+    "support": "Draagvlak",
```
VWSCoronaDashboard10, line 179: Do we need to update the line chart tile to also read the translations from this common block?

VWSCoronaDashboard10 (approved):

src/components/layout/SafetyRegionLayout.tsx
VWSCoronaDashboard4: I've copied this merely out of consistency with the rest of the sidebar.

There is room for improvement for this, but I'd leave that for a PR to have the sidebar use more styled system. The span is wrong anyway, since it contains block level elements.

src/locale/en.json
VWSCoronaDashboard4: It is either already done (for the toggle) or not necessary. We can move some texts perhaps from the linechart to the common block. I have moved all texts from the table that were not the intros / footnotes since these won't change between national and regional level.

PR #916, opened 2020-11-12
Infected age groups

VWSCoronaDashboard4:
## Summary

Schema for new infected age group.

Compared to the previous version:
* different main key: `infected_age_groups`;
* `last_value` was not properly used in the previous version; removed `last_value`
* adjustment of the keys inside `age_groups` objects.

### Example of new structure

NL.json

```json
{
   ...
  "infected_age_groups": {
      "values": [
         {
            "age_group_range": "0-9",
            "infected_percentage": 1,
            "age_group_percentage": 7,
            "date_of_insertion": 1605192514,
            "date_of_report_unix": 1605192514
         },
         {
            "age_group_range": "10-19",
            "infected_percentage": 20,
            "age_group_percentage": 17,
            "date_of_insertion": 1605192514,
            "date_of_report_unix": 1605192514
         },
        ....
      ]
   }
   ...
}
```


schema/national/infected_age_groups.json
```diff
@@ -1,20 +1,38 @@
 {
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "infected_age_groups.json",
+  "title": "infected_age_groups",
+  "type": "object",
+  "required": ["age_groups"],
+  "additionalProperties": false,
+  "properties": {
+    "age_groups": {
```
VWSCoronaDashboard1, line 9: I think in all of the other schema's these properties are called 'values', even when there's no last_value property. So I'd opt for keeping things consistent and simply calling this one 'values' as well.

VWSCoronaDashboard1 (changes requested):

schema/national/infected_age_groups.json
```diff
@@ -1,20 +1,38 @@
 {
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "infected_age_groups.json",
```
VWSCoronaDashboard8, line 3: $id is not being used so you can remove it.

schema/national/infected_age_groups.json
```diff
@@ -1,20 +1,38 @@
...
+  "title": "infected_age_groups",
```
VWSCoronaDashboard8, line 4: The newer schemas use a prefix in the title to make it unique, since the typescript types are generated from it. I would suggest `national_infected_age_groups` to keep it consistent.

schema/national/infected_age_groups.json
VWSCoronaDashboard8: Yes, we are not free to choose top-level properties like this. Everything uses values and/or last_value. With the exception being the difference keys.

schema/national/infected_age_groups.json
```diff
@@ -1,20 +1,38 @@
...
+  "properties": {
+    "age_groups": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/item"
+      }
+    }
+  },
   "definitions": {
     "item": {
```
VWSCoronaDashboard8, line 17: In the newer schemas this definition is called "value".  Don't forget to change `"$ref": "#/definitions/item"` then above.

schema/national/infected_age_groups.json
```diff
@@ -1,20 +1,38 @@
...
-      "title": "intake_share_age_groups_last_value",
+      "title": "infected_age_group",
```
VWSCoronaDashboard8, line 18: To make this unique and keep it consistent with newer schemas I suggest `national_infected_age_groups_value`

schema/national/infected_age_groups.json
```diff
@@ -1,20 +1,38 @@
...
       "additionalProperties": false,
       "type": "object",
       "required": [
-        "agegroup",
-        "infected_per_agegroup_increase",
+        "age_group",
+        "infected_percentage",
+        "national_percentage",
```
VWSCoronaDashboard8, line 24: This name is a little vague. Mabe `national_age_group_percentage` or `national_total_percentage`?

VWSCoronaDashboard8 (changes requested):

VWSCoronaDashboard1 (approved):

schema/national/infected_age_groups.json
```diff
@@ -1,20 +1,37 @@
...
       "properties": {
-        "agegroup": {
+        "age_group": {
           "type": "string"
         },
-        "infected_per_agegroup_increase": {
+        "infected_percentage": {
+          "type": "number"
+        },
+        "national_age_group_percentage": {
```
VWSCoronaDashboard8, line 34: I don't see a reason to prefix properties like this. We do not do that anywhere else. Properties are already scoped in their respective data context. This doesn't help to avoid naming collisions, it just makes the property name longer.

schema/national/infected_age_groups.json
```diff
@@ -1,20 +1,37 @@
...
+    "value": {
+      "title": "national_infected_age_groups_value",
       "additionalProperties": false,
       "type": "object",
       "required": [
-        "agegroup",
-        "infected_per_agegroup_increase",
+        "age_group",
+        "infected_percentage",
+        "national_age_group_percentage",
```
VWSCoronaDashboard8, line 23: See comment below.

src/types/data.d.ts
```diff
@@ -220,13 +220,13 @@ export interface InfectedPeopleDeltaNormalizedLastValue {
   date_of_report_unix: number;
   date_of_insertion_unix: number;
 }
-export interface IntakeShareAgeGroups {
-  values: IntakeShareAgeGroupsLastValue[];
-  last_value: IntakeShareAgeGroupsLastValue;
+export interface InfectedAgeGroups {
```
VWSCoronaDashboard8, line 223: Seems like you forgot to re-generate the types

VWSCoronaDashboard8 (changes requested):

VWSCoronaDashboard8 (approved):

src/domain/infected-people/age-demographic/age-demographic.tsx
```diff
@@ -0,0 +1,90 @@
...
+  data: NationalInfectedAgeGroups;
+}
+
+interface TooltipOptions {
+  left: number;
+  top: number;
+  value: NationalInfectedAgeGroupsValue;
+}
+
+const Wrapper = styled.div(
```
VWSCoronaDashboard8, line 23: I think Box also has position properties, so you can just reuse that.

PR #917, opened 2020-11-12
Fix gedrag line chart interaction

VWSCoronaDashboard10:
This PR fixes a nasty bug in relation to highcharts hover styles. I'm afraid I had to remove the hover style in order to get rid of a glitch where it became difficult to click an actual line.


src/domain/behavior/behavior-line-chart-tile.tsx
```diff
@@ -69,39 +74,37 @@ export function BehaviorLineChartTile({
...
+        <BehaviorTypeControl value={type} onChange={setType} />
+      </Box>
+
+      <Box
+        display="flex"
+        alignItems="baseline"
+        flexDirection={{ _: 'column', lg: 'row' }}
+      >
+        <Box flex="1" mr={{ lg: 2 }}>
+          <p>{text.basisregels.intro[type]}</p>
```
VWSCoronaDashboard4, line 93: This one needs to have a separate intro for the linechart

PR #920, opened 2020-11-13
Adjusted aligment of metadata

VWSCoronaDashboard4:
## Summary

Properly aligns all metadata in all places, not just (derivatives of) Tiles.

### DIFF
--- a/src/components-styled/metadata.tsx
+++ b/src/components-styled/metadata.tsx
@@ -29,7 +29,7 @@ function formatMetadataDate(date: number | [number, number]): string {
 export function Metadata({ date, source }: MetadataProps) {
   const dateString = date ? formatMetadataDate(date) : null;
   return (
-    <Box as="footer" mt={3} gridArea="metadata">
+    <Box as="footer" mt={3} mb={-3} gridArea="metadata">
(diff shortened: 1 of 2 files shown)


src/components-styled/metadata.tsx
```diff
@@ -29,7 +29,7 @@ function formatMetadataDate(date: number | [number, number]): string {
 export function Metadata({ date, source }: MetadataProps) {
   const dateString = date ? formatMetadataDate(date) : null;
   return (
-    <Box as="footer" mt={3} gridArea="metadata">
+    <Box as="footer" mt={3} mb={-3} gridArea="metadata">
```
VWSCoronaDashboard10, line 32: I think components in the `components-styled` directory should not have outside margins, especially negative margins. Instead this kind of layout should be controlled from the outside.

Also be careful with this change because it looks like the Metadata component is also used without the wrapping Tile component.

VWSCoronaDashboard1: I think Thijs has made comments about adding outside margins inside components. It's probably a good idea to use this as a rule of thumb.

VWSCoronaDashboard1 (approved):

PR #928, opened 2020-11-16
Add maatregelen page to landelijk and regional

VWSCoronaDashboard1:
## Summary

This adds the restrictions pages to the national and regional sections.
A new menu item is added to both national and regional sidebars that link to the respective pages.

## Motivation

Feature request

## Detailed design

The bulk of the logic to shape the data for the restrictions table is in the hook `useRestrictionsTable`. The result of this hook is used inside the RestrictionsTable component.

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added


src/assets/restrictions/Verzorgingstehuis.svg
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 17.1.0, SVG Export Plug-In . SVG Version: 6.00 Build 0) -->
+<svg width="32px" height="32px" viewBox="0 0 64 64">
```
VWSCoronaDashboard4, line 3: A lot of these icons contain things that are not needed or not handy. Eg: comments, empty (background?) rectangles, width/height attributes on <svg> (don't work well with resizing when embedded in HTML).

We could add a task to the package to clean up icon svgs (nice one for the tech debt backlog)

src/pages/landelijk/ziekenhuis-opnames.tsx
```diff
@@ -47,6 +47,11 @@ const IntakeHospital: FCWithLayout<INationalData> = (props) => {
         category={siteText.nationaal_layout.headings.ziekenhuizen}
         title={text.titel}
         Icon={Ziekenhuis}
+        iconAttrs={{
+          style: {
+            transform: 'scale(1.5)',
```
VWSCoronaDashboard4, line 52: Why is there need to apply this transform everywhere? Is there something wrong/different about the icons? Can this be fixed somewhere else, in the styling of this component?

It would be clearer and cleaner if we can avoid using these.

src/locale/en.json
```diff
@@ -688,6 +876,51 @@
...
+    "VR06": "https://www.vnog.nl/",
+    "VR07": "http://www.vggm.nl/vggm",
+    "VR08": "http://www.vrgz.nl/",
+    "VR09": "http://www.vru.nl/",
+    "VR10": "http://www.vrnhn.nl/",
+    "VR11": "https://www.veiligheidsregiozaanstreekwaterland.nl/",
+    "VR12": "http://www.vrk.nl/",
+    "VR13": "https://www.amsterdam.nl/veiligheidsregio/",
+    "VR14": "https://www.vrgooienvechtstreek.nl/",
+    "VR15": "Haaglanden",
```
VWSCoronaDashboard4, line 912: This is not a URL.

src/utils/useRestrictionsTable.ts
```diff
@@ -0,0 +1,88 @@
...
+) {
+  return useMemo(() => {
+    const table: RestrictionsTableData = {
+      rows: createRows(data),
+    };
+    return table;
+  }, [data]);
+}
+
+function createRows(
```
VWSCoronaDashboard4, line 63: Can you add a few comments for these methods?

src/pages/veiligheidsregio/[code]/maatregelen.tsx
```diff
@@ -0,0 +1,146 @@
...
+          </Text>
+        </Box>
+        <RestrictionsTable
+          data={restrictionsTable}
+          escalationLevel={escalationLevel.escalation_level}
+        />
+      </KpiSection>
+
+      <KpiSection display="flex" flexDirection={['column', 'row']}>
+        <Box borderRight={{ lg: '1px solid lightgrey' }} pr={[0, 2]}>
```
Comment (VWSCoronaDashboard4) [inline src/pages/veiligheidsregio/[code]/maatregelen.tsx:118]: This is not re-using the lightGray variable as defined in the styled system variables.

src/components/contentHeader/index.tsx
```diff
@@ -1,9 +1,11 @@
-import { Metadata } from './metadata';
+import { ReactNode } from 'react';
+import { Box } from '~/components-styled/base';
 import { TitleWithIcon } from '~/components/titleWithIcon';
 import styles from '../layout/layout.module.scss';
+import { Metadata } from './metadata';
 
-export function ContentHeader(props: IContentHeaderProps) {
-  const { category, Icon, title, subtitle, metadata, id } = props;
+export function GenericContentHeader(props: ISmallContentHeaderProps) {
```
VWSCoronaDashboard4, line 7: Instead of creating a styled system component next to an old component, you can create a styled system component in the `components-styled` folder with the same (or similar) name: `components-styled/content-header`. Then for the use new uses, just import that one. Then this file can be reverted.

VWSCoronaDashboard4 (changes requested):

src/assets/restrictions/Verzorgingstehuis.svg
VWSCoronaDashboard1: svgo/webpack is installed, so the optimizer get rid of all the garbage in the files.

src/locale/en.json
VWSCoronaDashboard1: fixed

src/utils/useRestrictionsTable.ts
VWSCoronaDashboard1: done

src/pages/veiligheidsregio/[code]/maatregelen.tsx
Comment (VWSCoronaDashboard1) [reply src/pages/veiligheidsregio/[code]/maatregelen.tsx:118]: Yes, for some reason I couldn't get the separate border attributes to work responsively unless I stuck them in a combined attr like here.

src/components/contentHeader/index.tsx
VWSCoronaDashboard1: I'm not sure I understand what you mean. I split up the ContentHeader component into two smaller components because on the restrictions page I needed only to show part of the ContentHeader component.

src/pages/landelijk/ziekenhuis-opnames.tsx
VWSCoronaDashboard1: The icons were cleaned up by one of the designers. Resulting in the icons not rendering in consistent sizes. I've been battling viewboxes and widths and heights for about a day, in the end this was the only way of getting everything to render in the same size.

src/scss/layout.scss
```diff
@@ -220,6 +220,13 @@
       min-height: 0;
     }
   }
+
+  .restrictions {
```
VWSCoronaDashboard8, line 224: Do we really need to add global CSS classes for newly built features? I suspect you can place this in a styled component.

src/types/data.d.ts
```diff
@@ -530,7 +532,7 @@ export interface RegionalRestrictions {
 export interface RegionalRestrictionValue {
   restriction_id: string;
   target_region: "nl" | "vr";
-  escalation_level: 0 | 1 | 2 | 3 | 4 | 41 | 401;
+  escalation_level: 0 | 1 | 2 | 3 | 4 | 41 | 401 | 402;
```
VWSCoronaDashboard8, line 535: Probably too late to change now, but strings would have been a better choice I think if we're going to use sub-levels. So we can write "4a" "4b" etc

src/utils/useEscalationColor.ts
```diff
@@ -0,0 +1,20 @@
+import { useMemo } from 'react';
+import { regionThresholds } from '~/components/choropleth/region-thresholds';
+import { ChoroplethThresholds } from '~/components/choropleth/shared';
+
+export function useEscalationColor(level: number | undefined) {
```
VWSCoronaDashboard8, line 5:
```suggestion
export function useEscalationColor(level?: number) {
```

src/utils/useEscalationColor.ts
```diff
@@ -0,0 +1,20 @@
...
+  return useMemo(() => {
+    if (level === undefined) {
```
VWSCoronaDashboard8, line 7: Might be worth doing an assert instead to fail hard. I think you might want to break the app if the data is not compatible with the escalation levels you can handle in the code.

src/utils/useRestrictionsTable.ts
```diff
@@ -0,0 +1,100 @@
+import { useMemo } from 'react';
```
VWSCoronaDashboard8, line 1: IMO utils are generic simple functions that are used commonly in different features. This is very specific to the restrictions feature, so there's nothing utilitarian about it. Do we already have that domain folder? I would suggest to move it there.

src/components/contentHeader/index.tsx
VWSCoronaDashboard4: This is now a mixed component, while for the styled system components there is a separate folder.

If we implement a new component in the styled system folder, we can leave the old one untouched.

src/utils/useRestrictionsTable.ts
```diff
@@ -0,0 +1,100 @@
...
+ * Creates unique category rows sorted based on the hardcoded rowOrder list.
+ * Each row has a category as its title and holds a list of columns
+ */
+function createRows(
+  data: (RegionalRestrictionValue | NationalRestrictionValue)[]
+) {
+  const uniqueCategories = data
+    .map((value) => value.category_id)
+    .filter((value, index, list) => list.indexOf(value) === index)
+    .sort((left, right) => rowOrder.indexOf(left) - rowOrder.indexOf(right));
```
VWSCoronaDashboard8, line 77: This code is a bit cryptic. Creating a unique array with the filter especially.

It should be possible to greatly simplify this code by combining it with the map logic below, using a single reduce function. But it's up to you if want to take this challenge ;)

VWSCoronaDashboard8 (changes requested):

src/types/data.d.ts
VWSCoronaDashboard1: not my choice, this has been a decision made somewhere by someone in the data analysis process

src/utils/useRestrictionsTable.ts
VWSCoronaDashboard1: I've created named functions, I think it's pretty straightforward now what is happening. Have a look.

src/pages/veiligheidsregio/[code]/maatregelen.tsx
Comment (VWSCoronaDashboard10) [reply src/pages/veiligheidsregio/[code]/maatregelen.tsx:118]: But then still that color won't match the theme's `lightGray`

src/types/data.d.ts
VWSCoronaDashboard8: But we aren't always a slave to their internal design choices I think :) If we define a schema for it won't they have to map their values to it?

src/pages/veiligheidsregio/[code]/maatregelen.tsx
Comment (VWSCoronaDashboard1) [reply src/pages/veiligheidsregio/[code]/maatregelen.tsx:118]: haha, yes, fair point, I'll change it

src/utils/useRestrictionsTable.ts
VWSCoronaDashboard8: To be honest I prefer the first version since this adds a lot more lines of code. I meant to suggest that a reduce could simplify and shrink the code.  But this is nitpicking so I'm fine if you leave it like that 👍

src/pages/landelijk/maatregelen.tsx
```diff
@@ -0,0 +1,81 @@
...
+          metricValueName="escalation_level"
+          onSelect={createSelectRegionHandler(router, 'maatregelen')}
+          tooltipContent={escalationTooltip(router)}
+        />
+      </ChoroplethTile>
+
+      <KpiSection display="flex" flexDirection="column">
+        <Heading level={3}>{restrictionInfo.extratoelichting.titel}</Heading>
+        <RestrictionsTable
+          data={restrictionsTable}
```
VWSCoronaDashboard10, line 71: Why didn't you pass `data.restrictions` to this component and run the `useRestrictionsX`-hooks in there?

src/utils/useRestrictionLevel.ts
```diff
@@ -0,0 +1,18 @@
...
+
+const codes = [1, 2, 3, 4, 401, 402, 41];
+
+export function useRestrictionLevel(
+  values: (RegionalRestrictionValue | NationalRestrictionValue)[]
+) {
+  return useMemo(() => {
+    const top = values.reduce((max, item) => {
+      return Math.max(max, codes.indexOf(item.escalation_level));
+    }, 0);
```
VWSCoronaDashboard10, line 15: nitpick; Sometimes a reduce is a bit overkill:
```tsx
const top = Math.max(
 ...values.map((item) => codes.indexOf(item.escalation_level))
)
```

src/components/contentHeader/index.tsx
```diff
@@ -17,21 +19,46 @@ export function ContentHeader(props: IContentHeaderProps) {
   return (
     <header id={id} className={layoutClasses.join(' ')}>
       {category && <p className={styles.category}>{category}</p>}
-      <TitleWithIcon Icon={Icon} title={title} as="h2" />
+      <TitleWithIcon Icon={Icon} iconAttrs={iconAttrs} title={title} as="h2" />
+      {children}
+      {!children && <Box mb={4} />}
```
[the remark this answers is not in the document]
VWSCoronaDashboard8: I don't see a problem with this if contentHeader is providing the logic for the layout it makes sense that margins are set in this context. I only see problems with components settings margins themselves without knowing for sure how and where they are used.

Could you elaborate on this @VWSCoronaDashboard10 ?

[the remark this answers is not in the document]
VWSCoronaDashboard8: Yeah I see why it's a bit dodgy. If the header needs an inside margin always it should not depend on whether it is rendering children / metadata.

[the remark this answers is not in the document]
VWSCoronaDashboard10: Applying a padding would be better, simply add a bottom margin on the last component might result in collapsing margins with sibling components which in turn can result in unexpected and hard to fix layout quirks.

src/components/restrictions/hooks/useRestrictionsTable.ts
```diff
@@ -0,0 +1,118 @@
...
+  const sortedRows = data
+    .map(selectCategory)
+    .filter(makeUniqueList)
+    .sort(sortByRowOrder);
+
+  return sortedRows.map<RestrictionsRowData>((category) => ({
+    categoryColumn: category,
+    restrictionsColumn: createColumn(data, category),
+  }));
+}
```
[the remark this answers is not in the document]
VWSCoronaDashboard10: Haha, now I see [an older comment](https://github.com/minvws/nl-covid19-data-dashboard/pull/928#discussion_r524998562) of @VWSCoronaDashboard8 which resulted in this verbose function 🤷  I guess its a nitpick thing so do with it whatever feels best :-)

schema/regional/__index.json
```diff
@@ -11,7 +11,7 @@
     "ggd",
     "nursing_home",
     "sewer",
-    "sewer_per_installation",
```
VWSCoronaDashboard8, line 14: Sewer per installation got tossed out? Did you also remove that json file then or is this a mistake?

src/domain/behavior/behavior-metric.tsx
```diff
@@ -13,6 +13,10 @@ interface BehaviorMetricProps {
 }
 
 export function BehaviorMetric({ data }: BehaviorMetricProps) {
+  if (!data?.last_value) {
```
VWSCoronaDashboard8, line 16: What's the use-case for not having a last_value? Can it happen according to the data source?

VWSCoronaDashboard8 (changes requested):
Looks to me that regional/sewer_per_installation got removed by accident.

src/utils/useRestrictionLevel.ts
```diff
@@ -6,7 +6,7 @@ import {
 
 const codes = [1, 2, 3, 4, 401, 402, 41];
 
-export function useRestrictionLevel(
+export function useEscalationLevel(
```
VWSCoronaDashboard8, line 9: The file is called useRestrictionLevel.

The filename should also be kebab-cased.

src/components/restrictions/restrictionsTable.tsx
```diff
@@ -1,7 +1,7 @@
 import { Fragment } from 'react';
```
VWSCoronaDashboard8, line 1: filename ☝️

src/components/restrictions/hooks/useRestrictionsTable.ts
```diff
@@ -13,7 +13,6 @@ const restrictionTexts: Record<string, string> = siteText.maatregelen.teksten;
  * is rendered in the table.
```
VWSCoronaDashboard8, line 13: filename ☝️

VWSCoronaDashboard8 (approved):

src/static-props/data-sorting.ts
```diff
@@ -1,7 +1,9 @@
 import { Municipal, National, Regionaal } from '~/types/data.d';
 
 export function sortNationalTimeSeriesInDataInPlace(data: National) {
-  const timeSeriesPropertyNames = getTimeSeriesPropertyNames(data);
+  const timeSeriesPropertyNames = getTimeSeriesPropertyNames(data).filter(
+    (propertyName) => propertyName !== 'restrictions'
```
VWSCoronaDashboard10, line 5: do we need to add a comment on why we filter `restrictions` out?

src/scss/layout.scss
VWSCoronaDashboard10: this ☝️

src/pages/veiligheidsregio/[code]/maatregelen.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+  const regioUrl = siteText.veiligheidsregio_maatregelen_urls[code];
+
+  const restrictionsTable = useRestrictionsTable(data.restrictions.values);
+
+  const escalationLevel = useEscalationLevel(data.restrictions.values);
+
+  // Colors etc are determined by the effective escalation level which is 1, 2, 3 or 4.
+  const effectiveEscalationLevel: 1 | 2 | 3 | 4 =
+    escalationLevel > 4 ? 4 : (escalationLevel as 1 | 2 | 3 | 4);
```
Comment (VWSCoronaDashboard10) [inline src/pages/veiligheidsregio/[code]/maatregelen.tsx:43]: it feels like when `escalationLevel > 4 ` happens there should be something wrong in the delivered data no?

src/pages/landelijk/maatregelen.tsx
VWSCoronaDashboard10: I still believe it makes sense to call the `useRestrictionsTable` hook from within the RestrictionsTable component. These two are tightly coupled and the data is not used outside of that table.

src/components-styled/escalation-level.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  return (
+    <Box
+      display="flex"
+      flexDirection="row"
+      justifyContent="flex-start"
+      alignItems="center"
+      mt={1}
+    >
+      <Text as="span" marginLeft="0 !important" marginRight=".5em !important">
```
VWSCoronaDashboard10, line 48: Are the `!important` really necessary? It's code smell..

src/components-styled/content-header/index.tsx
```diff
@@ -161,9 +167,9 @@ export function ContentHeader(props: ContentHeaderProps) {
 interface ContentHeaderProps {
   id?: string;
   title: string;
-  subtitle: string;
+  subtitle?: string;
   metadata?: MetadataProps;
-  reference: {
+  reference?: {
```
VWSCoronaDashboard10, line 172: Maybe @VWSCoronaDashboard4 can shine his light on this too, but he corrected me previously that this reference prop should never be optional.

src/components-styled/base/box.tsx
```diff
@@ -26,9 +34,13 @@ export type BoxProps = SpaceProps &
   ColorProps &
   PositionProps &
   TypographyProps &
-  BorderProps &
+  BordersProps &
   ShadowProps &
-  GridProps;
+  GridProps &
+  BackgroundImageProps &
```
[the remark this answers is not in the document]
VWSCoronaDashboard10: It also doesn't look like the background props are used in this PR, so I suggest to remove them again from the box component.. (and for future reference: there's also the combination of all separate styles, called `backgroundProps`)

src/assets/restrictions/Vinkje.svg
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
```
VWSCoronaDashboard10, line 1: filename lowercase

src/assets/restrictions/Winkelwagen.svg
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
```
VWSCoronaDashboard10, line 1: filename lowercase

src/assets/restrictions/Verzorgingstehuis.svg
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="utf-8"?>
```
VWSCoronaDashboard10, line 1: filename lowercase

src/assets/restrictions/Rugzak.svg
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
```
VWSCoronaDashboard10, line 1: filename lowercase, etc :-)

src/types/data.d.ts
VWSCoronaDashboard1: Yes, but it's a little late for that now I'm afraid...

schema/regional/__index.json
VWSCoronaDashboard1: This is very probably a merge mistake, I'll take a look

VWSCoronaDashboard1: I just checked my local file and "sewer_per_installation" is still in there. I think github is showing the changes a little wonky...

src/pages/veiligheidsregio/[code]/maatregelen.tsx
Comment (VWSCoronaDashboard1) [reply src/pages/veiligheidsregio/[code]/maatregelen.tsx:41]: No, there'is the escalation level which can be 1,2,3,4,401,402 or 41, on that level the list of restrictions is based. Then there's the effective level which is a simpler one: 1,2,3,4, which is communicated to the user.

Comment (VWSCoronaDashboard1) [reply src/pages/veiligheidsregio/[code]/maatregelen.tsx:41]: I've added a code comment that explains this

src/utils/useEscalationColor.ts
VWSCoronaDashboard1: I've added another assert as well, in case a color could not be found for the given level

PR #929, opened 2020-11-16
Styling adjustments for 1 KPI tile on a row

VWSCoronaDashboard4:
## Summary

Adjust TwoKpiSection/Tile components to display 1 Tile with the new design.

## Detailed design

* Use Tiles on all leftover pages that were previously blocked.
* Apply new layout when 1 Tile is used in TwoKpiSection: full width, description with a max width of 1 column (400px).
* Fix padding issue with reproduction number tile along the way.
* Add (temporary) `dataWarning` attribute to the Tile so we can use the new design on the Ziekenhuis-tiles as well.

## Unresolved questions

* Should we rename _TwoKpiSection_ to something different, now that it is optimized for 1 or 2 Kpi Tiles?

### DIFF
--- a/src/components-styled/kpi-tile.tsx
+++ b/src/components-styled/kpi-tile.tsx
@@ -1,13 +1,15 @@
 import { Box, Spacer } from './base';
-import { Text, Heading } from './typography';
+import { Heading } from './typography';
 import { Tile } from './layout';
 import { MetadataProps, Metadata } from './metadata';
+import { DataWarning } from '~/components/dataWarning';
 
 interface KpiTileProps {
   title: string;
   description?: string;
   children: React.ReactNode;
   metadata: MetadataProps;
+  dataWarning?: boolean /* TODO: remove this temporary attribute when it is not used anymore */;
 }
 
 /**
@@ -19,14 +21,18 @@ export function KpiTile({
   description,
   children,
   metadata,
+  dataWarning,
 }: KpiTileProps) {
   return (
     <Tile height="100%">
+      {dataWarning && <DataWarning />}
       <Heading level={3}>{title}</Heading>
       <Box>{children}</Box>
       {description && (
-        <Text
+        <Box
           as="div"
+          maxWidth="400px"
+          mt="3"
           dangerouslySetInnerHTML={{
             __html: description,
           }}
--- a/src/components-styled/kpi-tile.tsx
+++ b/src/components-styled/kpi-tile.tsx
@@ -9,7 +9,7 @@ interface KpiTileProps {
   description?: string;
   children: React.ReactNode;
   metadata: MetadataProps;
-  dataWarning?: boolean /* TODO: remove this temporary attribute when it is not used anymore */;
+  showDataWarning?: boolean /* TODO: remove this temporary attribute when it is not used anymore */;
 }
 
 /**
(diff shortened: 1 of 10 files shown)


VWSCoronaDashboard1 (approved):

src/components-styled/kpi-tile.tsx
```diff
@@ -1,13 +1,15 @@
...
 import { Tile } from './layout';
 import { MetadataProps, Metadata } from './metadata';
+import { DataWarning } from '~/components/dataWarning';
 
 interface KpiTileProps {
   title: string;
   description?: string;
   children: React.ReactNode;
   metadata: MetadataProps;
+  dataWarning?: boolean /* TODO: remove this temporary attribute when it is not used anymore */;
```
VWSCoronaDashboard8, line 12: This doesn't read like a boolean.
```suggestion
  showDataWarning?: boolean /* TODO: remove this temporary attribute when it is not used anymore */;
```

src/components-styled/kpi-tile.tsx
```diff
@@ -19,14 +21,18 @@ export function KpiTile({
...
     <Tile height="100%">
+      {dataWarning && <DataWarning />}
       <Heading level={3}>{title}</Heading>
       <Box>{children}</Box>
       {description && (
-        <Text
+        <Box
           as="div"
+          maxWidth="400px"
+          mt="3"
```
VWSCoronaDashboard8, line 35: I think this should be a number? Maybe styled system casts strings to numbers internally, but still a number would be correct I think.

src/components-styled/kpi-tile.tsx
```diff
@@ -19,14 +21,18 @@ export function KpiTile({
...
+  dataWarning,
 }: KpiTileProps) {
   return (
     <Tile height="100%">
+      {dataWarning && <DataWarning />}
       <Heading level={3}>{title}</Heading>
       <Box>{children}</Box>
       {description && (
-        <Text
+        <Box
```
VWSCoronaDashboard10, line 32: Did you change this component on purpose? Looks like the Text component has some default styling applied (e.g. fontSize=2) which is now lost by going for a Box.

VWSCoronaDashboard10 (approved):
Approved, only couple of questions.

src/components-styled/two-kpi-section.tsx
```diff
@@ -36,10 +36,14 @@ export function TwoKpiSection({ children, ...props }: TwoKpiSectionProps) {
       mr={{ _: -4, sm: 0 }}
       {...(props as any)}
     >
-      <Box flex="1 1 50%" mr={{ lg: 3 }} mb={{ _: 4, lg: 0 }}>
+      <Box flex={`1 1 ${childrenCount === 2 ? 50 : 100}%`} mb={{ _: 4, lg: 0 }}>
         {childrenArray[0]}
       </Box>
-      <Box flex="1 1 50%">{childrenArray[1]}</Box>
+      {childrenArray[1] && (
```
VWSCoronaDashboard8, line 42: I would prefer to use childrenCount here too, since that was extracted from the array and the intention is clearer then IMO

```suggestion
      {childrenCount == 2 && (
```

Might be worth creating a `hasTwo` boolean even, so you can use that in these two places.

VWSCoronaDashboard8 (changes requested):

VWSCoronaDashboard8 (approved):

PR #933, opened 2020-11-17
Clean up chart components

VWSCoronaDashboard10:
I ran into an issue where things like titles in combination with timeframes were rendered different across several implementations. This PR will solve these issues by rendering all charts with the same containing components.

These are the new components available for rendering charts:

```jsx
// plain chart:
<ChartTile showDialogWarning title description metadata>
  <SomeChart />
</ChartTile>

// chart with timeframe:
<ChartTileWithTimeframe showDialogWarning title description metadata timeframeOptions timeframeInitialValue>
  {(timeframe) => (
      <SomeChartWithTimeframe  {...} />
  )}
</ChartTileWithTimeframe>
```

Some charts already have a wrapping component, for example the LineChart, which in turn will now depend on one of the components above.


src/components-styled/chart-tile-container.tsx
```diff
@@ -0,0 +1,29 @@
+import { DataWarning } from '~/components/dataWarning';
+import { Spacer } from './base';
+import { Tile } from './layout';
+import { Metadata, MetadataProps } from './metadata';
+
+export function ChartTileContainer({
```
VWSCoronaDashboard8, line 6: Seems like this component is only to be used as part of the ChartTile. Maybe inline it there in the file or use folder structure to keep it local.

src/components-styled/chart-tile.tsx
```diff
@@ -0,0 +1,105 @@
...
+interface ChartTileProps {
+  children: React.ReactNode;
+  metadata: MetadataProps;
+  title: string;
+  description?: React.ReactNode;
+  showDataWarning?: boolean;
+}
+
+interface ChartTileWithTimeframeProps extends Omit<ChartTileProps, 'children'> {
+  children: (timeframe: TimeframeOption) => React.ReactNode;
```
VWSCoronaDashboard8, line 18: I think you can make TimeframeOption generic (using T). Then the component can derive it from the props. Also then the tile can be used for timeframes and other types of switches potentially.

You can than rename the component to something like `MultipleChartTile` or `SwitchableChartTile` or ...

src/components-styled/chart-tile.tsx
```diff
@@ -0,0 +1,105 @@
...
+  );
+
+  return (
+    <ChartTileContainer metadata={metadata} showDataWarning={showDataWarning}>
+      <ChartTileHeader
+        title={title}
+        description={description}
+        timeframe={timeframe}
+        timeframeOptions={timeframeOptions}
+        onTimeframeChange={setTimeframe}
```
VWSCoronaDashboard8, line 58: And this could be generic as well. `onChartChange` or something

VWSCoronaDashboard8 (approved):
Nice one! I think timeframe options should be abstracted further. But I suggest postponing that because this PR is already quite large and there are lots of other things to do in this sprint already.

VWSCoronaDashboard10:
> Nice one! I think timeframe options should be abstracted further. But I suggest postponing that because this PR is already quite large and there are lots of other things to do in this sprint already.

Nice suggestion, but as you suggested I'll pick that up in the future.

src/components-styled/chart-tile.tsx
```diff
@@ -0,0 +1,105 @@
...
+      {children}
+    </ChartTileContainer>
+  );
+}
+
+export function ChartTileWithTimeframe({
+  title,
+  description,
+  metadata,
+  timeframeOptions = ['all', '5weeks', 'week'],
```
VWSCoronaDashboard4, line 42:
```suggestion
  timeframeOptions = TimeframeOption[],
```

src/components-styled/chart-tile-container.tsx
VWSCoronaDashboard10: Nope, it's also used on other places (e.g. choropleth-tile) so I'll leave it here for now.

src/components-styled/chart-tile.tsx
```diff
@@ -0,0 +1,105 @@
...
+  description?: React.ReactNode;
+  timeframe?: TimeframeOption;
+  timeframeOptions?: TimeframeOption[];
+  onTimeframeChange?: (timeframe: TimeframeOption) => void;
+}) {
+  return (
+    <Box
+      mb={3}
+      display="flex"
+      flexDirection={['column', null, null, null, 'row']}
```
VWSCoronaDashboard4, line 82: why not use the named versions for the mediaqueries?
```suggestion
      flexDirection={{ xs: 'column', xl: 'row' }}
```

VWSCoronaDashboard4 (approved):
Some small questions/suggestions. Nice cleanup!

src/components-styled/chart-tile.tsx
VWSCoronaDashboard10: This is a default parameter instead of a type definition :-)

VWSCoronaDashboard1 (approved):

src/components-styled/chart-tile.tsx
VWSCoronaDashboard1: I accidentally merged this suggestion, I was getting tabs mixed up and thought I was looking at one of my own PRs haha. My apologies, please revert!

PR #959, opened 2020-11-18
Adjust colors of charts and figures

VWSCoronaDashboard10:
Updated the colors of choropleths/line-charts/bar-charts and the kpi/figures components.

- moved a lot of colors to the theme
- Prepared niveau/escalation level rendering for level 5 (was part of the design)


src/style/theme.ts
```diff
@@ -105,7 +126,7 @@ const theme: TDashboardTheme = {
   breakpoints: breakpoints as any,
   mediaQueries,
   space,
-  colors,
+  colors: (colors as unknown) as ThemeValue<CSS.Property.Color>,
```
VWSCoronaDashboard4, line 129: Perhaps assign this type to where colors is defined? ( `export const colors = {` )

VWSCoronaDashboard4 (approved):
:100:

src/components-styled/legenda.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+interface LegendProps {
+  items: Array<{ color: string; label: string; shape: Shape }>;
+}
+
+export function Legenda({ items }: LegendProps) {
+  return (
+    <List>
+      {items.map(({ label, color, shape = 'line' }, i) => (
+        <Item key={i}>
```
0x80, line 15: nit; I think the label generally makes more sense to use as the key. Index can be ok, but in a lot of cases it's not so I'd stay away from it.

src/style/theme.ts
VWSCoronaDashboard10: I'd rather export the colors as a constant type because that simplifies using that colors object with TS, downside is that our theme does not accept such a type, hence the type-cast.

src/components/choropleth/shared.d.ts
```diff
@@ -46,11 +46,11 @@ export type RegionGeoJSON = FeatureCollection<
...
 >;
 
-export type ChoroplethThresholds = {
-  thresholds: ChoroplethThresholdsValue[];
+export type ChoroplethThresholds<T extends number = number> = {
+  thresholds: ChoroplethThresholdsValue<T>[];
 };
 
-export type ChoroplethThresholdsValue = {
+export type ChoroplethThresholdsValue<T extends number = number> = {
```
0x80, line 53: Is this a default? What is it for? Wouldn't T always be defined by the input value of threshold?

src/components/choropleth/tooltips/region/escalation-tooltip.tsx
```diff
@@ -37,18 +32,7 @@ export const escalationTooltip = (router: NextRouter) => {
...
-              {thresholdInfo.threshold === 2 && (
-                <EscalationLevel2 color={thresholdInfo.color} />
-              )}
-              {thresholdInfo.threshold === 3 && (
-                <EscalationLevel3 color={thresholdInfo.color} />
-              )}
-              {thresholdInfo.threshold === 4 && (
-                <EscalationLevel4 color={thresholdInfo.color} />
-              )}
+              <EscalationLevelIcon level={thresholdInfo.threshold} />
```
0x80, line 35: Cleans up nicely 👍

src/components/lineChart/hooks/useRegionalSewerWaterChartOptions.ts
```diff
@@ -112,7 +113,7 @@ export function useRegionalSewerWaterChartOptions(
       name: text.average_label_text,
       description: text.average_label_text,
       showInLegend: true,
-      color: selectedRWZI ? '#A9A9A9' : '#3391CC',
+      color: selectedRWZI ? '#A9A9A9' : colors.chart.blue,
```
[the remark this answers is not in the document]
VWSCoronaDashboard10: That can be cleaned up within another PR, the gray colors are not touched at all.

src/components/lineChart/hooks/useRegionalSewerWaterChartOptions.ts
```diff
@@ -171,7 +172,7 @@ export function useRegionalSewerWaterChartOptions(
           description: replaceVariablesInText(text.daily_label_text, {
             name: selectedRWZI,
           }),
-          color: '#004277',
+          color: colors.chart.blueDark,
```
0x80, line 175: If we already have multiple levels of nesting I think the camel casing could as well be replaced by an extra level. Might read a bit nicer. 
```suggestion
          color: colors.chart.blue.dark,
```

0x80: Not sure if the chart sub-level is preferable. I don't expect the chart to have completely different blues from the other components. For now it's fine. We can also change it once we start duplicating colors.

src/style/theme.ts
```diff
@@ -79,15 +84,31 @@ const mediaQueries = {
...
   shadow: '#e5e5e5',
   lightGray: '#dfdfdf',
   annotation: '#595959',
-};
+
+  chart: {
+    blue: '#007BC7',
+    blueFill: 'rgba(0, 123, 199, .05)',
+    blueDark: '#154273',
+    blueScale: [
```
0x80, line 101: As I said earlier, I think nesting fill, dark, scale etc under blue might be a bit nicer. Especially once we start adding colors for magenta or other things. Then the structure can be mirrored.

0x80 (approved):

src/components-styled/legenda.tsx
VWSCoronaDashboard8, line 15: nit; I think the label generally makes more sense to use as the key. Index can be ok, but in a lot of cases it's not so I'd stay away from it.

src/components/choropleth/shared.d.ts
VWSCoronaDashboard8: Is this a default? What is it for? Wouldn't T always be defined by the input value of threshold?

src/components/choropleth/tooltips/region/escalation-tooltip.tsx
VWSCoronaDashboard8: Cleans up nicely 👍

src/components/lineChart/hooks/useRegionalSewerWaterChartOptions.ts
VWSCoronaDashboard8: If we already have multiple levels of nesting I think the camel casing could as well be replaced by an extra level. Might read a bit nicer. 
```suggestion
          color: colors.chart.blue.dark,
```

Not sure if the chart sub-level is preferable. I don't expect the chart to have completely different blues from the other components. For now it's fine. We can also change it once we start duplicating colors.

src/style/theme.ts
VWSCoronaDashboard8: As I said earlier, I think nesting fill, dark, scale etc under blue might be a bit nicer. Especially once we start adding colors for magenta or other things. Then the structure can be mirrored.

src/components-styled/legenda.tsx
VWSCoronaDashboard10: I understand, but in this case I'd like to keep it like this. Using the label could result in a react error (its not guaranteed unique), it would be necessary to concat the index with the label..

src/components/choropleth/shared.d.ts
VWSCoronaDashboard10: Yes it's the default type. The `ChoroplethThresholdsValue` type is not only used where we define the value, but also for example as function parameter type. The default type assignment keeps things simple, though dynamic enough to be typed as, e.g., `1 | 2 | 3`.

src/components/lineChart/hooks/useRegionalSewerWaterChartOptions.ts
VWSCoronaDashboard10: The charts use a different blue compared to the UI elements.. Only the KPI values will use the same blue as the chart, so I'd like to keep the chart level in there.

What would you have in mind as replacement for the `chart.blue` color?
`chart.blue.[normal | default | standard | whale | etc]`?

VWSCoronaDashboard10: brainfart, but we could rename `chart` to something like `metric` (suggestions are welcome), this allows us to move the `kpi` colors to that `metric.blue` object.

VWSCoronaDashboard8: Yes I like that idea, maybe `data` could be a more generic container than `metric`

src/style/theme.ts
```diff
@@ -79,15 +84,24 @@ const mediaQueries = {
...
+  data: {
+    primary: '#007BC7',
+    secondary: '#154273',
+    fill: 'rgba(0, 123, 199, .05)',
+    scale: {
+      blue: ['#8FCAE7', '#5BADDB', '#248FCF', '#0070BB', '#00529D', '#003580'],
+      red: ['#F6B4D1', '#D3719C', '#9E3A66', '#64032D', '#000000'],
+    },
+  },
+} as const;
```
VWSCoronaDashboard10, line 104: @VWSCoronaDashboard8 what do you think of this shape?

VWSCoronaDashboard1 (approved):

PR #962, opened 2020-11-18
Link to explanation from all metric pages

VWSCoronaDashboard4:
## Summary

Adds links to explanation from all content headers.

## Detailed design

* Extend content header component and co with `reference`, that should contain a `href` and `text` for the link.
* The link will only be shown if the href and text are non-zero-length strings.
* Extend `Collapsable` component to have a link to the anchor, this link is only visible when the collapsable is openend.


VWSCoronaDashboard10 (approved):

src/components/contentHeader/index.tsx
```diff
@@ -20,7 +20,12 @@ export function ContentHeader(props: IContentHeaderProps) {
       <TitleWithIcon Icon={Icon} title={title} as="h2" />
 
       <div className={styles.text}>
-        <p>{subtitle}</p>
+        <p>
+          {subtitle}{' '}
+          {reference.text && reference.href && (
```
VWSCoronaDashboard8, line 25: nit; I have a feeling that reference could enforce the existence of text and href together. Or are there reference texts that do not have a link associated with them?

src/components/contentHeader/index.tsx
```diff
@@ -42,6 +47,10 @@ interface IContentHeaderProps {
       text: string;
     };
   };
+  reference: {
+    href: string;
+    text: string;
```
VWSCoronaDashboard8, line 52: Looks like they are non-optional? Why the checks above then?

src/components/contentHeader_sourcesHack/index.tsx
```diff
@@ -29,7 +29,12 @@ export function ContentHeader_sourcesHack(props: IContentHeaderProps) {
       <TitleWithIcon Icon={Icon} title={title} as="h2" />
 
       <div className={styles.text}>
-        <p>{subtitle}</p>
+        <p>
+          {subtitle}{' '}
+          {reference.text && reference.href && (
```
VWSCoronaDashboard8, line 34: Same comment as before

src/components/contentHeader_weekRangeHack/index.tsx
```diff
@@ -31,7 +31,12 @@ export function ContentHeader_weekRangeHack(
       <TitleWithIcon Icon={Icon} title={title} as="h2" />
 
       <div className={styles.text}>
-        <p>{subtitle}</p>
+        <p>
+          {subtitle}{' '}
+          {reference.text && reference.href && (
```
VWSCoronaDashboard8, line 36: etc

src/locale/en.json
```diff
@@ -1159,6 +1247,10 @@
       "text": "",
       "href": ""
     },
+    "reference": {
+      "href": "/verantwoording",
+      "text": "Lees meer in de cijferverantwoording."
+    },
```
VWSCoronaDashboard8, line 1253: OK, I'm guessing the checks are there because the strings can be empty. 

I don't think it's worth guarding for in code. I would vote to clean up the code and assume these things are filled correctly. If a string or link is missing it can be caught in QA

VWSCoronaDashboard8 (changes requested):

VWSCoronaDashboard8 (approved):
I'm approving it because the behavior of links is according to the description, but I don't think I agree on hiding the links if strings are empty

src/components/contentHeader/index.tsx
VWSCoronaDashboard4: This check is merely to avoid empty strings for these values.

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard8 (approved):
💯

Issue #956, opened 2020-11-17
Large blobs in page loads, 60+% overhead


MMeent:
### Describe the bug, issue or concern

I noticed that a significant amount of data is stored in the `id=__NEXT_DATA__` script tag (some 515503 bytes, half a megabyte, out of 732471 total for `https://coronadashboard.government.nl`), and the data contained is mostly old (most recent timestamp = 2020-11-10T14:15:36Z).

It currently contains a significant, if not full, set of the daily infection data (accounting for >400kb of the 515kb), and I believe that it would be beneficial to either not include that data or include only the latest up-to-date values, and later request (and store locally) the historical data when required. This would save significantly on page load times, as the  `id=__NEXT_DATA__` script tag is positioned before the rest of the application javascript includes, blocking the page from going responsive. 

It also would save a lot on total traffic, as this data is injected in effectively every page and is therefore not usually cached by the browser.

### To Reproduce
check the size of the `__NEXT_DATA__` script tag 

### Expected behavior
Only the required data is included in the pages `__NEXT_DATA__` script tag

### Environment

**Desktop (please complete the following information):**
 - OS: Debian 10.9
 - Browser: FireFox
 - Version 83.0

### Governance
- [x] I've read the contributing document https://github.com/minvws/.github/blob/master/CONTRIBUTING.md
- [x] I've read and understand the Code of Conduct https://github.com/minvws/.github/blob/master/CODE_OF_CONDUCT.md
- [x] I understand that any contributions or suggestions I made may make it into the actual code. I've read the License https://github.com/minvws/nl-covid19-notification-app-coordination/blob/master/LICENSE.txt and the contributor license agreement https://github.com/minvws/nl-covid19-notification-app-coordination/blob/master/CLA.md

VWSCoronaDashboard10:
Thanks for bringing this up, we’ve had this issue on our list-of-things-to-improve for a while. It shouldn’t effect load time too much due to gzip but it’s worth looking into nonetheless. Will update this thread down the road.

VWSCoronaDashboard:
@MMeent 

First of all, thanks for this! We really appreciate it when people think with us and indicate improvement points in the codebase.

Question for you.

How and where do you measure half a megabyte? I've made a PR (see above comment) to reduce the payload we sent in the homepage, but that's a reduction from ~70KB to ~40KB. Relatively speaking still a big chunk of savings, but nowhere close to the 500KB you are mentioning.

[...]

PR #965, opened 2020-11-18
perf: Remove unused keys from homepages getStaticProps

VWSCoronaDashboard:
## Summary

Solves https://github.com/minvws/nl-covid19-data-dashboard/issues/956

To reduce payload of the homepages HTML file, we need to strip away all of the `values` except the `latest_values` from `NL.json`.

This saves us roughly **423KB deflated**, or **30KB gzipped**.

**Before**
before-detailed

**After**
after-detailed

## Motivation

Shipping smaller websites leads to better performance and less battery usage. 

## Detailed design

I'm doing a naive but perfectly efficient loop over `NL.json` and ~remove the `values`~ assigning an empty array to the values key. Feel free to improve on this method directly when doing a pull review if you can think of a cleaner way to do this.

```ts
  const data = JSON.parse(fileContents) as National;

  // Strip away unused data (values) from staticProps
  // keep last_values because we use them!
  // eslint-disable-next-line @typescript-eslint/no-unused-vars
  for (const [metricName, metric] of Object.entries(data)) {
    if (typeof metric === 'object' && metric !== null) {
      for (const [metricProperty, metricValue] of Object.entries(metric)) {
        if (metricProperty === 'values') {
          (metricValue as {
            values: Array<any>;
          }).values = [];
        }
      }
    }
  }
```

## Drawbacks

In the case of the homepage, I don't think there's a drawback. It adds a few lines of maintenance, but it's acceptable. As soon as these structures become difficult to maintain we should think twice about how we solve this.

### Using react-query and removing getStaticProps
Not a viable alternative. We purposefully ship static pages for fast loading speeds, good SEO and accessibility reasons. We don't want to ship a blank page while the JS is still loading.

## Unresolved questions

- [x] How do I keep `as National` behind the data import without TypeScript screaming at me for removing some keys? 

### Governance

- [X] Documentation is added


src/pages/index.tsx
```diff
@@ -189,9 +189,21 @@ export async function getStaticProps(): Promise<StaticProps> {
...
   const filePath = path.join(process.cwd(), 'public', 'json', 'NL.json');
   const fileContents = fs.readFileSync(filePath, 'utf8');
-  const data = JSON.parse(fileContents) as National;
-  const lastGenerated = data.last_generated;
+  const data = JSON.parse(fileContents);
+
+  // Strip away unused data (values) from staticProps
+  // keep last_values because we use them!
+  for (const [key, value] of Object.entries(data)) {
+    if (typeof value === 'object') {
```
VWSCoronaDashboard10, line 197:
```suggestion
    if (typeof value === 'object' && value !== null) {
```

src/pages/index.tsx
```diff
@@ -189,9 +189,21 @@ export async function getStaticProps(): Promise<StaticProps> {
 
   const filePath = path.join(process.cwd(), 'public', 'json', 'NL.json');
   const fileContents = fs.readFileSync(filePath, 'utf8');
-  const data = JSON.parse(fileContents) as National;
-  const lastGenerated = data.last_generated;
+  const data = JSON.parse(fileContents);
```
VWSCoronaDashboard4, line 192: I like the approach, and it can be scaled/adapted to different pages as well.

`data` will, after this procedure, have a different type than `National`; do we want to have an adjusted type as well, to guarantee type safety in the dynamic part of the page?

src/pages/index.tsx
```diff
@@ -189,9 +189,21 @@ export async function getStaticProps(): Promise<StaticProps> {
...
+  // keep last_values because we use them!
+  for (const [key, value] of Object.entries(data)) {
+    if (typeof value === 'object') {
+      for (const [key2] of Object.entries(data[key])) {
+        if (key2 === 'values') {
+          delete data[key].values;
+        }
+      }
+    }
+  }
```
VWSCoronaDashboard1, line 204: While I obviously agree with the bundle size-reduction, this change does potentially introduce runtime errors since you're changing the data shape at runtime. That is why you had to remove the `as National` cast.
So, to keep this type-safe you would either have to create an interface based on National, but with all of the 'values' properties taken out, or, instead of removing the values key altogether, setting it to an empty array.

VWSCoronaDashboard1 (changes requested):

VWSCoronaDashboard10 (changes requested):
I think we also need to define the data type without the `values`-properties, otherwise it could result in runtime errors not caught with the TS compiler.

src/pages/index.tsx
VWSCoronaDashboard: Aha! The first solution sounds like it's safest. I'll cook something up for this.

src/pages/index.tsx
VWSCoronaDashboard: Seems the new approach is to make a new type for this to ensure type safety.

src/pages/index.tsx
```diff
@@ -189,9 +189,21 @@ export async function getStaticProps(): Promise<StaticProps> {
...
+  // keep last_values because we use them!
+  for (const [key, value] of Object.entries(data)) {
+    if (typeof value === 'object' && value !== null) {
+      for (const [key2] of Object.entries(data[key])) {
+        if (key2 === 'values') {
+          delete data[key].values;
+        }
+      }
+    }
+  }
```
VWSCoronaDashboard8, line 204: I prefer to use named variables that give you a sense of what you're dealing with. Key and value do not tell you much and key2 even less. Maybe something like this :
```suggestion
  for (const [metricName, metric] of Object.entries(data)) {
    if (typeof metric === 'object' && metric !== null) {
      for (const [metricProperty] of Object.entries(data[metricName])) {
        if (metricProperty === 'values') {
          delete data[metricName].values;
        }
      }
    }
  }
```

VWSCoronaDashboard8 (approved):

src/pages/index.tsx
```diff
@@ -190,8 +190,23 @@ export async function getStaticProps(): Promise<StaticProps> {
...
+      for (const [metricProperty, metricValue] of Object.entries(metric)) {
+        if (metricProperty === 'values') {
+          (metricValue as {
+            values: Array<number>;
+          }).values = [];
+        }
+      }
+    }
+  }
+
```
VWSCoronaDashboard8, line 208: Small tweaks to make it more correct

```suggestion
  for (const metric of Object.values(data)) {
    if (typeof metric === 'object' && metric !== null) {
      for (const [metricProperty, metricValue] of Object.entries(metric)) {
        if (metricProperty === 'values') {
          (metricValue as {
            values: Array<unknown>;
          }).values = [];
        }
      }
    }
  }

```

VWSCoronaDashboard10 (approved):

PR #977, opened 2020-11-19
TitleWithIcon refactor and icons refactor

VWSCoronaDashboard1:
## Summary

Refactor titleWithIcon to a styled component and removed a bunch of obsolete functionality, also renamed TitleWithIcon to HeadingWithIcon.
Replace a number of svg icons with cleaned up versions.
Remove svgo optimizer from Next build.

## Motivation

Design request

## Detailed design

First of all, the svgo optimizations are turned off in the NextJS build. This per request from the designers, since the svg's will be delivered pre-optimized and some of the automatic optimizations led to inconsistent results in regard to width/height and viewbox attributes.
The TitleWithIcon (renamed to HeadingWithIcon) component is now a styled component that receives an Icon instance instead of a class. In that case tweaks like custom colors and such are easier to be set on a specific icon. The Icon prop is no longer optional, since the title of the component kind of suggested this ought to be mandatory ;)

## Drawbacks

n/a

## Alternatives

Let me know

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added


src/components-styled/title-with-icon.tsx
```diff
@@ -0,0 +1,66 @@
+import css from '@styled-system/css';
+import React, { useMemo } from 'react';
+import { Box } from './base';
+import { Heading } from './typography';
+
+interface IProps {
+  title: string;
+  Icon: React.ReactNode;
+  subtitle?: string;
+  as?: 'h2' | 'h3';
```
VWSCoronaDashboard8, line 10: I think you might as well call this headingLevel and use a number because you don't seem to pass the "as" value directly to a place where it is required as such. I think it will simplify the code because then you can omit some mappings

src/components/layout/NationalLayout.tsx
```diff
@@ -119,7 +120,11 @@ function NationalLayout(props: NationalLayoutProps) {
                     className={`last-developments-link ${getClassName('/')}`}
                   >
                     <TitleWithIcon
-                      Icon={Notification}
+                      Icon={
+                        <Notification
+                          color={(theme.colors as any).notification}
```
VWSCoronaDashboard8, line 125: This looks a bit dodgy. What is the problem you're solving with any?

src/pages/index.tsx
```diff
@@ -61,7 +62,7 @@ const Home: FCWithLayout<INationalHomepageData> = (props) => {
   return (
     <>
       <TitleWithIcon
-        Icon={Notification}
+        Icon={<Notification color={(theme.colors as any).notification} />}
```
VWSCoronaDashboard8, line 65: Same here, I hope we can find a better way

VWSCoronaDashboard8 (approved):

src/components/layout/NationalLayout.tsx
VWSCoronaDashboard1: I agree this is hacky. There's something dodgy going on with the theme typing here that I haven't been able to figure out. The colors are typed as CSS.Property.Color[] instead of the normal lookup that it is. I haven't been able to convince the compiler otherwise, so that's when I reached for the dreaded 'any' hack.

src/components-styled/title-with-icon.tsx
```diff
@@ -0,0 +1,66 @@
+import css from '@styled-system/css';
+import React, { useMemo } from 'react';
+import { Box } from './base';
+import { Heading } from './typography';
+
+interface IProps {
+  title: string;
+  Icon: React.ReactNode;
```
VWSCoronaDashboard8, line 8:
```suggestion
  icon: React.ReactNode;
```
I think what you're passing in is not a component constructor but an instance because <Icon/> in the calling context will already do the instantiation.

VWSCoronaDashboard8: Also make it React.FunctionComponent instead of ReactNode, because node can be anything including a plain string.

src/components-styled/title-with-icon.tsx
VWSCoronaDashboard1: Makes perfect sense, fixed

src/components-styled/title-with-icon.tsx
VWSCoronaDashboard1: Fixed, it's come to `JSX.Element`, only because the way icons are imported (`import GetestIcon from '~/assets/test.svg';`), the compiler only allows me to use this one.

src/components/layout/NationalLayout.tsx
VWSCoronaDashboard1: I have made a fix to the theme type where I re-declare the colors property with our own type. This solves the issue.

src/components-styled/heading-with-icon.tsx
```diff
@@ -0,0 +1,64 @@
...
+  title: string;
+  icon: JSX.Element;
+  subtitle?: string;
+  headingLevel?: 2 | 3;
+}
+
+export function HeadingWithIcon(props: IProps) {
+  const { icon, title, subtitle, headingLevel = 3 } = props;
+
+  const cssProps = useMemo(() => {
```
VWSCoronaDashboard4, line 16: Isn't useMemo a bit overkill?

VWSCoronaDashboard4 (approved):

src/components-styled/heading-with-icon.tsx
VWSCoronaDashboard1: fair enough, I'll kick that one out

VWSCoronaDashboard8: Yes, good catch. I'm a big fan of avoiding useMemo until you actually noticed a performance issue or it is clearly a lot of processing being done. In this case it will probably only add processing overhead, because useMemo is also not free to use.

src/style/theme.ts
```diff
@@ -101,7 +97,9 @@ export const colors = {
       magenta: ['#F6B4D1', '#D3719C', '#9E3A66', '#64032D', '#000000'],
     },
   },
-} as const;
+};
+
+export type ThemeColors = typeof colors;
```
VWSCoronaDashboard10, line 102: This type isn't used anywhere outside of this file, so an export is not necessary. Care to elaborate why the change from `as const` to `typeof colors`?

src/style/theme.ts
```diff
@@ -110,7 +108,9 @@ const shadows = {
 };
 
 type TDashboardTheme = ScaleThemeProperties &
-  ThemeBreakPoints & { mediaQueries: TMediaQueries };
+  ThemeBreakPoints & { mediaQueries: TMediaQueries } & {
+    colors: ThemeColors;
+  };
 
 const theme: TDashboardTheme = {
```
VWSCoronaDashboard8, line 115: Kunnen we dit eigenlijk niet gewoon typen met `as const` en dan `TDashboardTheme = typeof theme`?

src/components-styled/heading-with-icon.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard1: Already removed

src/pages/index.tsx
```diff
@@ -60,10 +61,10 @@ const Home: FCWithLayout<INationalHomepageData> = (props) => {
 
   return (
     <>
-      <TitleWithIcon
-        Icon={Notification}
+      <HeadingWithIcon
+        icon={<Notification color={theme.colors.notification} />}
         title={text.laatste_ontwikkelingen.title}
-        as="h2"
+        headingLevel={2}
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: I thought about it, but I thought this would leave no doubt about what it was. Its not the icon level, its the heading level

src/style/theme.ts
VWSCoronaDashboard1: Actually, it IS used outside this file. I use it to set the color on one of the icons, hence the typing.

src/style/theme.ts
VWSCoronaDashboard1: Tried it, but I get all sorts of weird compiler errors that I can't seem to fathom. Removing the const also removed those errors. I'm not sure what the deal is...

src/components-styled/heading-with-icon.tsx
```diff
@@ -0,0 +1,64 @@
...
+        flexDirection="row"
+        flexWrap="nowrap"
+        justifyContent="center"
+        alignItems="center"
+        padding={0}
+        marginRight={0}
+        css={css(cssProps)}
+      >
+        {icon}
+      </Box>
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: Nice suggestion, I made the changes

src/style/theme.ts
VWSCoronaDashboard10: Well, <kbd>command+f</kbd> on the diff doesn't yield any results 👀  which file do you import it?

PR #981, opened 2020-11-20
Improve choropleth UI and performance

VWSCoronaDashboard10:
## Summary

This PR removed the need of the `overlay` collection on choropleths. The main purpose of that layer was to render stuff like the afsluitdijk. These things are part of the country geo data and I've moved those features to the choropleth itself - it will always be rendered.

image

image

image

image

image

image

image


VWSCoronaDashboard1 (approved):
The overlays were also used to render the safety region borders on top of the municipalities. But I assume that throughout all of the design changes that requirement has disappeared altogether, so it makes sense now to get rid of the overlay. Which will definitely lighten the DOM load some more.

VWSCoronaDashboard10:
> The overlays were also used to render the safety region borders on top of the municipalities

I've discussed this with JM and he didn't want that anymore :-)

VWSCoronaDashboard (approved):

PR #985, opened 2020-11-20
Remove 'last week' option from GGD multiline chart

VWSCoronaDashboard1:
## Summary

As the title states

## Motivation

Change request

## Detailed design

The main target of this ticket is simply to remove the specified timeframe option.
I wanted to add some kind of regression test for this as well, so I took a look at the existing page tests and added a rudimentary check for this.

I've added ts-jest so that the page tests can be written in TS as well. Previously the test data was hard-coded into the test page, I'm now loading data from the `public/json` directory to keep things more in sync.

For now, the test is rather brittle. It basically selects all the elements on the page with the text 'Last week' in it, it then checks if only ONE instance exists. This is because currently there are three charts on the page, one of which is allowed to show the last week option.
The same goes for the simple tests that check for the existence of certain values. Again, its better than nothing, but it would be nicer to be able to select data from the page more contextually.

Obviously, this isn't extremely future-proof. It would be more robust if we select the specific timeframe groups and check their options, but easily dropping in page-specific test ids in combination with our styled component implementations isn't as straightforward as one would hope.
I would like to hear some suggestions on how we could tackle these issues in an easy way. Because I do feel we need to add more of these regression tests in the future.

## Drawbacks

The test implementation is currently rather brittle, I'm hoping through some discussion we can come to a more robust solution.

## Alternatives

Tell me! ;)

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added


VWSCoronaDashboard8 (approved):

src/tools/test/copy-fixtures.ts
```diff
@@ -0,0 +1,60 @@
...
+  const content = fs.readFileSync(filename, { encoding: 'utf8' });
+  const data = JSON.parse(content);
+
+  for (const prop in data) {
+    if (Array.isArray(data[prop])) {
+      data[prop] = data[prop].filter((item: any) => item.vrcode === vrcode);
+    }
+  }
+
+  fs.writeFileSync(filename, JSON.stringify(data, null, '\t'), {
```
VWSCoronaDashboard10, line 57: why `\t` instead of 2 whitespace (which is the default in our codebase?)

src/tools/test/copy-fixtures.ts
```diff
@@ -0,0 +1,60 @@
...
+
+function filterMunicipalities() {
+  const filename = path.join(fixturePath, 'MUNICIPALITIES.json');
+  const content = fs.readFileSync(filename, { encoding: 'utf8' });
+  const data = JSON.parse(content);
+
+  for (const prop in data) {
+    if (Array.isArray(data[prop])) {
+      data[prop] = data[prop].filter(
+        (item: any) => item.gmcode === municipalCode
```
VWSCoronaDashboard10, line 36: can we get rid of the any type?

src/pages-tests/veiligheidsregio/[code]/positief-geteste.mensen.test.tsx
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
```
Comment (VWSCoronaDashboard10) [inline src/pages-tests/veiligheidsregio/[code]/positief-geteste.mensen.test.tsx:1]: file name looks odd, did you mean `positief-geteste-mensen.test.tsx`?

src/pages-tests/landelijk/positief-geteste.mensen.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+        data={data as National}
+        lastGenerated="test"
+        text={{}}
+      />
+    );
+
+    getByText(/signaalwaarde/i);
+  });
+
+  it("should show not show 'last week' option for GGD charts", () => {
```
VWSCoronaDashboard10, line 42: To show or not to show, that's the question

src/pages-tests/landelijk/positief-geteste.mensen.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Currently there are three charts shown on the positief-geteste-mensen page,
+     * two of those charts should NOT have the 'Laatste week' option available since
+     * the data does not allow for it.
+     * This test therefore checks if there is only ONE radio element on the page
+     * with a value of 'week'.
+     * This test is rather brittle, but for now the only way to do this test as a
+     * way of a regression check.
+     */
```
VWSCoronaDashboard10, line 55: Isn't it possible to pass a test-name prop to the MultipleLineChartTile and query that exact component? Than you can test only that component tree.

src/test-utils/load-fixture.ts
```diff
@@ -0,0 +1,21 @@
+import fs from 'fs';
+import path from 'path';
+
+const publicFolder = path.join(__dirname, '../../../public/json/');
+const fixtureFolder = path.join(__dirname, '../../src/pages-tests/fixtures');
```
VWSCoronaDashboard10, line 5: with or without trailing slash?

VWSCoronaDashboard10 (commented):
Looking good! I think there's room for improvement with the actual tets itself. Counting all the radio-inputs on the page sounds like it could result in false positives too easy.

PR #986, opened 2020-11-20
Topic/choropleth legenda

VWSCoronaDashboard10:
Implement the new legend layout:
image

I've also cleaned up and simplified the legend-related code. A good next step would be to create a choropleth-context which holds thresholds and data. Inner components like the legend and choropleth itself can read the values from that context instead of reading it from an imported object using an injected key on props.


src/components-styled/choropleth-legenda.tsx
```diff
@@ -25,44 +44,29 @@ const List = styled.ul(
...
-      borderLeft: '1px solid lightgrey',
-    },
-    '&:last-child div:first-child': {
-      borderRight: '1px solid lightgrey',
-    },
   })
 );
 
-const LegendaItemBox = styled(Box)(
+const LegendaColor = styled.div<{
```
VWSCoronaDashboard1, line 51: Is there a specific why you're not inheriting from Box here? (No criticism, I'm just curious)

src/components-styled/choropleth-legenda.tsx
```diff
@@ -25,44 +44,29 @@ const List = styled.ul(
...
-    </Box>
-  );
-}
+const Label = styled.span(
+  css({
+    py: 1,
+    display: 'inline-block',
+    transform: 'translateX(-50%)',
+  })
+);
```
VWSCoronaDashboard1, line 72: Might it be better to inherit this from the Text component? Since that one has a lot of the text defaults that should be used everywhere.

VWSCoronaDashboard1 (approved):

src/components-styled/choropleth-legenda.tsx
VWSCoronaDashboard8: In the branch I'm working on I've added an InlineText export next to Text, which is simply Text but as a span, because the as property didn't seem to work on Text anymore plus I think it makes the markup more readable.

VWSCoronaDashboard8 (approved):

src/components-styled/choropleth-legenda.tsx
VWSCoronaDashboard10: Mainly because it's a local component and not a generic "building block", I don't need all the props/features of the Box component.

src/components-styled/choropleth-legenda.tsx
VWSCoronaDashboard10: When will that branch become part of develop?

src/components-styled/choropleth-legenda.tsx
VWSCoronaDashboard8: I think we intended the Box component to be used inside other components, because you would rarely use it without further abstraction. But I'm curious to know still if there would be any downsides in terms of performance.

Reusing the Box component locally could be nice to reduce code complexity a bit. In this case it would cut out 

```
const Label = styled.span(
  css({
  })
);
```

src/components-styled/choropleth-legenda.tsx
VWSCoronaDashboard8: It's part of the differences branch, so when the data for differences is verified and ready to be supplied on develop by BE

src/components-styled/choropleth-legenda.tsx
VWSCoronaDashboard10: But it would add complexity when writing jsx:
```tsx
{index > 0 && <Label>{threshold}</Label>}
```
vs
```tsx
{index > 0 && (
  <Box
    as="span"
    py={1}
    display="inline-block"
    css={css({ transform: 'translateX(-50%)' })}
  >
    {threshold}
  </Box>
)}
```

I must say that I prefer the `<Label>{threshold}</Label>` implementation ..

PR #987, opened 2020-11-20
Toe en afname indicators

VWSCoronaDashboard8:
This adds visual indicators for diff calculations. In the sidebar, we only show the icon, but in the tiles, we include a text depending on when the previous measurement was made.

When troubleshooting the data I ran into some oddities and unnecessary complexity WRT the static props and layout, so I cleaned that up in NL as well. A similar cleanup can later be applied to VR and GM.


src/components-styled/difference-indicator.tsx
```diff
@@ -0,0 +1,162 @@
...
+  value: DifferenceDecimal | DifferenceInteger;
+  isContextSidebar?: boolean;
+}
+
+export function DifferenceIndicator(props: DifferenceIndicatorProps) {
+  const { isContextSidebar, value } = props;
+
+  return isContextSidebar
+    ? renderSidebarIndicator(value)
+    : renderTileIndicator(value);
```
VWSCoronaDashboard1, line 31: why make these two functions?  Why not just implement them like components?
```
<SidebarIndicator value={value}>
<TileIndicator value={value}>
```

src/static-props/data-sorting.ts
```diff
@@ -94,9 +94,10 @@ function sortTimeSeriesValues(values: Timestamped[]) {
    * If none match we throw, since it means an unknown timestamp is used and we
    * want to be sure we sort all data.
    */
-  throw new Error(
-    `Unknown timestamp in value ${JSON.stringify(values[0], null, 2)}`
-  );
+  // throw new Error(
+  //   `Unknown timestamp in value ${JSON.stringify(values[0], null, 2)}`
+  // );
+  return values;
```
VWSCoronaDashboard1, line 100: Is this temporary until we have a method of executing this only on production builds?

VWSCoronaDashboard1 (approved):

src/components-styled/difference-indicator.tsx
VWSCoronaDashboard8: I considered it but eventually didn't like the added boilerplate. You need to define props and their interfaces, and parse values from the props. I find this a bit cleaner. Creating them as separate components you might be tempted to use them on their own, and until we need that I think this is fine.

src/static-props/data-sorting.ts
VWSCoronaDashboard8: This slipped in by accident :)

src/components-styled/difference-indicator.tsx
```diff
@@ -0,0 +1,162 @@
...
+   * @TODO discuss logic for this
+   *
+   * 6 days is more like a week
+   * and 13 days more like two weeks
+   *
+   * Think this way we can prevent rounding errors. In our data timespans should
+   * typically be a few days or a week or multiple weeks anyway.
+   */
+
+  if (days < 2) {
```
VWSCoronaDashboard4, line 147: Why are there relative dates for the sidebar? These were removed previously. There is an option for the dateformatter to output text similar to these, so the logic could be moved there, if we really want these relative dates.

src/components/layout/NationalLayout.tsx
```diff
@@ -23,27 +24,35 @@ import { ReproductionIndexBarScale } from '~/components/landelijk/reproduction-i
...
+export function getNationalLayout(
+  page: React.ReactNode,
+  pageProps: NationalPageProps
+) {
+  return (
+    <Layout
+      {...siteText.nationaal_metadata}
+      lastGenerated={pageProps.lastGenerated}
+    >
+      {/* {console.log('XXXXX', pageProps.data.difference)} */}
```
VWSCoronaDashboard4, line 49: Remove this comment

src/components/layout/NationalLayout.tsx
```diff
@@ -23,27 +24,35 @@ import { ReproductionIndexBarScale } from '~/components/landelijk/reproduction-i
...
-    return getSiteLayout(
-      siteText.nationaal_metadata,
-      pageProps.lastGenerated
-    )(<NationalLayout {...pageProps}>{page}</NationalLayout>);
-  };
+/**
+ * Using composition this can be a less confusing and more direct replacement
+ * for getSiteLayout.
+ *
+ * @TODO replace other use of getSiteLayout()
```
VWSCoronaDashboard4, line 38: In addition to this TODO, add it to the places where it needs to be replaced (regionallayout, municipallayout)

src/style/theme.ts
```diff
@@ -31,6 +31,12 @@ const fontSizes = [
   '2.02729rem',
 ];
 
+const fontWeights = {
+  normal: 400,
+  bold: 600,
+  heavy: 700,
```
VWSCoronaDashboard4, line 37: Does our font even have a bold and heavy variant that we use?

VWSCoronaDashboard4 (commented):
Some small comments.

Although the refactoring and renaming seem fine, it might be better to split it to a different PR where possible in the future for easier reviewing, it confused me at first.

src/style/theme.ts
VWSCoronaDashboard8: These are defined in place in current CSS. Not sure if the classes are actually used.

src/components-styled/difference-indicator.tsx
VWSCoronaDashboard8: The sidebar indicator doesn't use this logic. Only the KPI variant.

I will look at the date formatter to see if it is convenient to share or inherig this logic 👍

src/components-styled/difference-indicator.tsx
```diff
@@ -0,0 +1,161 @@
...
+    );
+  }
+
+  if (difference < 0) {
+    <Container>
+      <Span color="green">
+        <IconDown />
+      </Span>
+    </Container>;
+  }
```
VWSCoronaDashboard10, line 53:
```suggestion
  if (difference < 0) {
    return (
      <Container>
        <Span color="green">
          <IconDown />
        </Span>
      </Container>
    );
  }
```

src/components-styled/difference-indicator.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+    <Container>
+      <Span color="green">
+        <IconDown />
+      </Span>
+      <Span fontWeight="bold" mr="0.3em">
+        {splitText[0]}
+      </Span>
+      <Span>{`${splitText[1]} ${timespanText}`}</Span>
+    </Container>;
```
VWSCoronaDashboard10, line 103:
```suggestion
  return (
    <Container>
      <Span color="green">
        <IconDown />
      </Span>
      <Span fontWeight="bold" mr="0.3em">
        {splitText[0]}
      </Span>
      <Span>{`${splitText[1]} ${timespanText}`}</Span>
    </Container>
  );
```

src/components-styled/difference-indicator.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+    return (
+      <Container>
+        <Span color="red">
+          <IconUp />
+        </Span>
+        <Span fontWeight="bold" mr="0.3em">
+          {`${difference} ${splitText[0]}`}
+        </Span>
+        <Span color="annotation">{`${splitText[1]} ${timespanText}`}</Span>
```
VWSCoronaDashboard10, line 87: It would be nice to add a `title`-prop with the exact date.

src/components-styled/kpi-value.tsx
```diff
@@ -36,19 +39,21 @@ export function KpiValue({
   absolute,
   percentage,
   valueAnnotation,
-  ...props
+  difference,
+  ...otherProps
```
VWSCoronaDashboard10, line 43: I believe `otherProps` will always be an empty object, all the values of the interface `KpiValueProps` are used.

src/components-styled/difference-indicator.tsx
VWSCoronaDashboard8: Good catch!

src/components-styled/kpi-value.tsx
VWSCoronaDashboard8: I think it's required to pass on cy-data, but I'm wondering why the component would not complain about this extra prop. Probably it is defined somewhere in a global typedef.

VWSCoronaDashboard10: Aah .. I guess it would be better to not pass `data-X` props then, instead explicitly name it something like `cy` or `testId` or whatever ..

VWSCoronaDashboard8: Or data props are allowed on react components always...?

VWSCoronaDashboard10: Sounds like a very error-prone opaque approach to me. You can pass any data-X prop on any component and then just guess/hope something works like expected? I'd expect that it should be part of the typed props.

VWSCoronaDashboard8: Yes, I fully agree this needs to be looked at. Outside of the scope of this PR though IMO.

VWSCoronaDashboard8: I would like a discussion on how to deal with data-cy props in general, because I think we can avoid setting them explicitly everywhere maybe.

PR #998, opened 2020-11-24
Fix menu visibility toggle on homepage

VWSCoronaDashboard10:
When a user visits the homepage the app should render the menu instead of the "laatste ontwikkelingen" page. This already worked when landing initially on the homepage but didn't work when navigating from another page to the homepage.

I was able to replace the `useMenuState` with next router core features.


VWSCoronaDashboard4 (approved):
Nice solution and improvement! We might need to look into those blur events for accessibility, but that would be a different PR anyway.

VWSCoronaDashboard8 (approved):
💯 Great idea to move this state to the URL

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard10:
> Nice solution and improvement! We might need to look into those blur events for accessibility, but that would be a different PR anyway.

Thanks!
How is the interaction for accessibility changed compared to the previous implementation?

PR #999, opened 2020-11-24
Topic/warning tile develop

VWSCoronaDashboard10:
## Summary

Create the `<MessageTile />` and translation keys for the homepage and VR page.

I've changed the way we parse markdown keys in our translation files. There were several issues with the old implementation:
1.  Previously we defined per page which keys should be parsed, this meant there would still be markdown in the locale object because some keys were not parsed. Manually adding these keys would fix those issues, but it felt weird and error-prone to manually add them to several page's `getStaticProps`.
2. The keys in an imported locale file were mutated with parsed markdown keys. When reloading a page (during development) it would again mutate those keys. Already parsed markdown would be overwritten with an empty string because the markdown was already replaced with html.

I've created a utility which will walk through a list of keys where markdown can live. It will clone the entire object and replace those keys.


src/components-styled/message-tile.tsx
```diff
@@ -0,0 +1,31 @@
...
+      css={css({
+        borderLeft: '9px solid #FFd600',
+        backgroundColor: '#FFF4B9',
+      })}
+    >
+      <Box display="flex" alignItems="flex-start">
+        <Box mr={3} width={38} flexShrink={0}>
+          <Warning width={38} height={38} fill="black" />
+        </Box>
+        <Box maxWidth={450} css={css({ p: { m: 0 } })}>
```
VWSCoronaDashboard8, line 25: `css={css({ p: { m: 0 } })` looks like a bit of a hack. Wouldn't it be better to pass in the message as a string (with markdown maybe) and control how it is rendered (p or no p) in here? 

In other words does message really need to be "any type of react component", because it seems like you are already expecting a p element

src/pages/index.tsx
```diff
@@ -90,7 +96,18 @@ const Home: FCWithLayout<INationalHomepageData> = (props) => {
         </a>
       </article>
 
+      {text.regionaal_index.belangrijk_bericht && (
+        <MessageTile>
+          <div
+            dangerouslySetInnerHTML={{
```
VWSCoronaDashboard8, line 102: I think I would prefer to hide this logic inside the MessageTile

src/pages/veelgestelde-vragen.tsx
```diff
@@ -71,26 +53,18 @@ const Verantwoording: FCWithLayout<VeelgesteldeVragenProps> = (props) => {
...
-              )}
+              {text.over_veelgestelde_vragen.vragen.map((item) => {
+                const id = getSkipLinkId(item.vraag);
+                return (
+                  <Collapsable key={id} id={id} summary={item.vraag}>
+                    <div
+                      dangerouslySetInnerHTML={{
+                        __html: item.antwoord,
+                      }}
+                    ></div>
```
VWSCoronaDashboard8, line 64: nit
```suggestion
                    />
```

src/utils/parse-markdown-in-locale.ts
```diff
@@ -0,0 +1,44 @@
...
+  'regionaal_index.belangrijk_bericht',
+  'veiligheidsregio_index.selecteer_toelichting',
+  'veiligheidsregio_index.belangrijk_bericht',
+  'over_risiconiveaus.toelichting',
+  'over_beschrijving.text',
+  'over_veelgestelde_vragen.vragen[].antwoord',
+  'verantwoording.cijfers[].verantwoording',
+];
+
+export function parseMarkdownInLocale(text: TALLLanguages) {
```
VWSCoronaDashboard8, line 19: 💯

src/utils/parse-markdown-in-locale.ts
```diff
@@ -0,0 +1,44 @@
...
+  const textClone = JSON.parse(JSON.stringify(text)) as TALLLanguages;
```
VWSCoronaDashboard8, line 20: 😄 Maybe not the most efficient way to clone an object but I guess this is built-time only anyway.

src/utils/parse-markdown-in-locale.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  MARKDOWN_KEYS.forEach((key) => replaceWithMarkdown(textClone, key));
+
+  return textClone;
+}
+
+function replaceWithMarkdown(object: any, path: string) {
```
VWSCoronaDashboard8, line 27: Any can be stricter if you expect an object
```suggestion
function replaceWithMarkdown(object: Record<string, unknown>, path: string) {
```

VWSCoronaDashboard8 (approved):
Some minor suggestions

src/pages/index.tsx
```diff
@@ -90,7 +96,18 @@ const Home: FCWithLayout<INationalHomepageData> = (props) => {
...
+              __html: text.regionaal_index.belangrijk_bericht,
+            }}
+          />
+        </MessageTile>
+      )}
+
       <ChoroplethTile
+        showDataWarning
```
VWSCoronaDashboard10, line 110: remove this, was for testing.

VWSCoronaDashboard1 (approved):

src/utils/parse-markdown-in-locale.ts
VWSCoronaDashboard10: I thought the stringify/parse-method was one of the more performant clone strategies, but according to [this benchmark](https://jsben.ch/VLrZA) `lodash.deepClone` is a bit faster :-) But you're right about being built-time only so I'll leave it like this.

PR #1008, opened 2020-11-24
Add data download links to content header

VWSCoronaDashboard1:
## Summary

The content header is not only required to list the data sources, but, where possible, should also provide download links for the raw data that was used to populate the dashboard.
This PR adds this functionality.

## Motivation

Feature request

## Detailed design

I've consolidated the way that "bronnen" were defined in the locale files to one specific way. Before, there were properties called "bron" and/or "bronnen", sometimes both were used in the same page. So I've changed them all to use the "bronnen" shape.

The `ContentHeader_**hack` components are removed and all the shared functionality has been consolidated into `content-header`, which is now a styled component version.
the Metadata has been refactored into a styled component as well: `metadata`. (in the content-header directory, since its only used by that component)

The "bronnen" locale object now also contains a "download" key which can contain the download link from a data source. Only non-empty download links are rendered in the content header.

## Drawbacks

"bronnen" are now named, so if a datasource changes, this will probably mean a code change as well.

## Alternatives

Let me know

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added


src/components-styled/content-header-metadata.tsx
```diff
@@ -0,0 +1,160 @@
...
+export interface IContentHeaderMetadataProps {
+  dataSources: IDatasource[];
+  datumsText: string;
+  dateInfo: number | IWeekRange;
+  dateOfInsertionUnix: number;
+}
+
+const text = siteText.common.metadata;
+
+export function Metadata(props: IContentHeaderMetadataProps) {
```
VWSCoronaDashboard8, line 32:
```suggestion
export function Metadata(props: MetadataProps) {
```

src/components-styled/content-header-metadata.tsx
```diff
@@ -0,0 +1,160 @@
...
+  text: string;
+  download: string;
+}
+
+export interface IWeekRange {
+  weekStartUnix: number;
+  weekEndUnix: number;
+}
+
+export interface IContentHeaderMetadataProps {
```
VWSCoronaDashboard8, line 23:
```suggestion
export interface MetadataProps {
```

src/components-styled/content-header-metadata.tsx
```diff
@@ -0,0 +1,160 @@
...
+import DownloadIcon from '~/assets/download.svg';
+import siteText from '~/locale/index';
+import theme from '~/style/theme';
+import { formatDateFromSeconds } from '~/utils/formatDate';
+import { replaceVariablesInText } from '~/utils/replaceVariablesInText';
+import { Box } from './base';
+import { ExternalLink } from './external-link';
+import { Text } from './typography';
+
+export interface IDatasource {
```
VWSCoronaDashboard8, line 12: Liever geen I prefixes

src/components-styled/content-header-metadata.tsx
```diff
@@ -0,0 +1,160 @@
+import ClockIcon from '~/assets/clock.svg';
```
VWSCoronaDashboard8, line 1: File can be named simply `metadata` I think. If metadata is only to be used inside the content-header, consider making content-header a directory and nest the metadata in there.

src/components-styled/content-header-metadata.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+const text = siteText.common.metadata;
+
+export function Metadata(props: IContentHeaderMetadataProps) {
+  const { dataSources, datumsText, dateInfo, dateOfInsertionUnix } = props;
+
+  let dateText = '';
+  if (typeof dateInfo === 'number') {
+    const dateOfReport = formatDateFromSeconds(dateInfo, 'weekday-medium');
+    const dateOfInsertion = dateOfInsertionUnix
```
VWSCoronaDashboard8, line 38: I don't see dateOfInsertionUnix being optional

src/components-styled/content-header-metadata.tsx
```diff
@@ -0,0 +1,160 @@
...
+    );
+    dateText = replaceVariablesInText(datumsText, {
+      weekStart,
+      weekEnd,
+      dateOfInsertion,
+    });
+  }
+
+  const dataDownloads = dataSources
+    .filter((ds) => Boolean(ds.download.trim()))
```
VWSCoronaDashboard8, line 66: Instead of using vague abreviated variable names, I prefer to either write a real name, or in this case use a standard / deliberately vague short name that can be recognized as such. I prefer to use `x`.

With ds my mind immediately goes "what is ds?" with x I would just glance over it.

src/components-styled/content-header-metadata.tsx
```diff
@@ -0,0 +1,160 @@
...
+        label={text.source}
+      />
+
+      <MetadataItem
+        icon={
+          <DownloadIcon
+            aria-hidden
+            color={theme.colors.annotation}
+            width="14"
+            height="14"
```
VWSCoronaDashboard8, line 96: I think 1em would make the most sense, so it matches the text (I'm guessing). Or was that really too big?

src/components-styled/content-header-metadata.tsx
```diff
@@ -0,0 +1,160 @@
...
+          />
+        }
+        items={dataDownloads}
+        label={text.download}
+      />
+    </Box>
+  );
+}
+
+interface IMetadataItemProps {
```
VWSCoronaDashboard8, line 106: I prefix. Especially with the Props suffix I feel it's redundant.

src/components-styled/content-header.tsx
```diff
@@ -0,0 +1,69 @@
...
+    '&[id]': {
+      marginLeft: '-100vw',
+      paddingLeft: '100vw',
+    },
+    marginTop: category ? undefined : 4,
+    marginLeft: icon ? undefined : 5,
+  };
+
+  return (
+    <Box as="header" id={id} css={css(cssRules)}>
```
VWSCoronaDashboard8, line 24: nit; I think it would be cleaner / more consistent with other code if you created a custom styled component above instead of using box. So you can apply the specific css there straightaway. 

I think reusing Box is only worth it if it results in clean code. You can also write `const MyBox = styled(Box) { }` and then apply the css there.

src/components-styled/content-header.tsx
```diff
@@ -0,0 +1,69 @@
...
+      {category && (
+        <Text
+          fontSize={3}
+          fontWeight="600"
```
VWSCoronaDashboard8, line 28: I think this one is in the theme now (I've added it recently)

src/components-styled/content-header.tsx
```diff
@@ -0,0 +1,69 @@
...
+          color="category"
+          margin={0}
+          marginBottom={1}
+          marginLeft={5}
+        >
+          {category}
+        </Text>
+      )}
+      {icon && <HeadingWithIcon icon={icon} title={title} headingLevel={2} />}
+      {!icon && <Heading level={2}>{title}</Heading>}
```
VWSCoronaDashboard8, line 38: `{ icon ? <HeadingWithIcon ... /> : <Heading ... /> }`

VWSCoronaDashboard8 (approved):

src/components-styled/content-header-metadata.tsx
VWSCoronaDashboard1: I just took the size from the other icons, I'll try 1em, it makes sense to use that size.

src/components-styled/content-header/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+} from '~/components-styled/content-header/metadata';
+import { HeadingWithIcon } from '~/components-styled/heading-with-icon';
+import { Heading, Text } from '~/components-styled/typography';
+import { Box, BoxProps } from '../base';
+
+const HeaderBox = styled.header<BoxProps>({
+  '&[id]': {
+    marginLeft: '-100vw',
+    paddingLeft: '100vw',
+  },
```
VWSCoronaDashboard10, line 17: What kind of effect have these styles? Can we apply them using a named prop instead of the potential existence of an id attribute?
And are you sure the BoxProps are valid props? Because it's a `styled.header` instead of `styled(Box)`.

src/components-styled/content-header/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+    margin: 0,
+    marginBottom: 1,
+    marginLeft: 5,
+  })
+);
+
+const BodyBox = styled(Box)(
+  css({
+    display: [null, null, null, 'flex'],
+    marginLeft: [null, null, null, 5],
```
VWSCoronaDashboard10, line 54: Is it possible to use the object syntax?

```suggestion
    display: { lg: 'flex' },
    marginLeft: { lg: 5 },
```

src/components-styled/content-header/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+  })
+);
+
+const ReferenceBox = styled(Box)(
+  css({
+    maxWidth: '30em',
+    marginRight: 3,
+    flex: [null, null, null, 'flex: 1 1 60%;'],
```
VWSCoronaDashboard10, line 62: same

src/components-styled/content-header/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+  })
+);
+
+const MetadataBox = styled(Box)(
+  css({
+    flex: [null, null, null, 'flex: 1 1 40%;'],
```
VWSCoronaDashboard10, line 68: same

src/components-styled/content-header/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+        <MetadataBox>
+          <Metadata {...metadata} />
+        </MetadataBox>
+      </BodyBox>
+    </Header>
+  );
+}
+
+interface IContentHeaderProps {
```
VWSCoronaDashboard10, line 104: I saw earlier comments about not using an `I`-prefix on interfaces

src/components-styled/content-header/metadata.tsx
```diff
@@ -0,0 +1,161 @@
...
+export interface Datasource {
+  href: string;
+  text: string;
+  download: string;
+}
+
+export interface IWeekRange {
+  weekStartUnix: number;
+  weekEndUnix: number;
+}
```
VWSCoronaDashboard10, line 21: It doesn't look like these exported types aren't used anywhere outside this file. Also another `I`-prefix :-)

src/components-styled/content-header/metadata.tsx
```diff
@@ -0,0 +1,161 @@
...
+    const dateOfInsertion = formatDateFromSeconds(
+      dateOfInsertionUnix,
+      'weekday-medium'
+    );
+    dateText = replaceVariablesInText(datumsText, {
+      weekStart,
+      weekEnd,
+      dateOfInsertion,
+    });
+  }
```
VWSCoronaDashboard10, line 64: How about moving this block to a separate function in this module? Makes reading the component a bit easier.

src/components-styled/content-header/metadata.tsx
```diff
@@ -0,0 +1,161 @@
...
+        {icon}
+      </Box>
+      <Text margin={0}>
+        {label}:{' '}
+        {items.map((item, index) => {
+          if (index) {
+            return (
+              <>
+                {' '}
+                &amp;{' '}
```
VWSCoronaDashboard10, line 141: This can maybe also be written as `{' & '}`

src/components-styled/content-header/metadata.tsx
```diff
@@ -0,0 +1,161 @@
...
+                  key={`label_${index}_${item.href}`}
+                />
+              </>
+            );
+          }
+          return (
+            <ExternalLink
+              href={item.href}
+              text={item.text}
+              key={`label_${index}_${item.href}`}
```
VWSCoronaDashboard10, line 154: should be sufficient:
```suggestion
              key={index + item.href}
```

VWSCoronaDashboard10 (commented):
Good thing those hack components are gone now 💯

src/components-styled/content-header/index.tsx
VWSCoronaDashboard1: I'm not sure how this works, but it's some kind of hack to let skiplinks work correctly in IE11. I agree that a specific attribute is probably a better way to make this work, I'll make the change (and check if it keeps working in IE11...)

src/components-styled/content-header/index.tsx
VWSCoronaDashboard1: Nope, unfortunately not, I get compile errors if I use the object syntax.

src/components-styled/content-header/index.tsx
VWSCoronaDashboard1: same

src/components-styled/content-header/index.tsx
VWSCoronaDashboard10: I remember running into that as well .. we need to dive into that some day

src/components-styled/content-header/index.tsx
VWSCoronaDashboard10: Might be a good idea to add a comment about the IE11 stuff :-)

src/components-styled/content-header/index.tsx
VWSCoronaDashboard4: Shouldn't this be 

`flex: [null, null, null, '1 1 60%'],` ?

src/components-styled/content-header/index.tsx
VWSCoronaDashboard4: Shouldn't this be `flex: [null, null, null, '1 1 40%'],` ?

src/components-styled/content-header/metadata.tsx
VWSCoronaDashboard4: Perhaps a `items.map(() => { ... }).join(' & ')` would work here?

src/components-styled/content-header/index.tsx
VWSCoronaDashboard4: Yes, this is indeed for IE11; perhaps add a comment to this code in case this pops up in the future?

src/components-styled/content-header/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+const HeaderBox = styled.header<BoxProps>({
+  '&[id]': {
+    marginLeft: '-100vw',
+    paddingLeft: '100vw',
+  },
+});
+
+interface HeaderProps {
+  id?: string;
```
VWSCoronaDashboard4, line 21: Put optional properties as last?

VWSCoronaDashboard4 (approved):
Nice cleanup; some small comments

src/components-styled/content-header/metadata.tsx
VWSCoronaDashboard10: @VWSCoronaDashboard4 nope, you cannot concatenate react components like that

src/components-styled/content-header/index.tsx
VWSCoronaDashboard1: This was refactored, so the property has been changed entirely. The CSS mistakes gave been fixed too! :)

PR #1015, opened 2020-11-25
Remove behavior pages and components

VWSCoronaDashboard10:
Remove behavior from our app until further notice.


VWSCoronaDashboard4 (approved):

VWSCoronaDashboard1 (approved):

src/components/choropleth/shared.d.ts
```diff
@@ -15,7 +15,9 @@ export type TMunicipalityMetricType = ValueOf<
 >[number] &
   Partial<MunicipalityProperties>;
 
-export type TRegionMetricName = TMetricHolder<Omit<Regions, 'deceased'>>;
+export type TRegionMetricName = TMetricHolder<
+  Omit<Regions, 'deceased' | 'behavior'>
```
VWSCoronaDashboard8, line 19: I don't think this is necessary, also not for deceased. We control whether we render a choropleth and with what data.

VWSCoronaDashboard8 (approved):

src/components/choropleth/shared.d.ts
VWSCoronaDashboard10: Not with the current setup .. not touching this line means several TS errors.

PR #1026, opened 2020-11-25
Infected age groups barchart

VWSCoronaDashboard4:
## Summary

Adds Visx graph for pyramid barchart, including tooltip, as dedicated component for the age group data of infected people.


src/pages/landelijk/positief-geteste-mensen.tsx
```diff
@@ -188,27 +206,22 @@ const PositivelyTestedPeople: FCWithLayout<NationalPageProps> = ({ data }) => {
...
-          axisTitle={text.barchart_axis_titel}
-        />
+        <Text mt="0">
+          {replaceVariablesInText(
+            siteText.infected_age_groups.example,
+            ageGroupExample
+          )}
+        </Text>
+        <Box mx="-2rem">
+          <AgeGroupChartWrapper data={data.infected_age_groups} />
```
VWSCoronaDashboard10, line 223: Is it possible to import the `AgegroupChart` component instead of the wrapper? And inside the `AgeGroupChart` you can wrap the chart with the wrapper.

src/pages/landelijk/positief-geteste-mensen.tsx
```diff
@@ -28,35 +27,54 @@ import { FCWithLayout } from '~/components/layout';
...
 } from '~/static-props/nl-data';
 import { colors } from '~/style/theme';
-import { formatDateFromSeconds } from '~/utils/formatDate';
-import { formatNumber, formatPercentage } from '~/utils/formatNumber';
-import { replaceKpisInText } from '~/utils/replaceKpisInText';
 
 const text = siteText.positief_geteste_personen;
 const ggdText = siteText.positief_geteste_personen_ggd;
 
+function getAgeGroupExample(data: NationalInfectedAgeGroups) {
```
VWSCoronaDashboard10, line 49: does "example" mean that this will return temporary mock data..?

src/domain/infected-people/age-group/age-group-tooltip.tsx
```diff
@@ -0,0 +1,93 @@
...
+  return (
+    <Tooltip style={{ left: `${left}px`, top: `${top}px` }}>
+      <Text fontSize={3} fontWeight="bold" px={3} py={2} m="0">
+        {replaceVariablesInText(text.graph.age_group_range_tooltip, {
+          ageGroupRange: formatAgeGroupRange(value.age_group_range),
+        })}
+      </Text>
+      <Legend>
+        <LegendItem>
+          <b>{formatPercentage(value.age_group_percentage)}%</b>{' '}
```
VWSCoronaDashboard10, line 79: It has been a while ago I've seen the `<b />` tag, is it still a thing?

src/domain/infected-people/age-group/age-group-tooltip.tsx
```diff
@@ -0,0 +1,93 @@
...
+    },
+    '&.infected-percentage:before': {
+      background: '#3391cc',
+    },
+  })
+);
+
+export function AgeGroupTooltip({ value, left, top }: AgeGroupTooltipProps) {
+  return (
+    <Tooltip style={{ left: `${left}px`, top: `${top}px` }}>
```
VWSCoronaDashboard10, line 71:
```suggestion
    <Tooltip style={{ left, top }}>
```
React will automatically append `px` to the value

VWSCoronaDashboard10 (commented):
looking good!

src/domain/infected-people/age-group/age-group-tooltip.tsx
VWSCoronaDashboard4: It's still part of HTML5. While eg `<strong>` is used to put emphasis on something, `<b>` is there for cases whe you don't need emphasis. Alternative here would be to create a styled component with just the fontWeight to bold, but that seems like a different kind of oddity to me.

src/pages/landelijk/positief-geteste-mensen.tsx
VWSCoronaDashboard4: It compiles data to be used in an example. I'll see if I can improve the naming/add a comment to clarify this.

src/domain/infected-people/age-group/age-group-chart-wrapper.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const [tooltip, setTooltip] = useState<null | TooltipOptions>();
+
+  // @TODO move this tooltip logic elsewhere
+  // And combine with keyboard logic
+  const timer = useRef(-1);
+
+  const debounceSetTooltip = useCallback((options: TooltipOptions | null) => {
+    if (timer.current > -1) {
+      window.clearTimeout(timer.current);
+      timer.current = -1;
```
VWSCoronaDashboard1, line 38: is it really necessary to default to -1 when on the next line its being set by setTimeout()?

src/domain/infected-people/age-group/age-group-chart.tsx
```diff
@@ -0,0 +1,282 @@
...
+  closeTooltip,
+}: AgeGroupChartProps) {
+  const values = data.values.sort((a, b) => {
+    const aStart = parseInt(a.age_group_range, 10);
+    const bStart = parseInt(b.age_group_range, 10);
+    return bStart - aStart;
+  });
+
+  // Define the graph dimensions and margins
+  const isSmallScreen = parentWidth < 500;
```
VWSCoronaDashboard1, line 60: shouldn't we use our fancy useBreakpoints() hook here?

src/domain/infected-people/age-group/age-group-chart.tsx
```diff
@@ -0,0 +1,282 @@
...
+  // Then we'll create some bounds
+  const xMax = (width - margin.left - margin.right - ageRangeAxisWidth) / 2;
+  const yMax = height - margin.top - margin.bottom;
+
+  // We'll make some helpers to get at the data we want
+  const ageGroupPercentage = (d: NationalInfectedAgeGroupsValue) =>
+    d.age_group_percentage;
+  const infectedPercentage = (d: NationalInfectedAgeGroupsValue) =>
+    d.infected_percentage;
+  const y = (d: NationalInfectedAgeGroupsValue) => d.age_group_range;
```
VWSCoronaDashboard1, line 78:
```suggestion
  const yDomain = (d: NationalInfectedAgeGroupsValue) => d.age_group_range;
```
Just for clarity sake?

src/domain/infected-people/age-group/age-group-chart.tsx
```diff
@@ -0,0 +1,282 @@
...
+  });
+  const yScale = scaleBand({
+    range: [margin.top, height - margin.top],
+    round: true,
+    domain: values.map(y),
+    padding: 0.4,
+  });
+
+  // Compose together the scale and accessor functions to get point functions
+  const compose = (scale: any, accessor: any) => (
```
VWSCoronaDashboard1, line 108: instead of `compose` call this `createPoint`?

src/domain/infected-people/age-group/age-group-chart.tsx
```diff
@@ -0,0 +1,282 @@
...
+            }
+            onMouseLeave={closeTooltip}
+            css={css({
+              '&:hover .hoverbar': {
+                fill: '#E0EEF6',
+              },
+            })}
+          >
+            {/* This bar takes all width to display the background color on hover
+              The tranparent stroke is to capture mouse movements in between bars for the tooltip */}
```
VWSCoronaDashboard1, line 220:
```suggestion
              The transparent stroke is to capture mouse movements in between bars for the tooltip */}
```

VWSCoronaDashboard1 (approved):

src/domain/infected-people/age-group/age-group-chart.tsx
VWSCoronaDashboard4: I'll rename this to `ageGroupRange` then, `yDomain` is also not correct/clear.

src/pages/landelijk/positief-geteste-mensen.tsx
VWSCoronaDashboard4: Naming wise, how would you rename the current `AgeGroupChart` ?

VWSCoronaDashboard10: `AgeGroupChartInner`/`AgeGroupChartChart` / `AgeGroupChartActualChart`/dunno 🤷  it does not really matter imho, it's only local component

src/domain/infected-people/age-group/age-group-chart.tsx
VWSCoronaDashboard10: nit: I must say that I like `x` and `y` since it maps to the axes

src/domain/infected-people/age-group/age-group-chart.tsx
VWSCoronaDashboard10: And can we get rid of the any types?

src/domain/infected-people/age-group/age-group-chart.tsx
VWSCoronaDashboard10: `useBreapoints` will trigger on specific viewport size, not component sizes, so that's actually different. We're also using the component width to determine the position and dimenions of several elements.

It might be a good idea to rename the variable to `isSmallComponent` instead of `isSmallScreen`.

src/domain/infected-people/age-group/age-group-chart.tsx
VWSCoronaDashboard1: Then call is xAxis and yAxis ;)

VWSCoronaDashboard10: yeah or getX, getY

src/domain/infected-people/age-group/age-group-chart-wrapper.tsx
```diff
@@ -0,0 +1,90 @@
...
+    },
+    [debounceSetTooltip]
+  );
+
+  const closeTooltip = useCallback(() => {
+    debounceSetTooltip(undefined);
+  }, [debounceSetTooltip]);
+
+  return (
+    <Box mx="-2rem">
```
VWSCoronaDashboard10, line 68: Can we get rid of outside margins? Perhaps move this Box to the location where you render the AgeGroupChartWrapper?
Also, `-2rem` can be extracted from our theme's scale array with `mx={-4}`, right?

src/domain/infected-people/age-group/age-group-chart.tsx
```diff
@@ -0,0 +1,285 @@
...
+      event: MouseEvent<SVGGElement>,
+      value: NationalInfectedAgeGroupsValue
+    ) => { x: number; y: number }
+  ) => void;
+  closeTooltip: () => void;
+}
+
+const TickValue = ({ x, y, formattedValue }: TickRendererProps) => {
+  return (
+    <Text x={x} y={y} fill="#595959" fontSize={'1rem'} textAnchor="middle">
```
VWSCoronaDashboard10, line 38: Does it make sense to use the theme's color named "annotation" for this text color? (`fill='annotation'`)

src/domain/infected-people/age-group/age-group-chart.tsx
```diff
@@ -0,0 +1,285 @@
...
+}: AgeGroupChartProps) {
+  const values = data.values.sort((a, b) => {
+    const aStart = parseInt(a.age_group_range, 10);
+    const bStart = parseInt(b.age_group_range, 10);
+    return bStart - aStart;
+  });
+
+  // Define the graph dimensions and margins
+  const breakpoints = useBreakpoints();
+  const isSmallScreen = !breakpoints.md;
```
VWSCoronaDashboard10, line 62: Previously you compared against the actual component width. I think that makes more sense because we're rendering for example less ticks when there's less space. A breakpoint doesn't necessarily mean a smaller component, it depends on where it is rendered (a containing div could e.g. be 500px wide, while the viewport is 1300px wide).

In this case it doesn't really matter based on how we currently render charts, but I actually do like the `parentWidth` implementation and it will be more future proof. It is also more performant because we already receive the parentWidth thus we don't need the extra overhead of the `useBreakpoints` hook.

src/domain/infected-people/age-group/age-group-chart.tsx
VWSCoronaDashboard10: And can we use fontsizes from our theme? `fontSize={2}`

src/domain/infected-people/age-group/age-group-chart.tsx
```diff
@@ -0,0 +1,285 @@
...
+      aria-label={text.graph.accessibility_description}
+    >
+      <Text
+        textAnchor="end"
+        verticalAnchor="start"
+        y={0}
+        x={width / 2 - ageRangeAxisWidth / 2}
+        fill="#000"
+        fontWeight="bold"
+        fontSize={isSmallScreen ? '1rem' : '1.2rem'}
```
VWSCoronaDashboard10, line 170: same here about the fontsize `fontSize={isSmallScreen ? 2 : 3}` - only difference is that fontSize[3] maps to `1.42383rem` but that might be something to discuss with design.

src/domain/infected-people/age-group/age-group-chart.tsx
```diff
@@ -0,0 +1,285 @@
...
+        numTicks={numTicks}
+        stroke="#c4c4c4"
+      />
+
+      {values.map((d, i) => {
+        const ageGroupPercentageWidth = xMax - ageGroupPercentagePoint(d);
+        const infectedPercentageWidth = infectedPercentagePoint(d);
+        return (
+          <Group
+            key={`bar-${i}`}
```
VWSCoronaDashboard10, line 211: in this case ```key={`bar-${i}`}``` equals `key={i}`

src/domain/infected-people/age-group/age-group-tooltip.tsx
```diff
@@ -0,0 +1,93 @@
...
+    borderRadius: 1,
+    zIndex: 42,
+  })
+);
+
+const Legend = styled.ul(
+  css({
+    borderTop: '1px solid',
+    borderTopColor: 'lightGray',
+    m: '0',
```
VWSCoronaDashboard10, line 39:
```suggestion
    m: 0,
```

src/domain/infected-people/age-group/age-group-tooltip.tsx
```diff
@@ -0,0 +1,93 @@
...
+    px: 0,
+    py: 2,
+  })
+);
+
+const LegendItem = styled.li(
+  css({
+    listStyle: 'none',
+    m: '0',
```
VWSCoronaDashboard10, line 48:
```suggestion
    m: 0,
```

src/pages/landelijk/positief-geteste-mensen.tsx
```diff
@@ -28,35 +27,54 @@ import { FCWithLayout } from '~/components/layout';
...
-import { replaceKpisInText } from '~/utils/replaceKpisInText';
 
 const text = siteText.positief_geteste_personen;
 const ggdText = siteText.positief_geteste_personen_ggd;
 
+function getAgeGroupExampleData(data: NationalInfectedAgeGroups) {
+  const ageGroupRange = '20-29';
+
+  const value = data.values.find(
+    (x: NationalInfectedAgeGroupsValue) => x.age_group_range === ageGroupRange
```
VWSCoronaDashboard10, line 53: is it really necessary to type the parameter? Isn't `data.values` not already typed?
```suggestion
    (x) => x.age_group_range === ageGroupRange
```

src/domain/infected-people/age-group/age-group-chart.tsx
VWSCoronaDashboard1: This makes perfect sense actually...

src/domain/infected-people/age-group/age-group-chart-wrapper.tsx
```diff
@@ -0,0 +1,90 @@
+import css from '@styled-system/css';
```
VWSCoronaDashboard8, line 1: If we are going to place components in the domain folder, I think it would be good to have some discussion about how we want to use that folder. Otherwise, hooks, helper functions, and components will get mixed there at the root and I'm not sure that preferable.

Also, I doubt if the "age-groups" folder is required within infected-people, because you are still using that name as part of the component names too. Might be enough to create a folder "components" and place the things directly there.

src/domain/infected-people/age-group/age-group-chart-wrapper.tsx
```diff
@@ -0,0 +1,90 @@
...
+  value: NationalInfectedAgeGroupsValue;
+}
+
+const Wrapper = styled.div(
+  css({
+    position: 'relative',
+  })
+);
+
+export function AgeGroupChartWrapper({ data }: AgeGroupChartWrapperProps) {
```
VWSCoronaDashboard8, line 29: Nit; If this is the component that is used by 'the outside', in other words, if it's always the required interface, I think it makes sense to call this AgeGroupChart and the other thing AgeGroupChartChart or AgeGroupChartBase. I have no good name for it, but I suspect those are internal components. 

My point is, the public interface should have a clean name IMO

src/domain/infected-people/age-group/age-group-chart.tsx
```diff
@@ -0,0 +1,285 @@
...
+  parentWidth: number;
+  data: NationalInfectedAgeGroups;
+  openTooltip: (
+    event: MouseEvent<SVGGElement>,
+    value: NationalInfectedAgeGroupsValue,
+    getTooltipCoordinates: (
+      event: MouseEvent<SVGGElement>,
+      value: NationalInfectedAgeGroupsValue
+    ) => { x: number; y: number }
+  ) => void;
```
VWSCoronaDashboard8, line 32: This interface doesn't look right, as in way to complicated. What is openTooltip supposed to do functionally? 

Is is a handler? (in which case I think it should be called onOpenTooltip) or is it triggering the opening of a tooltip?

src/domain/infected-people/age-group/age-group-chart.tsx
```diff
@@ -0,0 +1,285 @@
...
+  const margin = { top: 35, bottom: 20, left: marginX, right: marginX };
+
+  const numTicks = isSmallScreen ? 3 : 4;
+
+  // Then we'll create some bounds
+  const xMax = (width - margin.left - margin.right - ageRangeAxisWidth) / 2;
+  const yMax = height - margin.top - margin.bottom;
+
+  // We'll make some helpers to get at the data we want
+  const ageGroupPercentage = (d: NationalInfectedAgeGroupsValue) =>
```
VWSCoronaDashboard8, line 76: These names do not indicate that they are functions

src/domain/infected-people/age-group/age-group-chart.tsx
```diff
@@ -0,0 +1,285 @@
...
+    d.infected_percentage;
+  const ageGroupRange = (d: NationalInfectedAgeGroupsValue) =>
+    d.age_group_range;
+
+  // And then scale the graph by our data
+  const domainPercentages = [
+    0,
+    Math.max(
+      ...values.map((d) =>
+        Math.max(ageGroupPercentage(d), infectedPercentage(d))
```
VWSCoronaDashboard8, line 88: If this is the only place they are used, the functions are really unnecessary and only bloating the code IMO

src/domain/infected-people/age-group/age-group-chart.tsx
```diff
@@ -0,0 +1,285 @@
...
+        top={height - margin.bottom}
+        numTicks={numTicks}
+        hideTicks={true}
+        hideAxisLine={true}
+        tickFormat={(a) => `${formatPercentage(a as number)}%`}
+        tickComponent={TickValue}
+      />
+    </svg>
+  );
+}
```
VWSCoronaDashboard8, line 285: This component is very large. Maybe you can extract some components and helper functions. We could also do it as part of a different PR because this is already a lot of code.

src/domain/infected-people/age-group/age-group-tooltip.tsx
```diff
@@ -0,0 +1,93 @@
...
+    pl: '2.75rem',
+    pr: 3,
+    py: 1,
+    position: 'relative',
+    '&:before': {
+      content: '""',
+      display: 'block',
+      width: '1rem',
+      height: '1rem',
+      background: '#c6c8ca',
```
VWSCoronaDashboard8, line 58: If colors are not part of the theme yet, we should probably add them.

src/domain/infected-people/age-group/age-group-tooltip.tsx
```diff
@@ -0,0 +1,93 @@
...
+      position: 'absolute',
+      left: 3,
+      top: '0.5rem',
+    },
+    '&.infected-percentage:before': {
+      background: '#3391cc',
```
VWSCoronaDashboard8, line 64: same for this one

src/pages/landelijk/positief-geteste-mensen.tsx
```diff
@@ -28,35 +27,54 @@ import { FCWithLayout } from '~/components/layout';
...
+  );
+
+  if (!value) {
+    return {};
```
VWSCoronaDashboard8, line 57: I suspect it makes sense to fail hard using an assert here. Is there a practical benefit to returning undefined values? Because the downside is of course that you need to handle those values down the line.

VWSCoronaDashboard8 (commented):
I think we might be able to postpone my points about abstractions and file composition to a later time.

src/domain/infected-people/age-group/age-group-chart.tsx
VWSCoronaDashboard4: The most impotant change that is triggered with the `isSmallScreen` is the alignment of the tooltip. It cannot flow outside the component on the left or right on smaller screens, and that is perfectly trigger-able with the breakpoint.

I can revert this change to the breakpoints, but we do have to accept the use of a certain fixed number to toggle this behavior anyway.

src/domain/infected-people/age-group/age-group-chart.tsx
VWSCoronaDashboard4: This is not part of styled system, it's a property provided by visx.

src/pages/landelijk/positief-geteste-mensen.tsx
VWSCoronaDashboard4: It's in a separate file, so not per se local

src/domain/infected-people/age-group/age-group-chart.tsx
VWSCoronaDashboard4: In this case it is hard; removing them will get complaints about the implicit any inferring; trying to set more explicit types is also a pain since it is composing certain data types.

I'll leave these `any` types for now.

src/domain/infected-people/age-group/age-group-chart.tsx
VWSCoronaDashboard10: Ah of course.

src/domain/infected-people/age-group/age-group-chart.tsx
VWSCoronaDashboard4: This file is calculating what the graph's coordinates are, and then render it. What should be moved elsewhere?

src/domain/infected-people/age-group/age-group-chart.tsx
VWSCoronaDashboard4: This code (& interface) is subject to changes with an upcoming PR for the keyboard navigation. Right now it does a loop from the graph to the tooltip, back to the graph mainly for performance reasons (otherwise the code in the tooltip needs to be re-evaluated each re-render)

src/domain/infected-people/age-group/age-group-chart.tsx
VWSCoronaDashboard4: Originally this would be the `x` function the retrieve the value's x value; but we use multiple values on the x axis.

Adding a `get` prefix or `Accessor` suffix would bloat the code too much in my opinion,

src/style/theme.ts
```diff
@@ -101,10 +101,12 @@ export const colors = {
   red: '#F35363',
   category: '#6b6b6b',
   border: '#c4c4c4',
+  lightBlue: '#E0EEF6',
 
   data: {
     primary: '#007BC7',
     secondary: '#154273',
+    default: '#C6C8CA',
```
VWSCoronaDashboard10, line 109: I'm not sure "default" is the right name to choose for this gray color, but on the other hand I can't really think of another name right now .. maybe tertiary ..?

src/domain/infected-people/age-demographic/age-demogrraphic-chart.tsx
```diff
@@ -0,0 +1,292 @@
+import css from '@styled-system/css';
```
VWSCoronaDashboard10, line 1: filename typo :-)

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard10 (approved):
Approved if you change the incorrect filename (`age-demogrraphic-chart.tsx`) 👍

PR #1035, opened 2020-11-26
Do some layout tweaks and introduce the Box spacing prop

VWSCoronaDashboard10:
## Summary

Implemented several style adjustments as discussed with design. I've also introduced two new props on the Box component for easier margin-control of child elements.

## Motivation

When several child components should have some space between them it's easier to set a single value on a parent component instead of applying margins on all child components. I've had good experience with a specific `SpacedChildren` component in the past, but this time I've decided to add it as a feature of the Box component.

## Detailed design

When a Box component received a `space={x}` prop, it will apply the following style definition:
```css
& > *:not(:last-child) { margin-bottom: x }
```

With the additional `spaceHorizontal`-boolean prop it will apply a `margin-right` instead of `margin-bottom`.


src/style/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+ * Styled-system's `css()` function does not support responsive object notation.
+ * This utility will transform a responsive style object notation to its array
+ * equivalent.
+ *
+ * input: `{ _: 2, md: 4, xl: 6 }`
+ * output: `[2, undefined, undefined, 4, undefined, 6]`
+ *
+ */
+export function getArrayValue<T>(
+  value: ResponsiveValue<T>
```
VWSCoronaDashboard1, line 16: the only criticism on this otherwise very useful method would be its name.
Something along the lines of `convertSyledObjectToArray` or something would probably be a bit more descriptive, although I do admit this is a tough one to name ;)

VWSCoronaDashboard1 (approved):

src/style/style-funtions/space-children.ts
```diff
@@ -0,0 +1,23 @@
+import css from '@styled-system/css';
+import { ResponsiveValue, styleFn } from 'styled-system';
+import { isDefined } from 'ts-is-present';
+import { SpaceValue } from '../theme';
+import { getArrayValue } from '../utils';
+
+export interface SpaceChildrenProps {
+  space?: ResponsiveValue<SpaceValue>;
```
VWSCoronaDashboard8, line 8: I think `spacing` would be a better name. When I read `space` I was thinking padding or margins. Spacing is more specific about the space in between elements I think.

src/style/style-funtions/space-children.ts
```diff
@@ -0,0 +1,23 @@
...
+  spaceHorizontal?: boolean;
```
VWSCoronaDashboard8, line 9:
```suggestion
  spaceHorizontally?: boolean;
```

src/style/utils.ts
VWSCoronaDashboard10: I'll try `getResponsiveValue` and wait for other criticism 🙈

src/style/style-funtions/space-children.ts
```diff
@@ -0,0 +1,23 @@
+import css from '@styled-system/css';
```
VWSCoronaDashboard8, line 1: Does it need to be called style-functions if it's already nested under style? Maybe just `style/functions/space-children`

src/style/theme.ts
```diff
@@ -12,6 +12,11 @@ const space = [
   '32rem',
 ];
 
+/**
+ * Valid space index values
+ */
+export type SpaceValue = 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8;
```
VWSCoronaDashboard8, line 18: If it's an index value why not call it that? `SpaceIndexValue` or `SpaceIndex`. I find it confusing if I don't read the comment above it. And if you fix the name the comment is redundant I think.

src/style/utils.ts
VWSCoronaDashboard8: Nice idea to write a util for this! 

I think `convertSyleObjectToArray` might be a little verbose if you want to use it as a util often. I would suggest `asResponsiveArray` or `asArray`. I think the "value" suffix is a little redundant because you call a getter function, so of course you will get a value from it :)

src/style/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+): ResponsiveValue<T> {
+  return isStyleObject(value) ? createArrayValue(value) : value;
+}
+
+type StyleObject<T> = {
+  [key in (ThemeValue<'breakpoints', DefaultTheme> & string) | number]?: T;
```
VWSCoronaDashboard8, line 22: I feel it might be simpler to take the keys from the responsiveKeys array you defined above. Also why are you allowing numbers as keys? 

Maybe something like this works:

`type StyleObject<T> = Record<value of responsiveKeys, T>`

src/style/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+};
+
+function isStyleObject<T>(value: ResponsiveValue<T>): value is StyleObject<T> {
+  return !Array.isArray(value) && value !== null && typeof value === 'object';
+}
+
+function createArrayValue<T>(value: StyleObject<T>) {
+  return responsiveKeys.map(
+    (key) => value[(key as unknown) as number] as T
```
VWSCoronaDashboard8, line 31: I don't see why you would have to cast the keys to numbers here. I suspect that if you use responsiveKeys to define the StyleObject you don't have to cast at all.

VWSCoronaDashboard8 (approved):

src/style/utils.ts
VWSCoronaDashboard8: Or simply create a type `type ResponseKeys = '_' | 'xs' | ...` etc and use that directly for both the array and this type.

VWSCoronaDashboard10: I should have add a comment here that this StyleObject is copy paste from SS's types - it is part of the `ResponsiveValue` type.

I've had a real struggle with typing these things to get rid of all the compilations errors, but I guess I'll just have to give it another try tomorrow morning with some fresh energy.

src/style/theme.ts
VWSCoronaDashboard10: Because these are the only possible values which can be mapped to the spaces array, I thought it wouldn't really matter to know from the outside that the value is actually an index of an array.

src/style/utils.ts
VWSCoronaDashboard10: You were totally right 👍

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard8 (approved):

PR #1036, opened 2020-11-26
Put absolute date on title attribute of relative times

VWSCoronaDashboard10:
Relative time elements will have a title attribute with the absolute date.

### DIFF
--- a/src/components-styled/difference-indicator.tsx
+++ b/src/components-styled/difference-indicator.tsx
@@ -14,6 +14,7 @@ import IconUp from '~/assets/pijl-omhoog.svg';
 import IconDown from '~/assets/pijl-omlaag.svg';
 import siteText from '~/locale/index';
 import { DifferenceDecimal, DifferenceInteger } from '~/types/data';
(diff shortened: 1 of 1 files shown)


src/components-styled/difference-indicator.tsx
```diff
@@ -136,6 +141,17 @@ const Container = styled.div(
   })
 );
 
+function TimespanText({ date }: { date: number }) {
+  const fullDate = formatDateFromMilliseconds(date * 1000, 'medium');
+  const text = getTimespanText(date);
+
+  return (
+    <Span suppressHydrationWarning title={fullDate}>
```
VWSCoronaDashboard8, line 149: Hmm so that did just work after all 🤔

VWSCoronaDashboard8 (approved):

src/components-styled/difference-indicator.tsx
VWSCoronaDashboard10: yep 🤷

PR #1041, opened 2020-11-30
Topic/elderly at home schemas

VWSCoronaDashboard10:
## Summary
Create schema's for "elderly at home" data.

This needs a thorough review because it's my first schema definition :-)

Adding the schema resulted in a TS error due to missing choropleth thresholds. When adding the thresholds, I ran into an issue with accessing a color in a theme color array on an undefined index. Exporting those colors `as const` will yield an error when you try to get a color at index X while it does not exist.

### DIFF
--- a/schema/municipal/elderly_at_home.json
+++ b/schema/municipal/elderly_at_home.json
@@ -1,0 +1,54 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "municipal_elderly_at_home",
+  "type": "object",
+  "properties": {
+    "values": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/value"
+      }
+    },
+    "last_value": {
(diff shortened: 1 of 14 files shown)


schema/municipal/elderly_at_home.json
```diff
@@ -0,0 +1,54 @@
...
+  "additionalProperties": false,
+  "definitions": {
+    "value": {
+      "title": "municipal_elderly_at_home_value",
+      "type": "object",
+      "properties": {
+        "positive_tested_daily": {
+          "type": "integer"
+        },
+        "positive_tested_daily_permillage": {
```
VWSCoronaDashboard8, line 26: Chique 😄

VWSCoronaDashboard8 (approved):
It's a bit more than just schema's but looking good 💯

VWSCoronaDashboard1 (approved):

PR #1042, opened 2020-11-30
Unify hospital schemas

VWSCoronaDashboard8:
## Summary

Unify and clean up schemas around hospital data.

This branch should only be merged to develop once the BE has updated the data and FE has updated the code to handle the new types / names.

### NL

- intake_hospital_ma moving_average_hospital => hospital
admissions_moving_average
- hospital_beds_occupied covid_occupied => hospital beds_occupied_covid
- difference intake_hospital_ma__moving_average_hospital => difference
hospital__admissions_moving_average
- difference hospital_beds_occupied__covid_occupied => difference
hospital__beds_occupied_covid

### VR

- results_per_region hospital_moving_avg_per_region => hospital
admissions_moving_average
- results_per_region hospital_total_counts_per_region => (delete)
- results_per_region hospital_increase_per_region => (delete)
- difference results_per_region__hospital_moving_avg_per_region => hospital__admissions_moving_average

### GM

- hospital_admissions moving_average_hospital =>
hospital admissions_moving_average
- difference hospital_admissions__moving_average_hospital => difference hospital__admissions_moving_average

PR #1051, opened 2020-11-30
Hospital schemas unify snowball

VWSCoronaDashboard8:
## Summary

This PR was aimed at altering the code to match the schema updates in #1042. Then it turned out I couldn't use a nested property on the GM level because the required changes for this type of structure were only made to VR at the time. While working on this I started defining stricter types, and things snowballed from there.

I decided to create a separate navigation map component for the GM index page, because the fact that it was used with and without data greatly complicated the code and the type definitions involved.

I also changed the way tooltips get served data. Previously a value was set on the "value" property, but also the value was spread across the context container. This PR changes it so that the value for the tooltip is always found on the value named property.

### DIFF
--- a/src/components/choropleth/hooks/use-municipality-data.ts
+++ b/src/components/choropleth/hooks/use-municipality-data.ts
@@ -9,6 +9,7 @@ import {
   TMunicipalityMetricType,
 } from '../shared';
 import set from 'lodash/set';
+import get from 'lodash/get';
 
 /**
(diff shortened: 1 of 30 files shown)


VWSCoronaDashboard1 (approved):

VWSCoronaDashboard8:
Converted to draft because merging this in the base would confuse BE probably. Let's first get the schemas implemented.

VWSCoronaDashboard8:
Closed in favor of #1065 where the refactoring work was executed separately from schema changes

PR #1058, opened 2020-12-01
Create elderly at home regional page

VWSCoronaDashboard10:
## Summary
Create the "elderly at home" regional page.

### DIFF
--- a/src/pages/veiligheidsregio/[code]/thuiswonende-ouderen.tsx
+++ b/src/pages/veiligheidsregio/[code]/thuiswonende-ouderen.tsx
@@ -1,0 +1,132 @@
+import ElderlyIcon from '~/assets/elderly.svg';
+import { ContentHeader } from '~/components-styled/content-header';
+import { KpiTile } from '~/components-styled/kpi-tile';
+import { KpiValue } from '~/components-styled/kpi-value';
+import { LineChartTile } from '~/components-styled/line-chart-tile';
+import { TwoKpiSection } from '~/components-styled/two-kpi-section';
+import { Text } from '~/components-styled/typography';
+import { FCWithLayout } from '~/components/layout';
+import { getSafetyRegionLayout } from '~/components/layout/SafetyRegionLayout';
+import { SEOHead } from '~/components/seoHead';
+import siteText from '~/locale/index';
+import {
+  getSafetyRegionData,
+  getSafetyRegionPaths,
+  ISafetyRegionData,
+} from '~/static-props/safetyregion-data';
+
+const text = siteText.thuiswonende_ouderen;
+
+const ElderlyAtHomeRegionalPage: FCWithLayout<ISafetyRegionData> = (props) => {
+  const elderlyAtHomeData = props.data.elderly_at_home;
+
+  return (
+    <>
+      <SEOHead
+        title={text.metadata.title}
+        description={text.metadata.description}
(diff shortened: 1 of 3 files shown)


VWSCoronaDashboard1 (approved):

src/pages/veiligheidsregio/[code]/thuiswonende-ouderen.tsx
```diff
@@ -0,0 +1,132 @@
...
+import { getSafetyRegionLayout } from '~/components/layout/SafetyRegionLayout';
+import { SEOHead } from '~/components/seoHead';
+import siteText from '~/locale/index';
+import {
+  getSafetyRegionData,
+  getSafetyRegionPaths,
+  ISafetyRegionData,
+} from '~/static-props/safetyregion-data';
+
+const text = siteText.thuiswonende_ouderen;
```
Comment (VWSCoronaDashboard8) [inline src/pages/veiligheidsregio/[code]/thuiswonende-ouderen.tsx:18]: I don't know if it's feasible to re-use the complete nl structure here. I think in a lot of cases the vr and gm pages have their own separate copy. So it might be worth duplicating and prefixing it with vr_ to scope it. I think we should probably discuss this in the team. If we always want to separate text like this I would opt to start prefixing the nl stuff with nl_ too.

src/pages/veiligheidsregio/[code]/thuiswonende-ouderen.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  return (
+    <>
+      <SEOHead
+        title={text.metadata.title}
+        description={text.metadata.description}
+      />
+
+      <ContentHeader
+        category={siteText.nationaal_layout.headings.kwetsbare_groepen}
```
Comment (VWSCoronaDashboard8) [inline src/pages/veiligheidsregio/[code]/thuiswonende-ouderen.tsx:31]: this is not part of the national layout but maybe that's ok. Just a misleading scope name maybe

VWSCoronaDashboard8 (approved):

src/pages/veiligheidsregio/[code]/thuiswonende-ouderen.tsx
Comment (VWSCoronaDashboard10) [reply src/pages/veiligheidsregio/[code]/thuiswonende-ouderen.tsx:18]: There are no differences when looking at the design, that's why I opted for using a single structure. We can duplicate it the moment it becomes necessary.

src/pages/veiligheidsregio/[code]/thuiswonende-ouderen.tsx
Comment (VWSCoronaDashboard10) [reply src/pages/veiligheidsregio/[code]/thuiswonende-ouderen.tsx:31]: whoops, no good catch, need to change this.

src/pages/veiligheidsregio/[code]/thuiswonende-ouderen.tsx
```diff
@@ -0,0 +1,132 @@
...
+  return (
+    <>
+      <SEOHead
+        title={text.metadata.title}
+        description={text.metadata.description}
+      />
+
+      <ContentHeader
+        category={siteText.veiligheidsregio_layout.headings.kwetsbare_groepen}
+        title={text.section_positive_tested.title}
```
Comment (VWSCoronaDashboard4) [inline src/pages/veiligheidsregio/[code]/thuiswonende-ouderen.tsx:32]: Titles for other metrics include the safety region name as part of the title

VWSCoronaDashboard4 (changes requested):

VWSCoronaDashboard4 (approved):

PR #1059, opened 2020-12-01
Create elderly at home sidebar menu items

VWSCoronaDashboard10:
## Summary

Create elderly at home sidebar menu items.

### DIFF
--- a/src/components/common/elderly-at-home-metric.tsx
+++ b/src/components/common/elderly-at-home-metric.tsx
@@ -8,13 +8,11 @@ import { replaceVariablesInText } from '~/utils/replaceVariablesInText';
 const text = siteText.common.metricKPI;
 const title = siteText.thuiswonende_ouderen.titel_kpi;
 
-export function ElderlyAtHomeMetric(props: {
-  data: NationalElderlyAtHomeValue | undefined;
-}) {
-  const { data } = props;
+interface ElderlyAtHomeMetricProps {
+  data: NationalElderlyAtHomeValue;
+}
 
-  if (data === undefined) return null;
-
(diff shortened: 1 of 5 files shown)


VWSCoronaDashboard1 (approved):

src/components/common/elderly-at-home-metric.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { NationalElderlyAtHomeValue } from '~/types/data.d';
+import { formatDateFromSeconds } from '~/utils/formatDate';
+import { formatNumber } from '~/utils/formatNumber';
+import { replaceVariablesInText } from '~/utils/replaceVariablesInText';
+
+const text = siteText.common.metricKPI;
+const title = siteText.thuiswonende_ouderen.titel_kpi;
+
+export function ElderlyAtHomeMetric(props: {
+  data: NationalElderlyAtHomeValue | undefined;
```
VWSCoronaDashboard8, line 12: Why would it be undefined? Are there VR regions that do not guarantee data?

VWSCoronaDashboard8 (approved):

src/components/common/elderly-at-home-metric.tsx
VWSCoronaDashboard10: Good question, it is copy-paste of another metric. The `last_value` is passed on the props so I guess it could/should never be undefined.

VWSCoronaDashboard10: I'll fix it for this specific component, but its obvious there's still some clean up work left.

VWSCoronaDashboard4 (approved):

PR #1062, opened 2020-12-02
Per 100k instead of 1000

VWSCoronaDashboard10:
## Summary

The positive tested relative value is per 100.000 instead of per 1000.

I assumed postfixing it with `_per_100k` would be more clear than the alternatives:
- `_per_100000`
- `_per_100_000`
- `_per_cent_mille`
- `_pcm`
- `_milli_percent`

Another approach would be to postfix it with `_relative`..

### DIFF
--- a/schema/national/elderly_at_home.json
+++ b/schema/national/elderly_at_home.json
@@ -23,7 +23,7 @@
         "positive_tested_daily": {
           "type": "integer"
         },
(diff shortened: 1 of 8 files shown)


VWSCoronaDashboard8 (approved):
Indeed clearer, although I liked the word permillage 😄

PR #1063, opened 2020-12-02
Fix build error due to strict timestamp check

VWSCoronaDashboard4:
## Summary

See title

## Detailed design

Removed the throwing of the error part.

### DIFF
--- a/src/static-props/data-sorting.ts
+++ b/src/static-props/data-sorting.ts
@@ -90,13 +90,7 @@ function sortTimeSeriesValues(values: Timestamped[]) {
     );
   }
 
-  /**
-   * If none match we throw, since it means an unknown timestamp is used and we
-   * want to be sure we sort all data.
-   */
-  throw new Error(
-    `Unknown timestamp in value ${JSON.stringify(values[0], null, 2)}`
-  );
+  return values;
 }
 
 type Timestamped = ReportTimestamped | WeekTimestamped | MeasurementTimestamped;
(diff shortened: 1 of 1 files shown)


src/static-props/data-sorting.ts
```diff
@@ -90,13 +90,19 @@ function sortTimeSeriesValues(values: Timestamped[]) {
...
+
   /**
    * If none match we throw, since it means an unknown timestamp is used and we
    * want to be sure we sort all data.
    */
   throw new Error(
     `Unknown timestamp in value ${JSON.stringify(values[0], null, 2)}`
   );
+
+  return values;
```
VWSCoronaDashboard10, line 105: this return statement is unreachable?

VWSCoronaDashboard10 (approved):

src/static-props/data-sorting.ts
```diff
@@ -50,7 +50,7 @@ export function sortMunicipalTimeSeriesInDataInPlace(data: Municipal) {
       >;
 
       nestedSeries.values = nestedSeries.values.map((x) => {
-        x.values = sortTimeSeriesValues(x.values);
+        x.values = sortTimeSeriesValues(x.values, propertyName);
```
VWSCoronaDashboard8, line 53: I don't see a reason pass the property down to the sort function which then decides if it needs to sort of not. You can do the whitelist check before calling this function I think, before line 52

VWSCoronaDashboard8 (changes requested):

src/static-props/data-sorting.ts
VWSCoronaDashboard10: I also thought of that improvement, but then I realised the current implementation allows whitelisted keys to be sorted when they *also* match one of the timestamp formats. I'm not sure if that's a relevant use-case though..

VWSCoronaDashboard8 (approved):

PR #1065, opened 2020-12-02
Refactor choropleths for stricter data handling

VWSCoronaDashboard8:
## Summary

I almost forgot how this started. Something with trying to render a nested property value in GM choropleths. I started making some parts stricter and things unraveled from there.

- Do not use an empty choropleth for VR and GM navigation. It complicates the code. This PR introduces a separate component for navigation.
- Define choropleth values explicitly, by pointing to the metric and the property everywhere. The same goes for thresholds. No more implicit fallbacks.
- Add a clear definition for data shapes, where the color values for the choropleth (which are always expected to be a number) are clearly separated from the other data.
- Tooltips do not use this color value but take whatever they need from the metric properties directly.

I probably forgot some things here.


src/components/choropleth/shared.d.ts
```diff
@@ -31,6 +31,7 @@ export interface SafetyRegionProperties {
 export interface MunicipalityProperties {
   gemnaam: string;
   gemcode: string;
+  gmcode: string;
```
VWSCoronaDashboard1, line 34: perhaps for future improvement, but gmcode and gemcode would preferably be consolidated into one. Most probably gmcode, is my guess.

VWSCoronaDashboard1 (approved):

src/components/choropleth/shared.d.ts
VWSCoronaDashboard8: I fully agree, but I didn't want to burn myself on that. It might involve changes from backend / schema's as well.

src/components/choropleth/tooltips/municipal/create-positive-tested-people-municipal-tooltip.tsx
```diff
@@ -28,9 +30,9 @@ export const createPositiveTestedPeopleMunicipalTooltip = (
     context && (
       <TooltipContent title={gemnaam} onSelect={onSelect}>
         <span>
-          {value !== undefined ? (
+          {positive_tested_people !== undefined ? (
             <>
-              <p className="info-value">{value} per 100.000</p>
+              <p className="info-value">{positive_tested_people} per 100.000</p>
```
VWSCoronaDashboard4, line 35: Hmmmm, this text should be moved to a translation; because in different languages the wording and the number formatting can be different.

VWSCoronaDashboard4 (commented):
Overall this looks very good! Some specific comments.

src/components-styled/content-header/metadata.tsx
```diff
@@ -92,27 +92,16 @@ function MetadataItem(props: MetadataItemProps) {
...
-          }
-          return (
-            <ExternalLink
-              href={item.href}
-              text={item.text}
-              key={index + item.href}
-            />
-          );
-        })}
+        {items.map((item, index) =>
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: .join(' & ') won't work with React component unfortunately. Your out put will become [Object] & [Object].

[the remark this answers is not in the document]
VWSCoronaDashboard8: I was mainly fixing a react warning about the key prop. I don't think you can place a key props on a fragment since it will not result in an element in the tree.

src/components/choropleth/hooks/use-safety-region-data.ts
```diff
@@ -1,17 +1,34 @@
...
+  [key: string]: unknown;
+}
+
+export interface RegionChoroplethValue extends RegionMetricValue {
+  __color_value: number;
+}
+
+export type GetRegionDataFunctionType = (id: string) => RegionChoroplethValue;
+
+type UseDataReturnValue = {
```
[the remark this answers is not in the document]
VWSCoronaDashboard8: Agree, I will see if I can improve this. It's the return value of both the useMunicipalityData and useRegionsData hook. That's why it has a generic name.

src/components/choropleth/tooltips/municipal/create-positive-tested-people-municipal-tooltip.tsx
VWSCoronaDashboard8: I will make it a separate task in trello

src/components-styled/content-header/metadata.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard4: @VWSCoronaDashboard8 Keyed fragments is what react provides in their docs when you return multiple elements in an iteration: https://reactjs.org/docs/fragments.html#keyed-fragments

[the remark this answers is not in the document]
VWSCoronaDashboard8: Nice! I didn't know

src/components-styled/content-header/metadata.tsx
```diff
@@ -92,27 +92,16 @@ function MetadataItem(props: MetadataItemProps) {
...
+          index > 0 ? (
+            <Fragment key={index + item.href}>
+              {' & '}
+              <ExternalLink href={item.href} text={item.text} />
+            </Fragment>
+          ) : (
+            <ExternalLink href={item.href} text={item.text} />
+          )
+        )}
```
VWSCoronaDashboard10, line 104: I think writing it like this is more readable:
```tsx
{items.map((item, index) =>
  <Fragment key={index + item.href}>
    {index > 0 && ' & '}
    <ExternalLink href={item.href} text={item.text} />
  </Fragment>
)}

File: src/components/choropleth/legenda/utils.ts [hunk 6]
```diff
@@ -1,20 +1,22 @@
...
+  const thresholds = metricProperty
+    ? get(thresholdData, [metricName, metricProperty])
+    : thresholdData[metricName];
 
-  return thresholdInfo as ChoroplethThresholdsValue[];
+  assert(
+    thresholds,
+    `No thresholds are defined for ${metricName}${
+      metricProperty ? `:${metricProperty}` : ''
+    }`
```
VWSCoronaDashboard10, line 18: nit:
```suggestion
    `No thresholds are defined for ${[metricName, metricProperty].filter(x => !!x).join(':')}`
```

src/components/choropleth/municipality-navigation-map.tsx
```diff
@@ -0,0 +1,97 @@
+import css from '@styled-system/css';
```
VWSCoronaDashboard10, line 1: Good one on making this navigation map a separate component 👍

src/components/choropleth/tooltips/region/escalation-tooltip.tsx
```diff
@@ -25,29 +23,29 @@ export const escalationTooltip = (router: NextRouter) => {
       );
     };
 
+    const escalationText = (text.escalatie_niveau.types as Record<
+      string,
+      unknown
+    >)[level] as { titel: string; valid_from: string };
```
VWSCoronaDashboard10, line 29: isn't it possible to write this as
```suggestion
    const escalationText = (text.escalatie_niveau.types as Record<
      string,
      { titel: string; valid_from: string }
    >)[level];
```

VWSCoronaDashboard10 (approved):

src/components/choropleth/legenda/utils.ts
VWSCoronaDashboard8: Funny. I wonder why you find that more readable code. I prefer what I had.

src/components-styled/content-header/metadata.tsx
VWSCoronaDashboard8: I keep getting comments on this piece of code that I only altered because it generated a react warning 😄

src/components/choropleth/tooltips/region/escalation-tooltip.tsx
VWSCoronaDashboard8: I gave it a try. TS says no

VWSCoronaDashboard8: This worked

```
const escalationText = ((text.escalatie_niveau.types as unknown) as Record<
      EscalationLevel,
      { titel: string; valid_from: string }
    >)[level];
```

src/components/choropleth/legenda/utils.ts
VWSCoronaDashboard8: Aaah. maybe with a simplified filter I can see it. 
```suggestion
    `No thresholds are defined for ${[metricName, metricProperty].filter(isDefined).join(':')}`
```

src/components/choropleth/tooltips/municipal/create-positive-tested-people-municipal-tooltip.tsx
VWSCoronaDashboard14: https://github.com/minvws/nl-covid19-data-dashboard/pull/1286

PR #1066, opened 2020-12-03
Behavior feature - remove 3 kpis in VR

VWSCoronaDashboard8:
## Summary

Removing 3 kpis that will never be used.


src/domain/behavior/behavior-line-chart-tile.tsx
```diff
@@ -39,7 +39,9 @@ export function BehaviorLineChartTile({
        * result in a "line" in our line-chart.
        */
       const hasEnoughData =
-        values.flatMap((x) => x[valueKey]).filter(Boolean).length > 1;
+        (values as NationalBehaviorValue[])
```
VWSCoronaDashboard4, line 42: Why cast this to `NationalBehaviorValue`? It can also be a `RegionalBehaviorValue`?

src/domain/behavior/behavior-line-chart-tile.tsx
```diff
@@ -104,7 +106,7 @@ export function BehaviorLineChartTile({
 
       <BehaviorLineChart
         values={behaviorIdentifierWithData.map(({ valueKey, label }) =>
-          values
+          (values as NationalBehaviorValue[])
```
VWSCoronaDashboard4, line 109: Why cast this to NationalBehaviorValue? It can also be a RegionalBehaviorValue?

src/domain/behavior/behavior-line-chart-tile.tsx
VWSCoronaDashboard8: It's a bit of a hack. I didn't know how else to make typescript happy. It was complaining in the flatMap if I left the value typed as a combination of the two.

src/domain/behavior/behavior-line-chart-tile.tsx
VWSCoronaDashboard8: Same here

PR #1068, opened 2020-12-03
Implement scroll to nav bar behavior

VWSCoronaDashboard10:
## Summary

When navigation through the app we'll scroll to the horizontal navigation bar instead of all the way to the top (unless your scrolltop is less than that nav-bar, in that case it will scroll to `0`).

I've also refactored all header-related components to be styled with styled-components.

Some notes:

- The next.js `<Link />` component should not be used anymore, I've created an eslint rule to enforce that. The default scroll behavior of that `<Link />` component is not desired, we'll scroll manually to a specific position.  
Instead of adding `scroll={false}` to every Link-instance it's easier and less error-prone to create an extra abstraction (located at `utils/link`) which will default that scroll prop to false.
- I ran into an issue during the initial render of a choropleth, the height was 0 untill a DOM element was measured. I've introduced an `<AspectRatio />` component which will set its height based on the width, using css.
- improved the `useBreakpoints` hook where an initial value can be read from the client


src/components-styled/aspect-ratio.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+interface AspectRatioProps {
+  /**
+   * ratio is expressed as width / height
+   */
+  ratio: number;
+  children: React.ReactNode;
+}
+
+export function AspectRatio({ ratio, children }: AspectRatioProps) {
```
VWSCoronaDashboard1, line 13: This component sounds intriguing, but I don't understand immediately what it does exactly. Could you perhaps add a small code comment that explains the usage?

VWSCoronaDashboard1: Aha, it becomes clearer when I saw it used in the choropleths ;)

VWSCoronaDashboard1 (approved):
Another batch of really nice cleanups!

.eslintrc.json
```diff
@@ -65,7 +65,11 @@
           {
             "argsIgnorePattern": "^_"
           }
-        ]
+        ],
+        "no-restricted-imports": ["error", {
+          "name": "next/link",
+          "message": "Please use the custom ~/utils/link instead"
```
VWSCoronaDashboard8, line 71: Nice I didn't think this was that easy. Maybe we can add one later for console/assert because I keep tripping over vscode automatically importing that one. Or maybe I should just find a way to tell vscode to not do that :)

VWSCoronaDashboard8 (approved):
Awesome 🔥

src/components-styled/aspect-ratio.tsx
VWSCoronaDashboard10: yep, it's a component that will have height based on its width, calculated with a ratio :-)

src/components-styled/max-width.tsx
```diff
@@ -0,0 +1,15 @@
...
+import styled from 'styled-components';
+import { Box } from '~/components-styled/base';
+
+export const MaxWidth = styled(Box)(
+  css({
+    maxWidth: 1200,
+    margin: '0 auto',
+    padding: '0 1em',
+
+    '@media (min-width: 1400px)': {
```
VWSCoronaDashboard4, line 11: Shouldn't this use the styled system breakpoints?

src/domain/site/components/top-navigation.tsx
```diff
@@ -0,0 +1,115 @@
...
+      </Link>
+    </li>
+  );
+}
+
+const NavLink = styled.a<{ isActive: boolean }>((x) =>
+  css({
+    whiteSpace: 'nowrap',
+    display: 'block',
+    px: ['0.75em', null, '1.5em'],
```
VWSCoronaDashboard4, line 74: Can the sizes used here and elsewhere in this file be used from styled system?

VWSCoronaDashboard4: Same for colors

src/domain/site/site-footer.tsx
```diff
@@ -0,0 +1,133 @@
...
+      content: '"›"',
+      mr: 3,
+    },
+  })
+);
+
+const linkStyle = css({
+  color: 'inherit',
+  textDecoration: 'none',
+  '&:hover': {
```
VWSCoronaDashboard4, line 127:
```suggestion
  '&:hover, &:focus': {
```

src/domain/site/skip-links.tsx
```diff
@@ -0,0 +1,49 @@
...
+    left: '50%',
+    width: '100%',
+    transform: 'translateX(-50%)',
+  })
+);
+
+const SkipLink = styled.a(
+  css({
+    position: 'absolute',
+    fontSize: '1.2em',
```
VWSCoronaDashboard4, line 30: can we use fontsizes/paddings from styled system here?

src/components-styled/aspect-ratio.tsx
```diff
@@ -0,0 +1,36 @@
+import css from '@styled-system/css';
+import React from 'react';
```
VWSCoronaDashboard4, line 2:
```suggestion
import { ReactNode } from 'react';
```

src/components-styled/aspect-ratio.tsx
```diff
@@ -0,0 +1,36 @@
...
+import styled from 'styled-components';
+
+interface AspectRatioProps {
+  /**
+   * ratio is expressed as width / height
+   */
+  ratio: number;
+  children: React.ReactNode;
```
VWSCoronaDashboard4, line 10:
```suggestion
  children: ReactNode;
```

VWSCoronaDashboard4 (commented):
Nice cleanup! Some nitpicks/suggestions/questions :-)

src/domain/site/skip-links.tsx
VWSCoronaDashboard1: I'm not even sure if we need a font size for this menu, I mean, it's only visible to screenreaders, right?

src/domain/site/components/top-navigation.tsx
VWSCoronaDashboard10: In theory yes, but the spaces defined in our theme are a bit different than the ones used in here. I wanted to avoid any feedback about different margins and paddings, I think we should pick that up together with a designer.

src/domain/site/site-footer.tsx
VWSCoronaDashboard10: Well spotted 👍  I went for an outline style.

src/domain/site/skip-links.tsx
VWSCoronaDashboard10: no, in theory it's also valuable for people who can see the tiniest details but are unable to control a mouse input device.

src/components-styled/max-width.tsx
VWSCoronaDashboard10: There's no breakpoint at 1400px and this breakpoint goes hand in hand with the maxWidth value, so I'd rather not touch it instead of changing the maxWidth size to 1200 or 1600 px (which are existing breakpoints).

PR #1069, opened 2020-12-03
Refactor sidebar kpi part 1

VWSCoronaDashboard8:
## Summary

The idea was to refactor the MetricKPI component used in the sidebar to make it generic so we don't have to keep repeating the same changes through many specialized component files like `components/landelijk/hospital_intake_metric` etc.

This PR snowballed a bit and eventually takes a more radical approach. In order to render the data using generic components, I have defined a configuration file where we can declare properties on each of the data values that we are trying to render. Then these definitions are used to lookup any specifics required by the components.

A declaration file like this gives us a clear overview of how the different data is used, and eventually, it will make all specialized data wrapper type components obsolete.

This PR only focussed on a PoC for 1 metric. The NL hospital intake bar scale and metric in the sidebar.

The next step is to replace all sidebar metrics with this new SidebarMetric component after we can agree on the approach.


src/components-styled/sidebar-metric/sidebar-kpi-value.tsx
```diff
@@ -0,0 +1,71 @@
...
+      <Box display="flex" alignItems="center" justifyContent="flex-start">
+        {isPercentage ? (
+          <InlineText fontSize={3} fontWeight="bold" margin="0" marginRight={1}>
+            ({formatPercentage(value)}%)
+          </InlineText>
+        ) : (
+          <InlineText fontSize={3} fontWeight="bold" margin="0" marginRight={1}>
+            {formatNumber(value)}
+          </InlineText>
+        )}
```
VWSCoronaDashboard10, line 49: The component is the same for both cases, could be written as:
```tsx
<InlineText fontSize={3} fontWeight="bold" margin="0" marginRight={1}>
  {isPercentage ? `(${formatPercentage(value)}%)` : formatNumber(value)}
</InlineText>;
```

Btw, you wrapped the percentage variant between haakjes, is that on purpose?

src/components-styled/sidebar-metric/sidebar-metric.tsx
```diff
@@ -0,0 +1,110 @@
...
+  scope,
+  data,
+  metricName,
+  metricProperty,
+  localeTextKey,
+  differenceKey,
+  showBarScale,
+  annotationKey,
+}: SidebarMetricProps<T>) {
+  const lastValue = get(data, [metricName, 'last_value']);
```
VWSCoronaDashboard10, line 34: What's wrong with vanilla js?
```suggestion
  const lastValue = data[metricName].last_value;
```

src/components-styled/sidebar-metric/sidebar-metric.tsx
```diff
@@ -0,0 +1,110 @@
...
+      metricProperty,
+    ]
+      .filter(isDefined)
+      .join(':')}`
+  );
+
+  const config = getDataConfig(scope, metricName, metricProperty);
+  const commonText = siteText.common.metricKPI;
+
+  const title = get(siteText, [localeTextKey, 'titel_kpi']);
```
VWSCoronaDashboard10, line 51:
```suggestion
  const title = siteText[localeTextKey].titel_kpi;
```

src/components-styled/sidebar-metric/sidebar-metric.tsx
```diff
@@ -0,0 +1,110 @@
...
+  localeTextKey,
+  differenceKey,
+  showBarScale,
+  annotationKey,
+}: SidebarMetricProps<T>) {
+  const lastValue = get(data, [metricName, 'last_value']);
+  const propertyValue = lastValue && lastValue[metricProperty];
+
+  assert(
+    propertyValue,
```
VWSCoronaDashboard10, line 38: Is it possible for this propertyValue to be `0`? In that case this assertion will yield a false positive. The same applies for other assertions further down this file.

VWSCoronaDashboard10 (commented):
I like the approach, but I wonder if it will result in errors (now and in the future) because our language files aren't bound to a specific interface..

src/components-styled/sidebar-metric/sidebar-kpi-value.tsx
VWSCoronaDashboard8: Ah no this is a mistake. In the original component the percentage is only shown as an extra behind the absolute value. We are not using percentages in the sidebar yet, so I think this can be removed altogether actually. But it raises in interesting problem, because when rendering a percentage after the absolute value, the component will need two metricPropery values to know which one is which. But I think I'll leave that for when we get to this point.

src/components-styled/sidebar-metric/sidebar-metric.tsx
VWSCoronaDashboard8: Of you are not sure that `data[metricName]` exists it can generate a runtime exception.

src/components-styled/sidebar-metric/sidebar-metric.tsx
VWSCoronaDashboard8: So, no.... :) In typescript of course you can write it as `siteText[localeTextKey]?.titel_kpi` but I prefer to not use that unless we define siteText as a dictionary (with partial results), because adding ? to places where typescript says it's not required only feels confusing and error prone.

src/components-styled/sidebar-metric/sidebar-metric.tsx
VWSCoronaDashboard8: Good point 👍

VWSCoronaDashboard8:
@VWSCoronaDashboard10 Errors in the future is up to us. And I think it's a good thing that errors are generated when working on new features, because it will force us to use the same keys in lokalize. 

It is possible that this approach is not compatible with all metrics, because we didn't implement the lokalize keys consistently (as I mentioned in the description) but I think most of them will work and we can fix the others as part of the refactoring.

src/components-styled/sidebar-metric/sidebar-metric.tsx
VWSCoronaDashboard10: yes, which still happens due to the assertion on the next few lines ;-)

src/components-styled/sidebar-metric/sidebar-barscale.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { BarScale } from '~/components/barScale';
+import siteText from '~/locale/index';
+import { assert } from '~/utils/assert';
+import { Box } from '../base';
+import { BarScaleConfig } from './data-config';
+
+interface SidebarBarScaleProps {
+  config: BarScaleConfig;
+  value: number;
+  localeTextKey: string;
```
VWSCoronaDashboard1, line 11:
```suggestion
  localeTextKey: keyof TALLLanguages;
```
That will save you the `assert(text, 'No text found for locale key ${localeTextKey}')` call

VWSCoronaDashboard1 (approved):

src/components-styled/sidebar-metric/sidebar-metric.tsx
VWSCoronaDashboard8: The assertion won't happen if this line tries to index from undefined. And then you will not get a nice message like in the assertion

PR #1073, opened 2020-12-04
Topic/infected age groups keyboard

VWSCoronaDashboard4:
## Summary

Adds keyboard navigation and a11y support for the infected people age demographic.

## Detailed design

* Added a separate file to calculate the coordinates used by the chart. The chart file now basically contains the buildup of the SVG using those coordinates.
* The coordinates are calculated and memorized by the main `age-demographic` component.
* The coordinates are used in a new `useTooltip` hook. That hook returns the event handlers and the read only state.
* These event handlers are passed on to the chart.
* A new Tooltip component uses the state to render the correct contents on the provided coordinates.
* `aria-live` attributes are tweaked to pick up the changes in the tooltip.
* Added generic help text for barcharts to help discover the keyboard navigation.


src/components-styled/tooltip.tsx
```diff
@@ -0,0 +1,139 @@
...
+    const KEY_ARROW_LEFT = 37;
+    const KEY_ARROW_RIGHT = 39;
+
+    if (event.which !== KEY_ARROW_LEFT && event.which !== KEY_ARROW_RIGHT) {
+      return;
+    }
+
+    const direction = event.which === KEY_ARROW_LEFT ? -1 : 1;
+
+    // The new index overfloews from zero to the last and vice versa
```
VWSCoronaDashboard1, line 85:
```suggestion
    // The new index overflows from zero to the last and vice versa
```

src/components-styled/tooltip.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+  const timer = useRef(-1);
+
+  const tooltipState: TooltipState<T> = {
+    visible,
+    coordinates,
+    value,
+  };
+
+  // This timeout smoothens the display of the tooltip for mouse users
```
VWSCoronaDashboard1, line 52:
```suggestion
  // This timeout smooths the display of the tooltip for mouse users
```

src/domain/infected-people/age-demographic/age-demographic.tsx
```diff
@@ -1,90 +1,68 @@
...
-        value: NationalInfectedAgeGroupsValue
-      ) => { x: number; y: number }
-    ) => {
-      const { x, y } = getTooltipCoordinates(event, value);
-      const options: TooltipOptions = { left: x, top: y, value };
-      debounceSetTooltip(options);
-    },
-    [debounceSetTooltip]
-  );
+  // Calculate graph's coordinates based on the data, the component width and eher we are on a small screen or not.
```
VWSCoronaDashboard1, line 26:
```suggestion
  // Calculate graph's coordinates based on the data, the component width and whether we are on a small screen or not.
```

VWSCoronaDashboard1 (approved):
Excellent!

src/components-styled/tooltip.tsx
VWSCoronaDashboard4: Smoothens is a word ;-)

VWSCoronaDashboard10 (commented):
My first thought is that it looks quite complex, I guess it's hard to wrap my head around all the calculations in the coordinates getter (and everything related to those calculations), but I guess that if it works it's good for now.  Couple of questions though, and I think I spotted a bug :-)

src/components-styled/tooltip.tsx
VWSCoronaDashboard1: haha, ok, ignore then ;)

src/domain/infected-people/age-demographic/age-demographic-chart.tsx
```diff
@@ -210,15 +126,13 @@ export function AgeDemographicChart({
...
+      {values.map((d: NationalInfectedAgeGroupsValue, i: number) => {
         const ageGroupPercentageWidth = xMax - ageGroupPercentagePoint(d);
         const infectedPercentageWidth = infectedPercentagePoint(d);
         return (
           <Group
             key={i}
-            onMouseMove={(event) =>
-              openTooltip(event, d, getTooltipCoordinates)
-            }
+            onMouseMove={(event) => openTooltip(event, d)}
```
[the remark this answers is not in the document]
VWSCoronaDashboard8: I agree, if you go for a generic shorthand name use `x` for anything else use a readable name.

src/domain/infected-people/age-demographic/age-demographic-coordinates.ts
```diff
@@ -0,0 +1,170 @@
...
+import {
+  NationalInfectedAgeGroups,
+  NationalInfectedAgeGroupsValue,
+} from '~/types/data';
+import { AGE_GROUP_TOOLTIP_WIDTH } from './age-demographic-chart';
+
+export interface AgeDemographicCoordinates {
+  width: number;
+  height: number;
+  numTicks: number;
```
VWSCoronaDashboard8, line 18: Reading this I wonder what ticks have to do with coordinates, or what it means in general

src/domain/infected-people/age-demographic/age-demographic-coordinates.ts
```diff
@@ -0,0 +1,170 @@
...
+  xMax: number;
+  yMax: number;
+  ageGroupPercentageScale: ValueOf<ScaleTypeToD3Scale<any, any, any>>;
+  infectedPercentageScale: ValueOf<ScaleTypeToD3Scale<any, any, any>>;
```
VWSCoronaDashboard8, line 22: So AgeDemographicCoordinates is tied to the infected people metric? Are you sure this kind of coupling is necessary?

src/domain/infected-people/age-demographic/age-demographic-coordinates.ts
```diff
@@ -0,0 +1,170 @@
...
+  ageGroupRangeScale: ScaleBand<string>;
+  ageGroupPercentagePoint: (d: NationalInfectedAgeGroupsValue) => any;
+  infectedPercentagePoint: (d: NationalInfectedAgeGroupsValue) => any;
+  ageGroupRangePoint: (d: NationalInfectedAgeGroupsValue) => any;
+  getTooltipCoordinates: GetTooltipCoordinates<NationalInfectedAgeGroupsValue>;
```
VWSCoronaDashboard8, line 27: Same here, I wonder why the tooltip coordinates have to care about the type of metric...

src/domain/infected-people/age-demographic/age-demographic-coordinates.ts
```diff
@@ -0,0 +1,170 @@
...
+  isSmallScreen: boolean;
+  margin: {
+    top: number;
+    right: number;
+    bottom: number;
+    left: number;
+  };
```
VWSCoronaDashboard8, line 34: It is surprising to see margin settings on a component interface. Maybe they can be handled by a parent/wrapping element in the composition?

But I have to admit at this point I have no idea what the component is doing, so you can ignore my comment if it seems essential to the inner workings

src/domain/infected-people/age-demographic/age-demographic-coordinates.ts
```diff
@@ -0,0 +1,170 @@
...
+
+export function getAgeDemographicCoordinates(
+  data: NationalInfectedAgeGroups,
+  isSmallScreen: boolean,
+  parentWidth: number
+): AgeDemographicCoordinates {
+  const values = data.values.sort((a, b) => {
+    const aStart = parseInt(a.age_group_range, 10);
+    const bStart = parseInt(b.age_group_range, 10);
+    return bStart - aStart;
```
VWSCoronaDashboard8, line 48: Sorting works directly on strings too 
```suggestion
    return bStart.localeCompare(aStart);
```

src/domain/infected-people/age-demographic/age-demographic-coordinates.ts
```diff
@@ -0,0 +1,170 @@
...
+  });
+  const ageGroupRangeScale = scaleBand({
+    range: [margin.top, height - margin.top],
+    round: true,
+    domain: values.map(ageGroupRange),
+    padding: 0.4,
+  });
+
+  // Compose together the scale and accessor functions to get point functions
+  const createPoint = (scale: any, accessor: any) => (
```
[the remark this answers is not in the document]
VWSCoronaDashboard8: Or, write a comment about why you need this escape hatch

src/domain/infected-people/age-demographic/age-demographic-coordinates.ts
```diff
@@ -0,0 +1,170 @@
...
+    ageGroupPercentagePoint,
+    infectedPercentagePoint,
+    ageGroupRangePoint,
+    getTooltipCoordinates,
+    isSmallScreen,
+    margin,
+    values,
+    ageGroupRange,
+    ageRangeAxisWidth,
+  };
```
VWSCoronaDashboard8, line 169: I have a bad feeling about this long and specific interface, but I think this is something to revisit later. We don't have proper abstractions or visx yet, and it will be hard to define them correctly in the beginning.

VWSCoronaDashboard8 (changes requested):

src/domain/infected-people/age-demographic/age-demographic-coordinates.ts
VWSCoronaDashboard4: This is not a CSS margin, its a margin inside the drawn graphic.

src/domain/infected-people/age-demographic/age-demographic-coordinates.ts
VWSCoronaDashboard4: It was a choice between providing the proper type for the value (and using a generic like this), or providing an `any` on all of the places this is processed.

src/domain/infected-people/age-demographic/age-demographic-tooltip-content.tsx
```diff
@@ -66,13 +47,14 @@ const LegendItem = styled.li(
...
-export function AgeDemographicTooltip({
+export function AgeDemographicTooltipContent({
   value,
-  left,
-  top,
-}: AgeDemographicTooltipProps) {
+}: AgeDemographicTooltipContentProps) {
+  if (!value) {
+    return null;
+  }
```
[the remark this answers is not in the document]
VWSCoronaDashboard4: No, I'll add a comment about this. The aria-live attribute would spell out every new display (additions), instead of only the last one (changes).

src/domain/infected-people/age-demographic/age-demographic-coordinates.ts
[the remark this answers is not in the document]
VWSCoronaDashboard4: I'll add a comment.

VWSCoronaDashboard8 (approved):

src/domain/infected-people/age-demographic/age-demographic.tsx
```diff
@@ -1,90 +1,68 @@
...
-          {(parent) => (
-            <AgeDemographicChart
-              parentWidth={parent.width}
-              data={data}
-              openTooltip={openTooltip}
-              closeTooltip={closeTooltip}
-            />
-          )}
+          {(parent) => {
+            setParentWidth(parent.width);
```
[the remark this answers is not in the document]
VWSCoronaDashboard4: I've added a comment in the code to clarify this a bit; I don't get an error and it's because of how the `ParentWidth` component handles this callback.

For now I suggest to leave this, as it does not give any troubles/errors. In theory it can be torn apart, resulting in more complex code / nesting of components.

PR #1083, opened 2020-12-07
Deceased pages

VWSCoronaDashboard10:
## Summary

Create deceased pages. The decease monitor section has been moved to a shared component because it looks exactly the same for national and regional.

One thing left to implement is to create the sidebar links.


src/domain/deceased/components/deceased-monitor-chart.tsx
```diff
@@ -0,0 +1,209 @@
...
+            hover: {
+              enabled: true,
+            },
+          },
+        },
+      },
+      {
+        type: 'line',
+        data: values.map((x) => [
+          new Date(x.week_start_unix * 1000),
```
VWSCoronaDashboard1, line 190: nit: we have a util for this: src/utils/createDate

VWSCoronaDashboard1 (approved):

src/domain/deceased/deceased-monitor-section.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { AnchorTile } from '~/components-styled/anchor-tile';
+import { Box } from '~/components-styled/base';
+import { ChartTile } from '~/components-styled/chart-tile';
+import { ContentHeader } from '~/components-styled/content-header';
+import { Legenda } from '~/components-styled/legenda';
+import DeceasedMonitor from '~/domain/deceased/components/deceased-monitor-chart';
+import siteText from '~/locale/index';
+import { NationalDeceasedCbs, RegionalDeceasedCbs } from '~/types/data';
+
+const sterftemonitorText = siteText.section_sterftemonitor;
```
VWSCoronaDashboard8, line 12: nit; if only one text source is used in this context maybe just call it text.  It would clean up the code below. This is a common pattern in other places too.

src/static-props/safetyregion-data.ts
```diff
@@ -49,33 +49,31 @@ interface IParams {
  * }
  * ```
  */
-export function getSafetyRegionData() {
-  return function ({ params }: IParams): IProps {
-    const { code } = params;
+export function getSafetyRegionData({ params }: IParams): IProps {
```
VWSCoronaDashboard8, line 52: I think eventually we should call this getSafetyRegionStaticProps to bring it inline with the recent NL refactoring on this, but that's something for a different PR I image.

VWSCoronaDashboard8 (approved):

src/domain/deceased/components/deceased-monitor-chart.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+  return {
+    chart: {
+      alignTicks: true,
+      animation: false,
+      backgroundColor: 'transparent',
+      borderColor: '#000',
+      borderRadius: 0,
+      borderWidth: 0,
+      className: 'undefined',
```
VWSCoronaDashboard4, line 56: this is adding the classname "undefined"; is that really what you want? if you don't want a classname, I guess this key/value can be omitted.

src/pages/landelijk/sterfte.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+      <ContentHeader
+        category={siteText.nationaal_layout.headings.besmettingen}
+        title={text.section_deceased_rivm.title}
+        icon={<CoronaVirusIcon />}
+        subtitle={text.section_deceased_rivm.description}
+        metadata={{
+          datumsText: text.section_deceased_rivm.datums,
+          dateInfo: dataRivm.last_value.date_of_report_unix,
+          dateOfInsertionUnix: dataCbs.last_value.date_of_insertion_unix,
```
VWSCoronaDashboard4, line 39: shouldn't this be both the `dataRivm` ?

src/pages/veiligheidsregio/[code]/sterfte.tsx
```diff
@@ -0,0 +1,97 @@
...
+        category={siteText.veiligheidsregio_layout.headings.besmettingen}
+        title={replaceVariablesInText(text.section_deceased_rivm.title, {
+          safetyRegion,
+        })}
+        icon={<CoronaVirusIcon />}
+        subtitle={text.section_deceased_rivm.description}
+        metadata={{
+          datumsText: text.section_deceased_rivm.datums,
+          dateInfo: dataRivm.last_value.date_of_report_unix,
+          dateOfInsertionUnix: dataCbs.last_value.date_of_insertion_unix,
```
Comment (VWSCoronaDashboard4) [inline src/pages/veiligheidsregio/[code]/sterfte.tsx:47]: same question here

src/components-styled/content-header/index.tsx
```diff
@@ -142,11 +142,13 @@ export function ContentHeader(props: ContentHeaderProps) {
         <ReferenceBox>
           <Text>
             {subtitle}{' '}
-            <Link href={reference.href}>
-              <Text as="a" href={reference.href}>
-                {reference.text}
-              </Text>
-            </Link>
+            {reference && (
```
VWSCoronaDashboard4, line 145: Not sure why this is made optional? In case this link was not present in the design of the new page, it still should be there (and this part of the props should still be required)

VWSCoronaDashboard4 (commented):
Looking good overall; some questions.

src/domain/deceased/components/deceased-monitor-chart.tsx
VWSCoronaDashboard10: No, it's copy paste and I didn't think about it. I'll check what it actually does.

src/pages/landelijk/sterfte.tsx
VWSCoronaDashboard10: Yep, absolutely!

src/components-styled/content-header/index.tsx
VWSCoronaDashboard10: Ah ok, so the design is not correct you say?

src/domain/deceased/components/deceased-monitor-chart.tsx
VWSCoronaDashboard10: The property can safely be removed. I've also removed it from the source where it came from :-)

PR #1092, opened 2020-12-08
Fix some errors related to hospital_admissions being optional

VWSCoronaDashboard1:
## Summary

As the title suggests.

## Motivation

Bug report

## Detailed design

n/a

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added


src/components/choropleth/shared.d.ts
```diff
@@ -16,7 +16,7 @@ export type TMunicipalityMetricType = ValueOf<
   Partial<MunicipalityProperties>;
 
 export type TRegionMetricName = TMetricHolder<
-  Omit<Regions, 'deceased' | 'behavior'>
+  Omit<Regions, 'deceased' | 'behavior' | 'hospital_admissions'>
```
VWSCoronaDashboard8, line 19: I wonder why these types are used. In a different PR (the one that stranded about unifying hospital data) I removed these types with deceased omit etc all together).  What is the point of these specific omits?

VWSCoronaDashboard10 (approved):

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard4 (approved):

src/components/choropleth/shared.d.ts
VWSCoronaDashboard1: Since they're optional, the TRegionMetricType type will fail since it tries to pluck the data type from an array type. Which only works if this array is guaranteed to exist. There might be some conditional typing magic possible here, but for now, this works...

PR #1093, opened 2020-12-08
Reimplement line chart in visx

VWSCoronaDashboard13:
## Summary

This PR implements a basic LineChart replacement using d3/visx instead of highcharts

## Motivation

The team wants to remove highcharts from the project

## Detailed design

I've added a folder `custom-line-chart` that houses the basic line chart replacement as well as a more general `chart` component. 

#### Custom line chart (chose an arbitrary name. open to suggestions!)
This component accepts the same props as the existing line chart and handles data filtering and manipulating (I'm replacing the date with a date object), determining the domains of the graph, and tooltip rendering. This is essentially a composition for the the specific graph that is to be rendered and allows the user to customize the contents of the tooltip. This also keeps tooltip rendering separate from the graph so the SVG doesn't re-render unnecessarily

#### Chart 
This component is supposed to act as a generic time based chart that currently accepts one trend, but is built with the intention to plot multiple trends via a `series` prop in the future. The base component uses the `yDomain` and `xDomain` to define the scales and a bisect function (determines which dataPoint is being hovered when given a trend and mouse position) and renders the axes, one optional benchmark value, and the trends that are abstracted into a separate component.

Trends Component
This component currently accepts one trend and its type. Currently there are only two types defined `line` `area` which are chosen depending on the existing `isFilled` prop that is passed to the line charts, but the hope is to add more types in the future. This renders the plotted trend as well as a `bar` to catch hover events which it passes to the `handleHover` prop that is passed from and defined in the chart composition. 

Decided to copy over and edit the functions/types from the existing line chart to the `CustomLineChart` so that they are entirely separate entities and have one variation (the tooltip formatting function)

## Drawbacks

- Hasn't gone through rigorous testing and I haven't checked every place that this component will be used 

## Unresolved questions

- How does this compare to the baseline of Highcharts accessibility? Is the implementation better, equal, or worse?

## Open bugs

- [ ] Date formatting in the tooltips seems off in various line charts http://localhost:3000/landelijk/thuiswonende-ouderen
- [ ] When a value is null, we should hide the line. (You can see this in the R-chart) http://localhost:3000/landelijk/reproductiegetal
- [ ] Not all dates are bold


VWSCoronaDashboard:
> The new chart accepts a formatting function for the tooltip, however the existing functions utilize features of highcharts to, for example, make some text bold. Now that we have more flexibility in regards to the content of our tooltips, how do we want to handle formatting them?

I would assume it can handle composition now. We're not using Highcharts anymore.

```
          <Tooltip>
            <div>
              <strong>x:</strong> 10}
            </div>
            <div>
              <strong>y:</strong> 20}
            </div>
          </Tooltip>
```

VWSCoronaDashboard:
> I'm not sure how this performs in terms of accessibility

That's not a drawback. I'll move it to unresolved questions. As long as it doesn't perform worse than Highcharts it's not a blocker.

VWSCoronaDashboard:
I've pushed a few commits that should solve some of the TS errors.
There's one big one left, and I know where to start solving it but I don't have the time and focus to solve it. 

The issue:

In `src/components/custom-line-chart/index.tsx` we have this type. I've naively made it myself when building the PoC instead of re-using one that already existed.

```tsx
export type ThresholdProps = {
  values: any[];
  width: number;
  height?: number;
  timeframe?: TimeframeOption;
  signaalwaarde?: number;
  formatTooltip: any;
  formatYAxis: any;
  showFill: boolean;
  valueAnnotation?: string;
};
```

I think you should be using the interface (including generics for the values) from `src/components/lineChart/index.tsx`

```tsx
export interface LineChartProps<T> {
  values: T[];
  signaalwaarde?: number;
  timeframe?: TimeframeOption;
  formatTooltip?: (value: T) => string;
  formatYAxis?: (y: number) => string;
  showFill?: boolean;
  valueAnnotation?: string;
}
```

I've solved the tooltip TypeScript errors by importing the Tooltip hook types from `@visx/tooltip/lib/hooks/useTooltip` and extending it with our `Value` type.

src/components/custom-line-chart/chart/trends.tsx
```diff
@@ -0,0 +1,96 @@
...
+  type = trendTypes.line,
+  color,
+  x,
+  y,
+  handleHover,
+  size,
+  isHovered,
+  bisect,
+}: TrendsProps) {
+  const onPointerMove = useCallback(
```
VWSCoronaDashboard8, line 36: Similar to my comment earlier about event handlers this should probably be called `handlePointerMove` because it's an event handling function.

src/components/custom-line-chart/chart/trends.tsx
```diff
@@ -0,0 +1,96 @@
...
+  area: 'area',
+};
+
+export type TrendsProps = {
+  isHovered: boolean;
+  trend: any;
+  type: string;
+  x: any;
+  y: any;
+  handleHover: any;
```
VWSCoronaDashboard8, line 16:
```suggestion
  onHover: any;
```

Apart from any not being specific enough, the name can be improved like I mentioned earlier

src/components/custom-line-chart/chart/trends.tsx
```diff
@@ -0,0 +1,96 @@
...
+      <Bar
+        x={0}
+        y={0}
+        width={size.width}
+        height={size.height}
+        fill="transparent"
+        rx={14}
+        onTouchStart={onPointerMove}
+        onTouchMove={onPointerMove}
+        onMouseMove={onPointerMove}
```
VWSCoronaDashboard8, line 91: See this is already not very nice to read and something like this is a little clearer I think:

```suggestion
        onMouseMove={handlePointerMove}
```

VWSCoronaDashboard8 (changes requested):

src/components/custom-line-chart/chart/tooltip.tsx
```diff
@@ -0,0 +1,112 @@
...
+  return (
+    <>
+      <Point style={{ top: y, left: x }} indicatorColor={primaryColor} />
+
+      <TooltipContainer
+        style={{
+          top: y,
+          left: x,
+          transform: `translate(${xTransform},${yTransform})`,
+          transition: 'left 0.075s, top 0.075s',
```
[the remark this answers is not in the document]
VWSCoronaDashboard13: yeah, I thought about moving it, but decided it made more sense to keep it with the dynamic styles it's associated with

src/components/custom-line-chart/chart/trends.tsx
```diff
@@ -0,0 +1,96 @@
+import { useCallback } from 'react';
+import { AreaClosed, LinePath, Bar } from '@visx/shape';
+import { localPoint } from '@visx/event';
+
+export const trendTypes = {
```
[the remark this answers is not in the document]
VWSCoronaDashboard13: ahh this is perfect! good to know about the enum thing

src/components/custom-line-chart/chart/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+      if (trend.length === 1) return trend[0];
+
+      const bisect = bisector((d: DataPoint) => d.date).left;
+      const date: Date = x.invert(mx - margin.left);
+      const index: number = bisect(trend, date, 1);
+
+      const d0: DataPoint = trend[index - 1];
+      const d1: DataPoint = trend[index];
+
+      return +date - +d0.date > +d1.date - +date ? d1 : d0;
```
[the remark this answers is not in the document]
VWSCoronaDashboard13: Just learned this from Jan: The sort function requires numbers so the + sign converts the dates to numbers

src/components-styled/line-chart-tile.tsx
```diff
@@ -34,7 +39,15 @@ export function LineChartTile<T extends Value>({
     >
       {(timeframe) => (
         <>
-          <LineChart {...chartProps} timeframe={timeframe} />
+          <ParentSize>
+            {(parent) => (
+              <CustomLineChart
```
VWSCoronaDashboard4, line 44: Is it possible to rename these to eg `LineChart` and `LineChartProps` before merging?

src/components/custom-line-chart/chart/index.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard13: The sort function requires numbers so the + sign converts the dates to numbers

src/components/custom-line-chart/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+      </Box>
+    </Box>
+  );
+}
+
+/**
+ * From all the defined values, extract the highest number so we know how to
+ * scale the y-axis
+ */
+function calculateYMax(values: Value[], signaalwaarde = -Infinity) {
```
VWSCoronaDashboard4, line 153: This method is also present in `src/components/lineChart/index.tsx` ; maybe just 1 of them is needed?

src/components/custom-line-chart/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+    tooltipData,
+    tooltipLeft = 0,
+    tooltipTop = 0,
+    showTooltip,
+    hideTooltip,
+  }: UseTooltipParams<T> = useTooltip();
+
+  const benchmark = useMemo(() => {
+    return signaalwaarde
+      ? { value: signaalwaarde, label: 'Signaalwaarde' }
```
VWSCoronaDashboard4, line 61: Label value is hardcoded?

VWSCoronaDashboard4 (commented):
Looking good, some nitpicks

src/components-styled/line-chart-tile.tsx
VWSCoronaDashboard13: done 👍

src/components/custom-line-chart/index.tsx
VWSCoronaDashboard13: should I be using this? `text.common.barScale.signaalwaarde` from `import text from '~/locale/index';`

src/components/custom-line-chart/index.tsx
VWSCoronaDashboard13: I've left it in this file and removed it from the other so that we don't have to move them if we decide to remove that file

VWSCoronaDashboard13: Ahh actually they are using different value types, so I'm going to leave it for now

src/components-styled/line-chart-tile.tsx
```diff
@@ -1,9 +1,14 @@
-import LineChart, { LineChartProps, Value } from '~/components/lineChart';
+import { ParentSize } from '@visx/responsive';
+
 import { TimeframeOption } from '~/utils/timeframe';
 import { ChartTileWithTimeframe } from './chart-tile';
 import { MetadataProps } from './metadata';
+import CustomLineChart, {
+  CustomLineChartProps,
+  Value,
+} from '~/components/custom-line-chart/index';
```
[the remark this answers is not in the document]
VWSCoronaDashboard13: yes! I organized them 👍

VWSCoronaDashboard8 (changes requested):
This is just a reminder that I've reviewed it thus far, but we can improve on the unknown/any types still

src/components/custom-line-chart/chart/trends.tsx
```diff
@@ -0,0 +1,96 @@
...
+    },
+    [handleHover, y, x, trend, bisect]
+  );
+
+  return (
+    <>
+      {type === trendTypes.line && (
+        <LinePath
+          data={trend}
+          x={(d: any) => x(d.date)}
```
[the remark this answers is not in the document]
VWSCoronaDashboard: Made d a Datapoint type.

src/components/custom-line-chart/index.tsx
VWSCoronaDashboard: Solved this.

src/components/custom-line-chart/index.tsx
VWSCoronaDashboard: As soon as everything is moved to the new line chart, we'll remove the old folder and see which types need to remain or not ;).

VWSCoronaDashboard8 (approved):

PR #1103, opened 2020-12-08
More cypress tests and some test refacors

VWSCoronaDashboard1:
## Summary

This adds some more page tests. This way the entire site is 'clicked through' by the tests, so should any page crash this will be on our radar a little quicker.
Furthermore, I refactored all of the helper functions as Cypress commands, since this is more 'the cypress way'.
Lastly, this also holds a regression fix where on the municipal positive tested page the selectHandler was erroneously removed from the choropleth.

## Motivation

More tests is good.

## Detailed design

Pretty much what the summary already tells. I've also added a test to each page that checks whether only _one_ H1  element is present on the page. If there are multiple it'll throw an error with a detailed description.
I've also upgraded to the latest version of Cypress and Typescript.

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added


cypress/support/commands.ts
```diff
@@ -0,0 +1,168 @@
...
+  (kpiTestInfo: Record<string, string | string[]>) => {
+    Object.entries(kpiTestInfo).forEach(([key, value]) => {
+      const element = cy.dataCy(key);
+      if (Array.isArray(value)) {
+        value.forEach((val) => {
+          element.contains(val);
+        });
+      } else {
+        element.contains(value);
+      }
```
VWSCoronaDashboard10, line 159: nit, but I think this can be written a bit simpler: 
```suggestion
      if (Array.isArray(value)) {
        [...value].forEach((val) => element.contains(val));
      }
```

VWSCoronaDashboard10 (approved):

cypress/support/commands.ts
VWSCoronaDashboard1: I'm not sure I follow your suggestion. you copy the array and then iterate? For now, I only removed the parentheses on the lambda, if that's the gist of what you mean.

VWSCoronaDashboard8 (approved):

cypress/support/commands.ts
VWSCoronaDashboard8: I think the point was to first convert whatever you have to an array (also if it's a single value) and then you can perform the same operation without needing the else statement to do something else.

I wonder if this works though. If value is a number for example wouldn't [...value] then throw, or is JS smart enough to convert it to a single value array?

VWSCoronaDashboard10: Ah yes that doesn't work .. my bad, and also the suggestion was buggy as hell. Guess I shouldn't do codereviews around 20:30.

What I actually had in mind was the following:

```tsx
[].concat(value).forEach(x => element.contains(val))
```

This will create an array and concat the `value` whether it's a string or an array of string:
```tsx
[].concat('plain string', ['string inside an array']) === ["plain string", "string inside an array"]

PR #1111, opened 2020-12-09
Age demographic QA fixes

VWSCoronaDashboard4:
## Summary

Age demographic QA fixes

## Detailed design

Fixes:
* tooltip started at index 1, not at index 0.
* alignment/size of headers was off
* `aria-live` works again when going through the content with a screenreader on (the fix was to calculate the size after a `requestAnimationFrame` since `ref` was not defined on init)

### DIFF
--- a/src/utils/use-element-size.ts
+++ b/src/utils/use-element-size.ts
@@ -14,7 +14,7 @@ export function useElementSize<T extends HTMLElement>(
   useEffect(() => {
     const debounceHandleResize = debounce(handleResize, 60);
 
-    handleResize();
+    requestAnimationFrame(handleResize);
     window.addEventListener('resize', debounceHandleResize);
 
     return () => window.removeEventListener('resize', debounceHandleResize);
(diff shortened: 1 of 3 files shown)


src/utils/use-element-size.ts
```diff
@@ -14,7 +14,7 @@ export function useElementSize<T extends HTMLElement>(
   useEffect(() => {
     const debounceHandleResize = debounce(handleResize, 60);
 
-    handleResize();
+    requestAnimationFrame(handleResize);
     window.addEventListener('resize', debounceHandleResize);
 
     return () => window.removeEventListener('resize', debounceHandleResize);
```
VWSCoronaDashboard1, line 20: Perhaps this logic needs to be in a useLayoutEffect hook?
https://kentcdodds.com/blog/useeffect-vs-uselayouteffect

That might remove the need for the requestANimationFrame hack

VWSCoronaDashboard1 (approved):

src/utils/use-element-size.ts
VWSCoronaDashboard4: Unfortunately, this does not give the desired result.

PR #1115, opened 2020-12-09
Fix escalation map tooltip text

VWSCoronaDashboard1:
## Summary

Fix correct rendering of valid from date in the escalation choropleth tooltip. (It went missing)

## Motivation

Bug report

## Detailed design

The reason why this text disappeared was an invalid locale value was given to the replaceVariablesInText method which silently swallowed this and just returned an empty string. I've changed the implementation now and added an assert for this value.
In the future, the build or e2e tests will break when this happens again.

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added

### DIFF
--- a/src/utils/replaceVariablesInText.ts
+++ b/src/utils/replaceVariablesInText.ts
@@ -1,3 +1,5 @@
+import { assert } from './assert';
+
 const curlyBracketRegex = /\{\{(.+?)\}\}/g;
 
 /**
@@ -17,10 +19,13 @@ const curlyBracketRegex = /\{\{(.+?)\}\}/g;
  */
 
 export function replaceVariablesInText(
-  translation?: string | undefined | null,
+  translation?: string,
   variables?: { [key: string]: string | number | undefined }
 ): string {
-  if (!translation) return '';
+  assert(
+    translation,
+    'translation placeholder text is not defined, perhaps a missing locale key?'
+  );
 
   return translation.replace(curlyBracketRegex, (_string, variableName) => {
     /**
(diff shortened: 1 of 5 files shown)


VWSCoronaDashboard4 (approved):

src/utils/replaceVariablesInText.ts
```diff
@@ -17,10 +19,13 @@ const curlyBracketRegex = /\{\{(.+?)\}\}/g;
  */
 
 export function replaceVariablesInText(
-  translation?: string | undefined | null,
+  translation?: string,
   variables?: { [key: string]: string | number | undefined }
 ): string {
-  if (!translation) return '';
+  assert(
+    translation,
```
VWSCoronaDashboard8, line 26: I think it's a little odd to define translation as an optional argument and then enforce it via an assert. Ideally, it would be non-optional and we'd get an error in the calling context if you try to use a key from the locale text that doesn't exist. But maybe there are "any" types involved in the code, so I guess it's fine for now.

VWSCoronaDashboard8 (approved):

src/utils/replaceVariablesInText.ts
VWSCoronaDashboard1: I'm refactoring some more. I've made the arguments required. Since locale keys are passed in there's still a small chance that @runtime an undefined is passed in, so I'm keeping the assert for now.

VWSCoronaDashboard1: Actually, an empty string will be OK for the compiler, but not for this function, so I think the assert is valid

PR #1120, opened 2020-12-09
QA feedback

VWSCoronaDashboard10:
## Summary

several QA thingies:

- choropleth should render grey lines when there's no fill or it is white
- remove rounder corners from percentage bars in behavior table
- change the alt color of the behavior tile
- remove incorrect MD from some localize keys
- don't scroll to top when route contains a hash

### DIFF
--- a/src/pages/_app.tsx
+++ b/src/pages/_app.tsx
@@ -24,9 +24,12 @@ export default function App(props: AppPropsWithLayout) {
 
   useEffect(() => {
     window.document.documentElement.classList.add('js');
-    const handleRouteChange = () => {
+    const handleRouteChange = (pathname: string) => {
       piwik.pageview();
-      scrollToTop();
+
+      if (!pathname.includes('#')) {
+        scrollToTop();
+      }
     };
 
     Router.events.on('routeChangeComplete', handleRouteChange);
--- a/src/components-styled/anchor-tile.tsx
+++ b/src/components-styled/anchor-tile.tsx
@@ -24,7 +24,7 @@ export function AnchorTile({
   shadow = false,
 }: AnchorTileProps) {
   return (
-    <Container shadow={shadow}>
+    <Container spacing={2} shadow={shadow}>
       <Content>
         <Heading level={3}>{title}</Heading>
         {children}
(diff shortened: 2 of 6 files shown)


VWSCoronaDashboard4 (approved):

src/pages/_app.tsx
```diff
@@ -24,9 +24,12 @@ export default function App(props: AppPropsWithLayout) {
 
   useEffect(() => {
     window.document.documentElement.classList.add('js');
-    const handleRouteChange = () => {
+    const handleRouteChange = (pathname: string) => {
       piwik.pageview();
-      scrollToTop();
+
+      if (!pathname.includes('#')) {
```
VWSCoronaDashboard1, line 30: this will fail on IE11 I suspect:
https://caniuse.com/?search=includes

VWSCoronaDashboard1 (changes requested):

src/pages/_app.tsx
VWSCoronaDashboard8: I think the typescript compiler with its target setting will take care of this

VWSCoronaDashboard8 (approved):

src/pages/_app.tsx
VWSCoronaDashboard1: I think we've had problems with this in past, but I could be mistaken

VWSCoronaDashboard1: AFAIK, the TS compiler doesn't transpile these things, it just assumes polyfills are present

VWSCoronaDashboard1: https://stackoverflow.com/questions/42564130/what-will-typescript-transpile-when-targeting-es5-es3

VWSCoronaDashboard8: Yeah you seem to be right about that. I never realized this because I always targeted evergreen browsers with TS. Shall we include a polyfill then? I think I'd prefer to stick to modern syntax and not think about it anymore from here.

VWSCoronaDashboard10: it's polyfilled by next I think 
image

VWSCoronaDashboard10: Here's a list with all polyfills: https://github.com/vercel/next.js/blob/canary/packages/next-polyfill-nomodule/src/index.js

PR #1129, opened 2020-12-09
Filter the behavior choropleth dropdown list

VWSCoronaDashboard1:
## Summary

Filter the dropdown for the behavior choropleth to not show certain rules

## Motivation

Demo feedback

## Detailed design

Some rules aren't being shown on VR level, so these are now removed from the dropdown.

## Drawbacks

For now, the values are hardcoded. Ideally, the dropdown is populated based on the given data, but that would require a more elaborate refactor for which currently I lack the time.

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added

### DIFF
--- a/src/domain/behavior/behavior-choropleth-tile.tsx
+++ b/src/domain/behavior/behavior-choropleth-tile.tsx
@@ -18,6 +18,12 @@ import { BehaviorTypeControl } from './components/behavior-type-control';
 
 const text = siteText.nl_gedrag;
 
+const unusedRules = [
+  'symptoms_stay_home',
+  'symptoms_get_tested',
+  'wear_mask_public_transport',
+];
+
 export function BehaviorChoroplethTile() {
   const [type, setType] = useState<BehaviorType>('compliance');
   const [currentId, setCurrentId] = useState<BehaviorIdentifier>('wash_hands');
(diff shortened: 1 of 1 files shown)


src/domain/behavior/behavior-choropleth-tile.tsx
```diff
@@ -18,6 +18,12 @@ import { BehaviorTypeControl } from './components/behavior-type-control';
 
 const text = siteText.nl_gedrag;
 
+const unusedRules = [
+  'symptoms_stay_home',
+  'symptoms_get_tested',
+  'wear_mask_public_transport',
+];
```
VWSCoronaDashboard10, line 25: Should't the data be filtered on the BE..?

VWSCoronaDashboard4 (approved):

src/domain/behavior/behavior-choropleth-tile.tsx
VWSCoronaDashboard1: The data is filtered on the backend. The dropdown itself is a hard coded list of values though. Ideally this would have been populated based on the data as well. But I get why I wasn't done like that because of the way the regions file is loaded.
That's something which will need to be looked at in the future, for now, this is the quickest fix.

PR #1144, opened 2020-12-10
Fix critital issue with behavior line chart

VWSCoronaDashboard10:
This PR will fix an issue with the behavior line chart. The entire app would free when changing the highlighted behavior, this only happened with a production build.

I was unable to find the root cause of this error but a hard remount of the chart will prevent the issue from happening.


VWSCoronaDashboard4 (approved):

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard10:
on hold - looks like its not only a bug on the behavior line chart

PR #1156, opened 2020-12-10
Fix crash due to multiple highchart instances

VWSCoronaDashboard10:
fix a nasty crash which is probably caused by loading multiple highchart instances.

### DIFF
--- a/src/components/common/highcharts.tsx
+++ b/src/components/common/highcharts.tsx
@@ -5,6 +5,10 @@ if (typeof HighchartsLibrary === 'object') {
   require('highcharts/highcharts-more')(HighchartsLibrary);
 }
 
-export function Highcharts({ options }: { options: Highcharts.Options }) {
+interface HighchartsWrapperProps {
+  options: Highcharts.Options;
+}
+
+export function HighchartsWrapper({ options }: HighchartsWrapperProps) {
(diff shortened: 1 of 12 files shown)


src/components/common/highcharts.tsx
```diff
@@ -0,0 +1,10 @@
+import HighchartsLibrary from 'highcharts';
+import HighchartsReact from 'highcharts-react-official';
+
+if (typeof HighchartsLibrary === 'object') {
+  require('highcharts/highcharts-more')(HighchartsLibrary);
+}
```
VWSCoronaDashboard1, line 6: Is this check enough? Shouldn't there be some sort of check that makes sure there is no highcharts instance present on the window object or something?

VWSCoronaDashboard4 (approved):

VWSCoronaDashboard1 (approved):

src/components/common/highcharts.tsx
VWSCoronaDashboard10: I have no idea actually, all I know it's not broken anymore 🤷  aand that it's time to rename highcharts to byecharts.

VWSCoronaDashboard1: oh wait, this is a single import, so this is executed only once. Forget my question.

PR #1167, opened 2020-12-11
Topic/visx barchart

VWSCoronaDashboard4:
## Summary

Work in progress: Visx version of the BarChart.

## Detailed design

* Most of this is copy-paste of the age demographic components, except with just one metric.
* Uses a generic interface (`BarChartValue`) to pass the values.
* Only pages that use the old BarChart are the Sewer water pages. The data formatting in the helper functions there is simplified a bit.

## Todo

Styling and alignment of the tooltips needs some love.


src/components-styled/bar-chart/bar-chart-chart.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  return (
+    <StyledSvg
+      width={width}
+      height={height}
+      role="img"
+      id="age-demographic-chart"
+      aria-label={chartLabel}
+      tabIndex={0}
+      onKeyUp={(event: any) => keyboardTooltip(event)}
```
VWSCoronaDashboard1, line 67:
```suggestion
      onKeyUp={(event: KeyboardEvent<HTMLElement>) => keyboardTooltip(event)}
```
nit pick ;)

src/components-styled/bar-chart/bar-chart-chart.tsx
```diff
@@ -0,0 +1,132 @@
...
+import { MouseEvent } from 'react';
+import styled from 'styled-components';
+import { colors } from '~/style/theme';
+import { BarChartCoordinates } from './bar-chart-coordinates';
+
+interface BarChartChartProps {
+  coordinates: BarChartCoordinates;
+  openTooltip: (event: MouseEvent<any>, value: any) => void;
+  closeTooltip: () => void;
+  keyboardTooltip: (event: any) => void;
```
VWSCoronaDashboard1, line 16:
```suggestion
  keyboardTooltip: (event: KeyboardEvent<HTMLElement>) => void;
```
nit ;)

src/components-styled/bar-chart/bar-chart-tooltip-contents.tsx
```diff
@@ -0,0 +1,14 @@
+import css from '@styled-system/css';
+import { Box } from '../base';
+
+interface BarChartTooltipContentProps {
+  value: any;
```
VWSCoronaDashboard1, line 5: should this be typed to something generic perhaps?

VWSCoronaDashboard1 (approved):

src/components-styled/bar-chart/bar-chart-chart.tsx
```diff
@@ -0,0 +1,132 @@
...
+      tabIndex={0}
+      onKeyUp={(event: any) => keyboardTooltip(event)}
+    >
+      {/* Vertical lines */}
+      <GridColumns
+        scale={xScale}
+        width={xMax}
+        height={yMax}
+        left={margin.left}
+        top={margin.top}
```
VWSCoronaDashboard8, line 75: I would expect left and top to set a position but I wonder why a margin setting is used to position them.

Also, where is a with and height setting being passed in but here yMax and xMax are used instead and the component has a confusing name, so it's hard to know what this does.

src/components-styled/bar-chart/bar-chart-chart.tsx
```diff
@@ -0,0 +1,132 @@
...
+          <Group key={index}>
+            <Text
+              textAnchor="end"
+              verticalAnchor="middle"
+              y={yPoint(value) + yScale.bandwidth() / 2}
+              x={margin.left - spacingLabel}
+              fill={colors.annotation}
+              fontSize="0.75rem"
+            >
+              {y(value)}
```
VWSCoronaDashboard8, line 105: y is a function doing what?

src/components-styled/bar-chart/bar-chart-chart.tsx
```diff
@@ -0,0 +1,132 @@
...
+        hideAxisLine={true}
+        tickFormat={(a) => `${a}`}
+        tickComponent={TickValue}
+        label={xAxisLabel}
+        labelClassName="bar-chart-x-axis-label"
+      />
+
+      {values.map((value, index) => {
+        return (
+          <Group key={index}>
```
VWSCoronaDashboard8, line 96: Are you sure you want to use index for a key? If values are unique you can use that or combine them with key.

src/components-styled/bar-chart/bar-chart-chart.tsx
```diff
@@ -0,0 +1,132 @@
...
+            <Text
+              textAnchor="end"
+              verticalAnchor="middle"
+              y={yPoint(value) + yScale.bandwidth() / 2}
```
VWSCoronaDashboard8, line 100: yPoint is not a clear name for a function since it doesn't have a verb in it. createYPoint, getYPoint? maybe?

Same goes for yScale.bandwidth() but maybe that's a visx thing?

src/components-styled/bar-chart/bar-chart-coordinates.ts
```diff
@@ -0,0 +1,113 @@
...
+import { GetTooltipCoordinates } from '../tooltip';
+
+export interface BarChartValue {
+  x: number;
+  y: string;
+  tooltip: string;
+  color: string;
+}
+
+export interface BarChartCoordinates {
```
VWSCoronaDashboard8, line 13: This name keeps throwing me off in the code. It seems to be much more than coordinates. I wonder why all these things are combined in a single interface like values together with margins.

src/components-styled/bar-chart/bar-chart-coordinates.ts
```diff
@@ -0,0 +1,113 @@
...
+  xScale: ValueOf<ScaleTypeToD3Scale<any, any, any>>;
+  yScale: ScaleBand<string>;
+  xMax: number;
+  yMax: number;
+  numTicks: number;
+  values: any[];
+  xPoint: (x: any) => number;
+  yPoint: (x: any) => number;
+  y: (x: any) => string;
+  getTooltipCoordinates: GetTooltipCoordinates<any>;
```
VWSCoronaDashboard8, line 32: Please don't use any

src/components-styled/bar-chart/bar-chart-coordinates.ts
```diff
@@ -0,0 +1,113 @@
...
+  const spacingLabel = 10;
+
+  const xMax = width - margin.left - margin.right;
+  const yMax = values.length * 40;
+  const height = yMax + margin.top + margin.bottom;
+
+  const numTicks = 10;
+
+  const x = (value: BarChartValue): number => value.x;
+  const y = (value: BarChartValue): string => value.y;
```
VWSCoronaDashboard8, line 66: These functions seem to be simply getting a specified property from the value. I wonder why you define functions for it, but in any case, I think they should be named accordingly like `getPropertyX()`

src/components-styled/bar-chart/bar-chart-coordinates.ts
```diff
@@ -0,0 +1,113 @@
...
+  });
+
+  const yScale = scaleBand({
+    range: [margin.top, height - margin.bottom],
+    round: true,
+    domain: values.map(y),
+    padding: 0.2,
+  });
+
+  const createPoint = (scale: any, accessor: any) => (value: BarChartValue) =>
```
VWSCoronaDashboard8, line 81: anys

src/components-styled/bar-chart/bar-chart-coordinates.ts
```diff
@@ -0,0 +1,113 @@
...
+    scale(accessor(value));
+
+  const xPoint = createPoint(xScale, x);
+  const yPoint = createPoint(yScale, y);
+
+  function getTooltipCoordinates(
+    event?: MouseEvent<any>,
```
VWSCoronaDashboard8, line 88: I don't see event being used in this function. And I wonder what it means to try to get tooltip coordinates for a non-existing value. In what valid scenario is there no value available?

src/components-styled/bar-chart/bar-chart.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { BarChartChart } from './bar-chart-chart';
+import { BarChartValue, useBarChartCoordinates } from './bar-chart-coordinates';
+import { BarChartTooltipContent } from './bar-chart-tooltip-contents';
+
+interface BarChartProps {
+  values: BarChartValue[];
+  xAxisTitle: string;
+  chartLabel: string;
+  valueAnnotation?: string;
+}
```
VWSCoronaDashboard8, line 15: Aaaah here is the public bar chart interface :) 

Wat is a chartLabel? Is it a title? I would instead call xAxisTitle => xAxisLabel as a title would be heading the chart and axis are labelled in my opinion.

PR #1169, opened 2020-12-14
Visx linechart improve types

VWSCoronaDashboard8:
Continuing on the draft for #1093, mainly improving typings.


src/components/custom-line-chart/chart/tooltip.tsx
```diff
@@ -22,8 +21,10 @@ export type Props = {
   bounds: Bounds;
 };
 
-// TODO: improve how bounds are used to keep tooltips within the chart
-function Tooltip({
+/**
+ * @TODO improve how bounds are used to keep tooltips within the chart
+ */
+export function Tooltip({
```
eileenmguo, line 27: Did removing memoization impact performance?

src/components-styled/line-chart-tile.tsx
```diff
@@ -1,12 +1,9 @@
...
-
+import { CustomLineChart } from '~/components/custom-line-chart/index';
 import { TimeframeOption } from '~/utils/timeframe';
 import { ChartTileWithTimeframe } from './chart-tile';
 import { MetadataProps } from './metadata';
-import CustomLineChart, {
-  LineChartProps,
-  Value,
-} from '~/components/custom-line-chart/index';
+import { LineChartProps, Value } from '~/components/lineChart';
```
eileenmguo, line 6: I had copied these types over to the CustomLineChart because I thought the plan was to remove this component all together. Is that not the intention?

src/components/custom-line-chart/index.tsx
```diff
@@ -51,20 +47,18 @@ export default function CustomLineChart<T extends Value>({
...
     hideTooltip,
-  }: UseTooltipParams<TooltipData> = useTooltip();
+  } = useTooltip<T>();
 
-  const benchmark = useMemo(() => {
-    return signaalwaarde
-      ? { value: signaalwaarde, label: 'Signaalwaarde' }
-      : null;
-  }, [signaalwaarde]);
+  const benchmark = signaalwaarde
```
eileenmguo, line 59: did removing memoization impact performance?

src/components/custom-line-chart/index.tsx
```diff
@@ -129,40 +123,20 @@ export default function CustomLineChart<T extends Value>({
...
           >
             {formatTooltip
-              ? formatTooltip({
-                  ...tooltipData,
-                  date: dateToValue(tooltipData.date),
-                })
-              : `${formatDateFromSeconds(
-                  dateToValue(tooltipData.date)
-                )}: ${formatNumber(tooltipData.value)}`}
+              ? formatTooltip(tooltipData)
```
eileenmguo, line 133: Since I've converted the dates values to date objects via `graphData`, to use the existing formatting function the date needs to be converted back. Since I was trying to keep the props the same, I did this. What would you prefer this to be handled?

src/components/lineChart/index.tsx
```diff
@@ -28,7 +29,7 @@ export interface LineChartProps<T> {
 function getChartOptions<T extends Value>(
   values: T[],
   signaalwaarde?: number,
-  formatTooltip?: (value: T) => string,
+  formatTooltip?: (value: T) => ReactNode,
```
eileenmguo, line 32: I haven't looked into the documentation myself, but my understanding is that the highCharts accepts a function that returns a string for the tooltip formatter. The CustomLineChart can accept a function that returns a node (which I believe is an improvement), but that means these charts will have different types. This is another reason why I opted to move the types to the new chart component

src/components-styled/line-chart-tile.tsx
VWSCoronaDashboard8: Yes, you're right, this was a merge conflict with my branch and I choose to import them instead. I think it's a little easier to import the types once the original code gets removed. That way you can't confuse imports, and you avoid having duplicate code.

src/components/custom-line-chart/chart/tooltip.tsx
VWSCoronaDashboard8: I have never seen memo being used this way, and it seems strange to me to memorize something if the props are constantly changing. So I would flip the question. Did this use of memo create a measurable performance benefit? 😄

src/components/custom-line-chart/index.tsx
VWSCoronaDashboard8: I think this is a good example of useMemo being complete overkill. The operation performed inside the useMemo is in no way heavy on the CPU. In fact the overhead of the useMemo call itself might actually impact performance more. No function is completely free, and if it doesn't hurt performance it still hurts code readability.

So I could flip the question again... did you apply this function because you perceived or measured a performance improvement?

src/components/custom-line-chart/index.tsx
VWSCoronaDashboard8: After fixing the types on the tooltip data, it seems like this conversion was not necessary. I think the `tooltipData` variable contains the data as it is passed in via `values` and that's also the format that the tooltip formatted function needs. 

It seemed to work when I tested it. Or did I miss something?

src/components/lineChart/index.tsx
VWSCoronaDashboard8: In highcharts we already use tooltip formatter functions that return a react node. The highchart types don't seem to allow this, but we were just tricking the compiler into thinking a string was returned. So the new line chart can have ReactNode as the return value and it would still be compatible.

src/components/lineChart/index.tsx
```diff
@@ -75,11 +76,14 @@ function getChartOptions<T extends Value>(
       backgroundColor: '#FFF',
       borderColor: '#01689B',
       borderRadius: 0,
-      formatter: function (): string {
-        if (formatTooltip) {
-          return formatTooltip(values[this.point.index]);
-        }
```
VWSCoronaDashboard8, line 81: @VWSCoronaDashboard13 this is where we tricked the compiler. If a formatTooltip function was passed in, we'd call that and declaring the result as string, but if you look at the code defining these formatTooltip functions, you'll see that a lot of them return ReactNode instead.

src/components/lineChart/index.tsx
```diff
@@ -75,11 +76,14 @@ function getChartOptions<T extends Value>(
...
-        return `${formatDateFromSeconds(this.x)}: ${formatNumber(this.y)}`;
+      formatter: function () {
+        /**
+         * Here we trick Highcharts in thinking we return a string, because it
+         * does render ReactNode but doesn't allow it as a return value here.
+         */
+        return formatTooltip
+          ? (formatTooltip(values[this.point.index]) as string)
```
VWSCoronaDashboard8, line 85: So here I'm going the same, just a little more explicit and clean.

src/components/custom-line-chart/chart/tooltip.tsx
VWSCoronaDashboard10: @VWSCoronaDashboard8 you might be confused, there's a `useMemo`-hook and a `memo`-HOC. Here the latter is used :-) The `memo`-HOC can be compared to a React.PureComponent.

it's a bad idea to wrap the Tooltip in a memo-HOC when props constantly update, it will impact performance negatively instead of positively.

src/components/custom-line-chart/index.tsx
VWSCoronaDashboard10: I disagree @VWSCoronaDashboard8, there's an actual difference right now. The `benchmark: { value,  label }` object is now re-created on every render and it's passed as a prop to the `<Chart />`. This Chart component will now always re-render only because of the new 'benchmark'-prop on every render. Previously it would receive the same object instance, now it'll be a new instance every render.

We could of course question ourself whether we want to create this object here and pass it as a single prop, or that we simply pass `signaalwaarde` and `label` as separate props, which would not negatively impact performance compared to the removed usememo..

VWSCoronaDashboard10: What I'd like to emphasise is that useMemo is not only useful for expensive calculations, but also to prevent expensive renders of child components receiving these props.

src/components/lineChart/index.tsx
VWSCoronaDashboard10: I didn't knew that was possible, can you show an example? I can't imagine highcharts compiles react..

src/components/custom-line-chart/chart/tooltip.tsx
VWSCoronaDashboard8: Yes, I was confused at first but, but in general memoizing anything that changes frequently doesn't make sense and that was my point :)

src/components/lineChart/index.tsx
VWSCoronaDashboard10: I really think eileen has a point and it will break if we for example have the following:
```
formatTooltip={x => <p>why hello there</p>}
```
where it's actually returning a ReactNode instead of a string

src/components/custom-line-chart/index.tsx
VWSCoronaDashboard8: @VWSCoronaDashboard10 You have a point there 😄 But looking at the props of `CustomLineChart`, isn't that the level where you'd want to avoid re-rendering? Because those props wouldn't change for a given dataset. If we memorize there, we could drop all the useMemo hooks in this component maybe?

src/components/custom-line-chart/chart/tooltip.tsx
VWSCoronaDashboard13: So I added memoization to the tooltip specifically for performance reasons but it is now accepting a bounds prop that gets recreated on every render and seems to be working fine so the memoization on the Tooltip is unnecessary. The chart, however, needs to the memoization along with the memoization of the benchmark (so a new object isn't triggering it to rerender) to keep it from rerendering and causing the interaction to lag. When memoized the chart only rerenders when the cursor enters and leaves the graph for the hover style of the line

src/components/custom-line-chart/index.tsx
VWSCoronaDashboard13: Well the chart still needs to rerender when the tooltip's position changes which will cause the svg part of the graph to still rerender unnecessarily so I think memoization of specific components within `CustomLineChart` is necessary for performance reasons

src/components/custom-line-chart/index.tsx
VWSCoronaDashboard13: It works when a formatting function isn't passed into the component, but this should break the `Percentage positieve testen over tijd` graph since those formatting functions are expecting the `value` instead of the date object. Converting that value back and forth is a bit strange, but since it's specific to this project I didn't want to change the existing props being passed to the line chart and I felt like the dates should be normalized before being passed to the generic chart. This is definitely something that needs to be improved, I just did this in the interim and I think how to handle this would be a good decision for the team to make together

src/components/lineChart/index.tsx
VWSCoronaDashboard13: Original Question (deleted because it was associated with the wrong username)

I haven't looked into the documentation myself, but my understanding is that the highCharts accepts a function that returns a string for the tooltip formatter. The CustomLineChart can accept a function that returns a node (which I believe is an improvement), but that means these charts will have different types. This is another reason why I opted to move the types to the new chart component

src/components/lineChart/index.tsx
VWSCoronaDashboard: That means the name `formatTooltip` is a misnomer. I expect Formatting to do nothing else but formatting a string.

It could be called returnTooltip. Or simply Tooltip?

VWSCoronaDashboard: Anyhow. Interesting discussion, but let's try to keep all of these things contained to 1 PR. This PR improves the types and that's it.

VWSCoronaDashboard (approved):

PR #1183, opened 2020-12-15
Clean up unused code

VWSCoronaDashboard10:
## Summary

- Cleaning up a lot of unused files
- Remove the export keyword from types that aren't used anywhere else


VWSCoronaDashboard8 (approved):

src/components/charts/index.tsx
```diff
@@ -2,4 +2,3 @@ import dynamic from 'next/dynamic';
 
 export const AreaChart = dynamic(() => import('~/components/areaChart'));
 export const BarChart = dynamic(() => import('~/components/barChart'));
-export const LineChart = dynamic(() => import('~/components/lineChart'));
```
VWSCoronaDashboard4, line 5: Why is this removed?

VWSCoronaDashboard4 (approved):

src/components/charts/index.tsx
VWSCoronaDashboard10: Because it's not used

PR #1210, opened 2020-12-16
Add stichting nice as bron for vr hospital choropleth

VWSCoronaDashboard8:


VWSCoronaDashboard10 (approved):

VWSCoronaDashboard10:
Closed -- discussed with client and they will update texts in lokalise. We can clean it up when we feel the need, but currently it's easier to have them set the correct strings.

PR #1221, opened 2020-12-17
Topic/actueel menu

VWSCoronaDashboard4:
## Summary

Implementation of new main menu with collapsable mobile menu.

## Detailed design

* Most styling put into styled components. I've tried using eg the box component but it became messy and certain layout things broke.
* For accessibility: aria roles, states and visually hidden texts are added.
* Cleaned up non-used lokalise keys


src/assets/close.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="36" height="36" viewBox="0 0 36 36" fill="currentColor" xmlns="http://www.w3.org/2000/svg">
```
VWSCoronaDashboard1, line 1: add focusable="false"

src/assets/menu.svg
```diff
@@ -0,0 +1 @@
+<svg width="36" height="36" viewBox="0 0 36 36" fill="none" xmlns="http://www.w3.org/2000/svg"><path d="M0 0h36v36H0z"/><path d="M28.852 7.826H7.148v4.07h19.67l2.034-4.07zM7.148 20.035h19.67l2.034-4.07H7.148v4.07zM26.817 28.174H7.147v-4.07h21.705l-2.035 4.07z" fill="currentColor"/></svg>
```
VWSCoronaDashboard1, line 1: add focusable="false"

VWSCoronaDashboard1 (changes requested):

VWSCoronaDashboard1 (approved):

src/domain/site/components/top-navigation.tsx
```diff
@@ -1,115 +1,198 @@
...
+import Close from '~/assets/close.svg';
+import theme from '~/style/theme';
+import { VisuallyHidden } from '~/components-styled/visually-hidden';
 
 export function TopNavigation() {
   const router = useRouter();
-  const breakpoints = useBreakpoints(true);
+
+  const [isSmallScreen, setIsSmallScreen] = useState(true);
+  const [isMenuOpen, setIsMenuOpen] = useState(true);
```
VWSCoronaDashboard8, line 19: Didn't the menu open/close state become part of the URL recently? I think it's not preferable that a menu like this holds state.

src/domain/site/components/top-navigation.tsx
```diff
@@ -1,115 +1,198 @@
...
+  const breakpoints = useBreakpoints();
+
+  useEffect(() => {
+    setIsSmallScreen(!breakpoints.md);
```
VWSCoronaDashboard8, line 23: It looks like you might not need the isSmallScreen state, because you can derive it as a boolean directly from the breakpoint.

src/domain/site/components/top-navigation.tsx
```diff
@@ -1,115 +1,198 @@
...
+      ml: 'auto',
+      mr: 0,
+      mt: 0,
+      pb: 1,
+      flex: '0 0 auto',
+    },
+  })
+);
+
+const NavList = styled.ol(
```
VWSCoronaDashboard8, line 135: nit; not sure if navigation can be considered an ordered list. It's not arbitrary, but it's not ordered according to some property either.

src/domain/site/components/top-navigation.tsx
```diff
@@ -1,115 +1,198 @@
...
+  css({
+    listStyle: 'none',
+    padding: 0,
+    margin: 0,
+    display: ['block', null, null, 'flex'],
```
VWSCoronaDashboard8, line 140: @VWSCoronaDashboard10 made a utility for this where you can use the object format and it'll convert it to the list. I would prefer we use that for these cases.

src/domain/site/components/top-navigation.tsx
```diff
@@ -1,115 +1,198 @@
...
+  })
+);
+
+// Lines in mobile menu between items
+const StyledListItem = styled.li(
+  css({
+    '& + &': {
+      borderTop: '1px solid rgba(255, 255, 255, 0.25)',
+      borderTopWidth: ['1px', null, null, 0],
```
VWSCoronaDashboard8, line 149: same comment about mq objects here

src/domain/site/components/top-navigation.tsx
```diff
@@ -1,115 +1,198 @@
...
+    padding: 0,
+    margin: 0,
+    display: ['block', null, null, 'flex'],
+  })
+);
+
+// Lines in mobile menu between items
+const StyledListItem = styled.li(
+  css({
+    '& + &': {
```
VWSCoronaDashboard8, line 147: What is this for? Styled components also lets you target children specifically. Maybe you can use that. See https://styled-components.com/docs/advanced#referring-to-other-components

src/domain/site/components/top-navigation.tsx
```diff
@@ -1,115 +1,198 @@
...
-    px: ['0.75em', null, '1.5em'],
-    py: '1em',
+    whiteSpace: 'nowrap',
     textDecoration: 'none',
-    fontSize: ['1em', '1.125em'],
+    fontSize: '1.1rem',
+    color: 'white',
+
+    // The span is a narrower element to position the underline to
+    '& span': {
```
VWSCoronaDashboard8, line 163: Maybe you can also target a specific child component here.

VWSCoronaDashboard8 (changes requested):
I think we can get rid of the use of state, but otherwise looks good 👍

src/domain/site/components/top-navigation.tsx
VWSCoronaDashboard10: This is a different kind of menu. The sidebar menu is toggled using menu state because it is used as a "page" on narrow devices. This new top level menu replaces the horizontal menu bar which is hidden with a hamburger icon. I think going for state is perfectly fine in this case.

src/domain/site/components/top-navigation.tsx
VWSCoronaDashboard10: Agree, `const isSmallScreen = !breakpoints.md` has the same effect.

src/domain/site/components/top-navigation.tsx
VWSCoronaDashboard4: I've tried that approach, it resulted in a different kind of spaghetti code because the `isActive` would need to be passed.

I'll update this to at least use a dedicated Component to look up the proper element in the selector.

src/domain/site/components/top-navigation.tsx
VWSCoronaDashboard4: This has nothing to do with child components. It's targeting the second, third etc use of this component in the list.

src/domain/site/components/top-navigation.tsx
VWSCoronaDashboard4: In this case it makes more sense to go for UL indeed

src/domain/site/components/top-navigation.tsx
VWSCoronaDashboard10: Yeah but that would at least be a kind of expected "static" spaghetti. Now there can happen unexpected runtime errors when someone renders another span in this component or when someone tries to change styling of that targeted span.. But do what feels best in this case 👍

src/domain/site/components/top-navigation.tsx
VWSCoronaDashboard8: Could you alternatively use a more understandable selector like `:not(:first-child)`?

VWSCoronaDashboard8 (approved):

PR #1231, opened 2020-12-17
Add support for Sanity

VWSCoronaDashboard:
**What?**

We're adding a CMS to the platform. 

**What should you review?**

- The monorepo setup in general
- The integration between Sanity and the front-end

Todo's

- [ ] Remove the sanity client in the component, or at least the preview mode, on production builds. All we really need for initial integration is a working client in getStaticProps
- [ ] Move all API calls to a central API file instead of writing all GROQ queries in the pages directly
- [ ] Make sure skipLinks returns human readable urls. So base ID's on title instead of _key.


src/pages/verantwoording.tsx
```diff
@@ -50,28 +91,27 @@ const Verantwoording: FCWithLayout<VerantwoordingProps> = (props) => {
...
-                ) : null;
+                );
               })}
             </article>
           </div>
         </MaxWidth>
       </div>
     </>
   );
+  ``;
```
VWSCoronaDashboard, line 114: This looks like an accident. I'll remove it.

src/types/sanity.d.ts
```diff
@@ -0,0 +1 @@
+declare module '@sanity/block-content-to-react';
```
VWSCoronaDashboard, line 1: Not sure if we still need this

src/pages/verantwoording.tsx
```diff
@@ -50,28 +91,27 @@ const Verantwoording: FCWithLayout<VerantwoordingProps> = (props) => {
       <div className={styles.container}>
         <MaxWidth>
           <div className={styles.maxwidth}>
-            <h2>{text.verantwoording.title}</h2>
-            <p>{text.verantwoording.paragraaf}</p>
+            <h2>{verantwoordingList.title}</h2>
+            <PortableText blocks={verantwoordingList.description} />
             <article className={styles.faqList}>
-              {text.verantwoording.cijfers.map((item) => {
-                const id = getSkipLinkId(item.cijfer);
```
VWSCoronaDashboard, line 57: The plan is to remove getSkipLink, because we get unique ID's from the database. Unless this method is used to human readable urls?

.env.local.example
```diff
@@ -1,2 +1,6 @@
 # This can be nl or en. If you don't specify it, we default to nl
-NEXT_PUBLIC_LOCALE=nl
\ No newline at end of file
+NEXT_PUBLIC_LOCALE=nl
+
+NEXT_PUBLIC_SANITY_PROJECT_ID=
```
VWSCoronaDashboard, line 4: Because of the NEXT_PUBLIC prefix, the SANITY_PROJECT_ID will be embedded in the bundle. That's the only way to use Sanity client-side. We could get away with only using Sanity in gSP, but that would probably impact the usage of next-sanity's image url builder.

src/lib/api.js
```diff
@@ -0,0 +1,72 @@
+import client, { previewClient } from './sanity';
```
VWSCoronaDashboard, line 1: This whole file is filled with code I copied from a random example.
The code as-is, is unreviewable.

However, the pattern of exposing awaitable exports here is quite nice. You can `await getUniquePosts` in getStaticProps and that's it. 

Question: Do we want to export all GROQ queries this way instead of writing them directly in page files?

src/pages/verantwoording.tsx
VWSCoronaDashboard1: AFAIK, there's just used to create unique anchors, so it should pose no problem

src/lib/api.js
VWSCoronaDashboard8: I think I would prefer to colocate the queries with the pages, because then you keep things in context and I suspect its easier to make changes that way. But we have some time to review that / figure it out I think.

src/pages/veelgestelde-vragen.tsx
```diff
@@ -5,31 +5,64 @@ import { FCWithLayout, getLayoutWithMetadata } from '~/components/layout';
...
+import {
+  getClient,
+  usePreviewSubscription,
+  PortableText,
+  localize,
+} from '~/lib/sanity';
+
+import { targetLanguage } from '../locale/index';
+
+interface IVraagEnAntwoord {
```
VWSCoronaDashboard8, line 22: nit; We don't prefix interfaces with I anymore

src/pages/veelgestelde-vragen.tsx
```diff
@@ -5,31 +5,64 @@ import { FCWithLayout, getLayoutWithMetadata } from '~/components/layout';
...
+}
+
 interface StaticProps {
   props: VeelgesteldeVragenProps;
 }
 
 interface VeelgesteldeVragenProps {
   text: TALLLanguages;
   lastGenerated: string;
+  faq: any;
```
VWSCoronaDashboard8, line 35:
```suggestion
  faq: Record<string, unknown>;
```
Something like this could probably work too?

src/pages/veelgestelde-vragen.tsx
```diff
@@ -51,17 +84,17 @@ const Verantwoording: FCWithLayout<VeelgesteldeVragenProps> = (props) => {
...
-                  <Collapsable key={id} id={id} summary={item.vraag}>
-                    <div
-                      dangerouslySetInnerHTML={{
-                        __html: item.antwoord,
-                      }}
-                    />
+                  <Collapsable
+                    key={item._key}
+                    id={item._key}
+                    summary={item.title}
```
VWSCoronaDashboard8, line 95: Odd that a summary is written in a title property

src/pages/verantwoording.tsx
```diff
@@ -6,30 +6,71 @@ import { FCWithLayout, getLayoutWithMetadata } from '~/components/layout';
...
+import {
+  getClient,
+  usePreviewSubscription,
+  PortableText,
+  localize,
+} from '~/lib/sanity';
+
+import { targetLanguage } from '../locale/index';
+
+interface ICijfer {
```
VWSCoronaDashboard8, line 21:
```suggestion
interface CijferVerantwoording {
```

src/pages/verantwoording.tsx
```diff
@@ -6,30 +6,71 @@ import { FCWithLayout, getLayoutWithMetadata } from '~/components/layout';
...
+}
+
 interface StaticProps {
   props: VerantwoordingProps;
 }
 
 interface VerantwoordingProps {
   text: TALLLanguages;
   lastGenerated: string;
+  cijferVerantwoording: any;
```
VWSCoronaDashboard8, line 34: really any?

src/pages/verantwoording.tsx
```diff
@@ -6,30 +6,71 @@ import { FCWithLayout, getLayoutWithMetadata } from '~/components/layout';
...
+
 export async function getStaticProps(): Promise<StaticProps> {
   const text = parseMarkdownInLocale((await import('../locale/index')).default);
 
   const filePath = path.join(process.cwd(), 'public', 'json', 'NL.json');
   const fileContents = fs.readFileSync(filePath, 'utf8');
   const lastGenerated = JSON.parse(fileContents).last_generated;
 
-  return { props: { text, lastGenerated } };
+  // @ts-ignore
```
VWSCoronaDashboard8, line 51: nit; Please add a comment about why you thought this was required.

VWSCoronaDashboard8 (approved):

src/pages/veelgestelde-vragen.tsx
VWSCoronaDashboard: Yup. I'm going to remove this one anyway. We've got a more generic Collapsible type for this that get's reused across different pages.

src/pages/veelgestelde-vragen.tsx
VWSCoronaDashboard: I think that's more of a misnomer of the collapsible component prop than an error in the CMS schema.

PR #1252, opened 2020-12-21
Convert project into a monorepo

VWSCoronaDashboard:
## Summary

This PR will add the foundations of a monorepo to the codebase 
## Motivation

We'll be adding the CMS to this repo soon. We want the CMS and front-end to live in different folders. The first step is to change the structure and ensure our pipeline is still working before we add the CMS.

## Detailed design

We're using [yarn workspaces](https://classic.yarnpkg.com/en/docs/workspaces/)

I've added an empty folder for the CMS with a small `.gitkeep` file. Just to reserve the folder/package name.

## Drawbacks

An alternative is to build the CMS in a different repository, but we prefer that our feature branches contain both the front-end and CMS changes needed to merge the feature.


package.json
```diff
@@ -1,143 +1,7 @@
 {
-  "name": "corona-dashboard",
-  "version": "2.9.0",
   "private": true,
```
VWSCoronaDashboard4, line 2: With this diff , this line caught my attention :-P Shouldn't this be `"private": false` ?

VWSCoronaDashboard4 (approved):
:100:

package.json
VWSCoronaDashboard8: I believe private true prevents you from accidentally publishing a package to npm. It has little to do with if your code is open source.

In a monorepo you would mark this package as private, so you can also have other packages that you do publish.

package.json
```diff
@@ -1,143 +1,7 @@
...
-    "cy:ci": "cypress run --headless --browser chrome --config-file ./cypress-ci.json",
-    "download": "download https://coronadashboard.rijksoverheid.nl/latest-data.zip -e -o public/"
-  },
-  "browserslist": [
-    "defaults",
-    "not ie < 11"
+  "workspaces": [
+    "corona-dashboard-cms",
+    "corona-dashboard-app"
   ]
```
VWSCoronaDashboard8, line 6: As I mentioned before, it's cleaner to have all packages live in a dedicated directory, and then this config can be:

```suggestion
 "workspaces": {
    "packages": [
      "packages/*"
    ]
  },
```

VWSCoronaDashboard8: I suspect also this is why workspaces isn't working, because the config structure differs

VWSCoronaDashboard8 (changes requested):

package.json
VWSCoronaDashboard: What 8 says. So this is resolved.

PR #1253, opened 2020-12-21
GGD daily + average

VWSCoronaDashboard8:
## Summary

GGD data will be delivered in two different sets. One calculated as moving average (over 7 days) like we have now, and the other data is reported over a single day.

Because one uses day timestamps and the other date range timestamps (week_start/week_end) I split them into two schema's, so each can have its own independent timestamps.

### DIFF
--- a/schema/national/__index.json
+++ b/schema/national/__index.json
@@ -24,6 +24,7 @@
     "hospital_beds_occupied",
     "intensive_care_beds_occupied",
     "ggd",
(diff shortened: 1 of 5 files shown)


VWSCoronaDashboard4 (approved):

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard8:
Closed in favor or #1309

PR #1265, opened 2020-12-21
Add landelijke toelichting to national actueel page

VWSCoronaDashboard1:
## Summary

This PR adds a tile that shows a photo and a small article describing the current covid situation in the country.

## Motivation

Feature request

## Detailed design

A tile with a photo and some text. Not much else to say really.

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added


src/components-styled/landelijke-toelichting.tsx
```diff
@@ -0,0 +1,54 @@
...
+import styled from 'styled-components';
+import ExternalLinkIcon from '~/assets/external-link.svg';
+import siteText from '~/locale';
+import theme from '~/style/theme';
+import { Box } from './base';
+import { ExternalLink } from './external-link';
+import { Tile } from './tile';
+import { Heading, Text } from './typography';
+
+export function LandelijkeToelichting() {
```
VWSCoronaDashboard10, line 11: I don't really like the dutch component name and by reading the name I don't think this should be part of our component-directory, it's not reusable. Can't it be renamed to something more generic? For example NewsMessage which receives props instead of extracting translation strings? I guess what I try to say is something like this:
```tsx
<NewsMessage
  image='images/toelichting-afbeelding.png'
  title={siteText.notificatie.titel}
  subtitle={siteText.notificatie.subtitel}
  publishedAt={siteText.notificatie.datum}
  anchor={siteText.notificatie.link.href}
  anchorText={siteText.notificatie.link.text}
/>
```
This way we create a reusable component and it is also more future-proof when we're moving towards a CMS.

src/components-styled/landelijke-toelichting.tsx
```diff
@@ -0,0 +1,54 @@
...
+  return (
+    <Tile
+      css={css({
+        bg: theme.colors.contextualContent,
+        flexDirection: [null, 'row'],
+      })}
```
VWSCoronaDashboard10, line 17: I believe Tile extends the Box component which means it should be possible to set props for `bg` and `flexDirecrion`
```suggestion
    <Tile
        bg={theme.colors.contextualContent}
        flexDirection={[null, 'row']}
```

src/components-styled/landelijke-toelichting.tsx
```diff
@@ -0,0 +1,54 @@
...
+    >
+      <Box flex="1 1 25%">
+        <img src="images/toelichting-afbeelding.png" />
+      </Box>
+      <Box flex="1 1 75%" pl={[null, 4]}>
+        <Text mt={0} color="gray">
+          {siteText.notificatie.subtitel}
+          {' \u2022 '}
```
VWSCoronaDashboard10, line 25: I think this is more readable:
```suggestion
          {' • '}
```

src/style/theme.ts
```diff
@@ -99,6 +99,7 @@ export const colors = {
   border: '#c4c4c4',
   lightBlue: '#E0EEF6',
   restrictions: '#CD0059',
+  contextualContent: '#e5eff8',
```
VWSCoronaDashboard10, line 102: ContextualContent is a bit vague to me, but I also don't know an alternative..

VWSCoronaDashboard10 (approved):
Approved, but I do have some suggestions - curious what others think.

src/components-styled/landelijke-toelichting.tsx
VWSCoronaDashboard1: I'd say this is premature optimization, to be honest. I find it incredibly difficult to anticipate re-usable components from the get-go and much more prefer refactors when several components have been made and specific overlap has crystalized.
But, I'll still do the refactor you suggest.

src/components-styled/landelijke-toelichting.tsx
VWSCoronaDashboard1: I tried those and that didn't work, the flexDirection was completely ignored...

src/style/theme.ts
VWSCoronaDashboard1: I chose this name because in the design description this color is being described as 'highlighting content that gives additional context'.

src/components-styled/news-message.tsx
```diff
@@ -0,0 +1,69 @@
...
+};
+
+export function NewsMessage(props: NewsMessageProps) {
+  const {
+    imageSrc,
+    title,
+    subtitle,
+    publishedAt,
+    href,
+    anchorText,
```
VWSCoronaDashboard8, line 26: nit; Anchor is a bit of a specific HTML term. I think linkText is more suitable.

VWSCoronaDashboard8 (approved):
Only one small suggestion

src/components-styled/landelijke-toelichting.tsx
VWSCoronaDashboard10: Yes perhaps, but than the LandelijkeToelichting component shouldn't be part of the components-styled directory imho :-)

src/pages/index.tsx
```diff
@@ -53,6 +54,16 @@ const Home: FCWithLayout<IHomeData> = (data) => {
           )}
         />
       </ChoroplethTile>
+
+      <NewsMessage
```
VWSCoronaDashboard4, line 58: The term `news` is more limited than it is going to be used for. Eg a `toelichting` is not news. Can the naming be made a bit more generic? Eg `Article`; then this can be an `ArticlePreview` perhaps?

src/components-styled/landelijke-toelichting.tsx
VWSCoronaDashboard10: Oh that's a shame, I think moving the Tile's default bg css style to the `attrs()` should fix that:
```
export const Tile = styled(Box).attrs({ as: 'article', bg: 'white' })(
  css({
    display: 'flex',
    flexDirection: 'column',
    p: 4,
    borderRadius: 1,
    boxShadow: 'tile',
  })
);
```

src/pages/index.tsx
VWSCoronaDashboard1: I don't think it's a preview either. I agree it might not be news, its more of a weekly update I guess. So, `ArticleMessage` perhaps?

VWSCoronaDashboard1: I'm leaving it for now, renaming can be done at a later moment

src/components-styled/landelijke-toelichting.tsx
VWSCoronaDashboard8: I agree that it would be smart to keep the components directory for reusable things and put domain specific stuff in the domain folder. But there is already lots of specific stuff in the components folder, so maybe we can make this part of a cleanup / refactor in the near future and define some domains.

PR #1280, opened 2020-12-22
Topic/actueel data driven text

VWSCoronaDashboard4:
## Summary

Data driven texts for the Actueel page.

## Detailed design

* New `DataDrivenText` component that borrows some approach from `SidebarMetric`
* New `getPluralizedText` to choose between 3 localize texts based on  provided count.
* New `replaceComponentsInText` utility similar to `replaceVariablesInText`, but it allows to pass `ReactNode`s; this is used to pass a difference indicator, bold text etc in Lokalise texts. This returns a `ReactNode`.
* New `RelativeDate` component to provide a tooltip with the full date.
* Added 1 Data driven text component usage to the Actueel page. More can be configured when marking up the page more.
* Adjusted the `formatDateFromSeconds` to distinguish the day before yesterday when using relative date formatting.
* Adjusted `DifferenceIndicator` to have support for a `context` prop. Added an extra context `inline`. This file might need some tweaking to make it more manageable, we could perhaps split this into 3 components and a utility?


packages/app/src/components-styled/data-driven-text.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  assert(
+    isDefined(differenceValue),
+    `Missing value for difference:${differenceKey}`
+  );
+
+  const baseText = getPluralizedText(baseTexts, propertyValue);
+  const additionalText = getPluralizedText(additionalTexts, differenceValue.difference);
+
+  return <p>
```
VWSCoronaDashboard1, line 53:
```suggestion
  return <Text>
```

packages/app/src/components-styled/data-driven-text.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { getLastFilledValue } from "~/utils/get-last-filled-value";
+import { getPluralizedText, PluralizationTexts } from "~/utils/get-pluralized-text";
+import { replaceComponentsInText } from "~/utils/replace-components-in-text";
+import { DifferenceIndicator } from "./difference-indicator";
+import { RelativeDate } from "./relative-date";
+
+interface DataDrivenTextProps {
+  data: National;
+  differenceKey: string;
+  metricName: string;
```
VWSCoronaDashboard1, line 15:
```suggestion
  metricName: keyOf National;
```

VWSCoronaDashboard1 (approved):

packages/app/src/utils/get-pluralized-text.ts
```diff
@@ -0,0 +1,12 @@
...
+export function getPluralizedText(texts: PluralizationTexts, count: number): string {
+  const absoluteCount = Math.abs(count);
+  if (absoluteCount === 0) {
+    return texts.zero;
+  }
+  if (absoluteCount <= 1) {
+    return texts.singular;
+  }
+  return texts.plural;
+}
```
VWSCoronaDashboard10, line 12: I think this function can be part of `/components-styled/data-driven-text.tsx`. It's only used in that file and also rather specific to the `{ zero: string, singular: string; plural: string }` interface.

If you'd like to keep it here, I guess to make it a "utility" the interface should look a bit more generic like this:
```tsx
function getPluralizedText(count: number, plural: string, singular: string, zero?: string): string {}
```
I think zero should be optional because when pluralising a word (`bike` vs `bikes`) you'd only give two options since there's no different word for `zero bikes`.

packages/app/src/locale/nl.json
```diff
@@ -747,11 +747,26 @@
...
   },
   "nationaal_actueel": {
     "data_sitemap_titel": "Alle landelijke cijfers op dit dashboard",
-    "data_sitemap_toelichting": "Op het corona dashboard worden de cijfers van meerdere onderwerpen gepresenteerd. In verloop van tijd zullen er nieuwe onderwerpen worden toegevoegd of bestaande ondeerwerpen worden uitgebreid. Alle onderwerpen zijn onderverdeelt in categorieen. Dit is een overzicht van de categorieen en hun onderwerpen."
+    "data_sitemap_toelichting": "Op het corona dashboard worden de cijfers van meerdere onderwerpen gepresenteerd. In verloop van tijd zullen er nieuwe onderwerpen worden toegevoegd of bestaande ondeerwerpen worden uitgebreid. Alle onderwerpen zijn onderverdeelt in categorieen. Dit is een overzicht van de categorieen en hun onderwerpen.",
+    "data_driven_texts": {
+      "infected_people_total": {
+        "base": {
+          "zero": "{{newDate}} zijn er geen nieuwe positieve tests gemeld.",
+          "singular": "{{newDate}} is {{propertyValue}} nieuwe positieve test gemeld.",
```
VWSCoronaDashboard10, line 760: In theory this could be `"{{newDate}} is 1/één nieuwe positieve test gemeld."` - wouldn't that make more sense instead of always replacing the variable `propertyValue` with `1`?

packages/app/src/components-styled/relative-date.tsx
```diff
@@ -0,0 +1,34 @@
...
+  isCapitalized?: boolean;
+}
+
+export function RelativeDate({ dateInSeconds, isCapitalized }: RelativeDateProps) {
+  const isoDate = formatDateFromSeconds(dateInSeconds, 'iso');
+  const fullDate = formatDateFromSeconds(dateInSeconds, 'medium');
+
+  // Relative dates only become relative when run in the browser with JS on.
+  // Non-JS versions are the day and month.
+  let displayDate = formatDateFromSeconds(dateInSeconds, 'relative');
```
VWSCoronaDashboard10, line 23: This will probably result in hydration errors and other unexpected UI quirks. I suggest to explicitly switch date rendering on mount to tell react what is our expected UI:
```tsx
const isMounted = useIsMounted(); // a hook which will be available when the actueel search is merged

return (
  <Time datetime={isoDate} title={fullDate}>
    {formatDateFromSeconds(dateInSeconds, isMounted ? 'relative' : 'medium')}
  </Time>
)

**Review by VWSCoronaDashboard10 — APPROVED** [review]
Looking good, couple of suggestions/questions :-)

File: packages/app/src/utils/replace-components-in-text.tsx [hunk 6]
```diff
@@ -0,0 +1,17 @@
+import { assert } from "~/utils/assert";
+import { ReactNode, Fragment } from "react";
+
+export function replaceComponentsInText(text: string, replacementMap: Record<string, ReactNode | string | number>): ReactNode {
```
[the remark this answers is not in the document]
VWSCoronaDashboard4: I've looked into this, but I don't know if we really want to replace/merge it. Biggest reason not to do it: this method requires returning a ReactNode, which is not suitable for a lot of cases we use `replaceVariablesInText` for (which require a string).

Allowing this method to also accept strings/numbers seems ok to me, so you can at least mix. (In an earlier dev version I passed a string value to it, it ended up not making the final version.)

We can always tweak this in a separate PR.

packages/app/src/locale/nl.json
VWSCoronaDashboard4: `propertyValue` can be between 0 and 1.

packages/app/src/utils/replace-components-in-text.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard10: Ah I see you're right. 

> Allowing this method to also accept strings/numbers seems ok to me

Sure, but the type ReactNode already implies it accepts strings and numbers.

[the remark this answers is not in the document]
VWSCoronaDashboard10: I mean: the type `ReactNode | string | number` is the same as e.g. `ReactElement | string | number | string | number`

packages/app/src/components-styled/relative-date.tsx
```diff
@@ -0,0 +1,34 @@
+/**
+ * The RelativeDate provides a relative date in a context in which it can be interpeted
+ * * as human readable relative date, if possible
+ * * a full date in a tooltip
+ * * machine readable iso dates
```
VWSCoronaDashboard8, line 5:
```suggestion
 * The RelativeDate provides a relative date in a context in which it can be interpreted as:
 * - a human readable relative date, if possible
 * - a full date in a tooltip
 * - machine readable iso dates
```

packages/app/src/locale/nl.json
VWSCoronaDashboard10: Would a number between 0 and 1 be plural?

packages/app/src/components-styled/sidebar-metric/sidebar-kpi-value.tsx
```diff
@@ -55,7 +55,7 @@ export function SidebarKpiValue(props: SidebarKpiValueProps) {
 
         {isDefined(difference) && (
           <Box fontSize={3} display="flex" alignItems="center" marginRight={1}>
-            <DifferenceIndicator value={difference} isContextSidebar={true} />
+            <DifferenceIndicator value={difference} format='sidebar' />
```
VWSCoronaDashboard8, line 58: The property was called `context` instead of `format` I believe? Which would be my preference for naming anyway

VWSCoronaDashboard8 (approved):

packages/app/src/utils/get-pluralized-text.ts
VWSCoronaDashboard4: I'll move this back into `data-driven-text.tsx` then for now.

packages/app/src/components-styled/sidebar-metric/sidebar-kpi-value.tsx
VWSCoronaDashboard4: I've changed it at some moment, and this one missed the update. Will fix this

packages/app/src/locale/nl.json
VWSCoronaDashboard4: This could be interesting; since it differs between Dutch and English. For now I'd move onward with this, and have a separate PR deal with a proper English formatting.

packages/app/src/components-styled/relative-date.tsx
VWSCoronaDashboard4: I'll add a Todo for now

VWSCoronaDashboard10 (approved):

PR #1287, opened 2020-12-23
Improve line chart abstraction

VWSCoronaDashboard8:
## Summary

This PR makes some major improvements to the LineChart abstraction. This also paves the way for implementing multi line charts in the near future.

- Pass data directly to the chart as-is
- Use privately scoped properties __date and __value to avoid clashing with original data properties
- Add config to tell the chart for each line what property to use for rendering and what the color should be
- Chart detects the type of timestamps used (daily vs week/range)
- Adds standard tooltip formatter for both types of timestamps, which also works for percentage values
- Adds default y-axis formatter for percentages
- Strict typing (removes almost all `any` types and unnecessary type casts)
- Improve code readability


packages/app/src/components-styled/line-chart/chart.tsx
```diff
@@ -62,46 +67,50 @@ export const Chart = memo(function Chart({
...
 
-      return +date - +d0.date > +d1.date - +date ? d1 : d0;
+      return +date - +d0.__date > +d1.__date - +date ? d1 : d0;
     },
-    [x, margin]
+    [margin, xScale]
   );
 
   return (
     <svg width={width} height={height}>
```
VWSCoronaDashboard1, line 102:
```suggestion
    <svg width={width} height={height} role='img'>
```

packages/app/src/components-styled/line-chart/helpers.ts
```diff
@@ -1,23 +1,167 @@
...
+    'Unable to determine timestamps if time series is empty'
+  );
+
+  return firstValue.week_end_unix !== undefined;
+}
+
 /**
  * From all the defined values, extract the highest number so we know how to
- * scale the y-axis
+ * scale the y-axis. We need to do this for each fo the keys that are used to
```
VWSCoronaDashboard1, line 50:
```suggestion
 * scale the y-axis. We need to do this for each of the keys that are used to
```

VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/line-chart/chart.tsx
```diff
@@ -1,16 +1,19 @@
...
+import { TrendValue } from './helpers';
+import { Trend, TrendType } from './trend';
 
 const NUM_TICKS = 3;
+
 export const defaultMargin = { top: 10, right: 20, bottom: 30, left: 30 };
+
 const defaultColors = {
   main: colors.data.primary,
   axis: '#C4C4C4',
```
VWSCoronaDashboard1, line 19: Do the rest of this hardcoded colors also belong in the theme, or do we want to leave them here?

VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/line-chart/trend.tsx
```diff
@@ -1,68 +1,75 @@
...
   type: TrendType;
-  x: any;
-  y: any;
+  /**
+   * I would like to type these as follows:
+   *
+   * xScale: ScaleTime<number, number>;
+   * yScale: ScaleLinear<number, number>;
+   *
+   * ... using the types from 'd3-scale' but the visx components used here do not
```
VWSCoronaDashboard4, line 18: For the barchart `ValueOf<ScaleTypeToD3Scale<any, any, any>>` is used for the linear scale. Maybe that/something similar can be used here as well? Not sure if the anys in it can be more defined without overspecifying.

packages/app/src/components-styled/line-chart/index.tsx
```diff
@@ -132,12 +154,47 @@ export function LineChart<T extends Value>({
...
+                  isPercentage
+                )}
           </Tooltip>
         )}
       </Box>
     </Box>
   );
 }
+
+function formatStandardTooltip<T extends Value & TrendValue>(
```
VWSCoronaDashboard4, line 168: I'd name this eg `formatDefaultTooltip`

packages/app/src/components-styled/line-chart/chart.tsx
```diff
@@ -62,46 +67,50 @@ export const Chart = memo(function Chart({
...
           numTicks={NUM_TICKS}
           stroke={defaultColors.axis}
         />
         <AxisBottom
-          scale={x}
-          tickValues={x.domain()}
-          tickFormat={formatXAxis}
+          scale={xScale}
+          tickValues={xScale.domain()}
+          tickFormat={formatXAxis as AnyTickFormatter}
```
VWSCoronaDashboard4, line 113: Or perhaps

```suggestion
          tickFormat={formatXAxis as TickFormatter<Date>}
```

packages/app/src/components-styled/line-chart/chart.tsx
```diff
@@ -112,12 +121,12 @@ export const Chart = memo(function Chart({
...
         />
         <AxisLeft
-          scale={y}
+          scale={yScale}
           numTicks={4}
           hideTicks
           hideAxisLine
           stroke={defaultColors.axis}
-          tickFormat={formatYAxis}
+          tickFormat={formatYAxis as AnyTickFormatter}
```
VWSCoronaDashboard4, line 129: Or perhaps

```suggestion
          tickFormat={formatXAxis as TickFormatter<number>}
```

VWSCoronaDashboard4 (commented):
Good improvements ; some small comments/suggestions

packages/app/src/components-styled/line-chart/chart.tsx
```diff
@@ -26,23 +29,25 @@ type Benchmark = {
...
   margin?: { top: number; right: number; bottom: number; left: number };
-  formatXAxis: (x: any) => string;
-  formatYAxis: (y: any) => string;
+  formatXAxis: TickFormatter<Date>;
+  formatYAxis: TickFormatter<number>;
 };
 
+type AnyTickFormatter = (value: any) => string;
+
 export const Chart = memo(function Chart({
```
VWSCoronaDashboard10, line 51: Don't we want to name this component `LineChart`?

packages/app/src/components-styled/line-chart/index.tsx
```diff
@@ -1,85 +1,97 @@
...
-  const xDomain = useMemo(() => extent(graphData.map((d) => d.date)), [
-    graphData,
-  ]);
-  const yDomain = useMemo(() => [0, calculateYMax(graphData, signaalwaarde)], [
-    graphData,
-    signaalwaarde,
-  ]);
+  const trendData = useMemo(
+    () => getTrendData(values, metricProperties[0], timeframe),
+    [values, metricProperties, timeframe]
```
VWSCoronaDashboard10, line 82: I suggest to wrap the metricProperties inside a useMemo hook too, otherwise that array is recreated on every render which defeats the purpose of  wrapping the trendData in a  useMemo-hook.

packages/app/src/components-styled/line-chart/index.tsx
```diff
@@ -1,85 +1,97 @@
...
     showTooltip,
     hideTooltip,
-  } = useTooltip<T>();
+  } = useTooltip<T & TrendValue>();
+
+  const metricProperties = linesConfig.map((x) => x.metricProperty) as string[];
 
   const benchmark = signaalwaarde
     ? { value: signaalwaarde, label: text.common.barScale.signaalwaarde }
     : undefined;
```
VWSCoronaDashboard10, line 78: You should wrap this one too inside a useMemo hook, otherwise the Chart's memo-HOC has no effect at all.

packages/app/src/components-styled/line-chart/index.tsx
```diff
@@ -132,12 +154,47 @@ export function LineChart<T extends Value>({
...
+      'short'
+    )} - ${formatDateFromSeconds(
+      (value as WeeklyValue).week_end_unix,
+      'short'
+    )}: ${
+      isPercentage
+        ? `${formatPercentage(value.__value)}%`
+        : formatNumber(value.__value)
+    }`;
+  }
```
VWSCoronaDashboard10, line 195: nit: I think the following is much more readable:
```tsx
if (isDailyValue([value])) {
  const date = formatDateFromSeconds((value as TrendValue).__date.getSeconds());
  const valueStr = isPercentage
    ? `${formatPercentage(value.__value)}%`
    : formatNumber(value.__value);

  return `${date}: ${valueStr}`;
}

if (isWeeklyValue([value])) { // <== type narrowing makes the `value as WeeklyValue` cast unnecessery
  const dateFrom = formatDateFromSeconds(value.week_start_unix, "short");
  const dateTo = formatDateFromSeconds(value.week_end_unix, "short");
  const valueStr = isPercentage
    ? `${formatPercentage(value.__value)}%`
    : formatNumber(value.__value);

  return `${dateFrom} - ${dateTo}: ${valueStr}`;
}
```

VWSCoronaDashboard10 (commented):
It's a bit too abstract too understand what's happening, but looking good overall 👍 
Couple of suggestions though, have a look.

packages/app/src/components-styled/line-chart/chart.tsx
VWSCoronaDashboard8: I've tried that but it didn't work

packages/app/src/components-styled/line-chart/trend.tsx
VWSCoronaDashboard8: I don't find that type very understandable and since it is based on any I don't expect it to make typing stricter

packages/app/src/components-styled/line-chart/chart.tsx
VWSCoronaDashboard8: The hardcoded colors were already part of the code. This PR was not focussed on that. I agree these need to be moved to the theme at some point though.

packages/app/src/components-styled/line-chart/chart.tsx
VWSCoronaDashboard8: Chart is a (local only) member of LineChart

VWSCoronaDashboard8: maybe if would be clearer is I moved all subcomponents into a `components` sub directory

packages/app/src/components-styled/line-chart/index.tsx
VWSCoronaDashboard8: @VWSCoronaDashboard10 I know the narrowing should make the casting unnecessary in theory, but I didn't get it to work in this case.

VWSCoronaDashboard8: I will add a comment about it

PR #1288, opened 2020-12-23
Apply new line chart abstractions to all

VWSCoronaDashboard8:
## Summary

This PR applies the new line chart abstraction from #1287 to all instances


packages/app/src/pages/landelijk/sterfte.tsx
```diff
@@ -84,10 +84,12 @@ const DeceasedNationalPage: FCWithLayout<NationalPageProps> = (props) => {
           description={
             text.section_deceased_rivm.line_chart_covid_daily_description
           }
-          values={dataRivm.values.map((value) => ({
-            value: value.covid_daily,
-            date: value.date_of_report_unix,
-          }))}
+          values={dataRivm.values}
+          linesConfig={[
```
VWSCoronaDashboard4, line 88: Reading more and more of these, I tend to prefer something like:

```
lines={[
 {
  values: dataRivm.values,
  metricProperty: 'covid_daily'
 }
]}
```

packages/app/src/pages/veiligheidsregio/[code]/ziekenhuis-opnames.tsx
```diff
@@ -119,14 +118,12 @@ const IntakeHospital: FCWithLayout<ISafetyRegionData> = (props) => {
             metadata={{ source: text.bronnen.rivm }}
             title={text.linechart_titel}
             description={text.linechart_description}
-            values={data.results_per_region.values
-              .filter((x) => isFilled(x.hospital_moving_avg_per_region))
```
Comment (VWSCoronaDashboard4) [inline packages/app/src/pages/veiligheidsregio/[code]/ziekenhuis-opnames.tsx:123]: cleans up nicely here :+1:

VWSCoronaDashboard4 (approved):
Neat!

One comment that probably belongs in the previous PR, and could be tweaked later as well.

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard10 (approved):

packages/app/src/pages/landelijk/sterfte.tsx
VWSCoronaDashboard8: @VWSCoronaDashboard4 That is similar to what @VWSCoronaDashboard suggested in Slack, but I tried to explain that this doesn't make sense for us, because all lines will always share the same values series in our graphs. It's the properties that are selected and make up the lines, but the source of values is shared.

PR #1298, opened 2020-12-28
Fix IE11 issues

VWSCoronaDashboard10:
## Summary

Fix IE11 issues:
- pin visx libraries to older versions which have better IE11 support.
- pin some d3-* sub-dependencies to older versions using resolutions. Those older versions have IE11 support.
- add an IE11 shadow DOM polyfill for nextjs to display runtime errors.
- remove visx/tooltip dependency which depends on react-use-measure which doesn't seem to work with ie11. I've added a simple custom hook to replace the `useTooltip` of `visx/tooltip`.


VWSCoronaDashboard1 (approved):

packages/app/next.config.js
```diff
@@ -55,6 +55,20 @@ const nextConfig = {
       })
     );
 
+    const originalEntry = config.entry;
+
+    config.entry = async () => {
```
VWSCoronaDashboard8, line 60: I think a comment would be helpful here

VWSCoronaDashboard8 (approved):

PR #1299, opened 2020-12-28
Add mini trend tiles

VWSCoronaDashboard1:
## Summary

Adds a tile to the landelijk actueel page that shows three columns that contain a title, data driven text and a mini trend line chart displaying data for positive tests, hospital intakes and IC admissions.

## Motivation

Feature request

## Detailed design

The mini trend line chart differ quite a bit from the defaults of the tile chart. For this a small override mechanism was implemented that allows for certain parts of the line chart (such as grid rows and axes) to be defined using call backs.

The marker (rounded dot) displayed on mouseover was removed from the tooltip and given its own logic. It now also optionally supports the rendering of a vertical line with an extra label.

LineChart now correctly calculates its yDomain.

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added


packages/app/src/components-styled/line-chart/components/chart.tsx
```diff
@@ -175,3 +217,66 @@ export const Chart = memo(function Chart({
...
+        result
+      ) : (
+        <GridRows {...(callbackInfo.configuration as any)} />
+      );
+    }
+    case 'AxisBottom':
+      return result !== undefined ? (
+        result
+      ) : (
+        <AxisBottom {...(callbackInfo.configuration as any)} />
```
VWSCoronaDashboard8, line 238: You can probably avoid the cast to any by adding a type guard (similar to isWeeklyData), but I don't think it's worth implementing now.

packages/app/src/components-styled/line-chart/index.tsx
```diff
@@ -70,6 +87,17 @@ export function LineChart<T extends Value>({
     hideTooltip,
   } = useTooltip<T & TrendValue>();
 
+  const [markerInfo, setMarkerInfo] = useState<
```
VWSCoronaDashboard8, line 90: nit; I don't like the word "info" so much. It makes it sound like the data is informative and somehow extra, but this configuration is defining all properties of the marker. Maybe just `[marker, setMarker]` will do? (see comment below also)

packages/app/src/components-styled/line-chart/index.tsx
```diff
@@ -166,6 +204,18 @@ export function LineChart<T extends Value>({
...
           </Tooltip>
         )}
+
+        {markerInfo && (
+          <Marker
+            x={markerInfo.xPosition}
+            y={markerInfo.yPosition}
+            data={markerInfo.data}
+            height={markerInfo.height}
+            margins={markerInfo.margins}
```
VWSCoronaDashboard8, line 214: nit; If you match the properties of the Marker component, then you can use the spread operator and the link is more clear. So xPosition and yPosition would become x and y, the rest can stay the same it seems.

packages/app/src/domain/topical/mini-trend-tile.tsx
```diff
@@ -0,0 +1,144 @@
...
+import text from '~/locale';
+import { formatNumber } from '~/utils/formatNumber';
+
+type MiniTrendTileProps<T> = {
+  icon: JSX.Element;
+  title: string;
+  text: ReactNode;
+  trendData: T[];
+  metricProperty: keyof T;
+} & BoxProps;
```
VWSCoronaDashboard8, line 23: Do we need to expose the box props? I don't remember seeing that on other components.

packages/app/src/domain/topical/mini-trend-tile.tsx
```diff
@@ -0,0 +1,144 @@
...
+            <ParentSize>
+              {(parent) => (
+                <LineChart
+                  width={parent.width}
+                  timeframe="5weeks"
+                  values={trendData}
+                  linesConfig={[{ metricProperty }]}
+                  componentCallback={componentCallback}
+                  showMarkerLine={true}
+                  formatTooltip={(value: T & TrendValue) =>
```
VWSCoronaDashboard8, line 48: This input type annotation should be redundant I think.

packages/app/src/pages/index.tsx
```diff
@@ -34,69 +37,122 @@ const Home: FCWithLayout<IHomeData> = (data) => {
...
-            <>
-              <span
-                dangerouslySetInnerHTML={{
-                  __html: text.risiconiveaus.selecteer_toelichting,
-                }}
+      <Tile>De actuele situatie in Nederland</Tile>
+      <Tile>
+        <Box display="flex" flexDirection="row">
+          <MiniTrendTile
+            flex="1 1 33%"
```
VWSCoronaDashboard8, line 50: I think it would be a little cleaner if you created a MiniTrendTileLayout / Container or something which handles all the flexbox stuff, instead of forwarding the flex property to the Box component inside the MiniTrendTile component.

It could even be a generic 3 column layout component maybe.

VWSCoronaDashboard8 (approved):

packages/app/src/components-styled/line-chart/components/chart.tsx
```diff
@@ -175,3 +217,66 @@ export const Chart = memo(function Chart({
...
+        numTicks: number;
+        stroke: string;
+      };
+    }
+  | {
+      type: 'AxisBottom';
+      configuration: {
+        scale: ScaleTime<number, number>;
+        tickValues: any[];
+        tickFormat: AnyTickFormatter;
```
VWSCoronaDashboard10, line 264: is it possible to get rid of any?

packages/app/src/components-styled/line-chart/components/chart.tsx
```diff
@@ -175,3 +217,66 @@ export const Chart = memo(function Chart({
...
+  componentCallback: ComponentCallbackFunction
+) {
+  const result = componentCallback(callbackInfo);
+  switch (callbackInfo.type) {
+    case 'GridRows': {
+      return result !== undefined ? (
+        result
+      ) : (
+        <GridRows {...(callbackInfo.configuration as any)} />
+      );
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: Actually, I implemented it this way to be able to turn a certain component 'off'. If the callback return `null`, simply nothing is rendered. If it returns `undefined`, it renders the default.

packages/app/src/components-styled/line-chart/components/chart.tsx
VWSCoronaDashboard1: This is actually a clash of versions still. I think, perhaps it can be solved by using the same resolutions trick for the typings as well

packages/app/src/domain/topical/mini-trend-tile.tsx
VWSCoronaDashboard1: I've added the box props in order for the `MiniTrendTile` to be able to receive styling for its root component (a `Box`).
This way, on the page I can add things like `flex="1 1 33%"`, I didn't want to hardcode that inside the component.

VWSCoronaDashboard1: Oh wait, I see your other comment about the layout component. That makes sense now.

packages/app/src/domain/topical/mini-trend-tile.tsx
```diff
@@ -0,0 +1,144 @@
...
+                    formatNumber(value.__value)
+                  }
+                  margin={{
+                    ...defaultMargin,
+                    left: 0,
+                  }}
+                />
+              )}
+            </ParentSize>
```
VWSCoronaDashboard10, line 57: What would you @VWSCoronaDashboard1 and @VWSCoronaDashboard8 think of a single function call to render child components, which receives a single context/config object? I think that would be a bit more clear on the outside what's happening ..

We can still provide component-specific default props on that context object but overwrite them with any of the information available on the chart context.

It could look something like this:

```tsx
<LineChart
  width={parent.width}
  timeframe="5weeks"
  values={trendData}
  linesConfig={[{ metricProperty }]}
  componentCallback={componentCallback}
  showMarkerLine={true}
  formatTooltip={(value: T & TrendValue) =>
    formatNumber(value.__value)
  }
  margin={{
    ...defaultMargin,
    left: 0,
  }}
>
  {context => {
    const xDomain = context.xScale.domain()
    const yDomain = context.yScale.domain()

    return (
      <>
        <Chart.GridRows
          {...context.defaultProps.gridRows}
          tickValues={[0, last(yDomain) / 2, last(yDomain)]}
        />

        <Chart.AxisBottom
          {...context.defaultProps.axisBottom}
          tickLabelProps={(value, index) => ({
            ...context.defaultProps.axisBottom.tickLabelProps(value, index),
            textAnchor: value === yDomain[0] ? 'start' 'end',
            dx: 0
          })}
          tickFormat={(value) => {
            if (value === yDomain[0]) {
              return context.defaultProps.axisBottom.tickFormat(value)
            }
            
            const days = Math.floor(
              (Date.now() / 1000 - value.valueOf() / 1000) / DAY_IN_SECONDS
            );

            if (days < 1) return text.common.vandaag;
            if (days < 2) return text.common.gisteren;
            return context.defaultProps.axisBottom.tickFormat(value)
          }
          }
        />

        <Chart.AxisLeft
          {...context.defaultPRops.axisLeft}
          tickValues={last(yDomain)}
          tickLabelProps={(value, index) => ({
            ...context.defaultProps.axisLeft.tickLabelProps(value, index),
            textAnchor: 'start',
            dx: 10,
            dy: -6
          })}
        />
      </>
    )
  }}
</Chart>
```

packages/app/src/domain/topical/mini-trend-tile.tsx
```diff
@@ -0,0 +1,144 @@
...
+                  linesConfig={[{ metricProperty }]}
+                  componentCallback={componentCallback}
+                  showMarkerLine={true}
+                  formatTooltip={(value: T & TrendValue) =>
+                    formatNumber(value.__value)
+                  }
+                  margin={{
+                    ...defaultMargin,
+                    left: 0,
+                  }}
```
VWSCoronaDashboard10, line 54: nit; I think it would look better to have 4 separate props for the margins

packages/app/src/pages/index.tsx
```diff
@@ -34,69 +37,122 @@ const Home: FCWithLayout<IHomeData> = (data) => {
...
+        header={text.quick_links.header}
+        links={[
+          { href: '/landelijk', text: text.quick_links.links.nationaal },
+          {
+            href: '/veiligheidsregio',
+            text: text.quick_links.links.veiligheidsregio,
+          },
+          { href: '/gemeentes', text: text.quick_links.links.gemeente },
+        ]}
+      ></QuickLinks>
```
VWSCoronaDashboard10, line 155: I think something went wrong with a merge conflict?

packages/app/src/domain/topical/mini-trend-tile.tsx
VWSCoronaDashboard10: Ah I guess one of the advantages of your current implementation is probably that all axis/rows/etc-components always get rendered which will reduce repetition of code..

packages/app/src/components-styled/line-chart/index.tsx
```diff
@@ -70,6 +87,17 @@ export function LineChart<T extends Value>({
...
+  const [markerInfo, setMarkerInfo] = useState<
+    | {
+        xPosition: number;
+        yPosition: number;
+        height: number;
+        data: T;
+        margins: ChartMargins;
+      }
+    | undefined
+  >(undefined);
```
VWSCoronaDashboard10, line 99: type `undefined` can be inferred by the undefined initial value:

```suggestion
  const [markerInfo, setMarkerInfo] = useState<{
    xPosition: number;
    yPosition: number;
    height: number;
    data: T;
    margins: ChartMargins;
  }>();
```

packages/app/src/components-styled/line-chart/components/marker.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { colors } from '~/style/theme';
+import { formatDateFromMilliseconds } from '~/utils/formatDate';
+import { TrendValue, Value } from '../helpers';
+import { ChartMargins } from './chart';
+
+type PointProps = {
+  indicatorColor: string;
+};
+
+const Label = styled.div`
```
VWSCoronaDashboard10, line 12: Why didn't you style any of these components in this file it with SS's `css()` function?

packages/app/src/components-styled/line-chart/index.tsx
```diff
@@ -41,12 +50,16 @@ export type LineChartProps<T> = {
...
   timeframe?: TimeframeOption;
   signaalwaarde?: number;
-  formatTooltip?: (value: T) => React.ReactNode;
+  formatTooltip?: (value: T & TrendValue) => React.ReactNode;
   formatXAxis?: TickFormatter<Date>;
   formatYAxis?: TickFormatter<number>;
   showFill?: boolean;
   valueAnnotation?: string;
   isPercentage?: boolean;
+  componentCallback?: ComponentCallbackFunction;
```
VWSCoronaDashboard10, line 59: What would you think of the name `getComponentProps`?

packages/app/src/components-styled/line-chart/index.tsx
```diff
@@ -41,12 +50,16 @@ export type LineChartProps<T> = {
...
+  showMarkerLine?: boolean;
+  formatMarkerLabel?: (value: T) => string;
```
VWSCoronaDashboard10, line 61: Can't all marker-props also be set using the componentCallback?

packages/app/src/pages/index.tsx
VWSCoronaDashboard10: Or maybe the diff on github was f*cked up :-)

packages/app/src/components-styled/line-chart/components/chart.tsx
```diff
@@ -61,6 +84,7 @@ export const Chart = memo(function Chart({
   benchmark,
   formatXAxis,
   formatYAxis,
+  componentCallback = (_callbackInfo: ComponentCallbackInfo) => undefined,
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: Otherwise the linter will complain

packages/app/src/components-styled/line-chart/components/chart.tsx
```diff
@@ -1,18 +1,36 @@
...
+import { NumberValue, ScaleLinear, ScaleTime } from 'd3-scale';
+import { memo, ReactNode, useCallback } from 'react';
 import { colors } from '~/style/theme';
 import { TrendValue } from '../helpers';
 import { Trend, TrendType } from './trend';
 
 const NUM_TICKS = 3;
 
-export const defaultMargin = { top: 10, right: 20, bottom: 30, left: 30 };
+export type ChartMargins = {
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: Hehe, you're right

packages/app/src/components-styled/line-chart/components/chart.tsx
```diff
@@ -175,3 +217,66 @@ export const Chart = memo(function Chart({
...
+  callbackInfo: ComponentCallbackInfo,
+  componentCallback: ComponentCallbackFunction
+) {
+  const result = componentCallback(callbackInfo);
+  switch (callbackInfo.type) {
+    case 'GridRows': {
+      return result !== undefined ? (
+        result
+      ) : (
+        <GridRows {...(callbackInfo.configuration as any)} />
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: Yea, like I said in my other comment, this is a typing clash. Perhaps I'll be able to fix it using the same resolution trick.

packages/app/src/components-styled/line-chart/components/chart.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard10: Would the linter complain if you remove the argument?
```suggestion
  componentCallback = () => undefined,
```

packages/app/src/components-styled/line-chart/index.tsx
VWSCoronaDashboard1: Well, it doesn't just get the props, it returns an actual instance of the component (this way you can override GridRows, for example, with an entirely different component)

packages/app/src/components-styled/line-chart/components/marker.tsx
VWSCoronaDashboard1: I thought we only require to use the `css()` function when we're using theme values. Am I missing something?

packages/app/src/components-styled/line-chart/components/chart.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard1: Now that the props are typed to `ComponentProps<...>` the `any` cast can indeed be removed

packages/app/src/components-styled/line-chart/components/chart.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard1: Indeed, that works.

packages/app/src/components-styled/line-chart/components/chart.tsx
VWSCoronaDashboard1: This `any` has been removed now

packages/app/src/domain/topical/mini-trend-tile-layout.tsx
```diff
@@ -0,0 +1,23 @@
+import { Children, ReactNode } from 'react';
+import { Box } from '~/components-styled/base';
+import { Tile } from '~/components-styled/tile';
+
+type MiniTrendTileLayoutProps = {
+  children: ReactNode;
+};
+
+export function MiniTrendTileLayout({ children }: MiniTrendTileLayoutProps) {
+  const miniTrendTiles = Children.toArray(children);
```
VWSCoronaDashboard8, line 10: nit-nitpick;

```suggestion
  const tiles = Children.toArray(children);
```

I think the parent components already provides enough context so tiles would suffice.

packages/app/src/domain/topical/mini-trend-tile.tsx
```diff
@@ -20,12 +19,12 @@ type MiniTrendTileProps<T> = {
...
   metricProperty: keyof T;
-} & BoxProps;
+};
 
 export function MiniTrendTile<T extends Value>(props: MiniTrendTileProps<T>) {
   const { icon, title, text, trendData, metricProperty, ...boxProps } = props;
   return (
-    <Box {...(boxProps as any)} display="flex" flexDirection="column">
+    <Box display="flex" flexDirection="column">
       <Box flexDirection="row" display="flex">
```
VWSCoronaDashboard8, line 28: nitnit; since row is the default you can omit it.

VWSCoronaDashboard8 (approved):

packages/app/src/components-styled/line-chart/index.tsx
VWSCoronaDashboard1: Perhaps, but the callback originates from the underlying chart component, so combining these might not be completely arbitrary, so for now, I'd like to keep it this way.

packages/app/src/components-styled/line-chart/components/marker.tsx
VWSCoronaDashboard10: Not that I was aware of .. A regex search in our codebase for ```styled.*` ``` yields only some results coming from VWSCoronaDashboard13, which indicates we missed those styles by that PR. I'd suggest to convert it to the SS `css()` implementation.

PR #1307, opened 2020-12-30
Add multiple line support to line chart

VWSCoronaDashboard1:
## Summary

image

## Motivation

Feature request

## Detailed design

Better separation of concerns is created between the components. There is now a base component called ChartAxes that only paints the axes (hence the name). LineChart then renders its lines as children into this base component.
The line configs are now used to determine the line colors, as well as the legend labels and line style (solid or dashed for now).
A legend has been added.
Support for multiple markers has been added along with a vertical tracker bar. The tooltip is painted above the nearest marker.

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

The left axis column needs to be sized according to its content, this isn't working properly yet.

### Governance

- [ ] Documentation is added

### DIFF
--- a/packages/app/src/components-styled/line-chart/components/marker.tsx
+++ b/packages/app/src/components-styled/line-chart/components/marker.tsx
@@ -2,6 +2,8 @@ import styled from 'styled-components';
 import { Text } from '~/components-styled/typography';
 import { colors } from '~/style/theme';
 import { formatDateFromMilliseconds } from '~/utils/formatDate';
+import { HoverPoint } from '..';
+import { TrendValue, Value } from '../helpers';
 import { ChartPadding } from './chart-axes';
 
 type ColorProps = {
@@ -9,12 +11,10 @@ type ColorProps = {
 };
 
 const Label = styled.div`
-  pointer-events: none;
   background-color: white;
 `;
 
 const DottedLine = styled.div<ColorProps>`
-  pointer-events: none;
   width: 1px;
   border-left-width: 1px;
   border-left-style: dashed;
@@ -22,7 +22,6 @@ const DottedLine = styled.div<ColorProps>`
 `;
 
 const Point = styled.div<ColorProps>`
-  pointer-events: none;
   position: relative;
   height: 18px;
   width: 18px;
@@ -43,7 +42,7 @@ const Point = styled.div<ColorProps>`
     position: absolute;
     height: 18px;
     width: 18px;
-    transform: translate(0, -50%);
+    transform: translate(0, -45%);
     border-radius: 50%;
     background: ${(props) => props.indicatorColor || 'black'};
     opacity: 0.2;
@@ -52,27 +51,26 @@ const Point = styled.div<ColorProps>`
 
 const MarkerContainer = styled.div`
   transform: translate(-50%, 0);
-  pointer-events: none;
   position: absolute;
   display: flex;
   flex-direction: column;
   align-items: center;
   flex-grow: 0;
   flex-shrink: 0;
-  min-width: 5em;
-  background-color: rgba(0, 0, 255, 0.05);
+  min-width: 26px;
+  background-color: rgba(0, 0, 0, 0.03);
 `;
 
-type MarkerProps = {
-  data: HoverPoint[];
+type MarkerProps<T> = {
+  data: HoverPoint<T>[];
   height: number;
   primaryColor?: string;
   padding: ChartPadding;
   showLine: boolean;
-  formatLabel?: (data: ChartValue) => string;
+  formatLabel?: (data: T & Value & TrendValue) => string;
 };
 
-export function Marker(props: MarkerProps) {
+export function Marker<T>(props: MarkerProps<T>) {
   const {
     primaryColor = colors.data.primary,
     data,
@@ -90,8 +88,8 @@ export function Marker(props: MarkerProps) {
         height: height - (padding.top + padding.bottom),
       }}
     >
-      {data.map((d) => (
-        <Point indicatorColor={primaryColor} style={{ top: d.y }} />
+      {data.map((d, index) => (
+        <Point indicatorColor={d.color} style={{ top: d.y - index * 18 }} />
       ))}
 
       {showLine && (
@@ -115,6 +113,6 @@ export function Marker(props: MarkerProps) {
   );
 }
 
-function defaultFormatLabel(data: ChartValue): string {
+function defaultFormatLabel<T>(data: T & Value & TrendValue): string {
   return formatDateFromMilliseconds(data.__date.getTime());
 }
--- a/packages/app/src/components-styled/line-chart/components/marker.tsx
+++ b/packages/app/src/components-styled/line-chart/components/marker.tsx
@@ -2,7 +2,7 @@ import styled from 'styled-components';
 import { Text } from '~/components-styled/typography';
 import { colors } from '~/style/theme';
 import { formatDateFromMilliseconds } from '~/utils/formatDate';
-import { ChartPadding, ChartValue, HoverPoint } from './chart';
+import { ChartPadding } from './chart-axes';
 
 type ColorProps = {
   indicatorColor: string;
@@ -93,6 +93,7 @@ export function Marker(props: MarkerProps) {
       {data.map((d) => (
         <Point indicatorColor={primaryColor} style={{ top: d.y }} />
       ))}
+
       {showLine && (
         <>
           <DottedLine
--- a/packages/app/src/components-styled/line-chart/components/marker.tsx
+++ b/packages/app/src/components-styled/line-chart/components/marker.tsx
@@ -1,0 +1,119 @@
+import styled from 'styled-components';
+import { Text } from '~/components-styled/typography';
+import { colors } from '~/style/theme';
+import { formatDateFromMilliseconds } from '~/utils/formatDate';
+import { ChartPadding, ChartValue, HoverPoint } from './chart';
+
+type ColorProps = {
+  indicatorColor: string;
+};
+
+const Label = styled.div`
+  pointer-events: none;
+  background-color: white;
+`;
+
+const DottedLine = styled.div<ColorProps>`
+  pointer-events: none;
+  width: 1px;
+  border-left-width: 1px;
+  border-left-style: dashed;
+  border-left-color: ${(props) => props.indicatorColor || 'black'};
+`;
+
+const Point = styled.div<ColorProps>`
+  pointer-events: none;
+  position: relative;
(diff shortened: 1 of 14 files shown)


packages/app/src/components-styled/line-chart/components/trend.tsx
```diff
@@ -20,92 +22,65 @@ export type TrendProps = {
...
-      onHover(
-        event,
-        pointData,
-        xScale(pointData.__date),
-        yScale(pointData.__value)
-      );
-    },
-    [onHover, yScale, xScale, trend, bisect]
-  );
+  const internalOnHover = (
```
VWSCoronaDashboard8, line 45: I keep the pattern of onSomething (external event callback passed in via props) and handleSomething (internal event handling). Then I think you don't need to be more explicit. And for definitions prefer named function declarations to help debugging especially with callbacks.

So I would suggest writing `function handleHover(...`

packages/app/src/components-styled/line-chart/helpers.ts
```diff
@@ -32,9 +35,9 @@ export function isDailyValue(timeSeries: Value[]): timeSeries is DailyValue[] {
   return firstValue.date_of_report_unix !== undefined;
 }
 
-export function isWeeklyValue(
-  timeSeries: Value[]
-): timeSeries is WeeklyValue[] {
+export function isWeeklyValue<T>(
```
VWSCoronaDashboard8, line 38: Maybe <T extends Value> makes more sense here? Then T can be used in place of Value. Although I must admit I don't fully understand the difference

packages/app/src/components-styled/line-chart/helpers.ts
```diff
@@ -52,20 +55,15 @@ export function isWeeklyValue(
  * values.
  */
 export function calculateYMax(
-  values: Value[],
-  metricProperties: string[],
+  values: TrendValue[][],
```
VWSCoronaDashboard8, line 58: I'm surprised that values is now an array of arrays. I was under the impression that all trend lines are always drawn from the same time series values array, and thus present different properties of the same value object.

packages/app/src/components-styled/line-chart/helpers.ts
```diff
@@ -117,12 +115,20 @@ export type TrendValue = {
...
 const timestampToDate = (d: number) => new Date(d * 1000);
 
-export function getTrendData(
-  values: Value[],
-  valueKey: string,
+export function getTrendData<T>(
+  values: (T & Value)[],
+  valueKeys: NumberProperty<T>[],
+  timeframe: TimeframeOption
+): (T & TrendValue & Value)[][] {
```
VWSCoronaDashboard8, line 122: I wonder if all these more complicated types were really necessary. I think this is related to my previous question/comment.

packages/app/src/components-styled/line-chart/index.tsx
```diff
@@ -26,27 +41,43 @@ const formatXAxis = (date: Date) =>
 const formatYAxisFn = (y: number) => y.toString();
 const formatYAxisPercentageFn = (y: number) => `${formatPercentage(y)}%`;
 
+// This type limits the allowed property names to those with a number type,
+// so its like keyof T, but filtered down to only the appropriate properties.
+export type NumberProperty<T> = {
+  [K in keyof T]: T[K] extends number | null ? K : never;
+}[keyof T];
```
VWSCoronaDashboard8, line 48: Fancy 😄

packages/app/src/components-styled/line-chart/index.tsx
```diff
@@ -26,27 +41,43 @@ const formatXAxis = (date: Date) =>
...
-      color?: string;
-    }
-  ];
+  linesConfig: LineConfig<T>[];
   width?: number;
   height?: number;
   timeframe?: TimeframeOption;
   signaalwaarde?: number;
-  formatTooltip?: (value: T) => React.ReactNode;
+  formatTooltip?: (value: (T & TrendValue & Value)[]) => React.ReactNode;
```
VWSCoronaDashboard8, line 64: What's the difference between a Value and a TrendValue? T is the original data but I can see that this is maybe not enough to render the tooltip, because you might want to know exactly which of the properties is being hovered, which can be found in the Value properties __value and __date, but what is TrendValue adding?

packages/app/src/components-styled/line-chart/index.tsx
```diff
@@ -83,44 +124,130 @@ export function LineChart<T extends Value>({
...
-      yPosition
+      hide: boolean,
+      hoverPoints?: HoverPoint<T>[],
+      nearestPoint?: HoverPoint<T>
     ) => {
-      if (event.type === 'mouseleave') {
+      if (hide) {
         hideTooltip();
-      } else {
+        setMarkerProps(undefined);
```
VWSCoronaDashboard8, line 185:
```suggestion
        setMarkerProps();
```

This would be considered the same

packages/app/src/components-styled/line-chart/index.tsx
```diff
@@ -83,44 +124,130 @@ export function LineChart<T extends Value>({
...
+
+      const hoverPoints = trendsList
+        .map((trends, index) => {
+          const trendValue = bisect(trends, point.x, xScale);
+          return trendValue
+            ? {
+                data: trendValue,
+                color: linesConfig[index].color,
+              }
+            : undefined;
```
VWSCoronaDashboard8, line 231:
```suggestion
          return bisect(trends, point.x, xScale) ? {
                data: trendValue,
                color: linesConfig[index].color,
              }
            : undefined;
```

packages/app/src/components-styled/line-chart/index.tsx
```diff
@@ -148,39 +274,76 @@ export function LineChart<T extends Value>({
...
+          </Box>
+        )}
       </Box>
     </Box>
   );
 }
 
 function formatDefaultTooltip<T extends Value & TrendValue>(
-  value: T,
+  values: T[],
```
VWSCoronaDashboard8, line 336: Again, surprised to see that this is an array now

VWSCoronaDashboard8 (commented):
I have some questions, but looking good overall! 👍

packages/app/src/components-styled/line-chart/helpers.ts
VWSCoronaDashboard1: Indeed, I added `T extends Value` to the `getTrendData` generic. That way I can use `Value` as the base type in all the other functions, cleans up a lot.

packages/app/src/components-styled/line-chart/index.tsx
VWSCoronaDashboard1: Thanks to Stack Overflow ;) Don't ask me to explain it ;)

packages/app/src/components-styled/line-chart/index.tsx
VWSCoronaDashboard1: `TrendValue` actually has the __value and __date properties. Value is merely a union of `DailyValue` and `WeeklyValue`.

packages/app/src/components-styled/line-chart/index.tsx
VWSCoronaDashboard1: The compiler disagrees I'm afraid :)

packages/app/src/components-styled/line-chart/index.tsx
VWSCoronaDashboard1: I don't understand this suggestion. The result of bisect is assigned to a property of the result instance when its not undefined.
In your suggestion, where is `trendValue` actually assigned?

packages/app/src/components-styled/line-chart/index.tsx
VWSCoronaDashboard1: The tooltip displays the values of the hovered points of all the current lines. Therefore it's an array.
What would the alternative be in your mind?

packages/app/src/components-styled/line-chart/index.tsx
VWSCoronaDashboard1: But, because T extends Value, I've simplified it `T & TrendValue` now

packages/app/src/components-styled/line-chart/index.tsx
VWSCoronaDashboard8: Yeah never mind, the effect is not the same :)

packages/app/src/components-styled/line-chart/index.tsx
VWSCoronaDashboard8: All of the data from all the lines on any point of hover are I think available in type T, because each line is rendered from a property on that object. That is at least how I imagined the data structure to be, but maybe I'm overlooking something

VWSCoronaDashboard1: Yes, but the data you're receiving here originates from the lines you're hovering. Based on the line config an array of data values for each line, in `handleHover` the data points for each line is determined based on the x/y coordinates. So therefore you receive one datapoint for each line that is drawn. So, the tooltip receives exactly those values.

VWSCoronaDashboard8 (approved):

packages/app/src/components-styled/line-chart/components/trend.tsx
```diff
@@ -20,92 +22,65 @@ export type TrendProps = {
...
+
+  const dashes = style === 'dashed' ? 4 : undefined;
 
   return (
     <>
-      {type === 'line' && (
-        <LinePath
+      {type === 'area' && (
+        <AreaClosed
+          style={{ pointerEvents: 'all' }}
```
VWSCoronaDashboard4, line 59: isn't this the default?

packages/app/src/components-styled/line-chart/components/marker.tsx
```diff
@@ -0,0 +1,148 @@
...
+  background-color: white;
+`;
+
+const DottedLine = styled.div<ColorProps>`
+  pointer-events: none;
+  width: 1px;
+  border-left-width: 1px;
+  border-left-style: dashed;
+  border-left-color: ${(props) => props.indicatorColor || 'black'};
+  position: relative;
```
VWSCoronaDashboard4, line 30: why is position absolute needed here? there is no top/left property set?

removing it's parent `position: relative` might clear up things, as multiple layout patterns are now in use (flex)

packages/app/src/components-styled/line-chart/helpers.ts
VWSCoronaDashboard4: what is the verdict on this?

VWSCoronaDashboard4 (approved):

PR #1315, opened 2021-01-04
Add sanity integration for FAQ page

VWSCoronaDashboard:
## Summary

This PR shows how to retrieve API data from Sanity for 1 page.

## Detailed design

There's 2 things to note here:

1. We're placing the GROQ query directly with the pages. Do you all agree?

E.g.

```
const faqQuery = groq`
  *[_type == 'veelgesteldeVragen']
  {
    ...
  }[0]
`;
```

2. Localization works by using field localization. The example is taken directly from https://github.com/sanity-io/sanity-template-translation-examples/tree/78b5a5cf994a1ae083584d6f17bf8889a7818b93

## Unresolved questions

DO NOT MERGE THIS YET


packages/app/src/lib/sanity.ts
```diff
@@ -0,0 +1,78 @@
...
+   * data always (potentially slightly slower and a bit more expensive).
+   * Authenticated request (like preview) will always bypass the CDN
+   **/
+};
+
+/**
+ * Set up a helper function for generating Image URLs with only the asset reference data in your documents.
+ * Read more: https://www.sanity.io/docs/image-url
+ **/
+export const urlFor = (source: any) =>
```
VWSCoronaDashboard8, line 32: Why is this any? Seems like this would apply to a very narrow set of types

packages/app/src/lib/sanity.ts
```diff
@@ -0,0 +1,78 @@
...
+});
+
+// Helper function for easily switching between normal client and preview client
+export const getClient = (usePreview: boolean) =>
+  usePreview ? previewClient : sanityClient;
+
+// Helper function for using the current logged in user account
+export const useCurrentUser = createCurrentUserHook(config);
+
+export function localize(value: any, languages: any): any {
```
VWSCoronaDashboard8, line 62: Can't you at least specify what will be returned? And maybe type value to a string or something?

Languages could maybe be `Record<string, unknown>`

packages/app/src/pages/veelgestelde-vragen.tsx
```diff
@@ -3,29 +3,46 @@ import Head from 'next/head';
...
 interface StaticProps {
   props: VeelgesteldeVragenProps;
 }
 
 interface VeelgesteldeVragenProps {
-  text: TALLLanguages;
+  text: {
+    title: string;
+    description: string;
+    content: Array<{ content: Array<any>, title: string}>
```
VWSCoronaDashboard8, line 25: Prefer to use `[]` syntax over `Array<>`. And can we get rid of any here? Also, an array called `content` with children that have another prop called `content` is not very clear.

packages/app/src/pages/veelgestelde-vragen.tsx
```diff
@@ -3,29 +3,46 @@ import Head from 'next/head';
...
+  };
   lastGenerated: string;
 }
 
-export async function getStaticProps(): Promise<StaticProps> {
-  const text = parseMarkdownInLocale((await import('../locale/index')).default);
+const faqQuery = groq`
```
VWSCoronaDashboard8, line 30: If this is a pattern we'll use on every page, I would suggest to standardize the name and just call it `query`. The faq is already implied by the context of the page. Same for faqData and faqList below.

packages/app/src/pages/veelgestelde-vragen.tsx
```diff
@@ -50,23 +67,20 @@ const Verantwoording: FCWithLayout<VeelgesteldeVragenProps> = (props) => {
...
+                const id = getSkipLinkId(item.title);
                 return (
-                  <Collapsable key={id} id={id} summary={item.vraag}>
-                    <div
-                      dangerouslySetInnerHTML={{
-                        __html: item.antwoord,
-                      }}
-                    />
+                  <Collapsable key={id} id={id} summary={item.title}>
+                    <PortableText blocks={ item.content } />
```
VWSCoronaDashboard8, line 78: Is prettier formatting not working?

packages/app/src/lib/sanity.ts
VWSCoronaDashboard: The code is copy pasted from an example. We'll update it to actual types later?

packages/app/src/pages/veelgestelde-vragen.tsx
VWSCoronaDashboard: Renamed `content` to `questions`.

VWSCoronaDashboard: We can not get rid of any.

packages/app/src/pages/veelgestelde-vragen.tsx
VWSCoronaDashboard: I'm thinking of moving this to src/lib/api.ts eventually, so I'm opting to keep them uniquely named for now.

PR #1328, opened 2021-01-05
Add new multi line chart to region page

VWSCoronaDashboard1:
## Summary

The region page still showed the highcharts based multiline chart. This is replaced with the new implementation and the highcharts component is removed from the codebase.

## Motivation

Cleanup

## Detailed design

Summary says it all

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added


packages/app/src/pages/veiligheidsregio/[code]/positief-geteste-mensen.tsx
```diff
@@ -267,41 +268,64 @@ const PostivelyTestedPeople: FCWithLayout<ISafetyRegionData> = (props) => {
...
-              week: {
-                start: value.week_start_unix,
-                end: value.week_end_unix,
-              },
-            })),
-          ]}
+          showFill={false}
+          showLegend
+          padding={{
+            ...defaultPadding,
```
Comment (VWSCoronaDashboard8) [inline packages/app/src/pages/veiligheidsregio/[code]/positief-geteste-mensen.tsx:278]: It seems strange that the default padding is imported, and then passed back in via props. Looks like this could be solved just as well by applying the default padding internally to any direction where padding isn't specified in the prop.

packages/app/src/pages/veiligheidsregio/[code]/positief-geteste-mensen.tsx
```diff
@@ -267,41 +268,64 @@ const PostivelyTestedPeople: FCWithLayout<ISafetyRegionData> = (props) => {
...
+                <br />
+                <span
+                  style={{
+                    height: '0.5em',
+                    width: '0.5em',
+                    backgroundColor: colors.data.primary,
+                    borderRadius: '50%',
+                    display: 'inline-block',
+                  }}
+                ></span>{' '}
```
Comment (VWSCoronaDashboard10) [inline packages/app/src/pages/veiligheidsregio/[code]/positief-geteste-mensen.tsx:324]:
```suggestion
                />{' '}
```

Comment (VWSCoronaDashboard8) [inline packages/app/src/pages/veiligheidsregio/[code]/positief-geteste-mensen.tsx:324]: I think we'll have to look at this later, to see if we can create some reusable building blocks for tooltips like this, because this is very specific and not something you'd want to specify for every instance of the chart

VWSCoronaDashboard8 (approved):
My comments are about refactorings that can be picked up later as far as I'm concerned.

packages/app/src/pages/veiligheidsregio/[code]/positief-geteste-mensen.tsx
```diff
@@ -267,41 +268,64 @@ const PostivelyTestedPeople: FCWithLayout<ISafetyRegionData> = (props) => {
...
-            ggdValues.map((value) => ({
-              value: value.infected,
-              date: value.week_unix,
-              week: {
-                start: value.week_start_unix,
-                end: value.week_end_unix,
-              },
-            })),
-          ]}
+          showFill={false}
```
Comment (VWSCoronaDashboard10) [inline packages/app/src/pages/veiligheidsregio/[code]/positief-geteste-mensen.tsx:275]: Might be unrelated to this PR, but should we flip this prop to `hideFill` if it's shown by default ..?

VWSCoronaDashboard10 (approved):

packages/app/src/pages/veiligheidsregio/[code]/positief-geteste-mensen.tsx
Comment (VWSCoronaDashboard8) [reply packages/app/src/pages/veiligheidsregio/[code]/positief-geteste-mensen.tsx:275]: Or keep the name but make false the default. I agree that passing false like this doesn't look quite right.

packages/app/src/pages/veiligheidsregio/[code]/positief-geteste-mensen.tsx
Comment (VWSCoronaDashboard1) [reply packages/app/src/pages/veiligheidsregio/[code]/positief-geteste-mensen.tsx]: I was thinking the same thing earlier, it was an easy thing to fix so I've made the change

packages/app/src/components-styled/line-chart/index.tsx
```diff
@@ -88,9 +88,17 @@ export function LineChart<T extends Value>({
...
+  padding = {},
   showLegend = false,
 }: LineChartProps<T>) {
+  const chartPadding: ChartPadding = useMemo(
+    () => ({
+      ...defaultPadding,
+      ...padding,
+    }),
+    [padding]
+  );
```
VWSCoronaDashboard10, line 100: I'd opt for renaming this const to `padding` and rename the incoming prop to something like paddingOverride/overridePadding .. eg:

```suggestion
  padding: overridePadding = {},
  showLegend = false,
}: LineChartProps<T>) {
  const padding: ChartPadding = useMemo(
    () => ({
      ...defaultPadding,
      ...overridePadding,
    }),
    [overridePadding]
  );
```

packages/app/src/components-styled/line-chart/index.tsx
```diff
@@ -88,15 +88,15 @@ export function LineChart<T extends Value>({
   isPercentage,
   showMarkerLine = false,
   formatMarkerLabel,
-  padding = {},
+  padding: overridePadding = {},
```
VWSCoronaDashboard10, line 91: You might have missed my second comment, but I think we should not have the default padding over here. It's not necessary anymore, but worse: it will trigger the padding `useMemo`-hook on every render when the padding prop is left empty.

VWSCoronaDashboard1: Fixed

PR #1330, opened 2021-01-05
Fix y-axis formatting and spacing

VWSCoronaDashboard4:
## Summary

Small Y-axis tweaks:
* make spacing bitter for larger numbers to fit them better
* default number formatting

### DIFF
--- a/packages/app/src/components-styled/line-chart/index.tsx
+++ b/packages/app/src/components-styled/line-chart/index.tsx
@@ -132,7 +132,7 @@ export function LineChart<T extends Value>({
   const yDomain = useMemo(() => [0, yMax], [yMax]);
 
   // Increase space for larger labels
-  padding.left = Math.max(yMax.toFixed(0).length * 10, defaultPadding.left);
+  padding = { ...padding, left: Math.max(yMax.toFixed(0).length * 10, defaultPadding.left) };
 
   const [markerProps, setMarkerProps] = useState<{
     height: number;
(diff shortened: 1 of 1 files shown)


packages/app/src/components-styled/line-chart/index.tsx
```diff
@@ -130,11 +124,22 @@ export function LineChart<T extends Value>({
...
-  const yDomain = useMemo(() => [0, calculateYMax(trendsList, signaalwaarde)], [
+  const yMax = useMemo(() => calculateYMax(trendsList, signaalwaarde), [
     trendsList,
     signaalwaarde,
   ]);
 
+  const yDomain = useMemo(() => [0, yMax], [yMax]);
+
+  // Increase space for larger labels
+  padding.left = Math.max(yMax.toFixed(0).length * 10, defaultPadding.left);
```
VWSCoronaDashboard10, line 135: please don't mutate, this is potentially very buggy. Instead, clone the padding object and apply the new padding on that instance, eg:
```tsx
padding = { ...padding, left: ... }

VWSCoronaDashboard10 (approved):
Approved, but please have a look at the mutation thingy :-)

packages/app/src/components-styled/line-chart/index.tsx
VWSCoronaDashboard1: I'm not sure whether we should try to adjust the left padding automatically. Isn't it just an option to adjust the margin manually for a specific instance of the chart?
I don't think throughout the lifetime of the data there will be fluctuations between 10 and 10.000.000 in the given data, so a hardcoded solution would probably be just as good.
Just a thought...

VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/line-chart/index.tsx
VWSCoronaDashboard4: good catch!

PR #1342, opened 2021-01-05
Add bold text to line graphs

VWSCoronaDashboard14:
## Summary

Wrapped the date (dates) in bold tags in the line graph tooltip.


packages/app/src/components-styled/line-chart/index.tsx
```diff
@@ -338,23 +338,34 @@ function formatDefaultTooltip<T extends Value>(
...
 
   if (isDaily) {
-    return `${formatDateFromMilliseconds(value.__date.getTime())}: ${
-      isPercentage
-        ? `${formatPercentage(value.__value)}%`
-        : formatNumber(value.__value)
-    }`;
+    return (
+      <>
+        <b>{formatDateFromMilliseconds(value.__date.getTime())}: </b>
```
VWSCoronaDashboard, line 343: The text is not of special importance, and it's not a bring-to-attention type of element. MDN advices to use css to make this bold with font-weight attributes.

Does

```
<Text fontWeight="bold">
```

do what you want here?

packages/app/src/components-styled/line-chart/index.tsx
```diff
@@ -338,23 +338,34 @@ function formatDefaultTooltip<T extends Value>(
...
-      ((value as unknown) as WeeklyValue).week_end_unix,
-      'short'
-    )}: ${
-      isPercentage
-        ? `${formatPercentage(value.__value)}%`
-        : formatNumber(value.__value)
-    }`;
+    return (
+      <>
+        <b>
```
VWSCoronaDashboard, line 352: Same here, do not use `<b>`

VWSCoronaDashboard (changes requested):

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard8 (commented):
Before the `<strong>` tag was used to indicate bold text in tooltips. I don't know if that is better but maybe check if it renders the same as the other bold text in tooltips.

Also, in this repo we try to keep the convention to start commit messages with a capital letter and a present-tense verb, for example "Add bold text..."

VWSCoronaDashboard (approved):

PR #1349, opened 2021-01-06
Adjust legenda alignment

VWSCoronaDashboard4:
## Summary

Adjust legenda alignment to lay them out horizontally everywhere, with proper spacing.

### DIFF
--- a/packages/app/src/domain/deceased/deceased-monitor-section.tsx
+++ b/packages/app/src/domain/deceased/deceased-monitor-section.tsx
@@ -1,6 +1,7 @@
 import { useTheme } from 'styled-components';
 import CoronaVirusIcon from '~/assets/coronavirus.svg';
 import { AnchorTile } from '~/components-styled/anchor-tile';
+import { Box } from '~/components-styled/base';
 import { ChartTile } from '~/components-styled/chart-tile';
 import { ContentHeader } from '~/components-styled/content-header';
 import { Legenda } from '~/components-styled/legenda';
@@ -66,6 +67,8 @@ export function DeceasedMonitorSection({
             },
           }}
         />
+
+      <Box pl='56px'>
         <Legenda
           items={[
             {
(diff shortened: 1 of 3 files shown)


VWSCoronaDashboard1 (approved):

packages/app/src/domain/deceased/deceased-monitor-section.tsx
```diff
@@ -66,6 +67,8 @@ export function DeceasedMonitorSection({
             },
           }}
         />
+
+      <Box pl='56px'>
```
VWSCoronaDashboard10, line 71: Is this 56px based on the width of the Y-axis? If so I think it can break because of the dynamic width the sidebar can have.. Isn't it just fine to align the legenda with the normal element flow, instead of aligning it with the chart?

VWSCoronaDashboard4: I've left an optimisation for this for a separate PR. There are now 2 way to show the legenda, the one using the `showLegenda` option takes the dynamic sizing into account, but needs tweaking to support certain kinds of legenda contents.

Aligning them with the graph doesn't seem like a stretch, design wise, to me.

VWSCoronaDashboard8 (approved):
The title of the PR will become the (default) commit message when it gets squashed (we merge every PR with squash and merge) so pay some attention to the title and preferably make it a normal sentence that starts with a capitalized present-tense verb, as I suggested in the other PR.

VWSCoronaDashboard10 (approved):

packages/app/src/domain/deceased/deceased-monitor-section.tsx
VWSCoronaDashboard10: No but it's a tad error prone I guess with the eyeballed paddings, but you'll have my approval anyway :-)

PR #1350, opened 2021-01-06
Fix for pre-commit and pre-push hooks

VWSCoronaDashboard4:
## Summary

Fixes hooks by moving them to the root package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,5 +25,11 @@
     "packages": [
       "packages/*"
     ]
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "yarn workspace @corona-dashboard/app lint-staged",
+      "pre-push": "yarn workspace @corona-dashboard/app run type-check"
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


VWSCoronaDashboard10 (approved):

VWSCoronaDashboard (approved):

package.json
```diff
@@ -25,5 +25,11 @@
     "packages": [
       "packages/*"
     ]
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "yarn workspace @corona-dashboard/app lint-staged",
+      "pre-push": "yarn workspace @corona-dashboard/app run type-check"
```
VWSCoronaDashboard8, line 32: This might take it back to the old functionality, but in a monorepo I think it makes sense to have this for all packages, not just app. But maybe something we can add later.

VWSCoronaDashboard8 (approved):

package.json
VWSCoronaDashboard4: Good one, I'll add a task for it.

PR #1362, opened 2021-01-06
Add pages that are similar to FAQ in CMS

VWSCoronaDashboard:
## Summary

The following pages are very similar to the FAQ page.

- Over dit dashboard
- Cijferverantwoording
- Risico niveaus

This PR adds them to the CMS.

I've also fixed the prettier config, so you'll see some small unrelated changes.

## Todo before merging:

- [x] Transfer all relevant content from Lokalise to the CMS.
- [x] Remove all relevant content from Lokalise (development dataset).


packages/app/package.json
```diff
@@ -115,7 +115,7 @@
     "start": "next start -H 0.0.0.0 -p 8080",
     "test": "jest --watch",
     "test:ci": "jest --ci",
-    "format": "prettier --write src/**/*.{js,ts,tsx} *.{js,json,md,yml}",
+    "format": "echo cwd && prettier --write src/**/*.{js,ts,tsx} *.{js,json,md}",
```
VWSCoronaDashboard10, line 118: why this echo?

packages/app/src/pages/over-risiconiveaus.tsx
```diff
@@ -3,35 +3,44 @@ import Head from 'next/head';
...
 import { MaxWidth } from '~/components-styled/max-width';
-import siteText, { TALLLanguages } from '~/locale/index';
-import { parseMarkdownInLocale } from '~/utils/parse-markdown-in-locale';
+import siteText, { targetLanguage } from '~/locale/index';
 import styles from './over.module.scss';
 import { Collapsable } from '~/components-styled/collapsable';
 import { getSkipLinkId } from '~/utils/skipLinks';
 
+import { groq } from 'next-sanity';
+import { getClient, localize, PortableText } from '~/lib/sanity';
```
VWSCoronaDashboard10, line 12: I believe we usually perform a <kbd>command</kbd>+<kbd>shift</kbd>+<kbd>P</kbd> `> Organize Imports`.

packages/app/package.json
VWSCoronaDashboard8, line 118: What's the reason for the echo? AFAIK the result isn't having any effect on this script because it's not piped to the next command

packages/app/src/pages/over-risiconiveaus.tsx
```diff
@@ -3,35 +3,44 @@ import Head from 'next/head';
...
 interface StaticProps {
   props: OverRisiconiveausProps;
 }
 
 interface OverRisiconiveausProps {
-  text: TALLLanguages;
+  data: {
+    title: string;
+    description: string | null;
+    collapsibleList: Array<{ content: Array<any>; title: string }>;
```
VWSCoronaDashboard10, line 22: can we get rid of the any type?

VWSCoronaDashboard8, line 22: nit; Prefer to stick to one type of array syntax: `[]`

Any should be avoided here. If not strongly typed then probably by using some shape of `unknown`

packages/app/src/pages/over-risiconiveaus.tsx
```diff
@@ -3,35 +3,44 @@ import Head from 'next/head';
...
+  };
   lastGenerated: string;
 }
 
-export async function getStaticProps(): Promise<StaticProps> {
-  const text = parseMarkdownInLocale((await import('../locale/index')).default);
+const risicoQuery = groq`
```
VWSCoronaDashboard8, line 27: I think `risico` doesn't read the same as `risiconiveaus`, and cutting off those few letters doesn't have a clear advantage, but it does add confusion for the reader.

If you prefer to keep things short, I would use short standardized names like query and data, but I already suggested that before.

packages/app/src/pages/over-risiconiveaus.tsx
```diff
@@ -52,26 +61,22 @@ const OverRisicoNiveaus: FCWithLayout<OverRisiconiveausProps> = (props) => {
...
-                ) : null;
-              })}
-            </article>
+            <h2>{data.title}</h2>
+
+            {data.description && <PortableText blocks={data.description} />}
+
+            {data.collapsibleList && (
+              <article className={styles.faqList}>
+                {data.collapsibleList.map((item: any) => {
```
VWSCoronaDashboard10, line 70: please type `collapsibleList` :-)

packages/app/src/pages/over-risiconiveaus.tsx
```diff
@@ -52,26 +61,22 @@ const OverRisicoNiveaus: FCWithLayout<OverRisiconiveausProps> = (props) => {
...
+                  const id = getSkipLinkId(item.title);
```
VWSCoronaDashboard8, line 71: See, here you are assuming that title and content exist on the collapsible items, so you can't type it as any without introducing the risk for runtime exceptions.

packages/app/src/pages/over.tsx
```diff
@@ -3,31 +3,41 @@ import Head from 'next/head';
...
+
 interface StaticProps {
   props: OverProps;
 }
 
 interface OverProps {
-  text: TALLLanguages;
+  data: {
+    title: string;
+    description: string | null;
```
VWSCoronaDashboard10, line 19: We tend to prevent null values. Can we type it `string | undefined` instead?

packages/app/src/pages/verantwoording.tsx
```diff
@@ -4,32 +4,43 @@ import path from 'path';
...
 interface StaticProps {
   props: VerantwoordingProps;
 }
 
 interface VerantwoordingProps {
-  text: TALLLanguages;
+  data: {
+    title: string;
+    description: string | null;
+    collapsibleList: Array<{ content: Array<any>; title: string }>;
```
VWSCoronaDashboard8, line 22: Same comments here

packages/cms/desk-structure.js
```diff
@@ -80,57 +80,57 @@ export default () =>
...
+              //   .options({ previewURL: `${previewURL}/over` })
+              //   .title("Web"),
+              // S.view
+              //   .component(IFrameMobilePreview)
+              //   .options({ previewURL: `${previewURL}/over` })
+              //   .title("Mobile"),
+              // S.view
+              //   .component(ColorblindPreview)
+              //   .options({ previewURL: `${previewURL}/over` })
+              //   .title("Color Blindness"),
```
VWSCoronaDashboard8, line 105: Do we need to commit these comments?

packages/app/src/pages/over.tsx
```diff
@@ -3,31 +3,41 @@ import Head from 'next/head';
...
+  };
   lastGenerated: string;
 }
 
-export async function getStaticProps(): Promise<StaticProps> {
-  const text = parseMarkdownInLocale((await import('../locale/index')).default);
+const overQuery = groq`
```
VWSCoronaDashboard10, line 24: Doesn't the file `pages/over.tsx` makes it clear enough we're querying the `over`-page? Might rename it to `query` or `pageQuery` or remove the variable entirely and move the ```grog`` ``` to line 33 (`.fetch(groq())`).

VWSCoronaDashboard4 (approved):

packages/cms/desk-structure.js
```diff
@@ -80,57 +80,57 @@ export default () =>
...
-      //       ])
-      //   ),
+      S.listItem()
+        .title("Over dit dashboard")
+        .schemaType("overDitDashboard")
+        .icon(GrCircleInformation)
+        .child(
+          S.editor()
+            .title("Over dit dashboard")
+            .schemaType("overDitDashboard")
```
VWSCoronaDashboard8, line 90: How are schema types written WRT casing? I am surprised that this a type name that starts with a lowercase letter yet all Typescript types are of course starting with uppercase. 

I feel this adds confusion. Can we either make it like typescript or choose a different format altogether like the snake-cased we used in the data schema properties?

packages/cms/desk-structure.js
```diff
@@ -80,57 +80,57 @@ export default () =>
...
+            .documentId("overDitDashboard")
```
VWSCoronaDashboard10, line 91: nit: I'd prefer english id's and schema names

packages/cms/desk-structure.js
```diff
@@ -80,57 +80,57 @@ export default () =>
...
-      //   ),
+      S.listItem()
+        .title("Over de risiconiveaus")
+        .schemaType("overRisicoNiveaus")
+        .icon(BsMap)
+        .child(
+          S.editor()
+            .title("Over de risiconiveaus")
+            .schemaType("overRisicoNiveaus")
+            .documentId("overRisicoNiveaus")
```
VWSCoronaDashboard8, line 117: I think at this point we have to be very careful to choose these names, because if we don't get it right at the start, it will create a mess like lokalize keys that are never going to be fixable again. 

This transition to a CMS is the one chance we have to clean up and use consistent naming.

For example, I can imagine that it is useful to prefix id/types that are about cijferverantwoording so that these things are related and easy to find in the large amount of types and ids we will have later. But since that is a long word it might be good to define/keep a list of short prefixes that we can use for the different categories of datatypes.

VWSCoronaDashboard8 (changes requested):

packages/cms/desk-structure.js
VWSCoronaDashboard4: If this is going to stay in Dutch (I'd also prefer these IDs in English), at least it should be written as `overRisiconiveaus`, since risiconiveaus is 1 word.

packages/app/package.json
VWSCoronaDashboard: Debugging leftover, I need to remove it!

packages/app/src/pages/over-risiconiveaus.tsx
VWSCoronaDashboard: We can see if [this example type definition](https://github.com/nhi/sanity-block-content-to-react-types/blob/master/block-content-to-react.d.ts) is working, but I'm hesitant as long as there are no types provided by the package or DefinitelyTyped.

packages/app/src/pages/over.tsx
VWSCoronaDashboard: Apparently you can not use `undefined` in getStaticProps. It will return a serializer error. That's why I had to make it an explicit `null` value.

packages/app/src/pages/over.tsx
VWSCoronaDashboard: Fair enough. Will do this.

packages/cms/desk-structure.js
VWSCoronaDashboard: Guess we're going to have to do our first official [migration](https://www.sanity.io/docs/migrating-data).

packages/cms/desk-structure.js
VWSCoronaDashboard: So to prevent confusion here, this is NOT a TypeScript schema.

It doesn't seem like Sanity adheres to an official casing convention, so we can change it if we want. https://www.sanity.io/docs/naming-things

packages/cms/desk-structure.js
VWSCoronaDashboard: For context, there's 3 things to keep in mind when evaluating sanity schemas.

1. Documents are meant to be queried directly. You can make these very specific.
2. Objects are meant to be re-used across other objects or documents. Think of them as building blocks. By definition, these should be more generic in their naming approach. `collapsibeList` is a good example of a building block that gets used in different contexts. It can be a list of questions, a list of definitions, or something else.
3. Some documents are designed to be singletons, like "overRisicoNiveaus". Technically you can make more instances of this type, but we limit that by designing the UI in such a way that this is protected.

I agree with getting it right, so I'll rename this one to `overRisiconiveaus`

packages/cms/desk-structure.js
VWSCoronaDashboard: At some point I want to enable this again. I can also just remove it completely and rely on source control?

packages/app/src/pages/over-risiconiveaus.tsx
VWSCoronaDashboard: Done!

packages/app/src/pages/over.tsx
VWSCoronaDashboard: Changed it to this pattern:

```jsx
  const query = groq`
  *[_type == 'cijferVerantwoording'][0]
`;
  const rawData = await getClient(false).fetch(query);
  const data = localize(rawData, [targetLanguage, 'nl']);
```

I believe this strikes a nice balance between readability and maintenance.

packages/app/package.json
VWSCoronaDashboard: The echo is solved, I removed it.

packages/app/src/pages/over-risiconiveaus.tsx
VWSCoronaDashboard: Yeah, I renamed this to be just query.

packages/app/src/pages/over.tsx
VWSCoronaDashboard: See also `packages/app/src/lib/sanity.ts`


```ts
export function localize(value: any, languages: any): any {
  if (Array.isArray(value)) {
    return value.map((v) => localize(v, languages));
  } else if (typeof value == 'object') {
    if (/^locale[A-Z]/.test(value._type)) {
      const language = languages.find((lang: string) => value[lang]);

      // React will trip if you return undefined, which could happen
      // if you don't fill in anything in the CMS
      return value[language] === undefined ? null : value[language];
    }

    return Object.keys(value).reduce((result, key) => {
      // @ts-ignore
      result[key] = localize(value[key], languages);
      return result;
    }, {});
  }
  return value;
}
```

packages/app/src/pages/over-risiconiveaus.tsx
VWSCoronaDashboard: Changed the syntax.
Changed any to unknown.

I also changed description to unknown[] since it renders the same kind of block content as collapsibleList.

packages/app/src/pages/over-risiconiveaus.tsx
VWSCoronaDashboard: Solved! Thanks!

packages/app/src/pages/veelgestelde-vragen.tsx
```diff
@@ -1,39 +1,39 @@
...
   const fileContents = fs.readFileSync(filePath, 'utf8');
   const lastGenerated = JSON.parse(fileContents).last_generated;
 
-  const faqData = await getClient(false).fetch(faqQuery);
-  const data = localize(faqData, [targetLanguage, 'nl']);
+  const query = groq`
+  *[_type == 'veelgesteldeVragen'][0]
+`;
+  const rawData = await getClient(false).fetch(query);
+  const data = localize(rawData, [targetLanguage, 'nl']);
```
VWSCoronaDashboard1, line 36: I'm guessing the hardcoded 'nl' locale isn't desirable?

VWSCoronaDashboard1 (approved):

packages/cms/desk-structure.js
VWSCoronaDashboard8: I don't think a Typescript schema is a thing :) But yes, let's follow up soon with a meeting about this naming things.

packages/cms/desk-structure.js
VWSCoronaDashboard8: If you know you want to enable this later I don't mind. I tend to do this as well.

VWSCoronaDashboard8 (approved):

packages/app/src/pages/over-risiconiveaus.tsx
```diff
@@ -1,37 +1,45 @@
...
 
 interface StaticProps {
   props: OverRisiconiveausProps;
 }
 
 interface OverRisiconiveausProps {
-  text: TALLLanguages;
+  data: {
+    title: string | null;
+    description: unknown[] | null;
```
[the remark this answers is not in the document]
VWSCoronaDashboard8: I now read your reply on the orginal comment so yeah, never mind.

packages/app/src/pages/over.tsx
VWSCoronaDashboard8: @VWSCoronaDashboard I can imagine getStaticProps should return something besides undefined, but I am surprised that this would be the case for child properties of data structures returned by getStaticProps. I think it's worth doing some more investigation in the near future, before we litter the types all over the place with nulls.

packages/app/src/pages/veelgestelde-vragen.tsx
VWSCoronaDashboard: It actually is! 'nl' is the fallback locale for when it can't find targetLanguage.



Example, when calling `const data = localize(rawData, ['en', 'nl'])`
it tries to resolve `data.title.en` first, but if it can't find the English translation it will return `data.title.nl`.

VWSCoronaDashboard: We could also remove the localize function and write our queries with coalesce.
That might be even better, as you don't need the rawData variable anymore. The only "downside" is you need to write out your data query more explicitly like in GraphQL.

```
* [_type == "veelgesteldeVragen"] {
  ...,
  "title": coalesce(title.en, title.nl, "Missing translation")
}[0]
```

See https://www.sanity.io/docs/localization

packages/app/src/pages/over.tsx
VWSCoronaDashboard: @VWSCoronaDashboard8 I would very much appreciate that!

packages/app/src/pages/over-risiconiveaus.tsx
VWSCoronaDashboard: Made it a nice type for you 👍

packages/cms/desk-structure.js
VWSCoronaDashboard: Actually, I'm going to keep it as-is for this PR, and we'll make a new PR once we're ready to rename some keys. That needs to be automated anyway, so I'd say it's not a part of this PR.

VWSCoronaDashboard: For what it's worth, we can group these schema's in folder so they're always co-located. The comparison with Lokalise is not very relevant.

PR #1369, opened 2021-01-07
Move cypress into a separate package of the monorepo

VWSCoronaDashboard1:
## Summary

As the title states

## Motivation

This will PR will hopefully streamline the build process a little bit:
During the build step of the application, the Cypress dependencies would be installed as well, while these aren't needed.
Then in the next step where the Cypress tests are built, all of the app's dependencies would be installed as well, adding unnecessary time to the build.
By moving Cypress into its own package these superfluous installs will be prevented.

A few new tests have been added as well. On the positief-geteste-mensen page the choropleth click-throughs for municipal and regions are tested.

## Detailed design

Not much to explain. All of the Cypress test files have been moved to packages/cypress. The cypress dependencies (well, just one really) have been removed from the app's package.json and added to the cypress/package.json.

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

After this PR is merged there will need to be made some changes to the yaml file on gitlab. Specifically, this block:
```
script:
    - echo -e "Cloning '${BRANCH_NAME}' from ${ORIGIN_REPO}"
    - git clone --depth 1 -b ${BRANCH_NAME} https://${ORIGIN_REPO} mirror
    - cd mirror/packages/app/
    - mkdir -p public/json
    - cp -r ../../../files_from_blobstorage/* public/json/ 
    - yarn && yarn cy:ci --env host=${HOSTNAME_NL},port=80
  artifacts:
    when: on_failure
    paths:
      - /builds/vws/cdb/frontend-development-deployment/mirror/packages/app/cypress/screenshots/*
      - /builds/vws/cdb/frontend-development-deployment/mirror/packages/app/cypress/videos/*
```

Will need to be re-written to this:
```
script:
    - echo -e "Cloning '${BRANCH_NAME}' from ${ORIGIN_REPO}"
    - git clone --depth 1 -b ${BRANCH_NAME} https://${ORIGIN_REPO} mirror
    - cd mirror/packages/app/
    - mkdir -p public/json
    - cp -r ../../../files_from_blobstorage/* public/json/ 
    - cd mirror/packages/e2e/
    - yarn && yarn e2e:ci --env host=${HOSTNAME_NL},port=80
  artifacts:
    when: on_failure
    paths:
      - /builds/vws/cdb/frontend-development-deployment/mirror/packages/e2e/screenshots/*
      - /builds/vws/cdb/frontend-development-deployment/mirror/packages/e2e/videos/*

```

### Governance

- [ ] Documentation is added


VWSCoronaDashboard4 (approved):

packages/app/src/components-styled/aspect-ratio.tsx
```diff
@@ -8,17 +8,18 @@ interface AspectRatioProps {
    */
   ratio: number;
   children: ReactNode;
+  dataCy?: string;
```
VWSCoronaDashboard8, line 11: nit; I don't like this name since it is unclear and strictly tied to Cypress (our app code should be impartial to what e2e testing library is being used). I'd rather use a more readable generic name in our app code like `testingHandle` or `testingLabel`, but maybe that's a discussion we should have in-person to we if we can agree.

packages/cypress/integration/gemeente/positief-geteste-mensen.spec.ts
```diff
@@ -1,4 +1,4 @@
-import { MunicipalContext } from 'cypress/integration/types';
+import { MunicipalContext } from '@/integration/types';
```
VWSCoronaDashboard8, line 1: Why are you using `@` here?

packages/cypress/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "name": "corona-dashboard-cypress-tests",
+  "version": "1.0.0",
+  "main": "index.js",
+  "license": "MIT",
+  "devDependencies": {
+    "cypress": "^6.2.1"
+  },
+  "scripts": {
+    "lint": "echo 'No linter configured for @corona-dashboard/cypress'",
+    "compile": "echo 'No compiler configured for @corona-dashboard/cypress'",
```
VWSCoronaDashboard8, line 11: I think these two can easily be defined and are valuable since we don't want to deploy / merge code with broken testing code.

packages/cypress/tsconfig.json
```diff
@@ -1,8 +1,9 @@
 {
   "compilerOptions": {
-    "baseUrl": "../",
+    "baseUrl": ".",
     "paths": {
-      "~/*": ["src/*"]
+      "~/*": ["../app/src/*"],
+      "@/*": ["/*"]
```
VWSCoronaDashboard8, line 6: I think this adds unnecessary confusion.

Packages can import from other packages explicitly using the workspace package name. I think `~/` should only mean "local (package) src directory" in every package.

I will look up how this was done in my other monorepo project...

VWSCoronaDashboard8 (changes requested):

packages/cypress/integration/gemeente/positief-geteste-mensen.spec.ts
VWSCoronaDashboard1: because there are two different aliases, one for testing code and one that points to the app sources (I use a little bit of shared code there, most important is the `data.d.ts` file, but I also use some of the util functions)

packages/cypress/tsconfig.json
VWSCoronaDashboard1: Please do, I'd be interested to see how it's supposed to be done. IMHO, the two separate aliases were clear enough, but if there's a better solution I'm all ears.

packages/cypress/package.json
VWSCoronaDashboard1: You're absolutely right, this is fixed now

packages/app/src/components-styled/aspect-ratio.tsx
VWSCoronaDashboard1: With that in mind, I've renamed the package from cypress to e2e, as well as the commands from `cy:open` to `e2e` and `cy:ci` to `e2e:ci`

PR #1386, opened 2021-01-08
Load choropleth data during build time

VWSCoronaDashboard10:
The choropleth data were the last bits which got loaded async during the runtime lifecylce of our app. This PR will load this data build-time and inject it on the initial props of the page component.

The current implementation works as follows:

all `getStaticProps` of the pages were already centralized to three files:
  - `getMunicipalityStaticProps`
  - `getSafetyRegionStaticProps`
  - `getNationalStaticProps`

These functions have been modified to receive optional choropleth settings. The choropleth settings contain `vr` and `gm` filters:

```tsx
// This example implementation will pick the `behavior` data for a VR- and GM-choropleth.
export const getStaticProps = getNationalStaticProps({
  choropleth: {
    gm: x => ({ behavior: x.behavior }),
    vr: x => ({ behavior: x.behavior }),
  },
});
```

The return type of the `getStaticProps` can now be used within the `FCWithLayout`-component-type:
```tsx
const IntakeHospital: FCWithLayout<typeof getStaticProps> = (props) => {
  assert(props.choropleth.gm.behavior !== undefined)
  assert(props.choropleth.vr.behavior !== undefined)
  // ...
}
```

The choropleth data is passed on the props and can simply be forwarded to a choropleth chart:

```tsx
<SafetyRegionChoropleth
  data={props.choropleth.vr}
  metricName="behavior" // the prop types will still validate that `behavior` is part of `props.choropleth.vr`
  metricProperty={metricValueName}
/>
```

Also the `getClient` will now always return the normal sanity client, this is a prettier api imho. The preview client can be obtained by calling `getPreviewClient`.

Another thing to note is the `getClient`'s returned client-instance which now requires a generic type on the fetch method, otherwise it will fallback to a `never` type:

```tsx
const rawData = await getClient().fetch<OverRisiconiveausData>(query);
```


VWSCoronaDashboard1 (approved):

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard4 (approved):
:100: 👯

packages/app/src/domain/layout/layout.tsx
```diff
@@ -15,7 +15,18 @@ interface LayoutProps {
...
+>['props'];
+
+export type FCWithLayout<
+  PropsOrGetStaticProps = void,
+  Props = PropsOrGetStaticProps extends (
+    ...args: any[]
+  ) => Promise<{ props: any }>
+    ? StaticProps<PropsOrGetStaticProps>
+    : PropsOrGetStaticProps
+> = React.FC<Props> & {
```
VWSCoronaDashboard8, line 29: Have you tried replacing any with unknown? This is complex stuff and surely hard to type strict, but unknown will at least indicate that we don't know what it is.

packages/app/src/pages/veiligheidsregio/index.tsx
```diff
@@ -110,23 +97,38 @@ const SafetyRegion: FCWithLayout<any> = (props) => {
   );
 };
 
-SafetyRegion.getLayout = getSafetyRegionLayout();
+SafetyRegion.getLayout = getSafetyRegionLayout() as any;
```
VWSCoronaDashboard8, line 100: Really any? Maybe it could be `JSX.Element` or something?

packages/app/src/static-props/municipality-data.ts
```diff
@@ -1,57 +1,27 @@
...
- * ```ts
- * const PositivelyTestedPeople: FCWithLayout<IMunicipalityData> = props => {
- *   // ...
- * }
- * ```
- */
-export function getMunicipalityData() {
-  return async function ({ params }: IParams): Promise<IProps> {
+export function getMunicipalityStaticProps<T1 = undefined, T2 = undefined>(
+  settings?: MunicipalityPagePropsSettings<T1, T2>
```
VWSCoronaDashboard8, line 22: Not sure "settings" is the best name here. Maybe "selectors"? You are passing in functions to actively select data like a lens type of thing.

Of course, I suggest renaming the type then too.

packages/app/src/static-props/nl-data.ts
```diff
@@ -2,54 +2,40 @@ import fs from 'fs';
...
- *
- * Example:
- * ```ts
- * const PositivelyTestedPeople: FCWithLayout<INationalData> = props => {
- *   // ...
- * }
- * ```
- */
+export type NationalPageProps = Await<
+  ReturnType<ReturnType<typeof getNationalStaticProps>>
```
VWSCoronaDashboard8, line 13: 😅

VWSCoronaDashboard8 (approved):

packages/app/src/domain/layout/layout.tsx
VWSCoronaDashboard10: Yep, but I was unable to get that right.. At least the types are as expected on the outside.

PR #1387, opened 2021-01-08
Add generated choropleth data description

VWSCoronaDashboard1:
## Summary

Generates a text that gives a small summary of what the choropleth is showing and adds this as the aria label to the choropleths.

## Motivation

Choropleths are currently completely inaccessible. This is the first step in making them a bit more useful for visually impaired users.

## Detailed design

A new hook has been introduced called `useChoroplethDataDescription`. It takes the raw choropleth data and the metric name and property. The metric name and property are used to retrieve the text templates from the locale file and those bits and pieces are used to assemble a description of the data.

An example of this generated text looks like this:
"Deze kaart toont dat 3 gemeentes tussen de 0 en 3 positief geteste mensen per 100000 inwoners hebben, 2 gemeentes tussen de 4 en 6 positief geteste mensen per 100000 inwoners hebben, 3 gemeentes tussen de 7 en 9 positief geteste mensen per 100000 inwoners hebben, 24 gemeentes tussen de 10 en 19 positief geteste mensen per 100000 inwoners hebben, 66 gemeentes tussen de 20 en 29 positief geteste mensen per 100000 inwoners hebben en 238 gemeentes meer dan 30 positief geteste mensen per 100000 inwoners hebben."

The text is then rendered into a span with its display set to none. Its assigned id is then used to reference by the svg element's `aria-labelledby` attribute.
I've chosen this alternative to setting it as a `<title>` element on the svg, because the contents of this element are show as a tooltip by the browser by default. (Which would interfere with the custom tooltips)

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

~~Right now, templates are still missing for the behavior choropleths. This is a considerable amount of work and will be added ina separate PR once this one has been merged.~~

### Governance

- [ ] Documentation is added

### DIFF
--- a/packages/app/src/components/choropleth/choropleth.tsx
+++ b/packages/app/src/components/choropleth/choropleth.tsx
@@ -48,7 +48,7 @@ type TProps<T1, T3> = {
   // This callback is invoked right before a tooltip is shown for one of the features in the featureCollection property.
   // The id is the value that is assigned to the data-id attribute in the featureCallback.
   getTooltipContent: (id: string) => ReactNode;
-  title?: string;
+  dataDescription?: string;
 };
 
 /**
@@ -116,7 +116,7 @@ const ChoroplethMap: <T1, T3>(
     onPathClick,
     setTooltip,
     hoverRef,
-    title,
+    dataDescription,
   } = props;
 
   const clipPathId = useRef(`_${Math.random().toString(36).substring(2, 15)}`);
@@ -138,7 +138,7 @@ const ChoroplethMap: <T1, T3>(
   return (
     <>
       <span id={titleId.current} style={{ display: 'none' }}>
-        {title}
+        {dataDescription}
       </span>
       <svg
         width="100%"
--- a/packages/app/src/components/choropleth/choropleth.tsx
+++ b/packages/app/src/components/choropleth/choropleth.tsx
@@ -120,7 +120,9 @@ const ChoroplethMap: <T1, T3>(
   } = props;
 
   const clipPathId = useRef(`_${Math.random().toString(36).substring(2, 15)}`);
-  const titleId = useRef(`_${Math.random().toString(36).substring(2, 15)}`);
+  const dataDescriptionId = useRef(
+    `_${Math.random().toString(36).substring(2, 15)}`
+  );
   const timeout = useRef(-1);
   const isTouch = useIsTouchDevice();
 
@@ -137,7 +139,7 @@ const ChoroplethMap: <T1, T3>(
 
   return (
     <>
-      <span id={titleId.current} style={{ display: 'none' }}>
+      <span id={dataDescriptionId.current} style={{ display: 'none' }}>
         {dataDescription}
       </span>
       <svg
@@ -149,7 +151,7 @@ const ChoroplethMap: <T1, T3>(
           isTouch ? undefined : createSvgMouseOutHandler(timeout, setTooltip)
         }
         onClick={createSvgClickHandler(onPathClick, setTooltip, isTouch)}
-        aria-labelledby={titleId.current}
+        aria-labelledby={dataDescriptionId.current}
       >
         <clipPath id={clipPathId.current}>
           <rect
--- a/packages/app/src/components/choropleth/choropleth.tsx
+++ b/packages/app/src/components/choropleth/choropleth.tsx
@@ -48,6 +48,7 @@ type TProps<T1, T3> = {
   // This callback is invoked right before a tooltip is shown for one of the features in the featureCollection property.
   // The id is the value that is assigned to the data-id attribute in the featureCallback.
   getTooltipContent: (id: string) => ReactNode;
+  title?: string;
 };
 
 /**
@@ -115,9 +116,11 @@ const ChoroplethMap: <T1, T3>(
     onPathClick,
     setTooltip,
     hoverRef,
+    title,
   } = props;
 
   const clipPathId = useRef(`_${Math.random().toString(36).substring(2, 15)}`);
+  const titleId = useRef(`_${Math.random().toString(36).substring(2, 15)}`);
   const timeout = useRef(-1);
   const isTouch = useIsTouchDevice();
 
@@ -134,6 +137,9 @@ const ChoroplethMap: <T1, T3>(
 
   return (
     <>
+      <span id={titleId.current} style={{ display: 'none' }}>
+        {title}
+      </span>
       <svg
         width="100%"
         height="100%"
@@ -143,6 +149,7 @@ const ChoroplethMap: <T1, T3>(
           isTouch ? undefined : createSvgMouseOutHandler(timeout, setTooltip)
         }
         onClick={createSvgClickHandler(onPathClick, setTooltip, isTouch)}
+        aria-labelledby={titleId.current}
       >
         <clipPath id={clipPathId.current}>
           <rect
--- a/packages/app/src/components/choropleth/choropleth.tsx
+++ b/packages/app/src/components/choropleth/choropleth.tsx
@@ -48,7 +48,7 @@ type TProps<T1, T3> = {
   // This callback is invoked right before a tooltip is shown for one of the features in the featureCollection property.
   // The id is the value that is assigned to the data-id attribute in the featureCallback.
   getTooltipContent: (id: string) => ReactNode;
-  dataDescription?: string;
+  description?: string;
 };
 
 /**
@@ -116,7 +116,7 @@ const ChoroplethMap: <T1, T3>(
     onPathClick,
     setTooltip,
     hoverRef,
-    dataDescription,
+    description,
   } = props;
 
   const clipPathId = useRef(`_${Math.random().toString(36).substring(2, 15)}`);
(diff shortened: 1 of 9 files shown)


packages/app/src/locale/en.json
```diff
@@ -2062,5 +2056,74 @@
...
     "VR24": "http://www.vrzuidlimburg.nl/",
     "VR25": "http://www.veiligheidsregioflevoland.nl/"
+  },
+  "choropleth": {
+    "verb": {
+      "single": "has",
+      "plural": "have"
+    },
+    "municipal": {
+      "single": "municipality",
```
VWSCoronaDashboard4, line 2066: The correct term is `singular`.

packages/app/src/locale/en.json
```diff
@@ -2062,5 +2056,74 @@
...
+      "plural": "municipalities"
+    },
+    "region": {
+      "single": "safetyregion",
+      "plural": "safetyregions"
+    },
+    "tested_overall": {
+      "infected_per_100k": {
+        "full_sentence": "This map shows that {{first}} and {{last}}.",
```
VWSCoronaDashboard4, line 2075: Shouldn't the `full_sentence_single ` key be present here as well?

packages/app/src/components/choropleth/safety-region-choropleth.tsx
```diff
@@ -135,6 +144,7 @@ export function SafetyRegionChoropleth<T>(
     <div ref={ref} css={css({ position: 'relative', bg: 'transparent' })}>
       <AspectRatio ratio={1 / ratio}>
         <Choropleth
+          dataDescription={dataDescription}
```
VWSCoronaDashboard4, line 147: I'd opt to name this property just `description`. The `data` prefix does not add so much and leaving it out makes this property more flexible.

packages/app/src/components/choropleth/hooks/use-municipality-data.ts
```diff
@@ -73,13 +80,20 @@ export function useMunicipalityData(
       {} as Record<string, MunicipalityProperties>
     );
 
+    const values =
+      (data?.[metricName] as any[])?.map((x) => ({
```
VWSCoronaDashboard4, line 84: I think we can get rid of this `any`. If we really cannot, tere should be a new interface to assign to `values` (containing the `code` and `value`).

packages/app/src/components/choropleth/hooks/use-choropleth-data-description.ts
```diff
@@ -0,0 +1,96 @@
+import { useMemo } from 'react';
+import siteText from '~/locale';
+import { replaceVariablesInText } from '~/utils/replaceVariablesInText';
+import { ChoroplethThresholdsValue } from '../shared';
+import { DataValue } from './use-municipality-data';
+
+/**
+ * This hooks generates a text that describes the distribution of the choropleth data for accessibility purposes.
+ */
+export function useChoroplethDataDescription<T>(
```
VWSCoronaDashboard4, line 10: Some more comments that describe the inner workings of this hook would be great.

packages/app/src/components/choropleth/hooks/use-choropleth-data-description.ts
```diff
@@ -0,0 +1,96 @@
...
+            }
+            return acc;
+          }, 0),
+        };
+        return result;
+      })
+      .filter((x) => x.count > 0);
+
+    if (ranges.length == 0) {
+      return '';
```
VWSCoronaDashboard4, line 54: Maybe assert this? If the graph does not contain data, that seems like an error to me.

Otherwise, perhaps use a default text stating the graph does not contain data?

VWSCoronaDashboard4 (commented):
Nice improvement!

packages/app/src/components/choropleth/hooks/use-choropleth-data-description.ts
VWSCoronaDashboard1: Actually, as a rule, the hook description (in CF projects, that is) was decided to only give a broad description of what the hook does. Not any implementation details.

packages/app/src/components/choropleth/hooks/use-choropleth-data-description.ts
VWSCoronaDashboard1: I think the assert should be done earlier, but the text seems like a solid idea

packages/app/src/components/choropleth/hooks/use-municipality-data.ts
VWSCoronaDashboard1: Unfortunately, there is no way here to prevent the 'any' cast (unless someone can enlighten me). There already is an interface for the mapped type though: `DataValue`.

packages/app/src/locale/en.json
VWSCoronaDashboard1: I'm not sure I understand your question.

packages/app/src/components/choropleth/hooks/use-municipality-data.ts
VWSCoronaDashboard8: There is always the possibility of using some form of unknown, which is I believe always an improvement over any. In this case your code below is counting on the gmcode property so that and a string indexed value, so you could define that. But I'm not saying it's worth it here. We should probably tackle the problem at the source; the typing of the data object.

packages/app/src/locale/en.json
```diff
@@ -2062,5 +2056,74 @@
...
+      "plural": "have"
+    },
+    "municipal": {
+      "singular": "municipality",
+      "plural": "municipalities"
+    },
+    "region": {
+      "singular": "safetyregion",
+      "plural": "safetyregions"
+    },
```
VWSCoronaDashboard8, line 2072: I started using vr and gm in more places. I would choose to do the same here for the keys.

VWSCoronaDashboard8 (approved):

packages/app/src/components/choropleth/hooks/use-choropleth-data-description.ts
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * This hook generates a text that describes the distribution of the choropleth data for accessibility purposes.
+ */
+export function useChoroplethDataDescription<T>(
+  thresholds: ChoroplethThresholdsValue[],
+  gmValues: DataValue[],
+  metricName: keyof T,
+  metricProperty: string,
+  area: 'region' | 'municipal',
+  gemcodes?: string[]
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: No, the gemcodes are optional, they're only used in the context of the municipality choropleth (and even there used only in the case of a selected municipality). In the region component this is not necessary so this param is never used.
Because of this I didn't see it necessary to create two separate hook for each flavor of choropleth.

packages/app/src/components/choropleth/hooks/use-choropleth-data-description.ts
```diff
@@ -0,0 +1,96 @@
...
+  }, [thresholds, gmValues, metricName, metricProperty, area, gemcodes]);
+}
+
+function getDynamicText<T>(metricName: keyof T, metricProperty: string) {
+  if (metricName !== 'behavior') {
+    return (siteText.choropleth as any)[metricName]?.[metricProperty];
+  } else {
+    const parts = metricProperty.split('_');
+    const restrictionKey = parts.slice(0, -1).join('_');
+    const restriction = (siteText.gedrag_onderwerpen as any)[restrictionKey];
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: Yes, because behavior is an exception in this case. I'm not sure how I could implement this in a separate hook and then choose a specific implementation within the choropleth. It would simply lift the problem one step higher, but wouldn't make it any more clear. IMHO, of course.

packages/app/src/components/choropleth/hooks/use-choropleth-data-description.ts
```diff
@@ -0,0 +1,96 @@
...
+          last: texts.slice(-1).join(''),
+        })
+      : replaceVariablesInText(dynamicTexts.full_sentence_single, {
+          first: texts[0],
+        });
+  }, [thresholds, gmValues, metricName, metricProperty, area, gemcodes]);
+}
+
+function getDynamicText<T>(metricName: keyof T, metricProperty: string) {
+  if (metricName !== 'behavior') {
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: Done

packages/app/src/locale/en.json
VWSCoronaDashboard1: Fixed

packages/app/src/components/choropleth/choropleth.tsx
```diff
@@ -115,9 +116,13 @@ const ChoroplethMap: <T1, T3>(
     onPathClick,
     setTooltip,
     hoverRef,
+    description,
   } = props;
 
   const clipPathId = useRef(`_${Math.random().toString(36).substring(2, 15)}`);
+  const dataDescriptionId = useRef(
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: I checked the utils folder but can find no such hook

PR #1388, opened 2021-01-11
Clean up schema file names and validator code

VWSCoronaDashboard8:
## Summary

This cleans up the file names that are used for defining and validating schemas. Also, it removes validation of locale files, since it seemed to be very incomplete and therefore pointless.


VWSCoronaDashboard10 (approved):

packages/cli/src/validator/schema-information.ts
```diff
@@ -1,50 +1,39 @@
...
+    vr: {
+      files: getFileNames(fileList, /^VR[0-9]+.json$/),
+      basePath: path,
     },
-    municipalities: { files: ['GM_COLLECTION.json'], basePath: jsonPath },
-    regions: { files: ['VR_COLLECTION.json'], basePath: jsonPath },
-    locale: {
-      files: filterFilenames(localeJsons, /[^.]+.json$/),
-      basePath: localeDirectory,
-      customValidations: [validatePlaceholders],
```
VWSCoronaDashboard1, line 43: Am I misinterpreting something, or are the locale files completely removed from the validations?

VWSCoronaDashboard1 (approved):

packages/cli/src/validator/schema-information.ts
VWSCoronaDashboard8: @VWSCoronaDashboard1 Yes, that's also what I tried to say in the PR description. Do you think this is problematic?

VWSCoronaDashboard10: I think they are, and I think that would also fix the current validation errors on develop.

VWSCoronaDashboard10: And there's also validation using the TS compiler which will complain when required keys are missing.

VWSCoronaDashboard8: @VWSCoronaDashboard10 Are you saying that the validations as they were are problematic, or the removal of the validations by this PR?

VWSCoronaDashboard1: Dev was failing because of the removal of verantwoording and other bits, that were moved to the CMS.
I've already discussed with @VWSCoronaDashboard8 that the placeholder validation definitely is warranted.
The validations that were there were put there out of necessity because in some cases Lokalise will change an array value into an object value. So these validations weren't put there for nothing, but now that those bits are moved to the CMS, they're no longer needed.

PR #1410, opened 2021-01-12
Reduce padding of the labels on smaller screens

VWSCoronaDashboard14:
Reduced the padding X on smaller screens so the labels fit on the screen.

### DIFF
--- a/packages/app/src/components-styled/radio-group.tsx
+++ b/packages/app/src/components-styled/radio-group.tsx
@@ -2,7 +2,7 @@ import { Fragment, useRef, useState } from 'react';
 import { Box } from './base';
 import { css } from '@styled-system/css';
 import styled from 'styled-components';
-
+import { asResponsiveArray } from '~/style/utils';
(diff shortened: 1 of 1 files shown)


VWSCoronaDashboard1 (approved):

VWSCoronaDashboard8 (approved):

packages/app/src/components-styled/radio-group.tsx
```diff
@@ -47,6 +47,10 @@ const StyledLabel = styled.label(
     '@media (min-width: 768px)': {
       fontSize: 2,
     },
+
+    '@media (max-width: 390px)': {
```
VWSCoronaDashboard8, line 51: I think it's probably better to use the theme defined breakpoints and set it using array (or mq object) syntax directly on the "p" property above. 

The same goes for the other @media above it. This approach is bypassing the styled-system way of doing things and the theme mq definitions

VWSCoronaDashboard8 (approved):

PR #1418, opened 2021-01-12
Show just the one line if only one installation exists in the sewerwater chart

VWSCoronaDashboard1:
## Summary

As the title states, just show the line, skip the scatter plot if only one installation exists.

## Motivation

Change request

## Detailed design

n/a

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added

### DIFF
--- a/packages/app/src/components/lineChart/hooks/useSewerWaterChartOptions.ts
+++ b/packages/app/src/components/lineChart/hooks/useSewerWaterChartOptions.ts
@@ -300,8 +300,6 @@ export function useSewerWaterChartOptions<
         formatter: function (): false | string {
           const tooltipType = tooltipTypes[this.series.name];
 
-          console.log('nee');
-
           if (!tooltipType) {
             return false;
           }
--- a/packages/app/src/components/lineChart/hooks/useSewerWaterChartOptions.ts
+++ b/packages/app/src/components/lineChart/hooks/useSewerWaterChartOptions.ts
@@ -1,3 +1,4 @@
+import { uniq } from 'lodash';
 import { useMemo } from 'react';
 import { colors } from '~/style/theme';
 import { formatDateFromSeconds } from '~/utils/formatDate';
@@ -113,14 +114,8 @@ export function useSewerWaterChartOptions<
       [text.secondary_label_text]: 'scatter',
     };
 
-    const installationNames = filteredScatterPlotValues.reduce<string[]>(
-      (acc, item) => {
-        if (acc.indexOf(item.rwzi_awzi_name) < 0) {
-          acc.push(item.rwzi_awzi_name);
-        }
-        return acc;
-      },
-      []
+    const installationNames = uniq(
+      filteredScatterPlotValues.map((x) => x.rwzi_awzi_name)
     );
 
     const series: (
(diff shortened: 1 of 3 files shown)


packages/app/src/components/lineChart/hooks/useSewerWaterChartOptions.ts
```diff
@@ -113,47 +113,62 @@ export function useSewerWaterChartOptions<
...
+    const installationNames = filteredScatterPlotValues.reduce<string[]>(
+      (acc, item) => {
+        if (acc.indexOf(item.rwzi_awzi_name) < 0) {
+          acc.push(item.rwzi_awzi_name);
+        }
+        return acc;
       },
-    };
+      []
+    );
```
VWSCoronaDashboard8, line 124: I think this is much simpler:

```
const installationNames = uniq(filteredScatterPlotValues.map(x => x.rwzi_awzi_name))
```

uniq is a lodash function

VWSCoronaDashboard8 (approved):

packages/app/src/components/lineChart/hooks/useSewerWaterChartOptions.ts
VWSCoronaDashboard1: Done

packages/app/src/pages/gemeente/[code]/rioolwater.tsx
```diff
@@ -53,6 +53,18 @@ const SewerWater: FCWithLayout<IMunicipalityData> = (props) => {
...
+   * sewer station
+   */
+  const enableScatterPlot = useMemo(() => {
+    if (sewerStationNames.length === 1) {
+      setSelectedInstallation(sewerStationNames[0]);
+    }
+
+    return sewerStationNames.length > 1;
+  }, [sewerStationNames]);
+
```
Comment (VWSCoronaDashboard10) [inline packages/app/src/pages/gemeente/[code]/rioolwater.tsx:67]: I don't really get this change:
- The expression `sewerStationNames.length > 1` is not that hard to calculate, why the move to a useMemo-hook?
- I don't think we should mutate state during the render of a component. Mutation can happen within callbacks or an effect-hook.

VWSCoronaDashboard10 (changes requested):

packages/app/src/pages/gemeente/[code]/rioolwater.tsx
Comment (VWSCoronaDashboard10) [reply packages/app/src/pages/gemeente/[code]/rioolwater.tsx:67]: I guess what you actually want is something like this:

```diff
- const [selectedInstallation, setSelectedInstallation] = useState<string>();
+ const [selectedInstallation, setSelectedInstallation] = useState<string>(sewerStationNames.length === 1 ? sewerStationNames[0] : undefined);

Comment (VWSCoronaDashboard10) [reply packages/app/src/pages/gemeente/[code]/rioolwater.tsx:67]: which would still be buggy the moment `data` is becoming dynamic (that would require an effect which runs when `sewerStationNames` mutates). But it is not, and we're on page-level, so I think it's safe enough.

Comment (VWSCoronaDashboard1) [reply packages/app/src/pages/gemeente/[code]/rioolwater.tsx]: Fixed

PR #1427, opened 2021-01-13
Renamed difference indicator text to make it more clear

VWSCoronaDashboard14:
Renamed the difference indicator text to make it more clear for the user.
So "3 dagen geleden" becomes "3 dagen ervoor" as example.

### DIFF
--- a/packages/app/src/locale/en.json
+++ b/packages/app/src/locale/en.json
@@ -1376,7 +1376,8 @@
         "enkelvoud": "week",
         "meervoud": "weeks"
       },
-      "hiervoor": "before"
+      "vorige_week": "last week",
+      "ervoor": "before"
     }
   },
   "aria_labels": {
--- a/packages/app/src/pages/landelijk/reproductiegetal.tsx
+++ b/packages/app/src/pages/landelijk/reproductiegetal.tsx
@@ -68,7 +68,7 @@ const ReproductionIndex: FCWithLayout<NationalPageProps> = (props) => {
               localeTextKey="reproductiegetal"
               differenceKey="reproduction__index_average"
               differenceStaticTimespan={
-                siteText.toe_en_afname.tijdverloop.hiervoor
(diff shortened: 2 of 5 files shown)


VWSCoronaDashboard10 (approved):

packages/app/src/locale/en.json
```diff
@@ -1376,7 +1376,8 @@
         "enkelvoud": "week",
         "meervoud": "weeks"
       },
-      "hiervoor": "before"
```
VWSCoronaDashboard8, line 1379: are you sure this key can be removed?

packages/app/src/pages/landelijk/reproductiegetal.tsx
```diff
@@ -68,7 +68,7 @@ const ReproductionIndex: FCWithLayout<NationalPageProps> = (props) => {
               localeTextKey="reproductiegetal"
               differenceKey="reproduction__index_average"
               differenceStaticTimespan={
-                siteText.toe_en_afname.tijdverloop.hiervoor
+                siteText.toe_en_afname.tijdverloop.vorige_week
```
VWSCoronaDashboard8, line 71: I don't think reproductie getal is reported weekly, so this would be a mistake

VWSCoronaDashboard8 (changes requested):

packages/app/src/locale/en.json
VWSCoronaDashboard8: looks like it will still be required for reproductiegetal

packages/app/src/pages/landelijk/reproductiegetal.tsx
VWSCoronaDashboard8: Maybe it is, I'm not completely sure. Never mind. I'll let QA decide

VWSCoronaDashboard8 (approved):

PR #1434, opened 2021-01-13
Export images from Sanity and resize them locally

VWSCoronaDashboard:
## Summary

This PR adds scripts to automatically [download images from Sanity](https://www.sanity.io/docs/migrating-data#1-export) to the filesystem and resize them with [Sharp](https://sharp.pixelplumbing.com/).

It adds:

1. A `yarn image:download` command
2. A `yarn image:resize` command
3. A `yarn image` command that calls download and resize sequentially through [npm-run-all](https://www.npmjs.com/package/npm-run-all)

## Motivation

In our CMS, content managers can upload images. Due to security and technical availability concerns, we want to serve these images from our own server behind Akamai. 

We also want to resize/optimize these images automatically, because serving original images will lead to very high bandwidth.

## Detailed design

**Image downloading**

First, I `rm -rf` all folders containing the raw export, original sanity images and resized images. 

We call `sanity dataset export development|production export` from the CMS folder. We do this through a little bash script that also unzips the file and moves the assets to `packages/app/public/sanity/`

**Image resizing**

Since Next.js choose Sharp for image resizing, I made the same choice. We could have chosen ImageMagick too, but Sharp is easier to install through NPM and seems to be faster and actively maintained.

We loop over all the files in `packages/app/public/sanity/`. For every file we find, we loop over the sizes array and write new files to the filesystem.

```js

// These are the sizes we want to resize our original images to
const sizes = [320, 640, 768, 1024, 1280, 1536, 2048];

// looping over files here, and for every file we do this
const ext = path.extname(file);
const filename = path
  .basename(file)
  .split(".")
  .slice(0, -1)
  .join(".");
console.log(`Now resizing: ${file}`);

sizes.forEach((size) => {
  const output = `../app/public/sanity/${filename}-${size}${ext}`;
  sharp(file)
    .resize({ width: size, withoutEnlargement: true })
    .toFile(output);
});
```

## Drawbacks

Not a drawback, but something I wanted to flag for you to review:

I originally wanted to put the scripts in `packages/cli`. However, it was easier to call the scripts directly in `packages/cms` for 2 reasons:

1. We already have the required packages installed there
2. We already have the required environment variables for authentication to Sanity configured through dotfiles. This is useful for local development.

## Alternatives

- On development servers we can talk directly to Sanity CDN. There's no need to download the assets and resize them. I still need to implement this.

- `next/image` would serve this out of the box, but it's not compatible with static exports. We would need to start running the production server on a Node environment.

## Unresolved questions / open items

Things we still need to implement while this PR is already being reviewed:

- [ ] Calling Sanity CDN in development and local environments from the front-end.
- [ ] Call the resize and download commands in the production pipeline
- [x] ~Either use `next/image` with a custom loader if that works with static exports, or~ build our own image component


packages/app/.gitignore
```diff
@@ -4,6 +4,9 @@
 /.next
 /out
 
+# Images that are processed by Sanity/Resizing
+public/sanity
```
VWSCoronaDashboard8, line 8: nit; for the outside world it doesn't matter what cms we use, so I think it would make sense to give this a more generic name so it doesn't become part of urls. And in theory a cms could change as well.

packages/cms/.gitignore
```diff
@@ -5,8 +5,11 @@
 /.pnp
 .pnp.js
 
-# build
+# build and data exports
 dist
+export
+development.tar.gz
+production.tar.gz
```
VWSCoronaDashboard8, line 12:
```suggestion
*.tar.gz
```

I think that works as well

packages/cms/package.json
```diff
@@ -11,7 +11,10 @@
     "test": "sanity check",
     "test:ci": "yarn run test",
     "compile": "echo 'Placeholder to support yarn compile in repo root, until we use TS'",
-    "lint": "echo '@TODO add eslint'"
+    "lint": "echo '@TODO add eslint'",
+    "images": "run-s images:download images:resize",
+    "images:download": "bash scripts/images.sh",
```
VWSCoronaDashboard8, line 16: If `scripts/images` is mainly for download, then I would call it download. Or call one file `images-download.sh` and the other `images-resize.sh`

packages/cms/scripts/images.sh
```diff
@@ -0,0 +1,17 @@
...
+sanity dataset export production production.tar.gz --overwrite
+tar -xzf *.gz 
+mv production-export* export
+
+# Rename sha1-widthxheight.extension to sha1.extension
+cd export/images
+for f in *; do mv "$f" "`echo $f | sed s/\-[0-9]*x[0-9]*//g`"; done;
+
+# Put images in the public images folder of application.
+mkdir ../../../app/public/sanity
```
VWSCoronaDashboard8, line 16: Is this relative to CWD or file location? I suspect the former, which could be problematic with these paths. 

A solution could be to pass in the path via a command line argument

packages/cms/scripts/images.sh
```diff
@@ -0,0 +1,17 @@
+# Start with a clean slate
```
VWSCoronaDashboard8, line 1: I would place a hashbang here to indicate what to run it with then you don't have to specify bash in the calling of the script.

I think the correct way would be: `#!/usr/bin/env bash`, which makes the OS find bash no matter where it is installed.

packages/cms/scripts/resize.js
```diff
@@ -0,0 +1,83 @@
+const sharp = require("sharp");
```
VWSCoronaDashboard8, line 1: I would prefer to keep all sources Typescipt. You can execute it with ts-node.

packages/cms/scripts/resize.js
```diff
@@ -0,0 +1,83 @@
...
+const fs = require("fs");
+const path = require("path");
+
+const walkPath = "../app/public/sanity";
+
+// These are the sizes we want to resize our original images to
+const sizes = [320, 640, 768, 1024, 1280, 1536, 2048];
+
+const walk = function (dir, done) {
```
VWSCoronaDashboard8, line 10: nit; Prefer (named) function declarations over function expression

packages/cms/scripts/resize.js
```diff
@@ -0,0 +1,83 @@
...
+  fs.readdir(dir, function (error, list) {
+    if (error) {
+      return done(error);
+    }
+
+    var i = 0;
+
+    (function next() {
```
VWSCoronaDashboard8, line 18: Why are you using an IIFE here? I'm guessing you copy-pasted this since it looks also like ES5 code with the vars and such ;)

packages/cms/scripts/resize.js
```diff
@@ -0,0 +1,83 @@
...
+  });
+};
+
+// optional command line params
+//      source for walk path
+process.argv.forEach(function (val, index, array) {
+  if (val.indexOf("source") !== -1) {
+    walkPath = val.split("=")[1];
+  }
+});
```
VWSCoronaDashboard8, line 63: I would suggest something like `mew` for managing your command line arguments

packages/cms/scripts/resize.js
```diff
@@ -0,0 +1,83 @@
...
+
+console.log("-------------------------------------------------------------");
+console.log(`We're going to resize images in ${walkPath}...`);
+console.log("-------------------------------------------------------------");
+
+// Walk through the sanity image directory and report
+// that we're done or throw an error
+walk(walkPath, function (error) {
```
VWSCoronaDashboard8, line 71: An error callback function is quite ancient NodeJs style. We can use promises and possibly async/await.

packages/app/.gitignore
VWSCoronaDashboard7: Agree on the above!

VWSCoronaDashboard: Agreed. I changed it to be `public/cms` to at least indicate where it's coming from.

packages/cms/package.json
VWSCoronaDashboard: Thanks, I did exactly that.

packages/cms/scripts/images.sh
VWSCoronaDashboard: Yes, this is all relative to CWD, which is pretty annoying.

packages/cms/scripts/images.sh
VWSCoronaDashboard: Done!

packages/cms/scripts/resize.js
VWSCoronaDashboard: Done!

packages/cms/scripts/resize.js
VWSCoronaDashboard: Agree, and fixed.

packages/cms/scripts/resize.js
VWSCoronaDashboard: Oh this is definitely copy-pasted! 🥇

packages/cms/scripts/resize.js
VWSCoronaDashboard: I copy pasted ancient Nodejs stuff. Do you want me to rewrite this to modern JS or are you happy to leave it as-is?

packages/cms/.gitignore
VWSCoronaDashboard: I think this has been resolved in the branch already.

packages/cms/scripts/resize.js
VWSCoronaDashboard: You know what, I simply removed this block of code completely. I don't think we'll miss it.

packages/cms/scripts/resize.js
VWSCoronaDashboard1: > I copy pasted ancient Nodejs stuff. Do you want me to rewrite this to modern JS or are you happy to leave it as-is?

That's a silly question to ask to @VWSCoronaDashboard8 ;)

packages/cms/scripts/images-resize.ts
```diff
@@ -0,0 +1,75 @@
...
+          const ext = path.extname(file);
+          const filename = path
+            .basename(file)
+            .split(".")
+            .slice(0, -1)
+            .join(".");
+          console.log(`Now resizing: ${file}`);
+
+          sizes.forEach((size) => {
+            const output = `../app/public/cms/${filename}-${size}${ext}`;
```
VWSCoronaDashboard1, line 44: I see this path sprinkled around the codebase quite a bit. Maybe it's an idea to stick this in a config object that you can import in all the appropriate places.

VWSCoronaDashboard: Eh, it's not worth the effort. All the instances are different.

I did however add a config.js object so I only declare the sizes array in one place!

VWSCoronaDashboard1 (approved):

packages/app/src/utils/findClosestSize.ts
```diff
@@ -0,0 +1,15 @@
+const config = require('~/config.js');
```
VWSCoronaDashboard8, line 1: Why is the config a js file? And I think you can then still use a normal import statement

packages/cms/scripts/images-download.sh
```diff
@@ -0,0 +1,20 @@
...
+tar -xzf *.gz 
+mv *-export* export
+
+# Rename sha1-widthxheight.extension to sha1.extension
+cd export/images
+for f in *; do mv "$f" "`echo $f | sed s/\-[0-9]*x[0-9]*//g`"; done;
+
+# Put images in the public images folder of application.
+mkdir ../../../app/public/cms
+mv * ../../../app/public/cms
```
VWSCoronaDashboard8, line 20: Fine for now but I think it makes sense to just rewrite this in TS later so that it can be more understandable and robust. Where we can use __dirname and other techniques that everyone is familiar with.

Then execute that script with `ts-node`

packages/cms/scripts/images-resize.ts
```diff
@@ -0,0 +1,77 @@
+const sharp = require("sharp");
+const fs = require("fs");
+const path = require("path");
+const config = require("../../app/src/config.js");
```
VWSCoronaDashboard8, line 4: A package should not have direct imports from other packages like this. It kind of defeats the idea of a monorepo. 

You can now place this in common maybe.

VWSCoronaDashboard8 (changes requested):

packages/cms/scripts/resize.js
VWSCoronaDashboard8: Well I am fine with simply having a cleanup / convert round at the end of this sprint, so doing it in a separate PR.

packages/app/src/utils/findClosestSize.ts
VWSCoronaDashboard: Guess it should be TS? 👍 
I'll try the import.

packages/cms/scripts/images-resize.ts
VWSCoronaDashboard: Good one. I'll try rewriting this.

packages/app/src/pages/over.tsx
```diff
@@ -1,23 +1,34 @@
...
+  };
+  imageContent: {
+    coverImage: SanityImageProps;
+  };
 }
 
 const query = `
-*[_type == 'overDitDashboard'][0]
+"over": *[_type == 'overDitDashboard'][0],
+"imageContent": *[_type == 'imagePipelineTest'][0]{
```
VWSCoronaDashboard1, line 28: This reeks of test code, are we sure we want this merged?

VWSCoronaDashboard1 (approved):

packages/app/src/pages/over.tsx
VWSCoronaDashboard: Good one. I can remove it and add code to the article pages to test out the image pipeline.

packages/app/src/components-styled/image.tsx
```diff
@@ -0,0 +1,46 @@
...
+/**
+ *
+ * This is our Image component. It returns a set of rezided images
+ * in a srcSet. Ideally we'll use next/image as soon as it's
+ * compatible with next export.
+ */
+export function Image(props: ImageProps) {
+  const { src, width, height, alt } = props;
+  const filename = src.split('.')[0];
+  const extension = src.split('.')[1];
```
VWSCoronaDashboard8, line 26:
```suggestion
  const [filename, extension] = src.split('.');
```

packages/app/src/components-styled/image.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  const srcSet = sizes
+    .map((size: number) => {
+      return `/cms${filename}-${size}.${extension} ${size}w`;
+    })
+    .join(', ');
+
+  const url = `/cms${filename}-${findClosestSize(width)}.${extension}`;
```
VWSCoronaDashboard8, line 34: Looks like you are attaching the word `cms` to the `filename`. Why not put a dash in between?

VWSCoronaDashboard8: If filename includes a leading slash then it's technically not a filename, but I would be ok with adding a comment to inform the reader.

packages/app/src/components-styled/image.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ *
+ * This is our Image component. It returns a set of rezided images
+ * in a srcSet. Ideally we'll use next/image as soon as it's
+ * compatible with next export.
+ */
+export function Image(props: ImageProps) {
+  const { src, width, height, alt } = props;
+  const filename = src.split('.')[0];
```
VWSCoronaDashboard4, line 25:
```
const [filename, extension] = src.split('.')
```

packages/app/src/pages/over.tsx
```diff
@@ -1,23 +1,34 @@
...
+import { createGetStaticProps } from '~/static-props/create-get-static-props';
 import {
   createGetContent,
   getLastGeneratedDate,
 } from '~/static-props/get-data';
-import { createGetStaticProps } from '~/static-props/create-get-static-props';
-import styles from './over.module.scss';
+import { Image } from '~/components-styled/image';
 import { PortableTextEntry } from '@sanity/block-content-to-react';
+import { SanityImageProps } from '~/types/cms.d.ts';
```
VWSCoronaDashboard8, line 13: nit; The ts part will be resolved just like other ts files.

```suggestion
import { SanityImageProps } from '~/types/cms.d';
```

packages/app/src/pages/over.tsx
```diff
@@ -47,10 +60,15 @@ const Over: FCWithLayout<typeof getStaticProps> = (props) => {
...
         <MaxWidth>
           <div className={styles.maxwidth}>
-            {content.title && <h2>{content.title}</h2>}
-            {content.description && (
-              <PortableText blocks={content.description} />
-            )}
+            {over.title && <h2>{over.title}</h2>}
+
+            <Image
+              src={`/${coverImage.assetId}.${coverImage.extension}`}
```
VWSCoronaDashboard8, line 66: If the leading slash in my previous comment is coming from here, I suggest to remove it here and add it in the other place. To me that makes more sense and is clear to the user. It's uncommon anyway to start a path with / unless you mean the unix root.

packages/cms/scripts/images-resize.ts
```diff
@@ -0,0 +1,76 @@
+const sharp = require('sharp');
+const fs = require('fs');
+const path = require('path');
+
+import { sizes } from '@corona-dashboard/common/src/config';
+
+const walkPath = '../app/public/cms';
+
+// These are the sizes we want to resize our original images to
```
VWSCoronaDashboard4, line 9: This comment and other comments in this file do not make much sense. Perhaps have a closer look at the comments and add some that clarify what is going on, where needed?

packages/app/src/utils/findClosestSize.ts
```diff
@@ -0,0 +1,15 @@
+import { sizes } from '@corona-dashboard/common/src/config';
+
+// find closest resized element
+export function findClosestSize(width: number) {
```
VWSCoronaDashboard8, line 4: This function looks like quite specific domain logic and not really a utility belt function in that sense. I would move it to the cms folder, maybe inside a `logic` subfolder or `logic.ts` file. Not sure if you were there when we discussed file structure on Friday. That was about components but I think a similar way of thinking can be applied here.

Also I think the function would be a little better if you passed in sizes as the first argument. Then it's immediately clear what it does from the calling context, and you also don't need the comment.

packages/app/src/utils/imageLoader.ts
```diff
@@ -0,0 +1,15 @@
...
+
+interface LoaderProps {
+  src: string;
+  width: number;
+}
+
+export function imageLoader(props: LoaderProps) {
+  const { src, width } = props;
+  const filename = src.split('.')[0];
+  const extension = src.split('.')[1];
```
VWSCoronaDashboard8, line 11: Same suggestion as previously

packages/cms/scripts/images-resize.ts
```diff
@@ -0,0 +1,76 @@
+const sharp = require('sharp');
+const fs = require('fs');
+const path = require('path');
+
+import { sizes } from '@corona-dashboard/common/src/config';
```
VWSCoronaDashboard8, line 5: If config is something to be imported from common, then the common package should export it. It's not ok for one package to directly import source files from another package. The public facing files of common live in `dist` and are exposed via the package definitions. 

You probably just need to add the config to the exports list in src/index.ts

```suggestion
import { sizes } from '@corona-dashboard/common';
```

packages/common/src/config.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * This is the list of sizes we use to generate resized images
+ * and srcSets in our own image component
+ */
+export const sizes = [320, 640, 768, 1024, 1280, 1536, 2048];
```
VWSCoronaDashboard8, line 5: Since `common` and `config` don't have a clear context, I think it makes sense to be a little more specific here and call them `imagesSizes` or `resizeImageSizes` even.

VWSCoronaDashboard8 (changes requested):

packages/app/src/pages/over.tsx
VWSCoronaDashboard1: You can leave me `.d` part as well

packages/common/src/config.ts
VWSCoronaDashboard: Sounds good to me!

packages/app/src/components-styled/image.tsx
VWSCoronaDashboard: Yup, it's because of a trailing slash. I'll add some comments here.

packages/app/src/pages/weekberichten/index.tsx
```diff
@@ -0,0 +1,12 @@
+import { getLayoutWithMetadata } from '~/domain/layout/layout';
```
VWSCoronaDashboard, line 1: I don't know why this file is here or what it's for. It must've happened when we change bases?

packages/app/src/utils/findClosestSize.ts
VWSCoronaDashboard: I agree with your second point and have changed the code to reflect that. 👍 

I disagree with your first point, as the utility function is closely coupled to the front-end application.

packages/cms/scripts/images-resize.ts
VWSCoronaDashboard: I've added the source of the code as a comment and explained it's unoptimized, but working code.

VWSCoronaDashboard8 (approved):

packages/app/src/pages/weekberichten/index.tsx
VWSCoronaDashboard1: I deleted this file, so it looks like a faulty merge. You can remove it as well.

VWSCoronaDashboard1 (approved):

PR #1454, opened 2021-01-14
Topic/actueel artikelen

VWSCoronaDashboard4:
## Summary

Adds Articles structure to the CMS + basic routing and interfaces for the frontend.

## Detailed design

* Adds schemes in CMS, including types to combine images/graphs in rich text.
* Rich text is limited for now, removing options for H1, H4-H6 and other elements we don't want to use there.
* Basic TS Interfaces to work with the Articles, with Todos added to add more specifics later on.
* Basic routing: `/artikelen` and `/artikelen/[slug]`

# Left for a separate PR

* Validation of uniqueness of the slugs
* More detailed TypeScript interfaces


packages/app/src/pages/artikelen/[slug].tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  const filePath = path.join(process.cwd(), 'public', 'json', 'NL.json');
+  const fileContents = fs.readFileSync(filePath, 'utf8');
+  const lastGenerated = JSON.parse(fileContents).last_generated;
+
+  const articleData = await getClient(true).fetch(articleQuery(params.slug));
+  const article = localize(articleData, [targetLanguage, 'nl']);
+
+  return { props: { text, lastGenerated, article } };
+}
```
Comment (VWSCoronaDashboard1) [inline packages/app/src/pages/artikelen/[slug].tsx:58]: I'm guessing this will need to be changed according to the changes made by @VWSCoronaDashboard10 concerning the data loading?

packages/app/src/pages/artikelen/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+`;
+
+export async function getStaticProps(): Promise<StaticProps> {
+  const text = parseMarkdownInLocale(
+    (await import('../../locale/index')).default
+  );
+
+  const filePath = path.join(process.cwd(), 'public', 'json', 'NL.json');
+  const fileContents = fs.readFileSync(filePath, 'utf8');
+  const lastGenerated = JSON.parse(fileContents).last_generated;
```
VWSCoronaDashboard1, line 32: There's a handy little util for this:
https://github.com/minvws/nl-covid19-data-dashboard/blob/topic/actueel-artikelen/packages/app/src/static-props/utils/load-json-from-data-file.ts

packages/cms/schemas/documents/artikel.js
```diff
@@ -0,0 +1,70 @@
...
+      title: "Content",
+      name: "content",
+      type: "localeRichContentBlock"
+    }
+  ],
+  preview: {
+    select: {
+      title: "title.nl",
+      subtitle: "publicationDate",
+    },
```
VWSCoronaDashboard1, line 68: Doesn't the preview also need an image?

VWSCoronaDashboard1 (approved):

packages/app/src/pages/artikelen/[slug].tsx
Comment (VWSCoronaDashboard4) [reply packages/app/src/pages/artikelen/[slug].tsx:58]: Yes!

packages/cms/schemas/documents/artikel.js
VWSCoronaDashboard4: This is the preview in the CMS, not in the frontend.

VWSCoronaDashboard1: Aha, makes sense then!

packages/cms/schemas/locale/locale-rich-content-block.js
```diff
@@ -0,0 +1,43 @@
+
+const supportedLanguages = [
+  { id: "nl", title: "Nederlands", isDefault: true },
+  { id: "en", title: "Engels" },
+];
+
+export default {
+  name: "localeRichContentBlock",
+  type: "object",
+  title: "Locale Rich Content Block",
```
VWSCoronaDashboard8, line 10: Other titles seem to be in Dutch, and apart from that I wonder what a locale rich content block means. Maybe add a comment to explain. I see it also can include chart configs?

packages/cms/schemas/objects/line-chart.js
```diff
@@ -0,0 +1,22 @@
+export default {
```
VWSCoronaDashboard8, line 1: Maybe add a comment to explain what this is for. A line chart config with metric name and property doesn't appear to be quite enough to instantiate an actual line chart.

packages/cms/schemas/schema.js
```diff
@@ -31,16 +34,19 @@ export default createSchema({
   types: schemaTypes.concat([
     //documents
```
VWSCoronaDashboard8, line 35: I prefer to use block comments for actual comments, so they are more visible and also distinguishable from stuff that is just commented out.

packages/cms/schemas/locale/locale-rich-content-block.js
VWSCoronaDashboard1: Perhaps `localizedRichContentBlock` makes more sense?

VWSCoronaDashboard:
Generic feedback without looking at your code:

1. Good call to put metadata content in a fieldset. I've got an open-graph model defined already, so we can use that one for metadata. I think we should check if we can make this UI better by adding the sanity-tabs plugin!

2. I'm missing alt text on the image. We need to add those.

3. I simultaneously love and worry about the "lijngrafiek" in content blocks. 

It's impressive that it can work like this, and I believe the authoring flow would be nice. However, because we repeat it for the english version it almost feels like we're opening up the editor too much.

Maybe a better approach is to use document level translations for artikelen, and keep using field level translations for other documents. Of course that introduces the new problem that you need to re-enter your slug and images for translated posts, which is also less than ideal.

Another thing is that it relies on front-end specific knowledge to couple the charts. Maybe we can improve that too. I can imagine that we keep a collection of links between the CMS and front-end ourselves as a collection of tiles/charts in the CMS and let the content manager choose one from a predefined list?

packages/app/src/pages/artikelen/[slug].tsx
```diff
@@ -0,0 +1,73 @@
...
+
+export async function getStaticProps({
+  params,
+}: {
+  params: Params;
+}): Promise<StaticProps> {
+  const text = parseMarkdownInLocale(
+    (await import('../../locale/index')).default
+  );
```
Comment (VWSCoronaDashboard) [inline packages/app/src/pages/artikelen/[slug].tsx]: Does the UI copy still need markdown parsing? If things are right, all rich text now comes from the CMS and is parsed by the Sanity Block component.

packages/app/src/pages/artikelen/[slug].tsx
```diff
@@ -0,0 +1,73 @@
...
+  lastGenerated: string;
+}
+
+const articlesQuery = groq`
+*[_type == 'artikel']
+`;
+
+const articleQuery = (slug: string) => groq`
+*[_type == 'artikel' && slug.current == '${slug}'][0]
+`;
```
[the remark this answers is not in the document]
Comment (VWSCoronaDashboard) [reply packages/app/src/pages/artikelen/[slug].tsx]: nit: To reduce bandwidth we could look into specifying all fields up front, which means you don't get the fields that you don't need. Should save a few bytes.

packages/app/src/pages/artikelen/[slug].tsx
```diff
@@ -0,0 +1,73 @@
...
+  return { props: { text, lastGenerated, article } };
+}
+
+const Artikelen: FCWithLayout<ArtikelProps> = (props) => {
+  const { article } = props;
+
+  return <h1>{article.title}</h1>;
+};
+
+const metadata = {
```
Comment (VWSCoronaDashboard) [inline packages/app/src/pages/artikelen/[slug].tsx:51]: Fine with todo's, but make sure to write them down in the topic/actueel-base PR.

packages/app/src/types/cms.d.ts
```diff
@@ -7,3 +7,41 @@ export type CollapsibleList = {
   };
   title: string;
 };
+
+export interface Article {
+  title: string;
+  slug: {
+    _key: string;
+    _type: string;
+    current: string;
```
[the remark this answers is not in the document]
VWSCoronaDashboard: This is how Sanity stores data that's of a `slug` type:

```
{
  "_type": "slug",
  "current": "this-is-the-title"
}
```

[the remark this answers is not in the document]
VWSCoronaDashboard4: This is the default sanity slug object, since in the sanity schema we use the build in slug component.

packages/app/src/types/cms.d.ts
```diff
@@ -8,6 +8,44 @@ export type CollapsibleList = {
...
+type RichContentBlock =
+  | Block
+  | RichContentImageBlock
+  | RichContentLineChartBlock;
+
+// @TODO more properties are needed
+interface Block {
+  _key: string;
+  _type: string;
+}
```
VWSCoronaDashboard, line 33: These are hard to type. We'll need to spend more time figuring out how to deal with dynamic block content.

packages/cms/schemas/objects/line-chart.js
VWSCoronaDashboard4: I'll add a Todo. This will indeed need to be extended in a separate PR.

packages/cms/schemas/locale/locale-rich-content-block.js
VWSCoronaDashboard: The other block is also localized, and also has rich content. This one has reduced formatting options and adds new allowed types.

You might need to call this for what it is, a Block for rticles.

`articleBlock`?

packages/cms/schemas/locale/locale-rich-content-block.js
```diff
@@ -0,0 +1,43 @@
...
+    of: [
+      {
+        type: "block",
+
+        // Only allow these block styles
+        styles: [
+          {title: 'Normal', value: 'normal'},
+          {title: 'H2', value: 'h2'},
+          {title: 'H3', value: 'h3'}
+        ],
```
VWSCoronaDashboard, line 24: This is great for SEO! 👏

packages/cms/schemas/locale/locale-rich-content-block.js
VWSCoronaDashboard4: This is perhaps food for a different/broader discussion.

The `title` is merely what is displayed in the CMS, I'll tweak that a bit.

The `name` is how it is referred to within other blocks/schema's. There this camelcasing is used. It is the structure that supports one locale (either EN or NL in this case). See the other files in this directory.

packages/app/src/types/cms.d.ts
```diff
@@ -7,3 +7,41 @@ export type CollapsibleList = {
...
+// @TODO more properties are needed
+interface RichContentImageBlock {
+  _key: string;
+  _type: 'image';
+}
+
+// @TODO more properties are needed
+interface RichContentLineChartBlock {
+  _key: string;
+  _type: 'lineChart';
```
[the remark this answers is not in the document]
VWSCoronaDashboard4: All other CMS types are following the camelCase structure so far.

packages/cms/schemas/objects/line-chart.js
VWSCoronaDashboard: It's also not _really_ a linechart. It serves as a connection between front-end values and the CMS.

This is probably one of the more difficult models for the CMS. Let's take the time to get it right, even if it means launching without charts initially?

packages/cms/schemas/objects/line-chart.js
```diff
@@ -0,0 +1,22 @@
...
+      title: "Metrieknaam (code)",
+      name: "metricName",
+      type: "string"
+    },
+    {
+      title: "Metriekwaarde (code)",
+      name: "metricProperty",
+      type: "string"
+    }
+  ],
```
VWSCoronaDashboard, line 22: Technically, this makes sense... I'll give you that.

packages/app/src/types/cms.d.ts
```diff
@@ -7,3 +7,41 @@ export type CollapsibleList = {
...
 };
+
+export interface Article {
+  title: string;
+  slug: {
+    _key: string;
+    _type: string;
+    current: string;
+  };
+  intro: Block | null;
```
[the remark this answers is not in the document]
VWSCoronaDashboard4: It's a sanity `Block` type

packages/cms/schemas/documents/artikel.js
VWSCoronaDashboard: You actually can add the image in the preview too. It will help recognize articles when listing them. I'm voting that we should add this.

packages/cms/schemas/locale/locale-rich-content-block.js
VWSCoronaDashboard: I don't think this block should ever be used outside of articles though. That's why I suggest you name it in reference to that.

You can also move the file to be part of a `articles` schema folder. Co-locate those things to make it clear they are related.

packages/app/src/pages/artikelen/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  const articlesData = await getPreviewClient().fetch(articlesQuery);
```
VWSCoronaDashboard: You should not use previewClient here. I'm seeing both the original article and a draft on this page.

VWSCoronaDashboard (commented):
Lot's of small comments and questions. No real blockers, especially if we agree that this needs more refinement afterwards on develop or actueel-base. Also, no real approval because this isn't _done_-done.

Happy to see this!

packages/app/src/pages/artikelen/[slug].tsx
```diff
@@ -0,0 +1,58 @@
...
+  getLastGeneratedDate,
+  createGetContent<Article[]>((context) => {
+    assert(context?.params?.slug, 'Slug required to retrieve article');
+    return groq`
+  *[_type == 'artikel' && slug.current == '${context.params.slug}'][0]
+  `;
+  })
+);
+
+const ArticleDetail: FCWithLayout<ArticleDetailProps> = (props) => {
```
Comment (VWSCoronaDashboard10) [inline packages/app/src/pages/artikelen/[slug].tsx:45]: You can infer the props from the getInitialProps:

```suggestion
const ArticleDetail: FCWithLayout<typeof getStaticProps> = (props) => {
```

packages/app/src/pages/artikelen/[slug].tsx
```diff
@@ -0,0 +1,58 @@
...
+  const { content } = props;
+
+  return <h1>{content.title}</h1>;
+};
+
+const metadata = {
+  title: 'TODO',
+  description: 'TODO',
```
Comment (VWSCoronaDashboard10) [inline packages/app/src/pages/artikelen/[slug].tsx:53]:
```suggestion
  title: '@TODO',
  description: '@TODO',
```
makes it easier to find them :-)

packages/app/src/pages/artikelen/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+);
+
+const ArticlesOverview: FCWithLayout<ArticlesOverviewProps> = (props) => {
+  const { content } = props;
+
+  return (
+    <>
+      <h1>Todo: Artikelen!</h1>
+      {content.map((article) => (
+        <div key={`article-${article.slug.current}`}>
```
VWSCoronaDashboard10, line 32:
```suggestion
        <div key={article.slug.current}>
```

packages/app/src/pages/artikelen/[slug].tsx
```diff
@@ -0,0 +1,73 @@
...
+*[_type == 'artikel']
+`;
+
+const articleQuery = (slug: string) => groq`
+*[_type == 'artikel' && slug.current == '${slug}'][0]
+`;
+
+export async function getStaticPaths() {
+  const articlesData = await getClient(true).fetch(articlesQuery);
+  const articles = localize(articlesData, [targetLanguage, 'nl']);
```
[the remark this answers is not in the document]
Comment (VWSCoronaDashboard) [reply packages/app/src/pages/artikelen/[slug].tsx]: In return you get localized data. Localize takes a look at your targetLanguage and returns only the fields that are either .en or .nl.

packages/cms/schemas/documents/artikel.js
```diff
@@ -0,0 +1,70 @@
+export default {
+  title: "Artikel",
+  name: "artikel",
```
VWSCoronaDashboard10, line 3: Not sure what the `name` property is meant for, but we might want them to be English?

VWSCoronaDashboard10: Looks like a lot of other `name` properties in this file are English

packages/app/src/pages/artikelen/[slug].tsx
```diff
@@ -0,0 +1,58 @@
...
+  content: Article;
+  lastGenerated: string;
+}
+
+const articlesQuery = groq`
+*[_type == 'artikel']
+`;
+
+export async function getStaticPaths() {
+  const articlesData = await getPreviewClient().fetch(articlesQuery);
```
Comment (VWSCoronaDashboard) [inline packages/app/src/pages/artikelen/[slug].tsx:24]: Do not use the preview client.

VWSCoronaDashboard (changes requested):

packages/app/src/static-props/get-data.ts
```diff
@@ -42,8 +42,14 @@ export function getLastGeneratedDate() {
...
 
-export function createGetContent<T>(query: string) {
-  return async () => {
+export function createGetContent<T>(
+  queryOrQueryGetter: string | ((context: GetStaticPropsContext) => string)
+) {
+  return async (context: GetStaticPropsContext) => {
+    const query =
+      typeof queryOrQueryGetter === 'function'
+        ? queryOrQueryGetter.call({}, context)
```
VWSCoronaDashboard10, line 51: Why not simply `queryOrQueryGetter(context)`?

VWSCoronaDashboard10 (approved):
Couple of small comments 👍

packages/cms/schemas/documents/artikel.js
VWSCoronaDashboard4: I'd agree we can add this. I'd opt to do that in a separate PR, to also scope what it should/shouldn't do layout wise.

VWSCoronaDashboard (approved):

PR #1456, opened 2021-01-14
Make sidebar clickable again with skipLinkAnchor issue

VWSCoronaDashboard14:
So since there are `skipLinkAnchor` properties and recent changes that increased the z-index from the main content.
There was an issue created with the IE11 fix `margin-left: -100vw; padding-left: 100vw;` on `skipLinkAnchor` elements, you couldn't click the sidebar anymore on that specific place.

## Current fix

First set on the `skipLinkAnchor` element the `pointer-events` attribute to none where the positioning is done, and then add an additional element that resets the pointer-events back to default.

Tested it on IE11 and everything works as expected.
The only drawback is that you can't select the element directly in your inspector anymore

### DIFF
--- a/packages/app/src/components-styled/content-header/index.tsx
+++ b/packages/app/src/components-styled/content-header/index.tsx
@@ -19,6 +19,10 @@ import { Box } from '../base';
 /*
   the left margin '-100w' and left padding '100w' hack ensures skip link anchors to have a (non visible) start at the left side of the screen.
   This fixes odd skip-link behavior in IE11
+
+  Since this hack makes the part of the sidebar unclickable because the padding is overlapping it.
+  This is fixed by first setting a pointer even none to the HeaderBox element if there is a skip-link and create 
+  a new child element with the PointerEventBox that resets the pointer-events again so it works as expected.
 */
 const HeaderBox = styled.header<{
   hasIcon: boolean;
@@ -84,7 +88,7 @@ const AriaInlineText = styled(InlineText)(
   })
 );
 
-const ReferenceBox = styled(Box)(
+const ReferenceBox = styled.div(
   css({
     maxWidth: '30em',
     marginRight: 3,
--- a/packages/app/src/components-styled/content-header/index.tsx
+++ b/packages/app/src/components-styled/content-header/index.tsx
@@ -28,7 +28,7 @@ const HeaderBox = styled.header<{
     mt: 0,
     ml: x.skipLinkAnchor ? '-100vw' : x.hasIcon ? undefined : 5,
     pl: x.skipLinkAnchor ? '100vw' : undefined,
-    pointerEvents: 'none',
+    pointerEvents: x.skipLinkAnchor ? 'none' : 'auto',
   })
 );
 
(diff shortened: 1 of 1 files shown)


VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/content-header/index.tsx
```diff
@@ -28,6 +28,7 @@ const HeaderBox = styled.header<{
     mt: 0,
     ml: x.skipLinkAnchor ? '-100vw' : x.hasIcon ? undefined : 5,
     pl: x.skipLinkAnchor ? '100vw' : undefined,
+    pointerEvents: 'none',
```
VWSCoronaDashboard10, line 31: This results in an unresponsive UI: users will not be able anymore to interact with the text which is an accessibility drawback. Also things like anchors are not clickable anymore

VWSCoronaDashboard10 (changes requested):
> The only drawback is that you can't select the element directly in your inspector anymore

this is a huge drawback imho, this implies users cannot interact anymore with the content (e.g. select content, click anchors)

packages/app/src/components-styled/content-header/index.tsx
VWSCoronaDashboard10: I'm not entirely sure I understand the current implementation, but perhaps its sufficient to replace this line with:
```suggestion
    pointerEvents: x.skipLinkAnchor ? 'none' : 'auto',
```

VWSCoronaDashboard14: I forgot to commit my last change! Please check again

packages/app/src/components-styled/content-header/index.tsx
```diff
@@ -83,13 +88,19 @@ const ReferenceBox = styled(Box)(
...
 );
 
 const MetadataBox = styled(Box)(
   css({
-    flex: asResponsiveArray({ md: '1 1 auto', lg: '1 1 40%' })
+    flex: asResponsiveArray({ md: '1 1 auto', lg: '1 1 40%' }),
+  })
+);
+
+const PointerEventsBox = styled(Box)(
```
VWSCoronaDashboard10, line 101: This can be a simple div
```suggestion
const PointerEventsBox = styled.div(
```

packages/app/src/components-styled/content-header/index.tsx
```diff
@@ -28,6 +28,7 @@ const HeaderBox = styled.header<{
     mt: 0,
     ml: x.skipLinkAnchor ? '-100vw' : x.hasIcon ? undefined : 5,
     pl: x.skipLinkAnchor ? '100vw' : undefined,
+    pointerEvents: x.skipLinkAnchor ? 'none' : 'auto',
```
VWSCoronaDashboard10, line 31: it's unclear by only reading the code what the pointerEvents thing is for, so it might be useful to add a comment somewhere in this file to explain the issue you tried to solve with it.

VWSCoronaDashboard10 (approved):

PR #1465, opened 2021-01-14
Reduce amount of ticks on age demographic graph on smaller screens

VWSCoronaDashboard14:
Reduce the amount of ticks on age demographic graph on smaller screens, when it is below breakpoint XS it is capped to 2 each axes.

### DIFF
--- a/packages/app/src/domain/infected-people/age-demographic/age-demographic-chart.tsx
+++ b/packages/app/src/domain/infected-people/age-demographic/age-demographic-chart.tsx
@@ -11,7 +11,6 @@ import { colors } from '~/style/theme';
 import { NationalTestedPerAgeGroupValue } from '~/types/data.d';
 import { formatPercentage } from '~/utils/formatNumber';
 import { AgeDemographicCoordinates } from './age-demographic-coordinates';
-import { useBreakpoints } from '~/utils/useBreakpoints';
 export const AGE_GROUP_TOOLTIP_WIDTH = 340;
 
 const text = siteText.infected_age_groups;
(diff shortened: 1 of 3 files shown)


packages/app/src/domain/infected-people/age-demographic/age-demographic-chart.tsx
```diff
@@ -64,6 +64,9 @@ export const AgeDemographicChart = memo<AgeDemographicChartProps>(
       ageGroupRange,
     } = coordinates;
 
+    const breakpoints = useBreakpoints();
+    const numTicksOnBreakpoint = breakpoints.xs ? numTicks : 2;
```
VWSCoronaDashboard, line 68:
```suggestion
    const { xs } = useBreakpoints();
    const numTicksOnBreakpoint = xs ? numTicks : 2;
```

VWSCoronaDashboard (approved):

VWSCoronaDashboard1 (approved):

packages/app/src/domain/infected-people/age-demographic/age-demographic-chart.tsx
```diff
@@ -64,6 +64,9 @@ export const AgeDemographicChart = memo<AgeDemographicChartProps>(
       ageGroupRange,
     } = coordinates;
 
+    const { xs } = useBreakpoints();
+    const numTicksOnBreakpoint = xs ? numTicks : 2;
```
VWSCoronaDashboard4, line 68: You get `numTicks` via `coordinates` which is calculated in the file `age-demographic-coordinates`. It is crucial to keep thiese calculations in that file for performance and tidyness. Other parts of the graph also refer to these coordinates (eg the tooltip), so if these are not in sync things might break.

VWSCoronaDashboard4 (changes requested):

VWSCoronaDashboard4 (approved):

PR #1478, opened 2021-01-15
Implement Article page

VWSCoronaDashboard10:
Here's the article page 🎉


packages/app/src/components-styled/cms/image-block.tsx
```diff
@@ -0,0 +1,43 @@
+import { Box } from '~/components-styled/base';
+import { ContentBlock } from '~/components-styled/cms/content-block';
+import { MaxWidth } from '~/components-styled/max-width';
+import { urlFor } from '~/lib/sanity';
+import { ImageBlock, RichContentImageBlock } from '~/types/cms';
+import { assert } from '~/utils/assert';
+
+export function Image({ node }: { node: ImageBlock | RichContentImageBlock }) {
+  const url = urlFor(node).toString();
```
VWSCoronaDashboard1, line 9: There's also this PR that deals with the images:
https://github.com/minvws/nl-covid19-data-dashboard/pull/1434
Maybe that needs to be merged first and then that implementation can be used here as well?

VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/cms/image-block.tsx
VWSCoronaDashboard10: I think it's easier to base that PR on these changes and pick what's necessary..

VWSCoronaDashboard1: Sounds fine with me!

packages/app/src/components-styled/article-detail.tsx
```diff
@@ -0,0 +1,49 @@
+import css from '@styled-system/css';
+import ArrowIcon from '~/assets/arrow.svg';
+import { Box } from '~/components-styled/base';
+import { ContentBlock } from '~/components-styled/cms/content-block';
+import { Image } from '~/components-styled/cms/image-block';
+import { Heading } from '~/components-styled/typography';
+import siteText from '~/locale';
+import { Article } from '~/types/cms';
+import { formatDateFromMilliseconds } from '~/utils/formatDate';
+import { RichContent } from '../domain/cms/rich-content';
```
VWSCoronaDashboard8, line 10: What's the difference between domain/cms and components-styled/cms?

I guess this is a question that can be answered later, because I suspect that no one has a clear answer yet.

packages/app/src/components-styled/max-width.tsx
```diff
@@ -4,7 +4,7 @@ import { Box } from '~/components-styled/base';
 
 export const MaxWidth = styled(Box)(
   css({
-    maxWidth: 1400,
+    maxWidth: 'maxWidth',
```
VWSCoronaDashboard4, line 7: What is this doing?

packages/cms/schemas/documents/article.js
```diff
@@ -54,11 +54,14 @@ export default {
...
       type: 'image',
+      options: {
+        hotspot: true
+      },
       fields: [
         {
           title: 'Omschrijving',
           description: 'alt-tekst',
-          name: 'description',
+          name: 'alt',
```
VWSCoronaDashboard4, line 64: Why is this name adjusted?

packages/app/src/components-styled/article-detail.tsx
```diff
@@ -0,0 +1,49 @@
...
+          {siteText.article_detail.back_link.text}
+        </LinkWithIcon>
+
+        <Box spacing={2}>
+          <Heading level={1} mb={0}>
+            {article.title}
+          </Heading>
+          <time css={css({ color: 'gray' })} dateTime={article.publicationDate}>
+            {formatDateFromMilliseconds(
+              new Date(article.publicationDate).getTime(),
```
VWSCoronaDashboard4, line 31: Parsing this date might not work well in older browsers, eg IE11/older phones. Perhaps good to verify?

packages/app/src/components-styled/article-detail.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { RichContent } from '../domain/cms/rich-content';
+import { LinkWithIcon } from './link-with-icon';
+
+interface ArticleDetailProps {
+  article: Article;
+}
+
+export function ArticleDetail({ article }: ArticleDetailProps) {
+  return (
+    <Box bg="white" py={{ _: 4, md: 5 }}>
```
VWSCoronaDashboard4, line 19: Part of this layout wrapping could be moved up in the hierarchy / to a layout; but that's fine for a different PR.

VWSCoronaDashboard4 (approved):

packages/app/src/components-styled/cms/image-block.tsx
```diff
@@ -0,0 +1,43 @@
...
+            display="block"
+            borderRadius={1}
+            boxShadow="tile"
+            src={url}
+            alt={node.alt}
+          />
+          {caption}
+        </Box>
+      </MaxWidth>
+    </Box>
```
VWSCoronaDashboard, line 34: We should make all of this responsive, but we can do that after merging this and the responsive images PR.

packages/cms/schemas/documents/article.js
```diff
@@ -54,11 +54,14 @@ export default {
       title: 'Afbeelding',
       name: 'cover',
       type: 'image',
+      options: {
+        hotspot: true
+      },
       fields: [
         {
           title: 'Omschrijving',
           description: 'alt-tekst',
```
VWSCoronaDashboard, line 63:
```suggestion
          description: 'Onderschrift bij afbeelding',
          name: 'description',

```

packages/app/src/components-styled/cms/image-block.tsx
```diff
@@ -0,0 +1,43 @@
...
+  assert(
+    url !== null,
+    `could not get url for node: ${JSON.stringify(node, null, 2)}`
+  );
+
+  const caption = 'caption' in node && node.caption && (
+    <figcaption>{node.caption}</figcaption>
+  );
+
+  return 'isFullWidth' in node && node.isFullWidth ? (
```
VWSCoronaDashboard8, line 19: These conditions are a little cryptic to me. Wouldn't `'caption' in node` return true if node.caption exists? I guess not because then node.caption would be a redundant check....

VWSCoronaDashboard (approved):
Only some nitpicks, approved.

packages/cms/schemas/documents/article.js
VWSCoronaDashboard10: Because it is the `alt` text of an image .. And also because `description` is used for a lot of other keys (e.g. collapsible content), the alt-text of an image would be something completely different.

packages/app/src/components-styled/article-detail.tsx
VWSCoronaDashboard10: Ah good one, I wasn't aware that was an issue in older browsers. I'll try it out.

packages/app/src/domain/cms/rich-content.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { PortableText } from '~/lib/sanity';
+import { Block, RichContentBlock } from '~/types/cms';
+
+export function RichContent({
+  blocks,
+}: {
+  blocks: Block | RichContentBlock[];
+}) {
+  return (
+    <Box fontSize="1.125rem">
```
VWSCoronaDashboard8, line 15: Our CMS base components are not fitting the current theme?

VWSCoronaDashboard8 (approved):

packages/cms/schemas/documents/article.js
VWSCoronaDashboard10: Ah I forgot that one. I was actually thinking of 'Alternatieve tekst (toegankelijkheid)' because it's the alt-text, not an onderschrift.

packages/app/src/components-styled/article-detail.tsx
VWSCoronaDashboard10: No, good one, the intention was to move this file to the styled-components directory.

packages/app/src/components-styled/cms/image-block.tsx
VWSCoronaDashboard10: It's to capture this situation:  `{ caption: '' }`

packages/app/src/domain/cms/rich-content.tsx
VWSCoronaDashboard10: I'm afraid not .. it's fontsize 18px. I think the theme variables need some attention to get aligned with design, instead of the other way around.

packages/app/src/components-styled/cms/image-block.tsx
VWSCoronaDashboard10: How is this currently not responsive?

packages/app/src/components-styled/cms/image-block.tsx
VWSCoronaDashboard8: Ah I see, you're relying on the fact that the && returns the result of it's right-hand side. I think it would be useful to write a simple helper function to do this if it becomes a regular pattern.

packages/app/src/components-styled/cms/image-block.tsx
VWSCoronaDashboard: Responsive as in using `srcSet`. 👍 
Don't bother for now, it's easier to do this when the image resize PR is merged.

PR #1497, opened 2021-01-18
Add language switcher to CMS

VWSCoronaDashboard:
## Summary

This PR adds a plugin to clean up the UI when working with multi-language content

**Before:** 

image

**After:** 

image

## Motivation

Duplicating input fields per language takes up too much screen estate. This keeps things clean.

## Detailed design

- I took [@sanity/language-filter](https://github.com/sanity-io/sanity/tree/next/packages/%40sanity/language-filter) as a starting point.
- Because I didn't want to mess around with builds, I decided to move the plugin to the plugins folder in the CMS package
- I created a new plugin with `sanity init` and started porting functionality.

## Drawbacks

- I don't like how it still shows the language you're editing as a title above the input field. I would really like to get rid of that.

## Alternatives

We could use [@sanity/language-filter](https://github.com/sanity-io/sanity/tree/next/packages/%40sanity/language-filter) without any alterations. This way content managers don't have to choose between languages but they can make their own selection of languages.

## Unresolved questions

- Can we hide the language names above the fields
- How can we (and should we?) use different locales for previews in collapsible content previews. Right now all previews are hardcoded to the dutch language.


packages/cms/plugins/translate/config.js
```diff
@@ -0,0 +1,9 @@
+export default {
+  supportedLanguages: [
+    { id: "en", title: "English" },
+    { id: "nl", title: "Dutch" },
+  ],
+  filterField: (enclosingType, field, selectedLanguageIds) =>
+    !enclosingType.name.startsWith("locale") ||
+    selectedLanguageIds.includes(field.name),
+};
```
VWSCoronaDashboard, line 9: Let's use this file instead of `packages/cms/parts/languageFilterConfig.js`

VWSCoronaDashboard8 (approved):

PR #1498, opened 2021-01-18
Allow the Search component to have an initial value

VWSCoronaDashboard10:
## Summary

The search component should be able to have an initial value.

There was a small issue for narrow devices where the search results would always show as long as there was a value. In the case of this initial value we don't want this behavior.
The fix is to listen for an initial focus event on the input element, as long as that event doesn't fire we won't show any search results. When the event gets triggered we can fallback to the default implementation.

The diff contains as lot of whitespace changes, I suggest to see the diff with the `?w=1` url parameter.

### DIFF
--- a/packages/app/src/domain/topical/components/search/context.tsx
+++ b/packages/app/src/domain/topical/components/search/context.tsx
@@ -75,7 +75,7 @@ function useSearchContextValue<T extends Element>(
   /**
    * By default narrow devices will show search-results when the input field
    * has a value. We don't want to show these results when the value is
-   * preset/prepolated.
+   * pre-set/pre-populated.
    * To fix this we will listen to an initial focus event on the input element,
    * if that initial event is recognized, we'll allow the results to show.
    */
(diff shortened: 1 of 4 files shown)


VWSCoronaDashboard4 (approved):

packages/app/src/domain/topical/components/search/context.tsx
```diff
@@ -67,6 +72,18 @@ function useSearchContextValue<T extends Element>(containerRef: RefObject<T>) {
   const [hasInputFocus, setHasInputFocus] = useState(false);
   const [hasHitFocus, setHasHitFocus] = useState(false);
 
+  /**
+   * By default narrow devices will show search-results when the input field
+   * has a value. We don't want to show these results when the value is
+   * preset/prepolated.
+   * To fix this we will listen to an initial focus event on the input element,
+   * if that initial event is recognized, we'll allow the results to show.
+   */
```
VWSCoronaDashboard8, line 81: Yay for comments 💯

VWSCoronaDashboard8 (approved):

PR #1507, opened 2021-01-18
Actueel risicokaart and latest articles layout 

VWSCoronaDashboard1:
## Summary

A load of layout tweaks to make the actueel page look more like the designs

## Motivation

Design request

## Detailed design

Add some domain specific components to domain/topical. The rest is css tweaks.

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added

### DIFF
--- a/packages/app/src/components-styled/article-teaser.tsx
+++ b/packages/app/src/components-styled/article-teaser.tsx
@@ -32,8 +32,8 @@ export function ArticleTeaser(props: ArticleLinkProps) {
       borderRadius={4}
       minHeight={'26rem'}
       maxHeight={'26rem'}
-      minWidth={{ _: '20rem', lg: '24rem' }}
-      maxWidth={{ _: '20rem', lg: '24rem' }}
+      minWidth={{ _: '100%', lg: '24rem' }}
+      maxWidth={{ _: '100%', lg: '24rem' }}
       overflow="hidden"
     >
       {<CoverImage image={cover} />}
--- a/packages/app/src/components-styled/article-teaser.tsx
+++ b/packages/app/src/components-styled/article-teaser.tsx
@@ -32,8 +32,7 @@ export function ArticleTeaser(props: ArticleLinkProps) {
       borderRadius={4}
       minHeight={'26rem'}
       maxHeight={'26rem'}
-      minWidth={{ _: '100%', lg: '24rem' }}
-      maxWidth={{ _: '100%', lg: '24rem' }}
+      width="100%"
       overflow="hidden"
     >
       {<CoverImage image={cover} />}
--- a/packages/app/src/components-styled/article-teaser.tsx
+++ b/packages/app/src/components-styled/article-teaser.tsx
@@ -32,7 +32,6 @@ export function ArticleTeaser(props: ArticleLinkProps) {
       borderRadius={4}
       minHeight={'26rem'}
       maxHeight={'26rem'}
-      width="100%"
       overflow="hidden"
     >
       {<CoverImage image={cover} />}
--- a/packages/app/src/domain/topical/article-list.tsx
+++ b/packages/app/src/domain/topical/article-list.tsx
@@ -25,7 +25,6 @@ export function ArticleList({ articleSummaries, hideLink }: ArticleListProps) {
       css={css({
         display: 'flex',
         flexDirection: 'column',
-        padding: 4,
       })}
     >
       <Heading level={2}>
--- a/packages/app/src/domain/topical/article-list.tsx
+++ b/packages/app/src/domain/topical/article-list.tsx
@@ -59,12 +59,7 @@ export function ArticleList({ articleSummaries, hideLink }: ArticleListProps) {
           )}
         </Box>
       </Box>
-      <Box
-        display="flex"
-        spacingHorizontal={breakpoints.lg}
-        flexDirection={{ _: 'column', lg: 'row' }}
-        flexWrap="wrap"
-      >
+      <Box display="block" margin={0} maxWidth="100%" mt={3}>
         {articleSummaries.map((summary) => (
           <ArticleBox key={summary.slug.current}>
             <ArticleTeaser
@@ -75,6 +70,14 @@ export function ArticleList({ articleSummaries, hideLink }: ArticleListProps) {
             />
           </ArticleBox>
         ))}
+        <ArticleBox key={'JA'}>
+          <ArticleTeaser
+            title={articleSummaries[0].title}
+            slug={articleSummaries[0].slug.current}
+            summary={articleSummaries[0].summary}
+            cover={articleSummaries[0].cover}
+          />
+        </ArticleBox>
       </Box>
     </Box>
   );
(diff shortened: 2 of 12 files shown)


packages/app/src/domain/topical/article-list.tsx
```diff
@@ -76,8 +76,16 @@ export function ArticleList({ articleSummaries, hideLink }: ArticleListProps) {
...
-    flex: '0 0 auto',
-    marginTop: 3,
-    marginRight: asResponsiveArray({ _: 0, lg: 4 }),
+    display: 'inline-block',
+    marginBottom: 4,
+    width: asResponsiveArray({
+      _: '100%',
+      sm: '50%',
+      md: 'calc(33% - 32px)',
+      lg: 'calc(33% - 32px)',
```
VWSCoronaDashboard8, line 85: Ideally this 32px (gap I assume?) would be coming from the theme I think.

packages/app/src/domain/topical/article-list.tsx
```diff
@@ -76,8 +76,16 @@ export function ArticleList({ articleSummaries, hideLink }: ArticleListProps) {
...
+    }),
+    '&:nth-child(3n+2)': {
+      mx: asResponsiveArray({ md: '48px', lg: '48px' }),
```
VWSCoronaDashboard8, line 88: 48px would not be in the theme, but it's still strongly related since it could be calculated as `2.5 * spacing.lg`

VWSCoronaDashboard8 (approved):

packages/app/src/components-styled/article-teaser.tsx
```diff
@@ -32,8 +32,7 @@ export function ArticleTeaser(props: ArticleLinkProps) {
       borderRadius={4}
       minHeight={'26rem'}
       maxHeight={'26rem'}
-      minWidth={{ _: '20rem', lg: '24rem' }}
-      maxWidth={{ _: '20rem', lg: '24rem' }}
+      width="100%"
```
VWSCoronaDashboard10, line 35: isn't width 100% by default?

packages/app/src/domain/topical/article-list.tsx
VWSCoronaDashboard10: I'm not sure I can follow you. You suggest we should be able to access our theme's `space`-values using both an index as named keys? Just like our Breakpoints type?
Or are you referencing a new `sizes` object of our theme? Because that's where the `width` property will try to resolve from.

Besides that, I don't think we will be able to access theme properties within a css value like this:
`width: calc(33% - md)`

Instead, using the plain string should work:
`width: md`

VWSCoronaDashboard10 (approved):

packages/app/src/domain/topical/article-list.tsx
VWSCoronaDashboard1: The biggest problem is that thew calc() method only takes pixel values, not ems, rems, or whatever

VWSCoronaDashboard8: I mean to say, create new variable with object shape in the theme, import that and reference the value by name in that way. But I wasn't aware that calc only takes px so that throws a little root in the food.

packages/app/src/domain/topical/topical-choropleth-container.tsx
```diff
@@ -0,0 +1,80 @@
...
+      <ChoroplethLegenda thresholds={legend.thresholds} title={legend.title} />
+    )) ??
+    legendComponent;
+
+  return (
+    <Box
+      display="flex"
+      flexDirection={{ _: 'column', lg: 'row' }}
+      m={0}
+      as="figure"
```
VWSCoronaDashboard4, line 48: Why is this a `<figure>` ? The content we put in here is way too broad to be put in a figcaption. For accessibility we should just have the proper label/described by associated with the choropleth. The rest of the content is just content, not a caption.

packages/app/src/domain/topical/article-list.tsx
VWSCoronaDashboard8: @VWSCoronaDashboard1 According to this source calc can deal with other units besides px https://css-tricks.com/a-complete-guide-to-calc-in-css/

```
There are many lengths of CSS though, and they can all be used with calc():

px
%
em
rem
in
mm
cm
pt
pc
ex
ch
vh
vw
vmin
vmax
```

packages/app/src/pages/index.tsx
```diff
@@ -172,35 +174,48 @@ const Home: FCWithLayout<typeof getStaticProps> = (props) => {
...
-                    __html: text.risiconiveaus.selecteer_toelichting,
-                  }}
-                />
-                <EscalationMapLegenda
+          <TopicalTile>
+            <>
+              <TopicalChoroplethContainer
+                title={text.risiconiveaus.selecteer_titel}
+                description={
+                  <span
```
VWSCoronaDashboard4, line 182: Markdown should be put on a block level element such as a `div`, since it contains other block level elements such as `<p>` and `<h2>` etc.

VWSCoronaDashboard4 (changes requested):

packages/app/src/domain/topical/article-list.tsx
VWSCoronaDashboard8: So to recap; what I suggest is:
```
import {spacing} from '~/theme'
...
width: calc(33% - spacing.md)
```

packages/app/src/domain/topical/topical-choropleth-container.tsx
VWSCoronaDashboard1: This is also the case in the 'regular' choropleth container. This component is largely a copy of that.

packages/app/src/domain/topical/topical-choropleth-container.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { useBreakpoints } from '~/utils/useBreakpoints';
+
+/**
+ * We could use strong typing here for the values and also enforce the data
+ * prop to be set. Might be a good idea.
+ *
+ * @TODO move to a shared location
+ */
+interface DataProps {
+  'data-cy'?: string;
```
VWSCoronaDashboard8, line 15: I think ideally we wouldn't even deal with setting data-cy from outside. We can base it on the component name + mertric property for example. Then it should be unique per page and we can target it from the tests.

VWSCoronaDashboard1: We'll leave that for another PR ;)

packages/app/src/pages/index.tsx
VWSCoronaDashboard8: It's really time to write a thing wrapper / component for this I think to avoid this type of mistakes. Also it would make the code cleaner since then we can hide the dangerouslySetInnerHTML part and only write it once in that component.

packages/app/src/domain/topical/article-list.tsx
VWSCoronaDashboard1: For me, the most important aspect is that this value is divisible by three

packages/app/src/domain/topical/topical-choropleth-container.tsx
VWSCoronaDashboard1: But I've removed here now

PR #1521, opened 2021-01-19
Add custom background feature to chart-axes

VWSCoronaDashboard1:
## Summary

This PR adds a new component callback type called 'CustomBackground' which enables a developer to add arbitrary shapes to the chart-axes background.

## Motivation

This new feature was necessary in order to implement this functionality on the hospital pages:

image

## Detailed design

I've moved chart-axes to its own directory since this can be used as the base component for scatter charts and other flavors as well.

A new component callback called 'CustomBackground' was added. This will add its results as the first elements inside the main group, so they will be drawn behind all the other chart elements.

A factory method called `addBackgroundRectangleCallback ` was added that enabled a rectangle to be drawn that spans a given date range.

This rectangle was implemented on all the hospital admissions charts (NL, VR and GM).

Legenditems were generated based on the line config by default. This use case required a legend item to be added for the background rectangle as well. So, the LineChart component was given an optional prop called legendItems where the developer can just pass in a list of legend items.

## Drawbacks

n/a

## Alternatives

Please let me know in the comments

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added


VWSCoronaDashboard4 (approved):

packages/app/src/components-styled/chart-axes/custom-background/add-background-rectangle-callback.tsx
```diff
@@ -0,0 +1,49 @@
+import { Bar } from '@visx/shape';
```
VWSCoronaDashboard8, line 1: Personally I don't see much value in moving these files into a separate chart-axis folder. Also here you are breaking the common pattern where a complex component has its subcomponents in a folder called `components`. 

I would just add this background rectangle thing next to the chart axes component or in a `components` folder. Also by moving this component to the level of components-styled I think you are assuming that this component is generic enough to be used outside the `line-chart` component? If not, I suggest to keep it co-located with line-chart until the api is stabilised.

VWSCoronaDashboard8 (approved):

packages/app/src/components-styled/chart-axes/custom-background/add-background-rectangle-callback.tsx
VWSCoronaDashboard1: For now, I'll put the chart-axes back in the linechart directory. Even though I still feel it needs to be separate from the line, since it'll be re-used by the bar and scatter plot charts. So maybe all the chart related components just need to live in a 'charts' folder in the components-styled fir. But we'll see about that later.
I don't agree with the supposed violation of of the subcomponent dir pattern. The custom-background directory contains a component callback, which is more of a factory function. So it's not a component that is instantiated in a regular way. I'll call the directory component-callbacks if that makes ore sense.

PR #1533, opened 2021-01-20
Add schema for vaccination support

VWSCoronaDashboard1:
## Summary

Adds a new property to the national schema called vaccinations which holds a week based time series that shows vaccination support percentages.

## Motivation

Feature request

## Detailed design

Summary says it all

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added


packages/app/schema/nl/__index.json
```diff
@@ -27,7 +27,8 @@
     "tested_ggd_daily",
     "tested_ggd_average",
     "tested_overall",
-    "tested_per_age_group"
+    "tested_per_age_group",
+    "vaccination"
```
VWSCoronaDashboard8, line 31: There will be multiple schema's from different sources about vaccination, so I think we need to use a common prefix. The other schema that already existed about this is in topic/vaccine-base. It contains a property `vaccine_availability` if I remember correctly.

This schema is specific about support, so I think it needs to be reflected in the root schema as well. So I suggest using `vaccine_support`. 

The property `support_percentage` can I think then simply be named `percentage` given the scope it's in.

packages/app/schema/nl/vaccination.json
```diff
@@ -0,0 +1,45 @@
...
+    }
+  },
+  "required": ["values", "last_value"],
+  "additionalProperties": false,
+  "definitions": {
+    "value": {
+      "title": "nl_vaccination_value",
+      "type": "object",
+      "properties": {
+        "support_percentage": {
```
VWSCoronaDashboard8, line 23: See earlier comment

packages/app/schema/nl/vaccination.json
```diff
@@ -0,0 +1,45 @@
...
+        },
+        "date_end_unix": {
+          "type": "integer"
+        },
+        "date_of_insertion_unix": {
+          "type": "integer"
+        }
+      },
+      "required": [
+        "support",
```
VWSCoronaDashboard8, line 37: This name doesn't match the used property

VWSCoronaDashboard8 (changes requested):

PR #1535, opened 2021-01-20
Fix IE11 issues by fixing transpilation issues

VWSCoronaDashboard10:
A couple of libraries should be transpiled for IE11 to fix potential syntax errors. One of the libraries didn't work well with the `next-transpile-modules` library, it cannot handle a package.json `exports` field. A fix for now is to manually remove that field in a post-install step.. It's not ideal, but a broken dashboard on IE11 isn't either.


VWSCoronaDashboard1 (approved):

packages/app/fix-transpilation-error.js
```diff
@@ -0,0 +1,16 @@
+/**
```
VWSCoronaDashboard8, line 1: nit; I suggest to place this in /scripts and trigger it with ts-node so you can keep the codebase TS only

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard10:
Closed in favor of https://github.com/minvws/nl-covid19-data-dashboard/pull/1535

PR #1578, opened 2021-01-24
Add editorial tile to actueel page

VWSCoronaDashboard1:
## Summary

This adds a tile to the actueel page that displays the latest editorial and one highlighted article teaser.

## Motivation

Feature request

## Detailed design

A new schema for editorials was added. Right now this is still a verbatim copy of the articles schema (as is the editorialdetail component). But since these editorials are supposed to differ from regular articles they're modeled separately since they're subject to change.
A page schema for actueel was added that allows content managers to choose the highlighted article.
A placeholder page for the editorial page was added, since no design for this exists yet.

In the sanity schemas, the `__experimentalActions` were removed since these are now deprecated. The allowed document actions are now configured using the added `resolveDocumentActions` function which uses the correct API to perform these tasks.

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added


packages/cms/resolve-document-actions.js
```diff
@@ -0,0 +1,31 @@
+import defaultResolve, {
+  DeleteAction,
+  DiscardChangesAction,
+  DuplicateAction,
+  PublishAction,
+  UnpublishAction,
+} from 'part:@sanity/base/document-actions';
```
0x80, line 7: What is this weird looking import doing?

packages/cms/schemas/documents/editorial.js
```diff
@@ -0,0 +1,25 @@
+import { ARTICLE_FIELDS } from './fields/ARTICLE_FIELDS';
```
0x80, line 1: I don't see a need to have the filename match the variable casing. We agreed to make all files kebab-case so these ones can be that too.

packages/cms/schemas/schema.js
```diff
@@ -1,38 +1,34 @@
...
-import localeRichContentBlock from './locale/locale-rich-content-block';
 import localeText from './locale/locale-text';
+//objects are building blocks, but not queryable in itself
+// import openGraph from "./objects/open-graph";
+import collapsible from './objects/collapsible';
+import lineChart from './objects/line-chart';
+import lockdown from './restrictions/lockdown.js';
+import restrictionCategory from './restrictions/restriction-category.js';
+import restrictionGroupLockdown from './restrictions/restriction-group-lockdown.js';
+// routekaart en maatregelen
```
0x80, line 28: Looks like some things can be cleaned up

0x80 (approved):

packages/app/src/utils/formatDate.ts
```diff
@@ -4,10 +4,10 @@
  * datetimeformat needs to be second
  * datetimeformat locale's last
  */
-import '@formatjs/intl-getcanonicallocales/polyfill';
-import '@formatjs/intl-datetimeformat/polyfill';
 import '@formatjs/intl-datetimeformat/locale-data/en';
 import '@formatjs/intl-datetimeformat/locale-data/nl';
+import '@formatjs/intl-datetimeformat/polyfill';
```
VWSCoronaDashboard4, line 9: Revert this please, see the comment above it. This will cause IE11 to break the app

packages/app/src/utils/formatDate.ts
```diff
@@ -135,3 +135,10 @@ export function formatDateFromMilliseconds(
 
   return DayMonth.format(milliseconds);
 }
+
+export function formatSanityDate(
```
VWSCoronaDashboard4, line 139: Sanity is the name of the source, not the name of the date format. Perhaps adjust the name to match what the format is?

packages/cms/schemas/documents/editorial.js
```diff
@@ -0,0 +1,25 @@
...
+    {
+      title: 'Metadata',
+      name: 'metadata',
+      options: {
+        collapsible: true,
+        collapsed: true,
+      },
+    },
+  ],
+  fields: [].concat(ARTICLE_FIELDS),
```
VWSCoronaDashboard4, line 17: Wouldn't it be handier to just have two separate configs? Then you can safely remove an item from one, or add one to the other, without messing anything up.

VWSCoronaDashboard4 (changes requested):

packages/app/src/utils/formatDate.ts
VWSCoronaDashboard1: Oh damn, my formatter automatically changes this...

packages/cms/schemas/documents/editorial.js
VWSCoronaDashboard1: No, in this case, I'm using composition exactly because these types of content shared these fields. In the future, there will be different flavors as well, such as blogposts, etc.

packages/app/src/utils/formatDate.ts
VWSCoronaDashboard1: True, but I reckoned that within our domain the sanity source implies that it's a string-based date...

VWSCoronaDashboard1: I've renamed it to `formatDateFromString`

PR #1594, opened 2021-01-25
Vaccine coverage schema fix

VWSCoronaDashboard8:
## Summary

Fix schema of vaccine coverage to be more in-line with other schemas

### DIFF
--- a/packages/app/schema/vr_collection/__index.json
+++ b/packages/app/schema/vr_collection/__index.json
@@ -16,7 +16,7 @@
     "elderly_at_home",
     "disability_care",
     "sewer",
-    "vaccine"
+    "vaccine_coverage"
   ],
   "properties": {
     "last_generated": {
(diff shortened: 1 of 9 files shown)


VWSCoronaDashboard10:
I think I based these schema’s on the sewer schema’s, so I guess those aren’t in line either 😛

VWSCoronaDashboard10 (approved):

VWSCoronaDashboard8:
@VWSCoronaDashboard10 You're right, not all schema's are perfect. Some things are new-ish but were left out of the schema migration since they didn't seem to hurt. For sewer we have `sewer` and `sewer_per_installation`. Those are ok I think because a choropleth will never use sewer_per_installation data, and the distinction between the two is pretty clear.

I think I asked you to use a `vaccine_` prefix, and this got merged before I had a chance to look at it 😄  But I forgive you 😉

VWSCoronaDashboard1 (approved):

PR #1607, opened 2021-01-26
Vaccine availability chart

VWSCoronaDashboard8:
## Summary

Screenshot 2021-01-27 at 12 55 47

Screenshot 2021-01-27 at 12 56 26

A few things to note:

- This is not connected to actual data yet. The chart is using mock data.
- I didn't manage to reuse anything from our existing charts. Either because it was not compatible, or I thought things could be improved. In that sense this is also meant for comparison. The chart implementation is not generic yet, because a lot of things, like hover handler and tick formatting are very specific. I did not want to waste time trying to come up with meaningful abstractions for these because I felt is was not the right time to do this yet. Making this chart more generic and constructing reusable abstractions is left as an exercise for later.
- The chart uses the Visx tooltip implementation, because it was unclear to me why we are using our own. The visx one is rendered in a portal and flips automatically when getting out of bounds. It also has a nice subtle drop shadow which the design also had.
- The tooltip has a timer / throttle so that it is less jittery when you move over the bars. The gaps in between would otherwise quickly toggle the tooltip visibility.
- I have tried to come up with some name improvements like DateSpan and tried to keep naming similar to the Visx documentation. For example xMax in Visx is not the maximum value of the series but the x axis spatial max size.


packages/app/schema/nl/vaccine_delivery.json
```diff
@@ -0,0 +1,65 @@
+{
+  "definitions": {
+    "value": {
+      "title": "nl_vaccine_delivery_value",
```
VWSCoronaDashboard4, line 4: Almost all other schemas are prefixed with `national_` instead of `nl`_ (the only one with nl being `nl_deceased_rivm_per_age_group`); and other levels are also prefixed with the full word. I guess we should stick to the `national_` prefix?

packages/app/src/components-styled/responsive-chart-tile.tsx
```diff
@@ -0,0 +1,141 @@
...
+    <Box
+      mb={3}
+      display="flex"
+      flexDirection={{ _: 'column', lg: 'row' }}
+      justifyContent="space-between"
+    >
+      <div css={css({ mb: [3, null, null, null, 0], mr: [0, 0, 2] })}>
+        <Heading level={3}>{title}</Heading>
+        {!description && (
+          <div css={css({ display: 'none' })} aria-labelledby={uniqueId}>
```
VWSCoronaDashboard4, line 117:
```suggestion
          <div css={css({ display: 'none' })}>
```

Maybe just remove all the labelledby if they are not properly implemented?

packages/app/src/components-styled/responsive-chart-tile.tsx
```diff
@@ -0,0 +1,141 @@
...
+            {ariaDescription}
+          </div>
+        )}
+        {description &&
+          (typeof description === 'string' ? (
+            <p aria-labelledby={uniqueId} css={css({ m: 0 })}>
```
VWSCoronaDashboard4, line 123:
```suggestion
            <p css={css({ m: 0 })}>
```

packages/app/src/pages/landelijk/vaccinaties.tsx
```diff
@@ -121,9 +121,110 @@ const VaccinationPage: FCWithLayout<typeof getStaticProps> = ({
...
           }}
         >
-          <Box pt={1}>
-            <VaccinatieBarChart />
-          </Box>
+          <ParentSize>
+            {({ width }) => (
+              <StackedChart
+                width={width}
+                // height={400}
```
VWSCoronaDashboard4, line 128:
```suggestion
```

packages/app/schema/nl/vaccine_delivery.json
VWSCoronaDashboard8: I remember you have made this comment before in another PR, but maybe you didn't read my reply then; nl_ will be the new prefix since the files have been renamed. Only the legacy titles still need to be renamed. It doesn't affect much just our Typescript types.

packages/app/src/components-styled/responsive-chart-tile.tsx
VWSCoronaDashboard8: This file needs to be removed.

packages/app/schema/nl/vaccine_delivery.json
VWSCoronaDashboard8: All the other vaccine schema's are already also using this nl_ prefix

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard4 (approved):

PR #1616, opened 2021-01-26
Add link with arrow in the trend tiles and the risk level indicator

VWSCoronaDashboard14:
It is now possible to click on the header to go to a certain page.
Since I wanted the arrow to follow the flow of the parent I had to wrap the last word including the arrow in a span if the string had at least 2 words. To prevent that the arrow is only jumping to the next line.

### DIFF
--- a/packages/app/src/components-styled/risk-level-indicator.tsx
+++ b/packages/app/src/components-styled/risk-level-indicator.tsx
@@ -59,15 +59,15 @@ export function RiskLevelIndicator(props: RiskLevelIndicatorProps) {
         <Stopwatch />
       </Box>
       <Heading level={3} as="h2" py={2} pl={{ _: '3.5rem' }}>
-      <LinkWithIcon
+        <LinkWithIcon
           href={href}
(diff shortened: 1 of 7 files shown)


VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/risk-level-indicator.tsx
```diff
@@ -55,7 +59,15 @@ export function RiskLevelIndicator(props: RiskLevelIndicatorProps) {
         <Stopwatch />
       </Box>
       <Heading level={3} as="h2" py={2} pl={{ _: '3.5rem' }}>
-        {title}
+      <LinkWithIcon
```
VWSCoronaDashboard4, line 62: Indenting is not correct here; does your formatter run properly?

VWSCoronaDashboard14: Seems like when writing a commit from Tower (I Installed and start using it yesterday) it doesn't use the format script. I'll have to look into that.

VWSCoronaDashboard4 (approved):

PR #1627, opened 2021-01-27
Generate chart aria id from title

VWSCoronaDashboard8:
## Summary

This prevent a warning about aria labelled-by id that changes between SSR and client side rendering. I'm assuming each chart has a unique title.

### DIFF
--- a/packages/app/src/components-styled/chart-tile.tsx
+++ b/packages/app/src/components-styled/chart-tile.tsx
@@ -6,15 +6,15 @@ import { ChartTileContainer } from './chart-tile-container';
 import { ChartTimeControls } from './chart-time-controls';
 import { MetadataProps } from './metadata';
 import { Heading } from './typography';
-import { useUniqueId } from '~/utils/useUniqueId';
 import { assert } from '~/utils/assert';
+import slugify from 'slugify';
 interface ChartTileProps {
   children: React.ReactNode;
   metadata: MetadataProps;
   title: string;
   description?: React.ReactNode;
   ariaDescription?: string;
-  uniqueId?: string;
+  ariaId?: string;
 }
 
 interface ChartTileWithTimeframeProps extends Omit<ChartTileProps, 'children'> {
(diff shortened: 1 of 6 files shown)


VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/chart-tile.tsx
```diff
@@ -109,17 +114,17 @@ function ChartTileHeader({
       <div css={css({ mb: [3, null, null, null, 0], mr: [0, 0, 2] })}>
         <Heading level={3}>{title}</Heading>
         {!description && (
-          <div css={css({ display: 'none' })} aria-labelledby={uniqueId}>
+          <div css={css({ display: 'none' })} aria-labelledby={ariaId}>
```
VWSCoronaDashboard4, line 117: Have you tested if this works properly?

I don't think the aria attributes are set up properly. The labeledby should be on the graph's (or whatever is described) node, pointing to the ID that is put on the node with the description.

packages/app/src/components-styled/chart-tile.tsx
```diff
@@ -88,15 +93,15 @@ function ChartTileHeader({
...
+  ariaId,
   ariaDescription,
 }: {
   title: string;
   description?: React.ReactNode;
   timeframe?: TimeframeOption;
   timeframeOptions?: TimeframeOption[];
   onTimeframeChange?: (timeframe: TimeframeOption) => void;
-  uniqueId?: string;
+  ariaId?: string;
```
VWSCoronaDashboard4, line 104: I'd suggest to make this required, not optional.

Perhaps name it `labeledById`? "AriaId" is not specific enough and point to the technique, not what it is used for.

VWSCoronaDashboard4 (changes requested):

packages/app/src/components-styled/chart-tile.tsx
VWSCoronaDashboard8: @VWSCoronaDashboard4 This PR is not changing anything about how the label is used, so I consider this out of scope. This PR was just about fixing the id which wasn't static.

packages/app/src/components-styled/chart-tile.tsx
VWSCoronaDashboard8: Yes I agree about the optional prop, but it was also not my intention to change this functionality :) 

I will rename the prop

VWSCoronaDashboard4 (approved):
OK to merge this now and improve the functionality of the descriptions in a future PR.

PR #1629, opened 2021-01-27
Add various styling fixes for the mini trend chart

VWSCoronaDashboard1:
## Summary

Mini trend component changes:

- Adjust axis label padding
- Fix marker text line breaking
- Fix mobile height
- Add mobile padding
- Increase line width
- Fix area opacity

Also: fix correct font in the searchbar placeholder text

## Motivation

Design requests

## Detailed design

n/a

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added


packages/app/src/domain/topical/mini-trend-tile.tsx
```diff
@@ -63,13 +66,17 @@ export function MiniTrendTile<T extends Value>(props: MiniTrendTileProps<T>) {
             width={parent.width}
             timeframe="5weeks"
             values={trendData}
-            height={140}
-            linesConfig={[{ metricProperty }]}
+            height={sm ? 180 : 140}
+            linesConfig={[
+              { metricProperty, areaFillOpacity: 0.25, strokeWidth: 3 },
```
VWSCoronaDashboard8, line 71: Was it necessary to expose these opacity and stroke properties? I would expect all mini trend components to look the same so in that sense it would be good to enforce these things to be conform instead of breaking it open. I don't see other values being used.

VWSCoronaDashboard1: The mini trends are defined in the mini trend tile. This is the only place where they are used. Hence, they all look the same.

VWSCoronaDashboard1: I have no idea how, apart from creating a complete copy of the LineChart, to otherwise implement this...

PR #1649, opened 2021-01-27
Add Topical vaccine tile

VWSCoronaDashboard4:
## Summary

Add Topical vaccine tile

### DIFF
--- a/packages/app/src/domain/topical/topical-vaccine-tile.tsx
+++ b/packages/app/src/domain/topical/topical-vaccine-tile.tsx
@@ -1,0 +1,59 @@
+import { Box } from '~/components-styled/base';
+import siteText from '~/locale';
+import Vaccinaties from '~/assets/vaccinaties.svg';
+import { Heading, Text } from '~/components-styled/typography';
+import { formatNumber } from '@corona-dashboard/common';
+import { replaceComponentsInText } from '~/utils/replace-components-in-text';
+
+export function TopicalVaccineTile() {
+  const text = siteText.nationaal_actueel.mini_trend_tiles.vaccinaties;
+
+  const data = {
+    administeredVaccines: parseFloat(siteText.vaccinaties.data.kpi_total.value),
+    expectedDelivery: parseFloat(
+      siteText.vaccinaties.data.kpi_expected_delivery.value
+    ),
+  };
+
+  return (
+    <Box position="relative" pb={3}>
+      <Box width="3.5rem" height="3.5rem" position="absolute" left={0} mr={1}>
+        <Vaccinaties />
+      </Box>
+      <Heading
+        level={3}
+        as="h2"
+        py={2}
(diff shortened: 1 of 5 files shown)


packages/app/src/domain/topical/topical-vaccine-tile.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+export function TopicalVaccineTile() {
+  const text = siteText.nationaal_actueel.mini_trend_tiles.vaccinaties;
+
+  const data = {
+    administeredVaccines: parseFloat(siteText.vaccinaties.data.kpi_total.value),
+    expectedDelivery: parseFloat(
+      siteText.vaccinaties.data.kpi_expected_delivery.value
+    ),
+  };
```
VWSCoronaDashboard1, line 19: Do we need some kind of fallback here? In the case where a wrong value is added in Lokalise, so when it errors the component won't render or something?

VWSCoronaDashboard1 (approved):

packages/app/src/locale/en.json
```diff
@@ -2153,6 +2153,12 @@
       },
       "ic_opnames": {
         "title": "Intensive care admissions"
+      },
+      "vaccinaties": {
```
VWSCoronaDashboard8, line 2157: I suggest to nest or scope this key clearly to "toegediende vaccins" in case we decide to show other vaccin data in the future.

packages/app/src/pages/index.tsx
```diff
@@ -153,27 +152,7 @@ const Home: FCWithLayout<typeof getStaticProps> = (props) => {
...
-                      text.data_driven_texts.intake_intensivecare_ma.difference
-                    }
-                  />
-                }
-                icon={<ArtsIcon />}
-                trendData={dataIntake.values}
-                metricProperty="admissions_moving_average"
-                href="/landelijk/intensive-care-opnames"
-              />
+              <TopicalVaccineTile />
```
VWSCoronaDashboard8, line 155: Not for this PR, but I think we need to look into how we can use composition here instead of creating a separate tile component for every type of content. The header, icon, content should be passed in as props/children I believe, so we can reuse things and it's clear what is being rendered.

VWSCoronaDashboard8 (approved):

packages/app/src/pages/index.tsx
VWSCoronaDashboard1: Add a tech debt ticket I guess

VWSCoronaDashboard8: Yes, done.

packages/app/src/domain/topical/topical-vaccine-tile.tsx
VWSCoronaDashboard4: This exact logic is also in use on the vaccine page itself.

PR #1664, opened 2021-01-27
Replace sanity url builder with images from the filesystem

VWSCoronaDashboard:
## Summary

Parts of the codebase are talking to Sanity through the URL builder.
This PR replaces some of them with our own image component that builds a srcSet and serves images from the local file system.

## Motivation

Avoiding the Sanity CDN limits.

## Unresolved questions
- I'm quite sure the types could be better. Because I'm using a wrapper around  it seems TS does not understand properties such as srcSet, loading and alt natively anymore. I don't mind postponing these enhancements until after an initial release.

### DIFF
--- a/packages/app/src/components-styled/article-detail.tsx
+++ b/packages/app/src/components-styled/article-detail.tsx
@@ -17,7 +17,8 @@ interface ArticleDetailProps {
 }
 
 export function ArticleDetail({ article }: ArticleDetailProps) {
-  const { coverAsset } = article;
+  const { asset } = article.cover;
+  console.log(article);
 
   return (
     <Box bg="white" py={{ _: 4, md: 5 }}>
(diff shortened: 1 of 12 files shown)


packages/app/src/components-styled/article-detail.tsx
```diff
@@ -40,8 +40,6 @@ export function ArticleDetail({ article }: ArticleDetailProps) {
           <PortableText blocks={article.intro} />
         </Box>
 
-        {/* <Image node={article.cover} /> */}
```
VWSCoronaDashboard1, line 43: remove comment

VWSCoronaDashboard1 (changes requested):
This PR isnt complete yet IMHO. There are still the ArticleTeaser and EditorialTeaser components that make use of the 'old' image functionality.
Wouldn't this PR break these?

packages/app/src/components-styled/article-detail.tsx
VWSCoronaDashboard: I don't see this in my codebase, weirdly enough.

VWSCoronaDashboard1: GREMLINS!

VWSCoronaDashboard:
> This PR isnt complete yet IMHO. There are still the ArticleTeaser and EditorialTeaser components that make use of the 'old' image functionality.
> Wouldn't this PR break these?

I think it would. Know that we know how to restructure the GROQ query it should be relatively easy to replace. Let me check that...

VWSCoronaDashboard:
Should be good to go now!

PR #1682, opened 2021-01-27
Improve tooltip layout + 100k scale

VWSCoronaDashboard8:
## Summary

- Improve the tooltip layout + number formatting
- Use 100k scale for y axis

This is not the fully proposed tooltip layout improvement yet, but a step in the right direction.

Screenshot 2021-01-27 at 17 59 30

### DIFF
--- a/packages/app/src/components-styled/stacked-chart/logic.ts
+++ b/packages/app/src/components-styled/stacked-chart/logic.ts
@@ -136,7 +136,7 @@ export function getSeriesData<T extends Value>(
   return metricValues.map(
     (x) =>
       ({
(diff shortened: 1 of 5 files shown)


packages/app/src/components-styled/stacked-chart/logic.ts
```diff
@@ -136,7 +136,7 @@ export function getSeriesData<T extends Value>(
   return metricValues.map(
     (x) =>
       ({
-        ...mapValues(pick(x, metricProperties), (v) => v / 1000000),
```
VWSCoronaDashboard1, line 139: nit: I think TS allows this noawadays: `1_000_000`

VWSCoronaDashboard1 (approved):

PR #1688, opened 2021-01-28
Update article design tweaks

VWSCoronaDashboard14:
CSS tweaks
Article alignment 
Max-width of the article page added with padding
Footer link landelijk fixed

### DIFF
--- a/packages/app/src/domain/topical/article-list.tsx
+++ b/packages/app/src/domain/topical/article-list.tsx
@@ -91,7 +91,6 @@ export function ArticleList({ articleSummaries, hideLink }: ArticleListProps) {
 
 const ArticleBox = styled.div(
   css({
-    display: 'inline-block',
     marginBottom: 4,
     width: asResponsiveArray({
       _: '100%',
--- a/packages/app/src/domain/topical/article-list.tsx
+++ b/packages/app/src/domain/topical/article-list.tsx
@@ -67,10 +67,11 @@ export function ArticleList({ articleSummaries, hideLink }: ArticleListProps) {
         </Box>
       </Box>
       <Box
-        display="block"
+        display="flex"
         alignItems="stretch"
         margin={0}
         maxWidth="100%"
+        flexWrap="wrap"
         mt={3}
       >
         {articleSummaries.map((summary) => (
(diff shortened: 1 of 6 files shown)


packages/app/src/pages/actueel/gemeente/[code].tsx
```diff
@@ -244,7 +245,7 @@ const TopicalMunicipality: FCWithLayout<typeof getStaticProps> = (props) => {
                 borderTopColor={colors.silver}
                 mx={{ _: -3, md: 0 }}
               />
-              <TopicalTile>
+              <TopicalTile css={css({ py: '0' })}>
```
Comment (VWSCoronaDashboard8) [inline packages/app/src/pages/actueel/gemeente/[code].tsx:248]: I think it makes more sense to adjust the TopicalTitle so that it can accept the `py` prop directly.

packages/app/src/pages/actueel/veiligheidsregio/[code].tsx
```diff
@@ -220,7 +220,7 @@ const TopicalSafetyRegion: FCWithLayout<typeof getStaticProps> = (props) => {
                 borderTopColor="silver"
                 mx={{ _: -3, md: 0 }}
               />
-              <TopicalTile>
+              <TopicalTile css={css({ py: '0' })}>
```
Comment (VWSCoronaDashboard8) [inline packages/app/src/pages/actueel/veiligheidsregio/[code].tsx:223]: ...especially if you're going to do this more than once.

VWSCoronaDashboard8 (approved):

packages/app/src/domain/topical/article-list.tsx
```diff
@@ -67,10 +67,11 @@ export function ArticleList({ articleSummaries, hideLink }: ArticleListProps) {
         </Box>
       </Box>
       <Box
-        display="block"
+        display="flex"
         alignItems="stretch"
         margin={0}
         maxWidth="100%"
+        flexWrap="wrap"
```
VWSCoronaDashboard4, line 74: with this change, I think the `display: inline-block` of the child ArticleBox can be removed; I think.

packages/app/src/pages/actueel/veiligheidsregio/[code].tsx
Comment (VWSCoronaDashboard14) [reply packages/app/src/pages/actueel/veiligheidsregio/[code].tsx:223]: Well, only this element is an exception, not for all the other topical tiles. The fact that it keeps repeating is that there are 3 separate actueel pages. I'll look into it.

PR #1698, opened 2021-01-28
Update sitemap grid and arrow from quicklinks

VWSCoronaDashboard14:
The problem was the sitemap didn't fit properly on medium-size screens. It is now breaking to a 4-2 row layout.

Another problem was that in the mini-trend-tile component on a long word the arrow still moves to the next line from a certain point.  Took some time to figure it out but I've set it to absolute only when there is 1 word in the title. Then set it properly and add extra padding to still make the arrow clickable. 

I've also added the span wrapping for the regular link with icons so that when there are 2 words or more and there is a line break the arrow follows it.

### DIFF
--- a/packages/app/src/components-styled/link-with-icon.tsx
+++ b/packages/app/src/components-styled/link-with-icon.tsx
@@ -2,6 +2,7 @@ import css from '@styled-system/css';
 import { ReactNode } from 'react';
 import { UrlObject } from 'url';
 import { Link } from '~/utils/link';
+import { Box } from './base';
 
 interface LinkWithIconProps {
   href: UrlObject | string;
@@ -10,6 +11,10 @@ interface LinkWithIconProps {
   iconPlacement?: 'left' | 'right';
   fontWeight?: 'bold' | 'normal';
   headingLink?: boolean | undefined;
+}
+
+interface IconProps {
+  icon: ReactNode;
 }
 
 export function LinkWithIcon({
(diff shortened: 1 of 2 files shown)


VWSCoronaDashboard4 (approved):

packages/app/src/components-styled/link-with-icon.tsx
```diff
@@ -12,6 +13,11 @@ interface LinkWithIconProps {
   headingLink?: boolean | undefined;
 }
 
+interface IconProps {
+  icon: ReactNode;
+  singleWord?: boolean;
```
VWSCoronaDashboard8, line 18: This is not a good name for a boolean, since its content is not a word. We prefix all booleans with is/has/can/should etc.

packages/app/src/components-styled/link-with-icon.tsx
```diff
@@ -21,6 +27,8 @@ export function LinkWithIcon({
   headingLink,
 }: LinkWithIconProps) {
   const splittedString = children.split(' ');
+  const firstWords = `${splittedString.slice(0, -1).join(' ')} `;
+  const singleWord = splittedString.length === 1;
```
VWSCoronaDashboard8, line 31: same here

packages/app/src/components-styled/link-with-icon.tsx
```diff
@@ -37,39 +45,59 @@ export function LinkWithIcon({
...
-              svg: {
-                height: '11px',
-                width: '13px',
-                mx: '3px',
-              },
-            })}
-          >
-            {icon}
-          </span>
+        {iconPlacement == 'right' && !headingLink && (
```
VWSCoronaDashboard8, line 48: In JS you always use `===` for comparison unless you want to allow implicit casting to happen but that is considered bad practice. In this case it can't hurt, but it's easier to just never use `==` at all.

packages/app/src/components-styled/link-with-icon.tsx
```diff
@@ -21,6 +27,8 @@ export function LinkWithIcon({
   headingLink,
 }: LinkWithIconProps) {
   const splittedString = children.split(' ');
```
VWSCoronaDashboard8, line 29: I think you can simply call this `words` it would make the code a little more terse.

packages/app/src/components-styled/link-with-icon.tsx
```diff
@@ -37,39 +45,59 @@ export function LinkWithIcon({
...
-          </>
+          </Box>
         )}
       </a>
     </Link>
   );
+
+  function IconSmall({ icon }: IconProps) {
+    return (
+      <span css={css({ svg: { height: '11px', width: '13px', mx: '3px' } })}>
```
VWSCoronaDashboard8, line 78: This looks a bit nasty since we don't rely on CSS selectors normally with styled components/system. Can we not control width and height via props on the Icon component?

packages/app/src/domain/topical/data-sitemap.tsx
```diff
@@ -28,8 +29,13 @@ export function DataSitemap() {
         <Box maxWidth={{ md: 'maxWidthText' }} mb={4}>
           <Text>{siteText.nationaal_actueel.data_sitemap_toelichting}</Text>
         </Box>
-        <Box display="flex" flexDirection="row" justifyContent="space-between">
-          <Box spacing={2}>
+        <Box
+          display="flex"
+          flexDirection="row"
```
VWSCoronaDashboard8, line 34: row is the default flex direction so I think this is redundant

packages/app/src/domain/topical/data-sitemap.tsx
```diff
@@ -175,3 +181,13 @@ const Item = styled.li(
...
   })
 );
+
+const LinkBox = styled(Box)(
+  css({
+    width: asResponsiveArray({ md: '25%', lg: 'auto' }),
+    marginTop: asResponsiveArray({ md: 4, lg: 0 }),
+    ':nth-child(-n+4)': {
+      marginTop: 0,
+    },
```
VWSCoronaDashboard8, line 191: What is this for? Is it for setting a margin some of the LinkBox elements? If so, this margin can also be set via props since the LinkBox components are not instantiated programatically I think?

If true, you don't actually need the LinkBox component because the other css properties can also be set as props on Box directly. But yeah that might be a little verbose, unless you create a LinkBox from Box by setting defaultProps and thus create it once. I don't mind much, just some ideas.

VWSCoronaDashboard8 (changes requested):

packages/app/src/domain/topical/data-sitemap.tsx
VWSCoronaDashboard14: Totally agree, it both does the job but I think your proposal is more in line with the standard we use across the dashboard.

PR #1709, opened 2021-01-28
Make tooltip fit on tiny screens

VWSCoronaDashboard8:
## Summary

The tooltip did not fit on the smallest screens. I have solved this by removing the default tooltip x-offset and making the text shorter (also by removing the kind of redundant word "totaal").

image

image

### DIFF
--- a/packages/app/src/components-styled/stacked-chart/stacked-chart.tsx
+++ b/packages/app/src/components-styled/stacked-chart/stacked-chart.tsx
@@ -296,19 +296,22 @@ export function StackedChart<T extends Value>(props: StackedChartProps<T>) {
           <Box mb={2}>
             <InlineText fontWeight="bold">
               {`${replaceVariablesInText(
-                siteText.vaccinaties.verwachte_leveringen.van_week_tot_week,
+                isTinyScreen
+                  ? siteText.vaccinaties.verwachte_leveringen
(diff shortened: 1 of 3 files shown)


VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/stacked-chart/stacked-chart.tsx
```diff
@@ -296,19 +296,22 @@ export function StackedChart<T extends Value>(props: StackedChartProps<T>) {
           <Box mb={2}>
             <InlineText fontWeight="bold">
               {`${replaceVariablesInText(
-                siteText.vaccinaties.verwachte_leveringen.van_week_tot_week,
+                isTinyScreen
```
VWSCoronaDashboard14, line 299: Isn't it better to name checks according to their breakpoint? E.g. `IsExtraSmallScreen` etc.

VWSCoronaDashboard4 (approved):

packages/app/src/components-styled/stacked-chart/stacked-chart.tsx
VWSCoronaDashboard8: @VWSCoronaDashboard14 we don't have a name for anything below xs, and that is what this boolean is about. I already use the name isExtraSmallScreen for < sm

PR #1717, opened 2021-01-29
Use webpack plugin to import specific locale

VWSCoronaDashboard:
## Summary

I saw in the bundle analyzer that we bundle both languages. This PR makes it so we only ship one.

## Motivation

Reduce bundle size.

## Detailed design

There's 2 things you need to check when reviewing:

- I've used webpacks NormalModuleReplacementPlugin to import a specific language file.
- I added a new TS file that mimicks the "normal" import, so it exposes the correct types.
- We only expose the type from the NL file. Never the EN file. As they should be 100% the same I don't think it's a problem.


VWSCoronaDashboard1 (commented):
So, this solution is great for bundle size. Yet, it also removes the compile-time check for missing keys between nl.json and en.json.
In the original file this union `TALLLanguages = TNLLocale | TENLocale` would ensure that all keys are present in both files.

VWSCoronaDashboard7:
> So, this solution is great for bundle size. Yet, it also removes the compile-time check for missing keys between nl.json and en.json.
> In the original file this union `TALLLanguages = TNLLocale | TENLocale` would ensure that all keys are present in both files.

If we want to reduce the bundle-size and have this check, that type-enforced check could be replaced by a custom schema validator since the locale files are also just a json file with some structure. Of course this is some more effort than that union approach. No sure if there are other trade-offs involved.

VWSCoronaDashboard1:
> If we want to reduce the bundle-size and have this check, that type-enforced check could be replaced by a custom schema validator since the locale files are also just a json file with some structure. Of course this is some more effort than that union approach. No sure if there are other trade-offs involved.

It would actually be possible to generate a json schema @build time from the nl.json file and use that schema to validate the en.json. That would probably be the fastest way of performing this check.

VWSCoronaDashboard1:
> It would actually be possible to generate a json schema @ build-time from the nl.json file and use that schema to validate the en.json. That would probably be the fastest way of performing this check.

Come to think of it, this would probably be a much simpler approach since we only want to compare the structure:
https://www.npmjs.com/package/json-diff

VWSCoronaDashboard1 (approved):

packages/app/src/locale/APP_TARGET.ts
```diff
@@ -0,0 +1,7 @@
+// This file is meant to trick TypeScript.
+// We're using a webpack plugin to rewrite some imports based on desired locale.
+// This file just exports the type from the NL file so all types still work.
+import localefile from './nl.json';
```
VWSCoronaDashboard4, line 4:
```suggestion
import locale from './nl.json';
```

packages/cli/src/validator/validate-locales.ts
```diff
@@ -0,0 +1,48 @@
+import chalk from 'chalk';
+import meow from 'meow';
+
+import jsonDiff from 'json-diff';
+
+import nl from '../../../app/src/locale/nl.json';
+import en from '../../../app/src/locale/en.json';
+
+function objectsHaveSameKeys(
+  obj1: Record<string, unknown>,
```
VWSCoronaDashboard4, line 10: Can we perhaps get rid of the `unknown`?

```suggestion
  obj1: Record<string, LocaleValue>,
```

if you define `type LocaleValue = string | Record<string, LocaleValue>;` or something similar.

VWSCoronaDashboard4 (approved):

package.json
```diff
@@ -32,7 +32,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "yarn workspace @corona-dashboard/app lint-staged",
+      "pre-commit": "yarn workspace @corona-dashboard/app lint-staged && yarn workspace @corona-dashboard/cli validate-locales",
```
VWSCoronaDashboard4, line 35: For future improvement: a way to run these async to speed up dev time.

packages/app/next.config.js
```diff
@@ -32,6 +32,24 @@ const nextConfig = {
       sitemap.generateSitemap(process.env.NEXT_PUBLIC_LOCALE);
     }
 
+    // To prevent importing two languages, we use the NormalModuleReplacementPlugin plugin
+    // We match any import that uses APP_TARGET and replace it with the value of
+    // process.env.NEXT_PUBLIC_LOCALE
+    // e.g. ~/src/locale/APP_TARGET.json becomes ~/src/locale/nl.json
```
VWSCoronaDashboard8, line 38: nit; I prefer to stick to jsdoc style comment starting with `/**` and ending with `*/` It makes it easier to re-format and add lines. Also I feel it makes the comments stand out better and distinguish themselves from commented-out code chunks.

packages/app/src/locale/APP_TARGET.ts
```diff
@@ -0,0 +1,7 @@
+// This file is meant to trick TypeScript.
+// We're using a webpack plugin to rewrite some imports based on desired locale.
+// This file just exports the type from the NL file so all types still work.
```
VWSCoronaDashboard8, line 3: Same comment here. BTW the VSCode extension "rewrap" helps with this kind of thing

packages/app/src/locale/APP_TARGET.ts
```diff
@@ -0,0 +1,7 @@
+// This file is meant to trick TypeScript.
```
VWSCoronaDashboard8, line 1: I don't see much value in naming this file upper-case since it's not a constant and I don't know what the meaning would be.

packages/app/src/locale/index.ts
```diff
@@ -1,21 +1,18 @@
...
 
 export type TLanguages = {
-  nl: TNLLocale;
-  en: TENLocale;
+  nl: TLocale;
+  en: TLocale;
 };
 
-export type TALLLanguages = TNLLocale | TENLocale;
+export type TALLLanguages = TLocale;
```
VWSCoronaDashboard8, line 10: Maybe for later, but now that we have one type, we can remove TALLLanguages and just to a find/replace with TLocale. In that sense you might as well name it Locale since we stopped prefixing types with T or I.

packages/cli/src/validator/validate-locales.ts
```diff
@@ -0,0 +1,48 @@
+import chalk from 'chalk';
+import meow from 'meow';
+
+import jsonDiff from 'json-diff';
+
+import nl from '../../../app/src/locale/nl.json';
+import en from '../../../app/src/locale/en.json';
```
VWSCoronaDashboard8, line 7: One package should not import files from another package directly so this not quite ok. But then again app is not set up to be a shared module, so I'll guess I can just ignore this for now.

packages/cli/src/validator/validate-locales.ts
VWSCoronaDashboard8: I don't know if there are still arrays in the locale file, but then they would have to be included as well I guess...

packages/cli/src/validator/validate-locales.ts
```diff
@@ -0,0 +1,48 @@
...
+
+function objectsHaveSameKeys(
+  obj1: Record<string, unknown>,
+  obj2: Record<string, unknown>
+) {
+  const objects = [obj1, obj2];
+  const allKeys: string[] = objects.reduce(
+    (keys: string[], object) => keys.concat(Object.keys(object)),
+    []
```
VWSCoronaDashboard8, line 16: If you type the second argument you don't have to type the function arguments. The object type is already known from objects.

```suggestion
  const allKeys: string[] = objects.reduce(
    (keys, object) => keys.concat(Object.keys(object)),
    [] as string[]
```

packages/cli/src/validator/validate-locales.ts
```diff
@@ -0,0 +1,48 @@
...
+  );
+
+  const union = new Set(allKeys);
```
VWSCoronaDashboard8, line 19: You could have used a Set in the reduce straight away.

packages/cli/src/validator/validate-locales.ts
```diff
@@ -0,0 +1,48 @@
...
+  return objects.every((object) => union.size === Object.keys(object).length);
+}
+
+if (objectsHaveSameKeys(nl, en)) {
```
VWSCoronaDashboard8, line 23: This `objectsHaveSameKeys` function seems to only check the top-level keys, so I doubt if this validation is useful. I think you better compile a list of strings where each nested property is listed `some.nested.property` and then compare those lists. This would require a recursive function.

VWSCoronaDashboard8: Maybe you can get around this by doing a jsonDiff straight-away and then look at the output to see if the diff is only values or also differences in key (add/remove). I suspect the diff format will tell you that.

VWSCoronaDashboard8 (changes requested):

packages/app/src/locale/APP_TARGET.ts
VWSCoronaDashboard: It is a constant, look at the Webpack config. Or at least it should look like a constant and be recognisable as one even if it's technically not true.

packages/cli/src/validator/validate-locales.ts
VWSCoronaDashboard: Updated this!

packages/cli/src/validator/validate-locales.ts
VWSCoronaDashboard: Done exactly that now. Json-diff is the single source of truth for equality. I also don't think this check is as important really, we're better off building in fallbacks in the front-end and checking for empty values in the CMS or Lokalise.

VWSCoronaDashboard1 (approved):

packages/app/next.config.js
```diff
@@ -32,6 +32,25 @@ const nextConfig = {
...
+    /** To prevent importing two languages, we use the NormalModuleReplacementPlugin plugin
+     *  We match any import that uses APP_TARGET and replace it with the value of
+     *  process.env.NEXT_PUBLIC_LOCALE
+     *  e.g. ~/src/locale/APP_TARGET.json becomes ~/src/locale/nl.json
+     */
+    var appTarget = process.env.NEXT_PUBLIC_LOCALE || 'nl';
+
+    config.plugins.push(
+      new webpack.NormalModuleReplacementPlugin(
+        /(.*)APP_TARGET(\.*)/,
```
VWSCoronaDashboard8, line 44: nit; I wonder why you choose to call it "app target" and not just "locale" or something. I associate target with compile target/platform, and the name locale would have been clear to me.

packages/app/src/locale/APP_TARGET.ts
```diff
@@ -0,0 +1,8 @@
+/** This file is meant to trick TypeScript. We're using a webpack plugin to
+ * rewrite some imports based on desired locale. This file just exports the
+ * type from the NL file so all types still work.
+ */
+import locale from './nl.json';
+
+export type TLocale = typeof locale;
+export default locale;
```
VWSCoronaDashboard8: Is it necessary to use a default export? We choose to avoid it in other places.

packages/app/src/locale/index.ts
VWSCoronaDashboard8, line 10: I would suggest doing a find/replace on `TALLLanguages` and name `TLocale` simply `Locale` while you're at it

packages/cli/src/validate-schema/validate-locales.ts
```diff
@@ -0,0 +1,49 @@
+import chalk from 'chalk';
+
+import jsonDiff from 'json-diff';
+
+import nl from '@corona-dashboard/app/src/locale/nl.json';
+import en from '@corona-dashboard/app/src/locale/en.json';
```
VWSCoronaDashboard8: I guess we could move this to the common package at some point. Together with the schema definitions.

packages/cli/src/validate-schema/validate-locales.ts
```diff
@@ -0,0 +1,49 @@
...
+
+function objectsHaveSameKeys(
```
VWSCoronaDashboard8: I find this function name a little odd, since it doesn't start with a verb. I suggest something like `matchObjectKeys` or `validateObjectKeysEquality`

packages/cli/src/validate-schema/validate-locales.ts
```diff
@@ -0,0 +1,49 @@
...
+  obj1: Record<string, unknown>,
+  obj2: Record<string, unknown>
+) {
+  // @ts-ignore
```
VWSCoronaDashboard8: What are you trying to ignore here? Maybe you can solve it by casting the arguments to `diff`?

packages/cli/src/validate-schema/validate-locales.ts
```diff
@@ -0,0 +1,49 @@
...
+  const diff = jsonDiff.diff(obj1, obj2, { keysOnly: true });
+
+  // jsonDiff returns undefined if there are no differences
+  if (diff) {
+    // which means if diff is truthy, we found differences and should throw an error
```
VWSCoronaDashboard8: I'd suggest to combine this in to a single JSDoc style comment for readability.

packages/cli/src/validate-schema/validate-locales.ts
```diff
@@ -0,0 +1,49 @@
...
+  );
+  process.exit(0);
+} else {
+  console.group();
+  console.error(chalk.bgRed.bold(`  Locale files have different keys  \n`));
+  console.log(
+    jsonDiff.diffString(
+      nl,
+      en,
+      {},
```
VWSCoronaDashboard8: It seems odd that you are passing something to diffString if the library doesn't accept it in its types. I think a comment would be useful here.

VWSCoronaDashboard8 (approved):

package.json
```diff
@@ -31,7 +31,7 @@
...
```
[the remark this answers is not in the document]
VWSCoronaDashboard: Afaik lokalize will not remove keys. But it doesn't matter, we've removed the custom validator and let TypeScript do an equality check again. If en !== nl it errors.

packages/app/next.config.js
VWSCoronaDashboard: If I remember correctly it's because I just followed the documentation blindly without thinking about the filename. Just wanted to have it working first. I can still rename this before I merge :).

packages/app/src/locale/APP_TARGET.ts
VWSCoronaDashboard: That's a discussion for another PR or ticket 👍  (but I agree with you)

packages/app/src/locale/APP_TARGET.ts
```diff
@@ -1,29 +1,33 @@
...
-export type TLocale = typeof locale;
+export type Locale = typeof locale;
 
 import en from './en.json';
 import nl from './nl.json';
 
-export type TNLLocale = typeof nl;
-export type TENLocale = typeof en;
+export type NLLocale = typeof nl;
+export type ENLocale = typeof en;
```
VWSCoronaDashboard8, line 17: I'm not 100% sure if we reached consensus on this, but in most other places were are camel-casing acronyms just like normal words. I personally believe it's the only way to stay consistent with types and instances. For example `jsonParser` for instance and `JsonParser` for type/class. Otherwise, what are you are going to call them. jSONParser and JSONParser? or jsonParser and JSONParser? It just easier to think of them as regular words.

```suggestion
export type NlLocale = typeof nl;
export type EnLocale = typeof en;
```

packages/app/next.config.js
VWSCoronaDashboard: Renamed it to APP_LOCALE. It's clear what it means and I like that it's an easy, explicit selector. Don't want to get it confused with other files called locale.

packages/app/src/locale/APP_TARGET.ts
VWSCoronaDashboard: I would go for camelCasing too. It's much easier to read.

PR #1754, opened 2021-02-01
Add article strips to all appropriate pages

VWSCoronaDashboard1:
## Summary

This PR adds a component that displays a number of article teasers to the following pages:
- behavior
- deceased
- hospital
- intensive care
- positive tests
- reproduction
- sewerwater
- vaccinations

They are added to national, safety regions and municipalities where appropriate.

## Motivation

Feature request

## Detailed design

In the CMS the following schemas have been added:
- behaviorPage
- deceasedPage
- hospitalPage
- intensiveCarePage
- positiveTestsPage
- reproductionPage
- sewerPage
- vaccinationsPage

All of the schemas now just have one field which is called 'articles', which is a list of references to an Article document.
Per schema, the maximum of references in the list differs though, so that's why the fields aren't shared.

These schemas will be added to in later stages, I'm sure. When a specialized schema is needed for safety regions and/or municipalities these can be added later and the queries can be combined in the static props phase.

To show the article teasers on the pages the `ArticleStrip` component was added, which is a simple list of clickable images and article titles.

CMS-wise, the desk-structure.js file has been cleaned up a little to prevent a lot of code duplication while creating the menu structure.

Lastly, the CMS has been upgraded to the latest Sanity versions.

## Drawbacks

n/a
## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added


packages/app/src/pages/landelijk/sterfte.tsx
```diff
@@ -1,32 +1,43 @@
...
 
 const text = siteText.sterfte;
 
 export const getStaticProps = createGetStaticProps(
   getLastGeneratedDate,
-  getNlData
+  getNlData,
+  createGetContent<{
+    articles?: ArticleSummary[];
+  }>(createPageArticlesQuery('deceasedPage'))
```
VWSCoronaDashboard10, line 33: I have the feeling that the `createGetContent` is not very scalable, we might want to discuss this during FE meeting.

VWSCoronaDashboard1: To be honest, I think it's fine. If `createGetContent `simply makes sure the query gets executed, then we can build all sorts of logic that assembles this query. But, yes, it's a good subject for discussion.

packages/app/src/components-styled/article-strip.tsx
```diff
@@ -0,0 +1,101 @@
...
+              slug={article.slug.current}
+            />
+          </Box>
+        ))}
+      </Box>
+    </Tile>
+  );
+}
+
+type ArticleStripItemProps = {
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: Oh, I've been using types exclusively haha

packages/app/src/components-styled/article-strip.tsx
```diff
@@ -0,0 +1,101 @@
...
+  slug: string;
+  cover: ImageBlock;
+  title: string;
+};
+
+const ArticleStripItem = function (props: ArticleStripItemProps) {
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: I blame lack of sleep ;)

packages/app/src/components-styled/article-strip.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+export function ArticleStrip(props: ArticleStripProps) {
+  const { articles } = props;
+
+  if (!articles?.length) {
+    return null;
+  }
+
+  return (
+    <Tile css={css({ bg: 'lightBlue' })}>
```
VWSCoronaDashboard8, line 27: nit; if tiles commonly need background colors I think it makes sense to support it as a prop (by including that part of styled system).

packages/cms/desk-structure.js
```diff
@@ -38,234 +39,138 @@ import 'sanity-mobile-preview/dist/index.css?raw';
 // };
 
 // hiddenDocTypes will filter out all of the content models
-// we expose through other sections in the CMS.
+// we expose through other sections in the CMStructureBuilder.
```
VWSCoronaDashboard8, line 42: Looks like some accidental find/replace

VWSCoronaDashboard8 (approved):

packages/app/src/components-styled/article-strip.tsx
```diff
@@ -0,0 +1,95 @@
...
+function ArticleStripItem(props: ArticleStripItemProps) {
+  const { slug, cover, title } = props;
+
+  return (
+    <Link passHref href={`/artikelen/${slug}`}>
+      <StyledLink>
+        <Box width={122} maxHeight={122} overflow="hidden">
+          <Image
+            {...getImageProps(cover)}
+            width={122}
```
VWSCoronaDashboard4, line 68: This should not be needed. You can pass the `122` as second argument to `getImageProps` and get the proper width and height props back

packages/cms/schemas/documents/pages/hospital-page.js
```diff
@@ -0,0 +1,14 @@
+export default {
+  title: 'Ziekenhuis opnames',
+  name: 'hospitalPage',
+  type: 'document',
+  fields: [
+    {
+      title: 'Uitgelichte artikelen',
```
VWSCoronaDashboard4, line 7: Wouldn't it make sense to create an object/type for this? Now these identical properties are copied over. If we want to adjust these articles, we can do it in 1 spot in the CMS and in 1 spot in the App.

packages/cms/schemas/documents/pages/positive-tests-page.js
```diff
@@ -0,0 +1,14 @@
+export default {
```
VWSCoronaDashboard4, line 1: Is this the positive test page on NL, VR or GM level? Or all?

VWSCoronaDashboard4 (changes requested):

packages/cms/schemas/documents/pages/hospital-page.js
VWSCoronaDashboard1: As stated in the PR description: "Per schema, the maximum of references in the list differs though, so that's why the fields aren't shared."
Because this deals with just one field, I don't really see how putting just the one field in a shared field helps us at this point. The field itself isn't really much subject to change.
We should avoid anything like inheritance in the schema's, that's an anti-pattern and discouraged by Sanity themselves.

VWSCoronaDashboard1: I think right now it seems like a lot of duplication since all the pages only share the article selection, but I think in the future they will become very much different from each other and this approach will make more sense.
Anyway, let's have that meeting ;)

packages/cms/schemas/documents/pages/positive-tests-page.js
VWSCoronaDashboard1: "When a specialized schema is needed for safety regions and/or municipalities these can be added later and the queries can be combined in the static props phase." hehehe.
These are global, indeed.

PR #1755, opened 2021-02-01
Add page 'toegankelijkheid'

VWSCoronaDashboard10:
## Summary

This PR adds the `toegankelijkheid`-page. One of the requirements was the possibility to link to uploaded assets.

### **Things done:**
- refactored the bash script responsible for sanity asset export to be able to export both images as files
- refactored image url building: this is now handled on a single location using a utility function: `getImageProps()`
- create `getImageSrc()` to get the url for an image
- create `getFileSrc()` to get the url for a file
- Add `inlineAttachment`-annotation to the `RichContentBlock` type

### **Breaking changes:**
- The `images`-script of the cms package.json has been renamed to `assets`
- the download script requires a `SANITY_DATASET` environment-variable.


packages/app/src/pages/artikelen/[slug].tsx
```diff
@@ -47,14 +45,22 @@ export const getStaticProps = createGetStaticProps(
   		    ...content.nl[]
 			    {
       	    ...,
-      	    "asset": asset->
+      	    "asset": asset->,
```
Comment (VWSCoronaDashboard4) [inline packages/app/src/pages/artikelen/[slug].tsx:48]: Indentation is not going well here

VWSCoronaDashboard1 (approved):

packages/app/src/pages/artikelen/[slug].tsx
```diff
@@ -80,9 +86,8 @@ ArticleDetailPage.getLayout = (page, props) => {
   const { cover } = props.content;
   const { asset } = cover;
 
-  const url = `https://coronadashboard.rijksoverheid.nl/cms/${
-    asset.assetId
-  }-${findClosestSize(1200, imageResizeTargets)}.${asset.extension}`;
+  const imgPath = getImageSrc(asset, 1200);
+  const url = `https://coronadashboard.rijksoverheid.nl${imgPath}`;
```
Comment (VWSCoronaDashboard4) [inline packages/app/src/pages/artikelen/[slug].tsx:90]: Shouldn't this go to the `government.nl` domain for the EN locale?

packages/app/src/pages/weekberichten/[slug].tsx
```diff
@@ -81,9 +87,8 @@ EditorialDetailPage.getLayout = (page, props) => {
   const { cover } = props.content;
   const { asset } = cover;
 
-  const url = `https://coronadashboard.rijksoverheid.nl/cms/${
-    asset.assetId
-  }-${findClosestSize(1200, imageResizeTargets)}.${asset.extension}`;
+  const imgPath = getImageSrc(asset, 1200);
+  const url = `https://coronadashboard.rijksoverheid.nl${imgPath}`;
```
Comment (VWSCoronaDashboard4) [inline packages/app/src/pages/weekberichten/[slug].tsx:91]: Same comment as the other; perhaps this full URL could be part of the utility as well?

VWSCoronaDashboard4 (approved):
Neat! Some small questions for improvements/whether something is a possible bug.

packages/app/src/pages/artikelen/[slug].tsx
Comment (VWSCoronaDashboard1) [reply packages/app/src/pages/artikelen/[slug].tsx:90]: Actually, do we need a full URL here? Wouldn't a relative one suffice?

VWSCoronaDashboard8 (approved):

packages/app/src/pages/artikelen/[slug].tsx
Comment (VWSCoronaDashboard10) [reply packages/app/src/pages/artikelen/[slug].tsx:90]: > Shouldn't this go to the government.nl domain for the EN locale?

I guess that would be prettier yes, but for now I'm not going to change it because:
- the absolute url is used for `twitterImage` and `ogImage`, I think those can be from another domain without issues
- I just found out that the domain is prefixed in the SeoHead component, which means it is currently broken on production (urls are like `https://coronadashboard.rijksoverheid.nlhttps://coronadashboard.rijksoverheid.nl/cms/3f9be5cf64303c2221de128b8cb83a04f7b1de89-1280.jpg`)

> Actually, do we need a full URL here? Wouldn't a relative one suffice?

IIRC the openGraph image requires an absolute url, relative urls won't work.

PR #1760, opened 2021-02-02
Add mobile menu animation

VWSCoronaDashboard14:
Added an animation when opening and closing the menu. It is animating with the same transition as the collapsible component.


VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/layout/components/top-navigation.tsx
```diff
@@ -19,19 +19,34 @@ export function TopNavigation() {
...
+  }, [isMenuOpen]);
+
   function toggleMenu() {
+    if (isAnimating) return;
+
+    setIsAnimating(!isAnimating);
     setIsMenuOpen(!isMenuOpen);
   }
 
+  function onTransitionEnd() {
```
VWSCoronaDashboard8, line 46: We follow a pattern were internal handlers are prefixed with `handle` and handler prop names with `on`. This makes it better to follow and handle something internally and then pass it on to a prop handler.

So this function should be named `handleTransitionEnd`

VWSCoronaDashboard8 (changes requested):

VWSCoronaDashboard8 (approved):
I'll approve this assuming you will change the handler name.

PR #1762, opened 2021-02-02
Add images to long text content pages and collapsible

VWSCoronaDashboard4:
## Summary

Add images to long text content pages and collapsible

## Detailed design

* Adjusts `localeBlock` to support images, ad used in about pages, collapsible component, intros of maatregelen, articles, weekberichten; which requires edits of the `groq`-queries.
* Adds support for SVG images
* Fix alignment and styling of images used in non-article-detail places


packages/app/src/components-styled/cms/content-image.tsx
```diff
@@ -32,12 +35,12 @@ export function ContentImage({ node }: ContentImageProps) {
       </MaxWidth>
     </Box>
   ) : (
-    <ContentBlock>
-      <Box as="figure" role="group" spacing={3} textAlign="center">
+    <ContentWrapper>
+      <Box as="figure" role="group" spacing={3} my={2} textAlign="center">
```
VWSCoronaDashboard10, line 39: please don't use outside margins :-)

packages/app/src/components-styled/cms/rich-content.tsx
```diff
@@ -1,41 +1,45 @@
...
-        PortableText.defaultSerializers.types?.block,
-        'PortableText needs to provide a serializer for block content'
-      );
-      return (
-        <ContentBlock>
-          {PortableText.defaultSerializers.types.block(props)}
-        </ContentBlock>
-      );
+export function RichContent({ contentWrapper, blocks }: RichContentProps) {
+  const ContentWrapper = contentWrapper ?? Fragment;
```
VWSCoronaDashboard10, line 18: I'm not sure I like this pattern used here but I understand the purpose and can't think of a better solution right now..

packages/app/src/pages/over-risiconiveaus.tsx
```diff
@@ -2,26 +2,64 @@ import Head from 'next/head';
...
+    ],
+    "en": [
+      ...description.en[]
+      {
+        ...,
+        "asset": asset->
+       },
+    ],
+  },
+  "collapsibleList": [...collapsibleList[]
```
VWSCoronaDashboard1, line 42: A bit of nit: But I've been writing my queries like this:
```
*[_type == 'overRisicoNiveaus']{
  ...,
  "description": {
    "_type": description._type,
      ...description.${targetLanguage}[]
      {
        ...,
        "asset": asset->
       },
```

That way only the required language is retrieved. IMHO, it simplifies the query a little bit and saves a sliver of bandwidth.
Take a look at https://github.com/minvws/nl-covid19-data-dashboard/blob/develop/packages/app/src/pages/artikelen/index.tsx for example.

packages/app/src/components-styled/cms/rich-content.tsx
VWSCoronaDashboard10: I think the RichContent was previously designed only with "body content" in mind, not to be rendered as "pieces of content" on different places. We might be able to refactor it with that knowledge in mind, but that's something for another PR.

packages/app/src/components-styled/cms/rich-content.tsx
```diff
@@ -1,41 +1,45 @@
...
+          </ContentWrapper>
+        );
+      },
+      image: (props: { node: ImageBlock | RichContentImageBlock }) => (
+        <ContentImage contentWrapper={contentWrapper} {...props} />
+      ),
     },
+    marks: {
+      inlineAttachment,
+    },
```
VWSCoronaDashboard10, line 38: Could you, since you're touching these lines, change this to:
```tsx
    marks: {
      inlineAttachment: InlineAttachment,
    },
```
and rename that function component accordingly? thanks 😇

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard10 (approved):

packages/app/src/components-styled/cms/content-image.tsx
VWSCoronaDashboard4: Not sure what the alternative should be, in this case? This gives margins equal to eg `<p>` on the same level.

VWSCoronaDashboard10: I'd say all children should be wrapped with a `<Box spacing={2} />`

packages/app/src/components-styled/cms/rich-content.tsx
VWSCoronaDashboard10: Or that might break with the imported type with the same name .. maybe rename the function to `InlineAttachmentMark`

PR #1765, opened 2021-02-02
Add highlight path on safety region on choropleth

VWSCoronaDashboard14:
The choropleth supports now a safety-region highlight code.
Had to draw 2 new paths (inner and outer) on top of the map to make them not clip with regular paths.
Create separation of the `StyledPath` and `StyledHoverPath` to make it easier to style them.
Introduced also a `riskLevelStyling` boolean to have separate styling for a municipality or a risk level choropleth


VWSCoronaDashboard1 (commented):
I think in a meeting about this it was suggested that perhaps the same effect could be attained by changing the stroke width of the region feature (making the underlying feature's stroke very wide and dark) and displaying the hover feature on top of it (with a narrower stroke and lighter color).

Was this solution researched and later discarded because it didn't work?

While the solution in this PR works, it does add another set of features to the map, making it heavier on the DOM. While the region map isn't that huge, should the same feature be required for the municipal map in the future, the solution would no longer be practical since that would add another 350 features to the map.

packages/app/src/components/choropleth/safety-region-choropleth.tsx
```diff
@@ -29,7 +29,8 @@ type SafetyRegionChoroplethProps<T, K extends RegionsMetricName> = {
   highlightSelection?: boolean;
   onSelect?: (context: SafetyRegionProperties) => void;
   tooltipContent?: (context: SafetyRegionProperties & T) => ReactNode;
-  isSelectorMap?: boolean;
+  highlightCode?: string;
```
VWSCoronaDashboard1, line 32: nit, the prop for the selected feature is called 'selected', perhaps name this one 'highlighted'.

VWSCoronaDashboard1 (approved):

packages/app/src/components/choropleth/choropleth.tsx
```diff
@@ -37,6 +37,12 @@ type TProps<T1, T3> = {
     path: string,
     index: number
   ) => ReactNode;
+
+  highightCallback?: (
```
VWSCoronaDashboard8, line 41: Typo => install a spell checker?

packages/app/src/components/choropleth/choropleth.tsx
```diff
@@ -189,6 +196,14 @@ const ChoroplethMap: <T1, T3>(
               />
             </g>
           )}
+
+          {highightCallback && (
+            <MercatorGroup
+              data={featureCollection.features}
+              render={highightCallback}
```
VWSCoronaDashboard8, line 203: If this is a render callback I suggest you give it a name that indicates that. We do it in other places too afaik. I `renderHighlight` for example. The fact that it is a callback is already inherent to the type function, so it's not so valuable to use the word callback in the name.

packages/app/src/components/choropleth/path.tsx
```diff
@@ -17,15 +16,34 @@ export function Path({
...
+      fill={fill}
+      stroke={stroke}
+      strokeWidth={strokeWidth}
+    />
+  );
+}
+
+export function HoverPath({
+  id,
+  d,
```
VWSCoronaDashboard8, line 36: `d` is not a good name for a variable, since it doesn't tell you anything. Maybe `pathData` or something...

packages/app/src/components/choropleth/path.tsx
```diff
@@ -35,33 +53,37 @@ export function Path({
...
-  (x) =>
-    css({
-      fill: x.fill || 'transparent',
-      stroke: x.stroke,
-      strokeWidth: x.strokeWidth || 0.5,
-      pointerEvents: 'none',
+}>((x) =>
+  css({
+    fill: x.fill || 'transparent',
+    stroke: 'x.stroke',
```
VWSCoronaDashboard8, line 60: This looks like a mistake. And if you observed things working with this, it might be redundant?

packages/app/src/components/choropleth/safety-region-choropleth.tsx
VWSCoronaDashboard8: 👍  It would be also be nice to rename selected to `selectedCode` while you're at it, since "selected" is not a clear variable name.

packages/app/src/components/choropleth/safety-region-choropleth.tsx
```diff
@@ -113,26 +115,50 @@ export function SafetyRegionChoropleth<T, K extends RegionsMetricName>(
     [getFillColor, hasData]
   );
 
+  const highightCallback = useCallback(
```
VWSCoronaDashboard8, line 118: More typo

packages/app/src/components/choropleth/safety-region-choropleth.tsx
```diff
@@ -52,11 +53,12 @@ export function SafetyRegionChoropleth<T, K extends RegionsMetricName>(
   const {
     data,
     selected,
-    highlightSelection = true,
     metricName,
     metricProperty,
     onSelect,
     tooltipContent,
+    highlightCode,
+    riskLevelStyling = false,
```
VWSCoronaDashboard8, line 61: This name doesn't indicate that it's a boolean

VWSCoronaDashboard8 (changes requested):

VWSCoronaDashboard10:
> I think in a meeting about this it was suggested that perhaps the same effect could be attained by **changing the stroke width of the region feature** (making the underlying feature's stroke very wide and dark) and displaying the hover feature on top of it (with a narrower stroke and lighter color).

I think this won't work because other adjacent layers could be rendered on top of the selected feature which would result in odd missing borders on some places.

> While the solution in this PR works, it does add another set of features to the map, making it heavier on the DOM. While the region map isn't that huge, should the same feature be required for the municipal map in the future, the solution would no longer be practical since that would add another 350 features to the map.

I think the impact on the DOM is minimal because the callback returns `null` for features that don't match the highlighted feature, or am I wrong?

packages/app/src/components/choropleth/choropleth.tsx
VWSCoronaDashboard10: I agree, but currently all render callbacks of the choropleth component are currently named like this. It can be fixed with this PR, but I think it's also fine to do this in another PR

packages/app/src/components/choropleth/path.tsx
VWSCoronaDashboard10: Ah yes I agree .. I initially named it `d` on the original Path component to mimic the SVG path element [`d`-attribute](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/d), but it's of course nice to have a readable prop :p

packages/app/src/components/choropleth/path.tsx
```diff
@@ -35,33 +53,37 @@ export function Path({
...
-      '&:hover': {
-        transitionDuration: '0ms',
-        fill: x.fill ?? 'none',
-        stroke: x.stroke ?? '#222',
-        strokeWidth: x.strokeWidth ?? 2,
-      },
-    })
+    transitionProperty: 'fill, stroke, stroke-width',
+    transitionDuration: '120ms, 90ms',
+    transitionTimingFunction: 'ease-out',
```
VWSCoronaDashboard10, line 66: I guess these are obsolete now, they were previously meant for the `:hover` styles but those have been moved to the StyledHoverPath component.

packages/app/src/components/choropleth/safety-region-choropleth.tsx
```diff
@@ -113,26 +115,50 @@ export function SafetyRegionChoropleth<T, K extends RegionsMetricName>(
...
+            stroke="#fff"
+            strokeWidth={8}
+          />
+          <Path
+            key={`${vrcode}-inside-border`}
+            d={path}
+            stroke="#000"
+            strokeWidth={2}
+          />
+        </>
```
VWSCoronaDashboard10, line 138: I think the `key`-props can be moved to the fragment
```suggestion
        <Fragment key={vrcode}>
          <Path
            d={path}
            stroke="#fff"
            strokeWidth={8}
          />
          <Path
            d={path}
            stroke="#000"
            strokeWidth={2}
          />
        </Fragment>
```

packages/app/src/pages/index.tsx
```diff
@@ -198,6 +198,7 @@ const Home: FCWithLayout<typeof getStaticProps> = (props) => {
                       tooltipContent={escalationTooltip(
                         createSelectRegionHandler(router, 'actueel')
                       )}
+                      riskLevelStyling={true}
```
VWSCoronaDashboard10, line 201: It looks like the risk level styling should be applied on the choropleths with the metricName `escalation_levels`, so I think we should be able to remove this boolean prop and instead do something like this inside the SafetyRegionChoropleth  function body:

```tsx
const riskLevelStyling = metricName === 'escalation_levels'

// or maybe a better name:
const isEscalationLevelTheme = metricName === 'escalation_levels'

packages/app/src/components/choropleth/choropleth.tsx
VWSCoronaDashboard14: I'll do it in another pull-request after this today, to keep things separate.

VWSCoronaDashboard4 (approved):

packages/app/src/pages/index.tsx
VWSCoronaDashboard10: @VWSCoronaDashboard14 what do you think of this suggestion?

VWSCoronaDashboard14: That was honestly pretty smart, I've already changed it but didn't resolve the conversation yet.

PR #1781, opened 2021-02-03
Make highlighted-articles field a shared field

VWSCoronaDashboard1:
## Summary

As the title explains

## Motivation

As it turns out, the maximum number of article strip items is the same across all pages, so I've refactored the highlighted-article field into a shared field that's used by the separate page schemas.

## Detailed design

n/a

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added

### DIFF
--- a/packages/cms/schemas/documents/fields/highlighted-article.js
+++ b/packages/cms/schemas/documents/fields/highlighted-article.js
@@ -1,0 +1,7 @@
+export const HIGHLIGHTED_ARTICLE = {
+  title: 'Uitgelichte artikelen',
+  name: 'articles',
+  type: 'array',
+  of: [{ type: 'reference', to: { type: 'article' } }],
+  validation: (Rule) => Rule.required().unique().max(2),
+};
--- a/packages/cms/schemas/documents/fields/highlighted-articles.js
+++ b/packages/cms/schemas/documents/fields/highlighted-articles.js
@@ -1,4 +1,4 @@
-export const HIGHLIGHTED_ARTICLE = {
+export const HIGHLIGHTED_ARTICLES = {
   title: 'Uitgelichte artikelen',
   name: 'articles',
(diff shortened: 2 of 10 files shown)


packages/cms/schemas/documents/fields/highlighted-article.js
```diff
@@ -0,0 +1,7 @@
+export const HIGHLIGHTED_ARTICLE = {
```
VWSCoronaDashboard4, line 1: This can be put with the other re-used objects in `packages/cms/schemas/objects/`

Naming wise, the others use`export default`; but we import them as camelcase. Perhaps stick to the camelcase if you want to name this?

VWSCoronaDashboard1: Actually no, these are shared fields, not shared objects, hence they are placed in the directory 'fields' ;)
I took the upper casing from the Sanity examples where this is used by convention. (I take it they've gleaned this from old relational database systems).

packages/cms/schemas/documents/fields/highlighted-articles.js
```diff
@@ -0,0 +1,7 @@
+export const HIGHLIGHTED_ARTICLES = {
```
VWSCoronaDashboard8, line 1: nit; I'm not sure it makes sense to write this as if it were a constant value.

VWSCoronaDashboard8 (approved):

packages/cms/schemas/documents/fields/highlighted-articles.js
VWSCoronaDashboard1: As I stated above, I took this convention from the examples that Sanity is giving

VWSCoronaDashboard10 (approved):

PR #1785, opened 2021-02-03
Add difference metric in pages

VWSCoronaDashboard14:
Add difference metric in pages that were already in the sidebar but not on the page.


packages/app/src/pages/veiligheidsregio/[code]/thuiswonende-ouderen.tsx
```diff
@@ -29,6 +29,7 @@ const ElderlyAtHomeRegionalPage: FCWithLayout<typeof getStaticProps> = (
 ) => {
   const { safetyRegionName, data } = props;
   const elderlyAtHomeData = data.elderly_at_home;
+  const difference = data.difference;
```
Comment (VWSCoronaDashboard8) [inline packages/app/src/pages/veiligheidsregio/[code]/thuiswonende-ouderen.tsx:32]: nit; These are a bit pointless, but you I see you just followed the existing code. I'd say using `data.elderly_at_home` directly is even more clear than having a variable `elderlyAtHomeData`.

VWSCoronaDashboard4 (approved):

PR #1793, opened 2021-02-03
Improve the download-speed of Sanity assets

VWSCoronaDashboard10:
## Summary
The way we download Sanity's assets has been refactored. Instead of downloading all raw images + resizing them locally, we now download resized images of sanity's CDN.

There's also support for incremental updates when you would like to update the local asset library, it will only download new assets.


packages/cms/scripts/download-assets.ts
```diff
@@ -0,0 +1,172 @@
...
+  ];
+
+  await deleteFiles(obsoleteFiles);
+
+  console.log(
+    `\n🥳 Successfully downloaded ${missingAssets.length} assets and removed ${obsoleteFiles.length} obsolete files!!1\n`
+  );
+})().catch((err) => console.log(err));
+
+interface LocalAsset {
```
VWSCoronaDashboard4, line 82: There is an interface here, in the middle of the file; and some at the bottom. Perhaps group them, or move them to their own file?

packages/cms/scripts/download-assets.ts
```diff
@@ -0,0 +1,172 @@
...
+  url: string;
+  filename: string;
+  directory: string;
+}
+
+async function downloadAssets(assets: LocalAsset[]) {
```
VWSCoronaDashboard4, line 88: :100:

packages/cms/scripts/download-assets.ts
```diff
@@ -0,0 +1,172 @@
...
+  if (dataset !== 'development' && dataset !== 'production') {
+    throw new Error(`Unknown dataset ${dataset}`);
+  }
+
+  const sanityAssets = await getSanityAssets({
+    projectId: SANITY_PROJECT_ID,
+    dataset,
+  });
+
+  const localAssets: LocalAsset[] = [
```
VWSCoronaDashboard8, line 39: I think this name is confusing, because by this time it doesn't represent the local assets. You're getting those in `getFilteredAssetFilePaths`. This seems to represent the assets as declared by the Sanity dataset, and the logic is about getting the local assets synchronized with those.

Maybe just call it `assets` or something? Or `sanityAssets`, but then replace that name with `sanityAssetsInfo` or something (since it's a different structure from what you're using in the logic)

packages/cms/scripts/download-assets.ts
```diff
@@ -0,0 +1,172 @@
...
+    ...sanityAssets.files.map((file) => ({
+      url: file.url,
+      filename: `${file.assetId}.${file.extension}`,
+      directory: `files`,
+    })),
+
+    ...sanityAssets.images.flatMap((image) =>
+      image.extension === 'svg'
```
VWSCoronaDashboard4, line 47: Thanks for clearing the unneeded SVG-resize up!

packages/cms/package.json
```diff
@@ -12,9 +12,7 @@
     "test:ci": "yarn run test",
     "compile": "echo 'Placeholder to support yarn compile in repo root, until we use TS'",
     "lint": "echo '@TODO add eslint'",
-    "assets": "run-s assets:download assets:image-resize",
-    "assets:download": "bash scripts/assets-download.sh",
-    "assets:image-resize": "ts-node scripts/images-resize.ts"
+    "download-assets": "ts-node scripts/download-assets.ts"
```
VWSCoronaDashboard4, line 15: I think it can still be named `assets`; since you are also removing assets, not only downloading them.

VWSCoronaDashboard4 (approved):

packages/cms/scripts/download-assets.ts
```diff
@@ -0,0 +1,172 @@
...
+            directory: `images`,
+          }))
+    ),
+  ];
+
+  const missingAssets = localAssets.filter(
+    (x) => !fs.existsSync(`${TARGET_DIR}/${x.directory}/${x.filename}`)
+  );
+
+  await downloadAssets(missingAssets);
```
VWSCoronaDashboard8, line 65: nit; I think the interface would be clearer if you passed the download target folder to the function, because at this point the reader has to wonder where these things are going to end up.

packages/cms/scripts/download-assets.ts
```diff
@@ -0,0 +1,172 @@
...
+}
+
+async function deleteFiles(files: string[]) {
+  files.forEach((file, i, list) => {
+    fs.unlinkSync(file);
+    console.log(`[${i + 1}/${list.length}]: 🗑  ${file}`);
+  });
+}
+
+function getFilteredAssetFilePaths(
```
VWSCoronaDashboard8, line 108:
```suggestion
function getFilteredAssetsFilePaths(
```
Since you're retrieving a list

packages/cms/scripts/download-assets.ts
```diff
@@ -0,0 +1,172 @@
...
+  directory: string,
+  comparator: (filename: string) => boolean
+) {
+  return fs
+    .readdirSync(directory)
+    .map((filename) =>
+      comparator(filename) ? `${directory}/${filename}` : undefined
+    )
+    .filter((x) => x !== undefined) as string[];
```
VWSCoronaDashboard8, line 117:
```suggestion
    .filter(isDefined);
```
I suspect you don't need the type cast if you use isDefined, and we use it in other places already of course

packages/cms/scripts/download-assets.ts
```diff
@@ -0,0 +1,172 @@
...
+}: {
+  projectId: string;
+  dataset: string;
+}) {
+  const url = `https://${projectId}.api.sanity.io/v1/data/export/${dataset}`;
+
+  const request = await fetch(url);
+  const response = await request.text();
+
+  // ndjson is new-line delimited json, therefore we'll split on \n characters
```
VWSCoronaDashboard8, line 132:
```suggestion
  // Sanity returns ndjson (new-line delimited json), therefore we'll split on \n characters
```

VWSCoronaDashboard1 (approved):

packages/cms/scripts/download-assets.ts
```diff
@@ -0,0 +1,172 @@
...
+  const lines = response.split(/\n/g).filter((x) => x.length > 0);
+
+  const assets = lines
+    .map((x) => JSON.parse(x))
```
VWSCoronaDashboard8, line 136: I don't quite understand why you are splitting the json and then parsing each line separately. If possible it seems more efficient to parse the full response at once and then work with the array of items.

VWSCoronaDashboard8 (approved):

packages/cms/package.json
VWSCoronaDashboard8: Maybe assets:sync? or sync-assets?

VWSCoronaDashboard10: Yes I'll go for `sync-assets`

packages/cms/scripts/download-assets.ts
```diff
@@ -0,0 +1,172 @@
+import { imageResizeTargets } from '@corona-dashboard/common/src/config';
```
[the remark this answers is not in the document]
VWSCoronaDashboard10: Good catch, vscode auto import didn't go well

packages/cms/scripts/download-assets.ts
VWSCoronaDashboard10: 😅 makes more sense indeed

packages/cms/scripts/download-assets.ts
VWSCoronaDashboard10: It isn't an array of objects :-) Every line is a separate json record.

more info about ndjson: http://ndjson.org/

packages/cms/scripts/download-assets.ts
VWSCoronaDashboard8: Never mind this comment. I think your original name already indicates a list.

packages/cms/scripts/download-assets.ts
VWSCoronaDashboard8: Ah I see! Maybe add that in the comment? Something like "ndjson is a streaming format in which each line is its own document"

VWSCoronaDashboard14 (approved):

VWSCoronaDashboard14 (approved):

VWSCoronaDashboard4 (approved):

VWSCoronaDashboard1 (approved):

packages/cms/scripts/sync-assets.ts
```diff
@@ -0,0 +1,150 @@
...
+
+async function cacheAssets(assets: LocalAsset[], cacheDirectory: string) {
+  console.log(`Ensuring ${assets.length} assets are part of local cache...`);
+
+  let count = 0;
+  for (const urls of chunk(assets, 20)) {
+    const promisedDownloads = urls.map(async ({ url, filename }) => {
+      if (await fs.pathExists(`${cacheDirectory}/${filename}`)) return;
+      await download(url, cacheDirectory, { filename });
+      if (count === 0) console.log('');
```
VWSCoronaDashboard8, line 56: What is this for?

VWSCoronaDashboard8 (approved):

packages/cms/scripts/sync-assets.ts
VWSCoronaDashboard10: it was to improve readability of console output but I've reverted it since readability of code is better in this case :-)

PR #1807, opened 2021-02-04
Strip unnecessary NPM properties from package jsons

VWSCoronaDashboard8:
## Summary

* Set all packages to private, because we don't intend to publish them to NPM
* Remove version info since it doesn't have a real meaning without publishing
* Remove different licenses and only keep one declaration in the root package

The versions were often overlooked when making a new release, and updating them is cumbersome and error prone. I believe we better omit them completely.

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -1,6 +1,5 @@
 {
   "name": "@corona-dashboard/app",
-  "version": "2.14.3",
   "private": true,
   "license": "EUROPEAN UNION PUBLIC LICENCE v. 1.2",
   "dependencies": {
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@corona-dashboard/app",
   "private": true,
+  "version": "0.0.0",
   "license": "EUROPEAN UNION PUBLIC LICENCE v. 1.2",
   "dependencies": {
     "@formatjs/intl-datetimeformat": "^2.6.7",
--- a/packages/cms/package.json
+++ b/packages/cms/package.json
@@ -1,9 +1,6 @@
 {
   "name": "@corona-dashboard/cms",
   "private": true,
-  "version": "2.9.0",
-  "description": "",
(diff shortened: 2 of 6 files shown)


packages/app/package.json
```diff
@@ -1,6 +1,5 @@
 {
   "name": "@corona-dashboard/app",
-  "version": "2.14.3",
```
VWSCoronaDashboard4, line 3: We need to make sure this version is not read out, before we can remove it.

packages/cms/package.json
```diff
@@ -1,9 +1,6 @@
 {
   "name": "@corona-dashboard/cms",
   "private": true,
-  "version": "2.9.0",
-  "description": "",
-  "license": "UNLICENSED",
```
VWSCoronaDashboard4, line 6: Shouldn't the `EUROPEAN UNION PUBLIC LICENCE v. 1.2` be copied over to all packages?

VWSCoronaDashboard4 (changes requested):

packages/cms/package.json
VWSCoronaDashboard8: Since the packages are not published I think it doesn't have a value. We mention it in the root package.

packages/app/package.json
VWSCoronaDashboard8: I'm assuming it will turn up as undefined or an empty string. We'll find out soon enough 🤠 But I'll notify our chef pipeline.

VWSCoronaDashboard1: I don't think it's being extracted anywhere. The tag is set manually, so I am pretty sure this version is here just for decoration.

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard10:
Build is failing: https://github.com/minvws/nl-covid19-data-dashboard/pull/1807/checks?check_run_id=1829268978

VWSCoronaDashboard4 (approved):

PR #1809, opened 2021-02-04
Add redirect from old landelijk route

VWSCoronaDashboard4:
## Summary

Add redirect from old landelijk route

### DIFF
--- a/packages/app/next.config.js
+++ b/packages/app/next.config.js
@@ -74,6 +74,16 @@ const nextConfig = {
 
     return config;
   },
+
+  async redirects() {
+    return [
+      {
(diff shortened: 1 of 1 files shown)


packages/app/next.config.js
```diff
@@ -74,6 +74,16 @@ const nextConfig = {
...
+
+  async redirects() {
+    return [
+      {
+        source: '/landelijk',
+        destination: '/landelijk/vaccinaties',
+        permanent: true,
+      },
+    ];
+  },
```
VWSCoronaDashboard10, line 86: I guess this won't work with our static page?

PR #1815, opened 2021-02-04
Vaccine support chart

VWSCoronaDashboard8:
## Summary

This PR adds the vaccine support chart, but makes some fundamental changes to the line chart component which affects all line charts.

* The gray hover marker span is now data driven, so that depending on the timeframe setting (last week, last 5 weeks etc) the span changes to match the bandwidth of the sample. The effect can be seen easily on the hospital admissions chart.
* This span is visible on all charts, also the day charts that did not previously have it.
* For all charts that render data span values (date_start/end timestamps) the marker is now placed in the middle. So when rendering data for a week, the point in the chart falls halfway in the week. This was a demand for the vaccine support chart, so I assume it makes sense to have that behavior in all charts. This mean that for example for the sewer water chart, where we render 3-day averages, the point will fall on the middle day.


packages/app/src/components-styled/line-chart/components/chart-axes.tsx
```diff
@@ -129,6 +131,7 @@ export const ChartAxes = memo(function ChartAxes({
             props: {
               scale: yScale,
               width: bounds.width,
+              // width: width,
```
VWSCoronaDashboard10, line 134: comment

packages/app/src/components-styled/line-chart/line-chart.tsx
```diff
@@ -352,36 +373,27 @@ function formatDefaultTooltip<T extends Value>(
...
-            'short'
-          )}{' '}
-          -{' '}
-          {formatDateFromSeconds(
-            ((value as unknown) as WeeklyValue).date_end_unix,
-            'short'
-          )}
-          :
-        </Text>{' '}
+          {`${formatDateFromSeconds(value.date_start_unix, 'short')} -
```
VWSCoronaDashboard10, line 394: Not sure formatting goes well here, you might want to render it as follows just to be safe:
```suggestion
          {`${formatDateFromSeconds(value.date_start_unix, 'short')}{' - '}
```
(assuming there should be a whitespace _after_ the dash)

packages/app/src/components-styled/stacked-chart/stacked-chart.tsx
```diff
@@ -332,36 +332,39 @@ export function StackedChart<T extends Value>(props: StackedChartProps<T>) {
...
+    ) {
+      const isLeave = event.type === 'mouseleave';
+
+      if (isLeave) {
+        tooltipTimeout = window.setTimeout(() => {
+          hideTooltip();
+        }, 300);
+        hoverTimeout = window.setTimeout(() => {
+          setHoveredIndex(NO_HOVER_INDEX);
+        }, 300);
```
VWSCoronaDashboard10, line 349: it's safer to validate the component is still mounted when the timeout ends (especially with these quite long durations), otherwise we'll have runtime errors due to mutating state of an unmounted component.

You could create this hook:
```tsx
export function useIsMountedRef() {
  const isMountedRef = useRef(false);

  useEffect(() => {
    isMountedRef.current = true;
    return () => {
      isMountedRef.current = false;
    };
  }, []);

  return isMountedRef;
}
```

And use it like this:
```suggestion
        tooltipTimeout = window.setTimeout(() => {
          if (isMountedRef.current) hideTooltip();
        }, 300);
        hoverTimeout = window.setTimeout(() => {
          if (isMountedRef.current) setHoveredIndex(NO_HOVER_INDEX);
        }, 300);
```

packages/app/src/pages/landelijk/vaccinaties.tsx
```diff
@@ -239,6 +245,51 @@ const VaccinationPage: FCWithLayout<typeof getStaticProps> = ({
...
         </ChartTile>
 
+        <ChartTile
+          title={text.grafiek_draagvlak.titel}
+          description={text.grafiek_draagvlak.omschrijving}
+          ariaDescription={
+            siteText.accessibility.grafieken.vaccinatie_draagvlak
+          }
+          metadata={{
+            date: 1611593522,
```
VWSCoronaDashboard10, line 255: what's this hardcoded date doing here?

packages/app/src/pages/landelijk/vaccinaties.tsx
```diff
@@ -239,6 +245,51 @@ const VaccinationPage: FCWithLayout<typeof getStaticProps> = ({
...
+                ariaLabelledBy="chart_vaccine_support"
+                values={data.vaccine_support.values}
+                linesConfig={[{ metricProperty: 'percentage_in_favor' }]}
+                formatTooltip={(values) => {
+                  const value = values[0];
+
+                  return (
+                    <Text m={0}>
+                      <span style={{ fontWeight: 'bold' }}>
+                        {`${formatDateFromSeconds(value.date_start_unix)} -
```
VWSCoronaDashboard10, line 273: same comment about formatting (`{' - '}`)

VWSCoronaDashboard10 (approved):
Spotted one issue with timeouts which could create runtime errors, but overall looking good.

packages/app/src/components-styled/line-chart/components/marker.tsx
```diff
@@ -142,6 +136,20 @@ export function Marker<T extends Value>(props: MarkerProps<T>) {
...
+      <DateSpanMarker
+        style={{
+          width: dateSpanWidth,
+          left: data[0].x,
+        }}
+      >
+        {data.map((d, index) => (
+          <Point
+            indicatorColor={d.color ?? colors.data.primary}
+            style={{ top: d.y - index * 18 }}
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: that's the height of the marker point

packages/app/src/components-styled/line-chart/line-chart.tsx
VWSCoronaDashboard8: I prefer formatting things in a single string. I don't quite see a point in breaking things up. To me a single string is more readable.

packages/app/src/pages/landelijk/vaccinaties.tsx
VWSCoronaDashboard8: These have been part of the vaccinatie pagina from the start, but you're right, now it should be connected to the actual data. Good catch 👍

VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/line-chart/line-chart.tsx
VWSCoronaDashboard10: oh wait, I didn't see there were nested string interpolations. In that case I guess rendering will be fine, but the syntax is IMHO not very readable..

VWSCoronaDashboard8: I think it would help if I first formatted the date strings.... I'll try that

packages/app/src/components-styled/line-chart/components/marker.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard8: I will give it a constant so that is clear

VWSCoronaDashboard1 (approved):

PR #1827, opened 2021-02-05
Move sanity source code to src directory

VWSCoronaDashboard10:
Move Sanity source code to a `src`-directory.


VWSCoronaDashboard4 (approved):

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard8:
@VWSCoronaDashboard10 I'm surprised to not see any changes to tsconfig. I think we should be able to simplify it by setting includes to "src" (like common package) and maybe also enable "~/" path resolve.

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard10:
> @VWSCoronaDashboard10 I'm surprised to not see any changes to tsconfig. I think we should be able to simplify it by setting includes to "src" (like common package) and maybe also enable "~/" path resolve.

Yep, I'll pick that up with another PR

PR #1840, opened 2021-02-05
Add vaccine delivery/administered and estimates schemas

VWSCoronaDashboard1:
## Summary

As the title states

## Motivation

These schemas are meant to feed this chart:
image

The four datapoints added to the national schema are indicated in the screenshot to show which part of the graph they'll feed.

## Detailed design

n/a

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added


packages/app/schema/nl/vaccine_administered.json
```diff
@@ -0,0 +1,64 @@
...
+      "type": "object",
+      "required": [
+        "pfizer",
+        "moderna",
+        "astra_zeneca",
+        "cure_vac",
+        "janssen",
+        "sanofi",
+        "date_of_insertion_unix",
+        "date_unix"
```
VWSCoronaDashboard8, line 14: Just checking; so this data will be reported daily?

packages/common/src/types/data.ts
```diff
@@ -127,6 +127,10 @@ export interface National {
   deceased_cbs: NationalDeceasedCbs;
   elderly_at_home: NationalElderlyAtHome;
   vaccine_delivery: NlVaccineDelivery;
+  vaccine_availability?: NlVaccineAvailability;
+  vaccine_availability_estimate?: NlVaccineAvailabilityEstimate;
+  vaccine_administered?: NlVaccineAdministered;
+  vaccine_administered_estimate?: NlVaccineAdministeredEstimate;
```
VWSCoronaDashboard8, line 133: Looks like you forgot to add these in the "required" list in __index

packages/common/src/types/data.ts
```diff
@@ -424,6 +428,66 @@ export interface NlVaccineDeliveryValue {
...
+  last_value: NlVaccineAvailabilityValue;
+}
+export interface NlVaccineAvailabilityValue {
+  pfizer: number;
+  moderna: number;
+  astra_zeneca: number;
+  cure_vac: number;
+  janssen: number;
+  sanofi: number;
+  total?: number;
```
VWSCoronaDashboard8, line 442: In the previous chart we would calculate the total. It was a different kind of total because it was per vaccine company so it spanned multiple time series items. I guess having the total supplied by backend is a good things here. But I wonder then why is it considered optional? I suspect you forgot to add it to the required list...

packages/common/src/types/data.ts
```diff
@@ -424,6 +428,66 @@ export interface NlVaccineDeliveryValue {
...
+  last_value: NlVaccineAvailabilityEstimateValue;
+}
+export interface NlVaccineAvailabilityEstimateValue {
+  pfizer: number;
+  moderna: number;
+  astra_zeneca: number;
+  cure_vac: number;
+  janssen: number;
+  sanofi: number;
+  total?: number;
```
VWSCoronaDashboard8, line 457: Same for this one

packages/common/src/types/data.ts
```diff
@@ -424,6 +428,66 @@ export interface NlVaccineDeliveryValue {
...
+  last_value: NlVaccineAdministeredValue;
+}
+export interface NlVaccineAdministeredValue {
+  pfizer: number;
+  moderna: number;
+  astra_zeneca: number;
+  cure_vac: number;
+  janssen: number;
+  sanofi: number;
+  total?: number;
```
VWSCoronaDashboard8, line 472: etc

packages/common/src/types/data.ts
```diff
@@ -424,6 +428,66 @@ export interface NlVaccineDeliveryValue {
...
+  last_value: NlVaccineAdministeredEstimateValue;
+}
+export interface NlVaccineAdministeredEstimateValue {
+  pfizer: number;
+  moderna: number;
+  astra_zeneca: number;
+  cure_vac: number;
+  janssen: number;
+  sanofi: number;
+  total?: number;
```
VWSCoronaDashboard8, line 487: etc

VWSCoronaDashboard8 (changes requested):

packages/app/schema/nl/vaccine_administered.json
VWSCoronaDashboard1: that's a good question actually, I forgot to check. Thanks!

packages/common/src/types/data.ts
VWSCoronaDashboard1: Indeed, this is definitely Friday afternoon work... Sorry...

packages/app/schema/nl/vaccine_administered.json
VWSCoronaDashboard1: Yea, just check the example, its definitely a week span... Fixed!

packages/app/schema/nl/__index.json
```diff
@@ -27,8 +27,12 @@
     "tested_ggd_average",
     "tested_overall",
     "tested_per_age_group",
+    "vaccine_delivery",
     "vaccine_support",
-    "vaccine_delivery"
+    "vaccine_availability",
+    "vaccine_availability_estimate",
```
VWSCoronaDashboard8, line 33: I avoided the term "availability" before because it can be interpreted in different ways. Is it current stock, (estimated) delivery or a combination of both? If it's a combination of both, I don't have a better name for it. But if it's one of the other two I would not use this name.

VWSCoronaDashboard8 (approved):

packages/app/schema/nl/__index.json
VWSCoronaDashboard1: That's a good question, I'll try to find out which one it is because I don't have an answer for that currently.

VWSCoronaDashboard1: After some digging we found out that this is indeed NOT the availability, but the *delivery* instead.
So, they have been renamed to vaccine_delivery and vaccine_delivery_estimate

PR #1841, opened 2021-02-05
Adds schemas for escalation levels

VWSCoronaDashboard4:
## Summary

Adds schemas for escalation levels


packages/app/schema/vr/__index.json
```diff
@@ -76,6 +76,15 @@
...
       "$ref": "elderly_at_home.json"
+    },
+    "escalation_level": {
+      "$ref": "escalation_level.json"
+    },
+    "tested_escalation_level": {
+      "$ref": "tested_escalation_level.json"
+    },
+    "hospital_escalation_level": {
+      "$ref": "hospital_escalation_level.json"
```
VWSCoronaDashboard10, line 87: These two KPI's will be rendered on the 'positief-geteste-mensen'- and 'ziekenhuis-opnames' pages, so what would you think of defining them in the `tested_overall` and `hospital_nice` schema's?

The fields could be named something like `infected_per_100k_moving_average` and `infected_per_100k_moving_average`.

The new escalation level page can simply grab these schema's to render the KPI's on that page.

VWSCoronaDashboard4: That does not combine well. Those are daily updates, these are weekly.

VWSCoronaDashboard4: The `_100k` and `_unix` are always at the end, since they tell something about the format; see the schemas/readme file.

VWSCoronaDashboard10: Aren't they also daily updated, but with a moving weekly average?. And it could of course be named `infected_moving_average_per_100k` :-)

VWSCoronaDashboard8: I think the fact that this data is used to display an escalation level is not neccesarily interesting from a data definition standpoint. I would prefer to describe _what_ the data is and not _how_ it is used, because that might change in the future.

Also we need to think carefully if we want to keep the shape of this data compatible with the value/values/timestamps of what we already have for normal timeseries or if this is something fundamentally different. For example

```
"date_unix",
        "escalation_level",
        "valid_from_unix",
        "date_of_insertion_unix"
```
This `valid_from_unix` is not a timestamp we have in other time series and I wonder what it means in relation to `date_unix`. Also of escalation-level data is different from normal time series (for example we will never display older values) then it might be warranted to diverge from the value/values format.

Whatever we choose, I think it needs to be a conscious decision. I suggest we'll have a chat about this with devs.

VWSCoronaDashboard8 (changes requested):

VWSCoronaDashboard1 (approved):

packages/app/schema/vr/__index.json
```diff
@@ -19,7 +19,10 @@
     "deceased_cbs",
     "elderly_at_home",
     "disability_care",
-    "behavior"
+    "behavior",
+    "escalation_level",
+    "tested_weekly",
```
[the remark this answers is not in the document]
VWSCoronaDashboard8: Also, what type of "tested" is this? Might be worth using `tested_overall` or `tested_ggd` as a prefix, depending on the context. So then it could be `tested_overall_average` if it is about all reported tests including ggd.

packages/app/schema/vr/escalation_level.json
```diff
@@ -0,0 +1,41 @@
...
+    }
+  },
+  "definitions": {
+    "value": {
+      "title": "vr_escalation_level_value",
+      "type": "object",
+      "additionalProperties": false,
+      "required": [
+        "date_unix",
+        "escalation_level",
```
VWSCoronaDashboard8: The context is already "escalation_level" so maybe just call this "level" to avoid confusion. I know this shape is the same as vr_collection.escalation_levels so I get why you might want to use that. But then again the metric property names don't fully match either.

packages/app/schema/vr/hospital_weekly.json
```diff
@@ -0,0 +1,37 @@
...
+      "type": "object",
+      "additionalProperties": false,
+      "required": [
+        "admissions_per_1m",
+        "date_start_unix",
+        "date_end_unix",
+        "date_of_insertion_unix"
+      ],
+      "properties": {
+        "admissions_per_1m": { "type": "integer" },
```
VWSCoronaDashboard8, line 30: Are you sure this will not be a decimal value? Per 1M sounds like it might be a float

VWSCoronaDashboard8 (changes requested):

packages/app/schema/vr/escalation_level.json
```diff
@@ -0,0 +1,41 @@
...
+        "valid_from_unix",
+        "date_of_insertion_unix"
+      ],
+      "properties": {
+        "date_unix": { "type": "integer" },
+        "escalation_level": {
+          "type": "number",
+          "minimum": 1,
+          "maximum": 4
```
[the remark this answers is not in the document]
VWSCoronaDashboard4: This is not related to the metrics; this is just the escalation level (copied from the choropleth json to the VR json).

packages/app/schema/vr/hospital_weekly.json
VWSCoronaDashboard4: Yes, this should be an integer.

VWSCoronaDashboard8 (approved):

PR #1844, opened 2021-02-05
Convert cms package to typescript

VWSCoronaDashboard10:
## Summary

The cms package has been converted to typescript. A couple of files required a bit more work in order to fix all type errors. Some issues have been fixed using the `any` type, mainly in our schema definitions, but I guess we'll find a proper way to type these in the future.


packages/cms/src/desk-structure.ts
```diff
@@ -1,4 +1,6 @@
-import StructureBuilder from '@sanity/desk-tool/structure-builder';
+import { StructureBuilder } from '@sanity/structure';
+import { ReactNode } from 'react';
+// import StructureBuilder from '@sanity/desk-tool/structure-builder';
```
VWSCoronaDashboard4, line 3: commented out code

packages/cms/src/previews/assemble-preview-url.ts
```diff
@@ -0,0 +1,16 @@
...
+  options: { previewURL: string };
+}) {
+  const { previewURL } = options;
+
+  if (!previewURL) {
+    console.warn('Missing previewURL', { previewURL });
+    return '';
+  }
+
+  return `${previewURL}`;
```
VWSCoronaDashboard4, line 15: no need to do this templating, especially since previewURL is a string?

VWSCoronaDashboard4 (approved):
Only minor comments :-)

packages/cms/src/previews/resolve-production-url.ts
```diff
@@ -1,4 +1,9 @@
 const previewSecret = process.env.SANITY_STUDIO_PREVIEW_SECRET;
-export default function resolveProductionUrl(document) {
+
+interface IncompleteDocument {
+  slug?: { current: string };
+}
+
+export default function resolveProductionUrl(document: IncompleteDocument) {
```
VWSCoronaDashboard8, line 7: nit; I don't think this qualifies as "resolving" since its only formatting a string without using an algorithm. Maybe call it `format/create/makeProductionUrl`

packages/cms/tsconfig.json
```diff
@@ -2,10 +2,10 @@
   "extends": "../../tsconfig.base.json",
   "include": [
     "./node_modules/@sanity/base/types/**/*.ts",
```
VWSCoronaDashboard8, line 4: This is very odd. Have you tried removing it?

packages/cms/src/components/icons/icon.tsx
```diff
@@ -1,48 +1,60 @@
...
 
-import PatchEvent, { set, unset } from 'part:@sanity/form-builder/patch-event';
-
-import { restrictionIcons } from './icons';
-
-const createPatchFrom = (value) =>
-  PatchEvent.from(value === '' ? unset() : set(String(value)));
-
-export function Icon(props) {
+function createPatchFrom(value: string) {
```
VWSCoronaDashboard8, line 19: Not your fault, but this function and its name seem odd. If I understand correctly something like `createPatchEventFromValue` would be clearer.

packages/cms/src/components/icons/icon.tsx
```diff
@@ -1,48 +1,60 @@
...
+    title: string;
+    description: string;
+  };
+  value: RestrictionIcon;
+  onChange: (event: PatchEvent) => void;
+}
+export function Icon(props: IconProps) {
   const [isOpen, setIsOpen] = useState(false);
   const onClose = useCallback(() => setIsOpen(false), []);
   const onOpen = useCallback(() => setIsOpen(true), []);
```
VWSCoronaDashboard8, line 33: Also not your fault, but these should be called handleClose and handleOpen to make it in-line with the handle/on naming pattern we're using.

VWSCoronaDashboard8 (approved):

packages/cms/src/desk-structure.ts
VWSCoronaDashboard10: I've mainly renamed files to `*.ts` and tried to avoid changing the content of files.. I guess we can clean it up with another PR.

packages/cms/src/previews/assemble-preview-url.ts
VWSCoronaDashboard10: Yeah probably, this code is coming from @VWSCoronaDashboard so he might be able to verify it is indeed _always_ a string. Like I said earlier: I haven't really changed code, I've only done the necessary changes for TS :-)

packages/cms/tsconfig.json
VWSCoronaDashboard10: Good one, it works fine without this include.

PR #1851, opened 2021-02-08
Add a switch for a highlight article or custom message from sanity

VWSCoronaDashboard14:
On the topical page, there is a toggle that lets the user choose between a custom message or a highlighted article.
Also added a new `custom-content-teaser` component. Initially, I wanted to make the article-teaser a bit more flexible to accept multiple properties, but create a separation that is easier to follow and makes more sense.


VWSCoronaDashboard1 (changes requested):
I would simplify this schema by a lot. I would remove the 'isArticle' logic and just let the content manager add an image, text, and a link to a specific article if an article link is needed.

VWSCoronaDashboard1 (approved):

PR #1869, opened 2021-02-09
Add schema and added data for vaccine KPI tiles

VWSCoronaDashboard14:
Add schema for vaccine KPI tiles
And integrating it with the data from the backend 
Cleaned the locale vaccine key
Added the data on the sidebar and the topical homepage


packages/app/schema/nl/__index.json
```diff
@@ -28,7 +28,8 @@
     "tested_overall",
     "tested_per_age_group",
     "vaccine_support",
-    "vaccine_delivery"
+    "vaccine_delivery",
+    "vaccine_dose"
```
VWSCoronaDashboard8, line 32: This name is not clear to me. vaccine_dose could also be interpreted as vaccine_delivery since that is also about doses. And we already have schema's for vaccine_administered in this branch so I think this PR conflicts with that work.

https://github.com/minvws/nl-covid19-data-dashboard/pull/1840/files

packages/app/schema/nl/vaccine_dose.json
```diff
@@ -0,0 +1,57 @@
...
+        },
+        "calculated_ggd": {
+          "type": "number"
+        },
+        "calculated_hospitals": {
+          "type": "number"
+        },
+        "calculated_care_institutions": {
+          "type": "number"
+        },
```
VWSCoronaDashboard8, line 28: calculated is not a clear term to me. Is it an estimate? 

Then I would suggest something like `estimate_total`, `estimate_ggd` etc.

packages/app/schema/nl/vaccine_dose.json
```diff
@@ -0,0 +1,57 @@
...
+        "reported": {
```
VWSCoronaDashboard8, line 29: If this is a total then I suggest to add the _total suffix for clarity.

VWSCoronaDashboard8 (changes requested):

packages/app/schema/nl/vaccine_administered_per_authority.json
```diff
@@ -0,0 +1,57 @@
...
+        },
+        "reported_total": {
+          "type": "number"
+        },
+        "reported_ggd_ghor": {
+          "type": "number"
+        },
+        "reported_lnaz": {
+          "type": "number"
+        }
```
VWSCoronaDashboard8, line 37: I'm missing date timestamps here. The format is values/last_value so even though they are displayed in KPI tiles, I would expect data to be timestamped. Use date_start/end if data is spanning multiple days.

VWSCoronaDashboard8 (changes requested):

packages/app/schema/nl/vaccine_administered_total.json
```diff
@@ -0,0 +1,45 @@
...
+        "date_of_insertion_unix"
+      ],
+      "additionalProperties": false,
+      "properties": {
+        "estimated_total": {
+          "type": "number"
+        },
+        "reported_total": {
+          "type": "number"
+        },
```
VWSCoronaDashboard8, line 19: The context is already `_administered_total` so the total suffix is redundant here.

packages/app/schema/nl/__index.json
```diff
@@ -28,7 +28,13 @@
...
     "tested_per_age_group",
     "vaccine_support",
-    "vaccine_delivery"
+    "vaccine_delivery",
+    "vaccine_administered_care_institutions",
+    "vaccine_administered_ggd_ghor",
+    "vaccine_administered_ggd",
+    "vaccine_administered_hospitals",
+    "vaccine_administered_lnaz",
+    "vaccine_administered_total"
```
VWSCoronaDashboard8, line 37: Ooof I didn't realize the split approach was going to add this many schema's  👀  Now I feel things would have been better grouped by reporting source with possible null values to fill blanks. But whatever, I think we probably have more important things to focus on.

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard8 (approved):

packages/app/schema/nl/__index.json
```diff
@@ -28,7 +28,14 @@
     "tested_overall",
     "tested_per_age_group",
     "vaccine_support",
-    "vaccine_delivery"
+    "vaccine_delivery",
+    "vaccine_administered_care_institutions",
+    "vaccine_administered_general_practitioner",
```
0x80, line 33: We call this doctor for simplicity in other schema's. It's technically not the right word, but most people still refer to it by "doctor". And since you made the others plural I'd suggest using `_doctors`

packages/app/schema/nl/vaccine_administered_general_practitioner.json
```diff
@@ -0,0 +1,41 @@
+{
```
0x80, line 1: Same for these files / names of course.

0x80 (approved):

packages/app/src/domain/topical/topical-vaccine-tile.tsx
```diff
@@ -8,17 +8,19 @@ import { Heading, Text } from '~/components-styled/typography';
...
-    ),
-    expectedDelivery: parseFloat(
-      siteText.vaccinaties.data.kpi_expected_delivery.value
-    ),
-  };
+  // const data = {
+  //   expectedDelivery: parseFloat(
+  //     siteText.vaccinaties.data.kpi_expected_delivery.value
+  //   ),
+  // };
```
VWSCoronaDashboard8, line 23: comment

packages/app/src/domain/topical/topical-vaccine-tile.tsx
```diff
@@ -61,13 +63,13 @@ export function TopicalVaccineTile() {
...
 
-      <Text mt={0}>
+      {/* <Text mt={0}>
         {replaceComponentsInText(text.supply, {
           expectedDelivery: (
             <strong>{formatNumber(data.expectedDelivery)}</strong>
           ),
         })}
-      </Text>
+      </Text> */}
```
VWSCoronaDashboard8, line 72: I think this can go?

packages/app/src/pages/index.tsx
```diff
@@ -146,7 +146,11 @@ const Home: FCWithLayout<typeof getStaticProps> = (props) => {
                 href="/landelijk/ziekenhuis-opnames"
               />
 
-              <TopicalVaccineTile />
+              <TopicalVaccineTile
+                estimatedAmount={
+                  data.vaccine_administered_total.last_value.estimated
+                }
+              />
```
VWSCoronaDashboard8, line 153: nit; I think any number is usually about an amount, so it doesn't add much value to have that in the name.

packages/app/src/pages/landelijk/vaccinaties.tsx
```diff
@@ -92,122 +92,119 @@ const VaccinationPage: FCWithLayout<typeof getStaticProps> = ({
...
-                              >
-                                {item.report_date}
-                              </InlineText>
-                            </Text>
-                          )}
-                        </Fragment>
-                      )
-                    )}
+                    <VaccineAdministeredItem
+                      amount={
```
VWSCoronaDashboard8, line 123: nit; Maybe call this `value` instead. It's a little more generic.

packages/app/src/pages/landelijk/vaccinaties.tsx
```diff
@@ -420,6 +415,37 @@ VaccinationPage.getLayout = getNationalLayout;
 
 export default VaccinationPage;
 
+interface VaccineAdministeredProps {
+  amount: number;
+  date: number;
+  text: string;
```
VWSCoronaDashboard8, line 421: Maybe you can be more specific about what type of text this is. Is it a description for example?

PR #1875, opened 2021-02-09
Update and fix IE11 bug in the marker line-graph

VWSCoronaDashboard14:
The dotted line on the marker caused a crash by IE11 on the  `height: calc(100% - ${topY})px` (also wasn't visible on anymore on the dev branch). And the date wasn't positioned correctly at the bottom of the graph when hovering.
Revert and changed some styles to position and show them correctly again.


VWSCoronaDashboard10 (approved):
Approved code-wise, but let's not merge this yet until @VWSCoronaDashboard8 has confirmed these changes won't break another edge-case.

VWSCoronaDashboard4 (approved):

VWSCoronaDashboard8:
@VWSCoronaDashboard14 I can't remember why I made these changes. I probably assumed things could be simplified and only tested in on the normal line charts in non-IE browsers. If things appear to be working still in the normal line charts, when I'm fine with this.

VWSCoronaDashboard8 (approved):

PR #1881, opened 2021-02-10
Upgrade visx dependencies and make it ie11 compatible

VWSCoronaDashboard10:
## Summary

By transpiling all d3-* packages we can now use the more recent visx packages.


packages/app/next.config.js
```diff
@@ -3,10 +3,8 @@ const LodashModuleReplacementPlugin = require('lodash-webpack-plugin');
 const sitemap = require('./generate-sitemap.js');
 
 const withTM = require('next-transpile-modules')([
-  '@visx/scale',
-  '@visx/event',
-  '@visx/tooltip',
-  '@corona-dashboard/common',
+  'internmap',
```
VWSCoronaDashboard8, line 6: For what part of the code do we need internmap?

VWSCoronaDashboard10: It's a dependency of `d3-array`. I'll add a comment.

packages/app/next.config.js
```diff
@@ -49,6 +49,31 @@ const nextConfig = {
...
     });
 
+    /**
+     * All d3-* packages needs to be transpiled to make them ie11 compatible.
+     * For some reason `next-transpile-modules` doesn't pick them up properly.
+     */
+    config.module.rules.push({
+      test: /\.js$/,
+      loader: defaultLoaders.babel,
+      include: /[\\/]node_modules[\\/](d3-.*)[\\/]/,
```
VWSCoronaDashboard4, line 59: What does this do with the file size? Would it be worth splitting certain chunks for IE/modern?

VWSCoronaDashboard4 (approved):

packages/app/next.config.js
VWSCoronaDashboard10: It might have some impact compared to the non-transpiled code, but I think the difference with the previous d3 versions, which were already transpiled, is minimal.

PR #1884, opened 2021-02-10
Add dynamic hidden accessibility text for metadata sources

VWSCoronaDashboard14:
Add accessibility text for screenreaders on the metadata links in tiles and content-headers.
It takes the title and adds before the link a hidden text that yields what the link contains.

The only drawback is that some titles are called "Verloop over tijd" which could be a little confusing.
I'm not sure the titles should be more descriptive or I need to take a second look and add for example the current page. Would love to know your opinion.

Tested it with Voiceover from Apple:
Screenshot 2021-02-10 at 13 25 47


VWSCoronaDashboard4 (approved):

packages/app/src/components-styled/chart-tile-container.tsx
```diff
@@ -5,9 +5,11 @@ import { Metadata, MetadataProps } from './metadata';
 export function ChartTileContainer({
   children,
   metadata,
+  title,
 }: {
   children: React.ReactNode;
   metadata?: MetadataProps;
+  title: string;
```
VWSCoronaDashboard10, line 12: wouldn't `subject` make more sense? We're giving it the title, but it's somehow the subject of what the link is about.

packages/app/src/locale/nl.json
```diff
@@ -1952,6 +1952,8 @@
     }
   },
   "accessibility": {
+    "link_source": "Bekijk de databron van \"{{title}}\" op",
+    "link_download": "Download de databron van \"{{title}}\" op",
```
VWSCoronaDashboard10, line 1956: For flexibility I think it's good to also make the source part of this string:

`"Bekijk de databron van \"{{subject}}\" op {{source}}"`

packages/app/src/components-styled/chart-tile-container.tsx
VWSCoronaDashboard8: I agree that in this context subject might be more fitting. Of even something specific like screenReaderSubject if all we're going to be using this data for is the extra feature for screen readers. If this is passed as title you would expect it to be rendered as some sort of title.

In the case of ChartTileContainer it is not so bad, because in the parent tile the text is still used as a title, but later on you are passing it as "title" to Metadata, and that is really confusing because metadata doesn't have the concept of a title.

packages/app/src/components-styled/chart-tile-container.tsx
```diff
@@ -16,7 +18,7 @@ export function ChartTileContainer({
       {metadata && (
         <>
           <Spacer m="auto" />
-          <Metadata {...metadata} />
+          <Metadata {...metadata} title={title} />
```
VWSCoronaDashboard8, line 21: ...so here we should really find a better name.

VWSCoronaDashboard8: Also it looks like you are just passing title to ChartTileContainer to then pass it on the Metadata, but the props for metadata are already passed on. So I think you should be able to set the "title" data directly on the metadata prop.

packages/app/src/components-styled/chart-tile.tsx
```diff
@@ -45,7 +45,7 @@ export function ChartTile({
   const ariaLabelledBy = slugify(title);
 
   return (
-    <ChartTileContainer metadata={metadata}>
+    <ChartTileContainer metadata={metadata} title={title}>
```
VWSCoronaDashboard8, line 48: ☝️ ...here is what I mean. If you inject the "title" (under a better name) in metadata, when it is passed on automatically to the internal metadata component.

packages/app/src/components-styled/chart-tile-container.tsx
VWSCoronaDashboard8: Maybe `accessibilitySubject` is also an option, since eventually the data ends up in `accessibilityText`

packages/app/src/components-styled/content-header/metadata.tsx
```diff
@@ -82,10 +97,11 @@ interface MetadataItemProps {
     href: string;
     text: string;
   }[];
+  accessibilityText: string;
```
VWSCoronaDashboard8, line 100: I wouldn't mind shortening with a11y, since accessibility is such an annoying word to write 😅

VWSCoronaDashboard8 (changes requested):

packages/app/src/locale/nl.json
VWSCoronaDashboard14: We are using the source to display it on the screen. If I would do this it would read twice the source, or I need to disable the displayed one by wrapping it and add a `aria-hidden="true" ` not a fan of that.

packages/app/src/components-styled/content-header/metadata.tsx
```diff
@@ -66,7 +66,7 @@ export function Metadata(props: MetadataProps) {
         label={text.source}
         accessibilityText={replaceVariablesInText(
           siteText.accessibility.link_source,
-          { title }
+          { accessibilitySubject, source: 'test' }
```
VWSCoronaDashboard8, line 69: nit;
```suggestion
          { subject: accessibilitySubject, source: 'test' }
```
Because within the context of lokalize I don't think it has value to have this variable name spelled out with accessibility. Subject is clear enough in that text.

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard7:
> The only drawback is that some titles are called "Verloop over tijd" which could be a little confusing. I'm not sure the titles should be more descriptive or I need to take a second look and add for example the current page. Would love to know your opinion.


The graphs should not be titled "Verloop over tijd" but the editorial team should make them more descriptive as far as I'm concerned. The title should stand on its own and not rely on other parts of the page to make sense. Also for accessibility reasons.

PR #1887, opened 2021-02-10
Last value validation

VWSCoronaDashboard8:
## Summary

For every time series in data, try to match last_value with the last value in the values array. As part of this work I have unified the code around timestamps and data sorting to all use the same common functions. Also the type guards and logic have been improved in some places.

The script can be run with `yarn validate-last-values` and optionally had the flags -e (fail early, on first errors) and -v for verbose output.

`sortTimeSeriesInDataInPlace` has been made context agnostic, to replace functions like `sortMunicipalTimeSeriesInDataInPlace`

I encountered an odd exception around `nl_deceased_rivm_per_age_group_value` because it doesn't have a time stamp. Even though we don't currently use it in the code, I think it makes sense to just give this data a timestamp, since it does originate from some place in time, and it avoids writing exceptions like that (although now I've improved the code to auto-detect timestamped series). I have added the timestamp but in a backwards-compatible way so we don't have to wait for backend to implement it before merging.


packages/app/schema/nl/deceased_rivm_per_age_group.json
```diff
@@ -15,8 +15,10 @@
   "definitions": {
     "value": {
       "title": "nl_deceased_rivm_per_age_group_value",
-      "additionalProperties": false,
+      "//1": "Set to false once BE added date_unix timestamp",
```
VWSCoronaDashboard10, line 18: I think you should be able to include comments with the `$comments` keyword: https://github.com/json-schema-org/json-schema-spec/pull/359

packages/cli/src/validate-last-values/index.ts
```diff
@@ -0,0 +1,157 @@
...
+  console.log('Fail early?', cli.flags.failEarly);
+
+  const files = getFilesWithTimeSeries(directory);
+
+  const allFailures: Failure[] = [];
+
+  for (const file of files) {
+    const data = readJsonFile(path.join(directory, file));
+
+    sortTimeSeriesInDataInPlace(data);
```
VWSCoronaDashboard10, line 68: I don't really like the mutation pattern used here, but that refactor should probably not be part of this PR.

packages/cli/src/validate-last-values/index.ts
```diff
@@ -0,0 +1,157 @@
...
+
+      const promisedOperations = timeSeriesDataArray.map((metricData) => {
+        const installationName = metricData.rwzi_awzi_name;
+
+        const success = validateLastValue(metricData);
+        return {
+          success,
+          metricName: `sewer_per_installation.${installationName}`,
+        };
+      });
```
VWSCoronaDashboard10, line 119: It doesn't look like the map is returning a promise

packages/cli/src/validate-last-values/logic.ts
```diff
@@ -0,0 +1,60 @@
...
+  isDateSeries,
+  isDateSpanSeries,
+  isTimeSeries,
+} from '@corona-dashboard/common';
+
+export function getTimeSeriesMetricNames(object: UnknownObject) {
+  return Object.keys(object).filter((key) => isTimeSeries(object[key]));
+}
+
+export function readJsonFile(filePath: string): UnknownObject {
```
VWSCoronaDashboard10, line 16: `UnknownObject` is a bit too narrow, parsed JSON could for example also be an array. Maybe add a generic type?

packages/common/src/data-sorting.ts
```diff
@@ -0,0 +1,121 @@
...
+  return Object.entries(data).reduce(
+    (acc, [propertyKey, propertyValue]) =>
+      isTimeSeries(propertyValue) ? [...acc, propertyKey as keyof T] : acc,
+    [] as (keyof T)[]
+  );
+}
+
+export function sortTimeSeriesValues(values: TimestampedValue[]) {
+  /**
+   * There are 3 ways in which time series data can be timestamped. We need to
```
VWSCoronaDashboard10, line 44: you're handling 2 instead of 3

packages/common/src/utils/assert.ts
```diff
@@ -0,0 +1,5 @@
+export function assert(condition: any, msg: string): asserts condition {
```
VWSCoronaDashboard10, line 1: You might want to add a TSDoc comment about the `condition` matching any falsy value (e.g. passing `0` would fail the assertion)

VWSCoronaDashboard10 (approved):

VWSCoronaDashboard4 (approved):
Neat refactor!

packages/cli/src/validate-last-values/index.ts
VWSCoronaDashboard8: I know you don't like mutating 😄 But yeah definitely not going to do that as part of this PR. I don't have a problem with it in this case. I imaging not mutating will be more hassle than I think is worth it. All data is only mutated/touched once before getting used.

packages/cli/src/validate-last-values/index.ts
VWSCoronaDashboard8: Good catch!

packages/cli/src/validate-last-values/logic.ts
VWSCoronaDashboard8: We don't read files with array as their root. Adding it in the types only makes us having to program around it. I'm not sure that it adds any value then?

packages/common/src/utils/assert.ts
VWSCoronaDashboard8: It is standard for 0 to fail an assertion AFAIK

VWSCoronaDashboard8: I remember having used a second assert specifically for undefined like `assertDefined` but I think it's easier to have just 1 function and write `assert(isDefined(condition),...)`

VWSCoronaDashboard8: I will add a comment about falsy values.

packages/cli/src/validate-last-values/logic.ts
VWSCoronaDashboard8: Alternatively I could rename the function readObjectFromJsonFile ;) @VWSCoronaDashboard10

VWSCoronaDashboard1 (approved):
I would have prefered to see this implemented as a custom keyword in the JSON schema validation. Akin to the `equalsRootProperty` keyword. Granted, it would have been a little harder to share the sorting function because the AJV code is still JS. But it might have been worth porting that logic to TS and then implementing this as a keyword in order to keep all of the validation in one place instead of disparate scripts.
But this works too, so I won't make too big of a fuss over it.

VWSCoronaDashboard8:
@VWSCoronaDashboard1 I didn't realize that was an option. I'm not familiar with the capabilities of AJV. I suspect it might have been tricky to validate sewer per installation that way.

VWSCoronaDashboard1:
> @VWSCoronaDashboard1 I didn't realize that was an option. I'm not familiar with the capabilities of AJV. I suspect it might have been tricky to validate sewer per installation that way.

It would have been a bit trickier, but from what I can quickly glean, it wouldn't be impossible. I didn't notice in time you were working on this, otherwise, I could have discussed it beforehand with you. Probably not worth the re-write since there's plenty of other things still on our list.

PR #1888, opened 2021-02-10
Add husky to root of repo

VWSCoronaDashboard:
## Summary

Husky's commit and pre-push hooks should be called from the root of the repo.

## Motivation

- Make sure the hooks run again
- In the future, run more hooks in other packages?


VWSCoronaDashboard8:
Added as tech dept to Trello

PR #1894, opened 2021-02-11
Remove images from bundle and serve them from public folder

VWSCoronaDashboard:
## Summary

Recently two images were added to the bundle by importing them and using them as React components. Seeing as these images are simply images without any interactivity or a11y information, I've moved them to the public folder and decided to load these as normal images.

This will reduce the bundle size by around 400KB raw.

## Questions

- TS does not understand that a styled(Box).attr({as: 'img'}) components supports the src attribute. Is there a better way to tell TS it's an img type than just making a new type with src: string?


packages/app/src/components-styled/image.tsx
```diff
@@ -2,7 +2,15 @@ import css from '@styled-system/css';
 import styled from 'styled-components/';
 import { Box } from '~/components-styled/base';
 
-export const Image = styled(Box).attrs({ as: 'img' })(
+type ImageProps = {
+  src: string;
+};
+
+export const Image = styled(Box).attrs({
```
VWSCoronaDashboard8, line 9: I think basing an image on the styling of a the generic Box/div is a bit odd since an image is much more than just a box layout. Maybe just start with styled.img and work with that.

VWSCoronaDashboard8 (approved):

packages/app/src/components-styled/image.tsx
VWSCoronaDashboard: That maybe makes it easier with TS as well yeah. I just took the component that was already available, but we can refactor this. I'll try it out.

VWSCoronaDashboard: Made this an img, which solved the TS issues. 👍

PR #1897, opened 2021-02-11
Accept responsive image srcSet configuration for Sanity images

VWSCoronaDashboard14:
Since the images only get used on article detail pages it might be better to cap them for now. There is no need to use bigger images on larger screens with a resolution of 1536, 2048. 

I've updated the source-set, but later when we maybe could use full-screen images we can always think of a better solution and revert this change.


VWSCoronaDashboard:
I think this is an improvement on what we have now, so approved from my end.

For the future, I think there's alternative approaches we can think about:

1. Providing srcSet as a prop.

When you want to extend or change the srcSet behavior for a specific image, we can open up the component to accept a srcSet prop. Then we should skip generating the sourceSet automatically in sanity.tsx.

2. Providing srcSet instructions through configuration. 

Maybe we can think of an optional  configuration object where we map the common sizes of resized images to expected widths of the image. Sanity.tsx takes this object and still loops over all the sizes and automatically outputs the filenames, but takes into account that you want to overwrite some target sizes of the images.

packages/app/src/lib/sanity.tsx
```diff
@@ -94,7 +100,10 @@ export function getImageProps<T extends ImageBlock>(
     asset.extension === 'svg'
       ? undefined
       : imageResizeTargets
-          .map((size) => `${getImageSrc(asset, size)} ${size}w`)
+          .map(
+            (size) =>
+              `${getImageSrc(asset, size >= 1024 ? 1024 : size)} ${size}w`
+          )
```
VWSCoronaDashboard8, line 106: Seems like an odd choice to limit the return values from imageResizeTargets since that is a configuration specifying what sizes we need/want. I think it makes more sense to change that config. Or am I missing something?

VWSCoronaDashboard8 (dismissed):

packages/app/src/lib/sanity.tsx
```diff
@@ -94,7 +100,10 @@ export function getImageProps<T extends ImageBlock>(
     asset.extension === 'svg'
       ? undefined
       : imageResizeTargets
-          .map((size) => `${getImageSrc(asset, size)} ${size}w`)
+          .map(
+            (size) =>
+              `${getImageSrc(asset, size >= 1024 ? 1024 : size)} ${size}w`
```
VWSCoronaDashboard4, line 105: the 1024 could be an optional property passed to `getImageProps`? That would keep this method generic. Downside is, the 1024 would need to be duplicated on a bunch of places.

VWSCoronaDashboard8: Seems more like a system wide setting to me, so therefor I would expect it to be placed in some global config file not set via props. But without further comment I'll stick to my point that this capping should be done by changing the images sizes setting and not as part of this logic.

VWSCoronaDashboard: The problem is you might need the 2048 for retina screens, I'm hesitant to remove it from the configuration completely. I'll take a look at this PR and see how/if I would improve this logic.

packages/app/src/lib/sanity.tsx
VWSCoronaDashboard: This logic has been removed, so resolving this conversation.

packages/app/src/lib/sanity.tsx
VWSCoronaDashboard: Resolving this conversation because I pushed a new approach to this problem in this branch.

VWSCoronaDashboard:
After seeing your feedback I decided to change the approach of this PR. The new logic is as follows:

The Sanity Image Props getter `getImageProps` now accepts an options object. In this object you can overrule the default width of the image from the original asset width. It also accepts a `sizes` key with a nested array. In this array you can map viewport widths to image widths.

**How do you use it?**

```ts
getImageProps(imageNode, {
  "sizes": [
    [320, 320],
    [1024, 768]
  ]
})
```

what would happen:

- At 320vw we load a 320px image
- at 1024vw you only need to load a 768px image.

**What needs to be improved eventually?**

I've only applied the change to cover images in article pages now, just as an example. For every image that comes from the CMS we could provide these options which will lead to some more prop drilling. I can live with it, but maybe we can improve this at some point.

PR #1900, opened 2021-02-11
Bundling common with Rollup

VWSCoronaDashboard8:
## Summary

This is a first attempt a bundling common with Rollup. The idea was to have tree shaking and multiple compile target outputs (CommonJS and ES modules) and maybe a watch task with conversion of TS path aliases.

If you run `yarn rollup` it does create output, but there seems to still be a lot of stuff in index.js that shouldn't be there. Mainly the whole of Lodash again :(


VWSCoronaDashboard:
I think we should take a look at https://github.com/formium/tsdx. Hopefully that'll work out of the box.

PR #1907, opened 2021-02-12
Remove hyperlinks in the tiles

VWSCoronaDashboard14:
Remove all the hyperlinks in the tiles and changed the order of the hidden text in content headers.


packages/app/src/components-styled/content-header/metadata.tsx
```diff
@@ -119,21 +118,11 @@ function MetadataItem(props: MetadataItemProps) {
...
-                css={css({
-                  display: 'inline-block',
-                  my: '0',
-                  fontSize: 'inherit',
-                })}
-              >
-                {item.text}
-              </Text>
-            )}
+            {!item.href && <>{item.text}</>}
```
VWSCoronaDashboard8, line 125: Why is text wrapped in a fragment?

packages/app/src/components-styled/chart-tile-container.tsx
```diff
@@ -18,7 +16,7 @@ export function ChartTileContainer({
       {metadata && (
         <>
           <Spacer m="auto" />
-          <Metadata {...metadata} accessibilitySubject={accessibilitySubject} />
+          <Metadata {...metadata} />
```
VWSCoronaDashboard8, line 19: Seems like this was carried over from an older PR?

packages/app/src/components-styled/metadata.tsx
```diff
@@ -42,31 +33,7 @@ export function Metadata({
...
-                  display: 'inline-block',
-                  my: '0',
-                  fontSize: 'inherit',
-                })}
-              >
-                {source.text}
-              </Text>
-            )}
-          </>
+          <>{`${locale.common.metadata.source}: ${source.text}`}</>
```
VWSCoronaDashboard8, line 36:
```suggestion
          `${locale.common.metadata.source}: ${source.text}`
```
Again, I don't think the fragment has a function here.

VWSCoronaDashboard8 (approved):
It's a bit convoluted, because the title doesn't tell me why the accessiblitySubject is being removed, but I'm just going to assume this is all intended.

VWSCoronaDashboard10 (approved):

VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/chart-tile-container.tsx
VWSCoronaDashboard14: I've created a new PR because I merged the other one in development already (before there was some talk about the right way to implement this).

PR #1908, opened 2021-02-12
chore: Update the sync-assets script to retrieve webp images

VWSCoronaDashboard13:
## Summary

Updated the sync-assets script to retrieve webp and added a `SanityImage` type to use the webp format of images when supported 

## Detailed design

Updated the script to retrieve a webp version of every image. 

Replaced all instances of the `Image` component with `SanityImage` which checks for a srcSet (which is only available for images that are not svgs) and uses the picture element to define srcSets and a fallback image

Decided to keep them a separate components as one builds upon the other, but it could make sense to just export one?


packages/app/src/components-styled/image.tsx
```diff
@@ -8,3 +9,26 @@ export const Image = styled.img(
...
+};
+
+export function SanityImage(props: SanityImageProps) {
+  const { src, srcSet, ...imageProps } = props;
+
+  if (!srcSet) {
+    return <Image {...props} />;
+  }
+
+  const extension = src.split('.')[1];
```
VWSCoronaDashboard8, line 26: nit; This doesn't seem like a very robust way to determine a file extension. Maybe take the last item instead of assuming there to be one dot / two components, that would already make it more robust.
```suggestion
  const extension = src.split('.').pop();
```

packages/app/src/lib/sanity.tsx
```diff
@@ -83,6 +83,7 @@ export function getImageProps<T extends ImageBlock>(
   node: T,
   desiredWith = node.asset.metadata.dimensions.width
 ) {
+  console.log(node);
```
VWSCoronaDashboard8: Was this left in on purpose?

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard (approved):
Looks nice and clean to me. Good improvement! 

I like the name SanityImage, because images from Sanity are the only images that possibly have webp. The only thing I question about is the css prop and the TS declaration, but one of the other developers will probably have opinions on that.

packages/app/src/components-styled/image.tsx
VWSCoronaDashboard: You can also get the sha1hash from sanity and manually add the width and webp extension behind it. That's even safer, because this approach will break whenever an image is uploaded with a dot in the filename. But I'll admit, it's a bit more complicated and we know Sanity returns sha1 hashes anyway. There's no real danger of that ever happening.

Another possibility:

`const [filename, extension] = src.split('.')`

packages/app/src/components-styled/image.tsx
```diff
@@ -8,3 +9,26 @@ export const Image = styled.img(
     height: 'auto',
   })
 );
+
+type SanityImageProps = {
+  src: string;
+  srcSet?: string;
+  css?: CssFunctionReturnType;
```
VWSCoronaDashboard10, line 16: We usually do not define `css` props on custom function components, I also have the feeling it might cause some unexpected bugs (not sure why, but I think I've experienced that in the past).
It might be better to accept a `className` prop and pass it on to the Image component.

Implementing a className prop allows us to style the SanityImage component as follows:
```tsx
styled(SanityImage)(css({ ... }))
```

packages/app/src/components-styled/image.tsx
```diff
@@ -8,3 +9,26 @@ export const Image = styled.img(
...
+};
+
+export function SanityImage(props: SanityImageProps) {
```
VWSCoronaDashboard10, line 19: How about moving this `SanityImage` component to the subdirectory `src/components/cms/sanity-image.tsx`?

packages/app/src/components-styled/image.tsx
VWSCoronaDashboard10: I have the feeling this logic should be moved to the `getImageSrc` utility somehow ..

VWSCoronaDashboard13: i opted for the suggestion to get the last item 👍

VWSCoronaDashboard13: yeah, this can be provided by getImageProps since they will always be used together 👍

packages/app/src/components-styled/image.tsx
VWSCoronaDashboard13: looks like I don't even need to add className to the props when styled that way!

packages/app/src/components-styled/image.tsx
VWSCoronaDashboard13: That makes sense 👍

PR #1912, opened 2021-02-12
Load IE11 polyfills in _document.js

VWSCoronaDashboard:
## Summary

Currently all Intl polyfills are loaded by default.
In this PR I'm trying to change it so polyfills are only loaded for browsers that need it. This saves us around 44KB for the common bundle, which is loaded on ALL pages.

I've done this by moving all polyfills related to Intl and Intl Locale Data to polyfill.io. It introduces an extra round-trip to polyfill.io, but that can't be slower than embedding 45KB of mostly unused JavaScript.

image

## Motivation

The projects bundle size needs to be reduced. It's getting heavier and heavier, impacting performance on our visitors devices.

# Browser check

It's a big change, so let's make sure to test this properly on all browsers.
Tested browsers:
- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge
- [x] Edge 18
- [x] Edge 17
- [x] IE 11

# Open questions

- ~Do we need to do anything special to support polyfill.io on production? CSP?~ No changes needed, we only use this for IE11 which doesn't understand CSP.


VWSCoronaDashboard1 (approved):

packages/app/src/pages/_app.tsx
```diff
@@ -9,6 +9,15 @@ import * as piwik from '~/lib/piwik';
 import { GlobalStyle } from '~/style/global-style';
 import theme from '~/style/theme';
 
+if (typeof window !== 'undefined') {
+  if (process.env.NODE_ENV === 'development') {
+    /**
+     * this polyfill allows next.js to show runtime errors in IE11
+     */
+    require('@webcomponents/shadydom');
```
VWSCoronaDashboard8, line 17: We only want runtime errors in development?

VWSCoronaDashboard8 (approved):

packages/app/src/pages/_app.tsx
VWSCoronaDashboard: I didn't change this from the current implementation. I think we need shadydom for the error overlays in Next.js, which are only visible in development mode?

VWSCoronaDashboard1 (approved):

packages/app/src/pages/_app.tsx
VWSCoronaDashboard10: > I think we need shadydom for the error overlays in Next.js, which are only visible in development mode?
this is correct 👍

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard1 (approved):

PR #1926, opened 2021-02-15
Run common build as postinstall script

VWSCoronaDashboard8:
## Summary

Now that more packages like CLI are depending on common, I think it makes sense to build common in the NPM postinstall phase, so that we are sure it is always ready to be used.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,6 @@ RUN yarn
 ARG SANITY_AUTH_TOKEN
 ARG SANITY_DATASET
 ARG SANITY_ASSETS_CACHE_DIR
-RUN yarn workspace @corona-dashboard/common build
 RUN yarn workspace @corona-dashboard/cli validate-json
 # We need to have an ENV var switch to only enable this in production, because
 # BE does not guarantee consistent data in develop.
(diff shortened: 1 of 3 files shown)


VWSCoronaDashboard (approved):

VWSCoronaDashboard10 (commented):
Are you sure the build keeps working after this change? All stages are `COPY`ing stuff from the `react-build-base`-stage, and I have the feeling we now should also do a copy of the common package ..

but keep in mind my docker knowledge is really limited, I might be wrong :-)

VWSCoronaDashboard8:
@VWSCoronaDashboard10 yes, good point. I might have overlooked that.

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard:
The plan is now to:

1. Merge the following PR and put most of the build logic outside of Docker https://github.com/minvws/nl-covid19-data-dashboard/pull/2267
2. Make a new PR where we switch to Nodejs in Docker

As such, this PR is outdated.
I do like the postinstall addition though, if even just to make sure a fresh clone is automatically ready to be used.

PR #1929, opened 2021-02-15
Stop forwarding styled component props to dom elements in base components

VWSCoronaDashboard13:
## Summary

Uses the `shouldForwardProps` config provided by emotion to stop forwarding styled component props to dom elements when using "base" components.

## Motivation

Declutter dom elements. Styled component props won't be uselessly "forwarded" to the base elements; they'll just be applied via the classname. 

## Detailed design

Added `shouldForwardProps` config to exported reusable styled components that resided in their own file. This means this only applies to components that are built from these "base" components. 

Seems that there is a type mismatch somewhere, so the `shouldForwardProp` is getting cast to a defined type and is exported from the `styledShouldForwardProp`file


VWSCoronaDashboard8 (approved):

packages/app/src/components-styled/table.ts
```diff
@@ -17,6 +17,7 @@ import {
   width,
   WidthProps,
 } from 'styled-system';
+import { StyledShouldForwardProp } from '~/utils/styledShouldForwardProp';
```
VWSCoronaDashboard8, line 20: We write all file names in kebab-case. Also it seems like you are exporting the function starting with a capital letter as if it was a type, but it is used as a value here. So I think this should be 

```suggestion
import { styledShouldForwardProp } from '~/utils/styled-should-forward-prop';
```

VWSCoronaDashboard8 (changes requested):

packages/app/src/utils/styledShouldForwardProp.ts
```diff
@@ -0,0 +1,5 @@
+import shouldForwardProp from '@styled-system/should-forward-prop';
+
+type StyledShouldForwardPropType = (prop: string | number) => boolean;
+
+export const StyledShouldForwardProp = shouldForwardProp as StyledShouldForwardPropType;
```
VWSCoronaDashboard8, line 5: What you are exporting is a (typed) value, not a type itself.

```suggestion
export const styledShouldForwardProp = shouldForwardProp as StyledShouldForwardPropType;
```

packages/app/src/utils/styledShouldForwardProp.ts
```diff
@@ -0,0 +1,5 @@
+import shouldForwardProp from '@styled-system/should-forward-prop';
+
+type StyledShouldForwardPropType = (prop: string | number) => boolean;
```
VWSCoronaDashboard8, line 3: Once you rename the exported function, this doesn't need the Type suffix. 

```suggestion
type StyledShouldForwardProp = (prop: string | number) => boolean;
```

packages/app/src/components-styled/table.ts
VWSCoronaDashboard13: right, this makes sense. I just glanced a the other util files so I missed that there was a mix of casings. The requested updates have been made 👍

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard1:
The changes seem fine to me, but there are still a bunch of compile errors happening. So, run `yarn compile` in the project root and have a look at what's going wroing.

VWSCoronaDashboard13:
> The changes seem fine to me, but there are still a bunch of compile errors happening. So, run `yarn compile` in the project root and have a look at what's going wroing.

Yeah, it looks like defining the StyledShouldForwardProp type is causing the compilation errors. I have a feeling trying to fix this probably isn't worth the effort and it would make sense to revert to casting to any?

VWSCoronaDashboard1:
> > The changes seem fine to me, but there are still a bunch of compile errors happening. So, run `yarn compile` in the project root and have a look at what's going wroing.
> 
> Yeah, it looks like defining the StyledShouldForwardProp type is causing the compilation errors. I have a feeling trying to fix this probably isn't worth the effort and it would make sense to revert to casting to any?

if all else fails ;)

VWSCoronaDashboard8 (approved):

PR #1937, opened 2021-02-16
Add stacked area chart for vaccination delivery/estimates

VWSCoronaDashboard1:
## Summary

This adds a stacked area chart that displays 4 separate datasets as one graph: Delivered vaccines, estimated delivered vaccines, administered vaccines, and estimated administered vaccines.

## Motivation

Feature request

## Detailed design

image

The chart largely follows the same patterns as the line chart. Differences are in the way the datasets are passed in. On the outside Trend and Area descriptors are passed in which are internally shaped into Trend and Area configs.
Gathering all of the data into a manageable chunk of data that needs to be passed to the tooltip handling proved to be quite a challenge too. It turned out rather elaborate, but for now this works but probably requires some discussion over how we'll handle this type of thing in all of the charts.

## Drawbacks

n/a

## Alternatives

n/a

## Unresolved questions

n/a

### Governance

- [ ] Documentation is added


packages/app/src/components-styled/area-chart/area-chart-graph.tsx
```diff
@@ -0,0 +1,240 @@
...
+          })}
+        />
+
+        <Group>
+          {areas.map((area, index) => (
+            <AreaStack
+              key={index}
+              keys={area.displays.map((x) => x.metricProperty) as string[]}
+              data={area.values}
+              x={(d) => xScale(d.data.__date) ?? 0}
```
VWSCoronaDashboard4, line 161: are the fallbacks here (`?? 0` and `|| 0`) preventable by tighter typing higher up in the chain?

packages/app/src/components-styled/area-chart/area-chart-graph.tsx
```diff
@@ -0,0 +1,240 @@
...
+              y0={(d) => yScale(d[0]) ?? 0}
+              y1={(d) => yScale(d[1]) ?? 0}
+            >
+              {({ stacks, path }) =>
+                stacks.map((stack) => (
+                  <path
+                    key={`area-chart-stack-${stack.key}`}
```
VWSCoronaDashboard4, line 168: if this is a loop in a loop, there will be more paths with the same key?

packages/app/src/components-styled/area-chart/area-chart-graph.tsx
```diff
@@ -0,0 +1,240 @@
...
+    '&:not(:root)': {
+      overflow: 'visible',
+    },
+    '&:focus': {
+      outline: 'none',
+    },
+    '& .area-chart-x-axis-label': {
+      fontSize: 1,
+      fill: 'annotation',
+      fontFamily: '"RO Sans", Calibri, sans-serif',
```
VWSCoronaDashboard4, line 237: you can import the font family via the theme

packages/app/src/components-styled/difference-indicator.tsx
```diff
@@ -188,7 +188,7 @@ function renderTileIndicator(
 type SpanProps = SpaceProps & ColorProps & TypographyProps;
 
 const Span = styled.span<SpanProps>(compose(color, space, typography));
-const IconContainer = styled(Span)(
+const IconContainer = (styled as any)(Span, { stuff: 1 })(
```
VWSCoronaDashboard4, line 191: stuff?

packages/app/src/pages/landelijk/vaccinaties.tsx
```diff
@@ -326,4 +390,79 @@ const VaccinationPage: FCWithLayout<typeof getStaticProps> = ({
 
 VaccinationPage.getLayout = getNationalLayout;
 
+export type TooltipValue = (
```
VWSCoronaDashboard4, line 393: I'd order this file a bit handier; interfaces more to the top; maybe put helper methods in a `domain/vaccines` folder?

packages/app/src/components-styled/area-chart/area-chart-graph.tsx
```diff
@@ -0,0 +1,240 @@
...
+          width={bounds.width}
+          numTicks={numTicks}
+          stroke={defaultColors.axis}
+        />
+
+        <AxisBottom
+          scale={xScale}
+          tickValues={xScale.domain()}
+          tickFormat={formatXAxis as AnyTickFormatter}
+          top={bounds.height}
```
VWSCoronaDashboard4, line 129: I've commented about this in the multi line chart from @VWSCoronaDashboard8 : does this work well if you adjust the `padding.top`? Maybe for a different PR

packages/app/src/components-styled/area-chart/area-chart-graph.tsx
VWSCoronaDashboard1: Correct, I had already fixed this in the patterns loop, but it needs to be fixed here as well, good catch

packages/app/src/components-styled/area-chart/area-chart-graph.tsx
VWSCoronaDashboard1: I don't think so, xScale can always return an undefined (even though at runtime this really wouldn't happen)

packages/app/src/components-styled/difference-indicator.tsx
VWSCoronaDashboard1: haha, sorry, this was a bit of test code. I'll get rid of it.

packages/app/src/components-styled/area-chart/area-chart-graph.tsx
VWSCoronaDashboard1: The bounds are calculated according to the padding, so this ought to work correctly, yes

packages/app/src/components-styled/area-chart/area-chart-graph.tsx
```diff
@@ -0,0 +1,240 @@
...
+  displays: AreaDisplay<T>[];
+};
+
+const defaultColors = {
+  axis: colors.silver,
+  axisLabels: colors.data.axisLabels,
+  benchmark: colors.data.benchmark,
+};
+
+type AreaChartGraphProps<T extends TrendValue, K extends TrendValue> = {
```
VWSCoronaDashboard8, line 47: Discussion for later; Looking at this component in relation to AreaChart I wonder what the goal is of the abstraction. The axes are also in here, and I think I concluded in the stacked chart that you might as well place the axes in the root component. 

I suspect composition will be easier / more transparent without this abstraction in between.

packages/app/src/components-styled/area-chart/area-chart.tsx
```diff
@@ -0,0 +1,382 @@
...
+type AreaChartProps<T extends TimestampedValue, K extends TimestampedValue> = {
+  formatTooltip?: (
+    values: HoverPoint<
+      (T & TimestampedTrendValue) | (K & TimestampedTrendValue)
+    >[],
+    isPercentage?: boolean
+  ) => React.ReactNode;
+  width: number;
+  trends: TrendDescriptor<T>[];
+  areas: AreaDescriptor<K>[];
```
VWSCoronaDashboard8, line 88: Discussion for later; I find it odd that trends and area data plus configuration are passed in separately. I would expect the data to be passed in as-is (in the case of the specific vaccination chart a consolidation of multiple schema data values) and then one configuration to determine whether each value time series should be rendered as a line or an area.

The fact that they are also not the same time (T and K) puzzles me as well. It all feels a bit convoluted like this.

packages/app/src/components-styled/area-chart/area-chart.tsx
```diff
@@ -0,0 +1,382 @@
...
+  height: number,
+  padding: ChartPadding,
+  xScale: ScaleTime<number, number>
+) {
+  const dates = areas.map((area) => area.values[0].__date);
+  dates.shift();
+
+  return dates.map((date) => {
+    const x = xScale(date);
+    return x !== undefined ? (
```
VWSCoronaDashboard8, line 301: nit; This is a bit easier to follow I think.

```suggestion
    return x === undefined ? null :
 
```

packages/app/src/components-styled/area-chart/components/tooltip.tsx
```diff
@@ -0,0 +1,75 @@
+import css from '@styled-system/css';
```
VWSCoronaDashboard8, line 1: I would have chosen to use the visx tooltip since I think it looks and behaves better than our current home grown one. It also saves you from committing all of the code below.

packages/app/src/components-styled/area-chart/hooks/use-area-configs.ts
```diff
@@ -0,0 +1,89 @@
...
+        ),
+        displays: [
+          ...descriptor.displays.map((x) => ({ id: createUniqueId(), ...x })),
+        ],
+      })
+    );
+  }, [areaDescriptors, timeframe]);
+
+  return areaConfigs;
+}
```
VWSCoronaDashboard8, line 38: This is all a little cryptic to me. I don't quite get what a display and a descriptor is, and like I said earlier, I mainly wonder why this area is being treated differently from trend data.

packages/app/src/components-styled/area-chart/hooks/use-chart-hover.ts
```diff
@@ -0,0 +1,162 @@
...
+    const trendValue = bisect(config.values, point.x, xScale);
+    return trendValue
+      ? {
+          data: trendValue,
+          color: config.color,
+          label: config.legendLabel,
+        }
+      : undefined;
+  };
+}
```
VWSCoronaDashboard8, line 30: Maybe you can add a comment about what this abstraction does. Because from just reading the name "createTrendValue" in combination with the arguments like bisect and scale, I am completely lost.

packages/app/src/pages/landelijk/vaccinaties.tsx
VWSCoronaDashboard1: Done

packages/app/src/components-styled/area-chart/area-chart.tsx
VWSCoronaDashboard1: The data comes from 4 different datasets, that don't share common types. so, having one generic wouldn't really work, I'm afraid.
I'd be very interested to see how you would specify the different line and area visualizations in one separate configuration without that becoming convoluted as well. Definitely worthy of discussion.

packages/app/src/components-styled/area-chart/hooks/use-area-configs.ts
VWSCoronaDashboard1: Mainly because the data shape that the AreaStack expects is completely different from what an array of trends expects.

packages/app/src/components-styled/area-chart/hooks/use-chart-hover.ts
VWSCoronaDashboard1: I've placed this functions underneath the useChartHover function, which is the main entry point of this file (hence the file name). In that function are comments that ought to clarify better what is being performed. I've renamed some of these functions as well, since their implementation had changed over the course of development.

packages/app/src/utils/calculate-distance.ts
```diff
@@ -0,0 +1,12 @@
+import { TimestampedValue } from '@corona-dashboard/common';
+import { Point } from '@visx/point';
+import { HoverPoint } from '~/components-styled/line-chart/components';
+
+export function calculateDistance(
+  point1: HoverPoint<TimestampedValue>,
+  point2: Point
+) {
```
VWSCoronaDashboard10, line 8: the api of this util looks weird, with two different point types. Does something like this work?
```suggestion
export function calculateDistance<T extends { x: number; y: number }>(
  point1: T,
  point2: T
) {
```

packages/app/src/components-styled/line-chart/logic/index.ts
```diff
@@ -29,24 +34,17 @@ export type AnyFilteredValue = Record<string, number>;
...
-      .filter(isPresent) // omit null values
-      .reduce((acc, value) => (value > acc ? value : acc), -Infinity)
-  );
-
-  const overallMaximum = Math.max(...peakValues);
+export function calculateYMax(values: TrendValue[], signaalwaarde = -Infinity) {
+  const peakValues = values
+    .map((x) => x.__value)
+    .filter(isPresent) // omit null values
+    .reduce((acc, value) => (value > acc ? value : acc), -Infinity);
```
VWSCoronaDashboard10, line 41: I might be mistaken, but isn't this reduce doing the same as Math.max(...values)?

packages/app/src/components-styled/area-chart/area-chart.tsx
```diff
@@ -0,0 +1,396 @@
...
+  signaalwaarde?: number;
+  isPercentage?: boolean;
+  divider?: DividerConfig;
+};
+
+const dateToValue = (d: { valueOf(): number }) => d.valueOf() / 1000;
+const formatXAxis = (date: Date | { valueOf(): number }) =>
+  formatDateFromSeconds(dateToValue(date), 'axis');
+const formatYAxisFn = (y: number) => formatNumber(y / 1000000);
+const formatYAxisPercentageFn = (y: number) => `${formatPercentage(y)}%`;
```
VWSCoronaDashboard10, line 105: why the `Fn`-suffix?

packages/app/src/components-styled/area-chart/area-chart.tsx
```diff
@@ -0,0 +1,396 @@
...
+  );
+
+  const bounds: ChartBounds = {
+    width: width - padding.left - padding.right,
+    height: height - padding.top - padding.bottom,
+  };
+
+  function getDate(x: TrendValue) {
+    return x.__date;
+  }
```
VWSCoronaDashboard10, line 201: this can be moved outside this function body

packages/app/src/components-styled/area-chart/area-chart.tsx
```diff
@@ -0,0 +1,396 @@
...
+              ? `${formatPercentage(value.data.__value)}%`
+              : formatNumber(value.data.__value)}
+          </Box>
+        ))}
+      </Box>
+    );
+  } else if (isDateSpanValue(data)) {
+    const dateStartString = formatDateFromSeconds(
+      data.date_start_unix,
+      'short'
```
VWSCoronaDashboard10, line 374: this will break when you've merged master in your branch. `short` has been replaced with `day-month`.

packages/app/src/components-styled/area-chart/components/area-chart-graph.tsx
```diff
@@ -0,0 +1,244 @@
...
+
+const StyledSvg = styled.svg(
+  css({
+    '&:not(:root)': {
+      overflow: 'visible',
+    },
+    '&:focus': {
+      outline: 'none',
+    },
+    '& .area-chart-x-axis-label': {
```
VWSCoronaDashboard10, line 238: I cannot find this className in this PR, leftover?

packages/app/src/components-styled/area-chart/hooks/use-chart-hover.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    result[0].values.pop();
+
+    return result;
+  }, [trends]);
+
+  const _areas = useMemo(() => {
+    const result = areas.map((x) => ({
+      ...x,
+      values: x.values.map((x) => ({ ...x })),
```
VWSCoronaDashboard10, line 47: why is it necessary to copy the value instances?

packages/app/src/components-styled/area-chart/hooks/use-chart-hover.ts
```diff
@@ -0,0 +1,188 @@
...
+    }));
+    result[0].values.pop();
```
VWSCoronaDashboard10, line 49: could you add a comment about why the pop is necessary?

VWSCoronaDashboard10 (commented):
its a lot to grasp and review at once.. Some pieces look really complex, but I think we can refactor that in another sprint.

packages/app/src/components-styled/line-chart/logic/index.ts
VWSCoronaDashboard1: Not sure, I didn't write the original code, I just refactored it a bit

packages/app/src/components-styled/area-chart/area-chart.tsx
VWSCoronaDashboard1: I copied this from another part of the code base

packages/app/src/components-styled/area-chart/area-chart.tsx
VWSCoronaDashboard1: Why would I merge master into my branch? Why is there a change in master that is not in develop?

packages/app/src/components-styled/area-chart/components/area-chart-graph.tsx
VWSCoronaDashboard1: Indeed

packages/app/src/components-styled/area-chart/hooks/use-chart-hover.ts
VWSCoronaDashboard1: Because I'm manipulating the array. I'm removing an element that is a duplicate, the duplicate is needed in rendering the graph to make the adjacent parts connect, but when determining the hovered points these duplicates cause problems.

packages/app/src/components-styled/area-chart/hooks/use-chart-hover.ts
VWSCoronaDashboard1: I believe there is a very large comment above these two memos...

packages/app/src/components-styled/area-chart/area-chart.tsx
VWSCoronaDashboard1: Aha, I see you actually meant develop ;)

VWSCoronaDashboard10: Yep haha

packages/app/src/components-styled/line-chart/line-chart.tsx
```diff
@@ -56,14 +58,14 @@ export type LineChartProps<T extends TimestampedValue> = {
   height?: number;
   timeframe?: TimeframeOption;
   signaalwaarde?: number;
-  formatTooltip?: (value: (T & TrendValue)[]) => React.ReactNode;
+  formatTooltip?: (value: (T & TimestampedTrendValue)[]) => React.ReactNode;
```
VWSCoronaDashboard8, line 61: This looks wrong to me. T is already a TimestampedValue (by extends). The definition of TrendValue was about adding __date and __value for rendering the trend data.

I think the introduction of `TimestampedTrendValue` as part of this PR might be a mistake.

packages/app/src/components-styled/line-chart/logic/index.ts
VWSCoronaDashboard8: Yes I think outcome is the same the same, but using reduce to calculate max is not a problem IMO, and the function fits the flow of data.

packages/app/src/components-styled/stacked-chart/logic.ts
```diff
@@ -39,15 +39,17 @@ export function calculateSeriesMaximum(series: SeriesValue[]) {
 export function getValuesInTimeframe<T extends TimestampedValue>(
   values: T[],
   timeframe: TimeframeOption
-) {
+): T[] {
```
VWSCoronaDashboard8, line 42: nit; I think this annotation is redundant if you use `as T[]` on the return values

packages/app/src/pages/landelijk/vaccinaties.tsx
```diff
@@ -217,39 +237,80 @@ const VaccinationPage: FCWithLayout<typeof getStaticProps> = ({
...
-              }}
-            />
-          }
-          ariaDescription={
-            siteText.accessibility.grafieken.vaccin_levering_en_prikken
-          }
+          description={text.grafiek.titel}
           metadata={{
-            date: 1612375710,
+            date: 1611593522,
```
VWSCoronaDashboard8, line 242: Looks like this still needs to be connected to the data timestamp

packages/app/src/pages/landelijk/vaccinaties.tsx
```diff
@@ -217,39 +237,80 @@ const VaccinationPage: FCWithLayout<typeof getStaticProps> = ({
...
-            ) : (
-              <Image
-                src="/images/vaccines_administered_chart_en.svg"
-                loading="lazy"
-                decoding="async"
-              />
-            )}
-          </AspectRatio>
+          <Box>
+            {' '}
```
VWSCoronaDashboard8, line 247: This looks unintended

packages/app/src/pages/landelijk/vaccinaties.tsx
```diff
@@ -217,39 +237,80 @@ const VaccinationPage: FCWithLayout<typeof getStaticProps> = ({
...
+            <ParentSize>
+              {({ width }) => (
+                <AreaChart<
+                  NlVaccineDeliveryValue | NlVaccineDeliveryEstimateValue,
+                  | NlVaccineAdministeredValue
+                  | NlVaccineAdministeredEstimateValue
```
VWSCoronaDashboard8, line 253: I feel these should all have been joined into one overarching type with nullable values before passing things into the area chart. I think it could have simplified a lot, because I think the chart should not have to know about the fragmented nature of the supplied data. Not if it is to be a generic area chart.

But this will be a study for later...

packages/app/src/pages/landelijk/vaccinaties.tsx
```diff
@@ -217,39 +237,80 @@ const VaccinationPage: FCWithLayout<typeof getStaticProps> = ({
...
+                  }}
+                  trends={[
+                    {
+                      values: vaccineDeliveryValues,
+                      displays: [
+                        {
+                          metricProperty: 'total',
+                          color: colors.data.emphasis,
+                          legendLabel: text.data.vaccination_chart.delivered,
+                        },
```
VWSCoronaDashboard8, line 272: The trends/area configurations with multiple display properties is where you lost me in this PR.

VWSCoronaDashboard8 (approved):
I feel the introduction of TimestampedTrendValue is a mistake, and the API interface with the configs and their types is way to complicated, but I'm approving this assuming we will revisit this at a later point in time.

packages/app/src/components-styled/line-chart/line-chart.tsx
VWSCoronaDashboard1: There's a TimestampedValue coming which is then transformed into a TrendValue, leaving the original properties as well, Hence, they become a combination of the two, a TimestampedTrendValue, all of the properties are used throughout the graph.
The __date property for instance can't be used to display the date in a tooltip because that value might have been manipulated to be in the middle of a datespan in order to render the rectangles.

packages/app/src/pages/landelijk/vaccinaties.tsx
VWSCoronaDashboard1: We discussed this very matter when the schemas were defined. It's a tad frustrating to read that it's being circled back on now...

packages/app/src/pages/landelijk/vaccinaties.tsx
VWSCoronaDashboard1: If you look at the chart, half of it is dashed/hatched, so there needs to be a separate display config for each. I'm not sure how else this could have been tackled without moving the complexity somewhere else.

packages/app/src/components-styled/line-chart/line-chart.tsx
VWSCoronaDashboard8: Yes, I believe the __date  should not be used in the in the formatting of the Tooltip. the __ named properties were only really meant to be private to the chart (and like you noticed can also be a mutation in case of date), but I used them wrong in the first implementation as well. 

The tooltips I made after that all get the original value T passed and use the original timestamps there.

packages/app/src/pages/landelijk/vaccinaties.tsx
VWSCoronaDashboard8: @VWSCoronaDashboard1 that's not the point I'm trying to make. I think the schema's make sense from a data perspective and can stay as they are. 

I believe the chart, especially if it were to be a generic area chart, should not have to deal with the data being fragmented like that. So in order for the chart API/implementation to be simplified, I think we should massage the data into a single type that the chart can then handle. 

But let's revisit this later.

packages/app/src/components-styled/legenda.tsx
```diff
@@ -2,24 +2,37 @@ import css from '@styled-system/css';
...
 
-export type LegendShape = 'line' | 'square' | 'circle';
+export type LegendShape = 'line' | 'square' | 'circle' | 'custom';
 
-export type LegendItem = { color: string; label: string; shape: LegendShape };
+export type LegendItem = {
+  color: string;
+  label: string;
+  shape: LegendShape;
+  ShapeComponent?: any;
```
VWSCoronaDashboard10, line 11: Prettier imho:

```suggestion
  shapeComponent?: ReactNode;
```

and in the calling context you can pass the shape with `shapeComponent: <HatchedSquare />`

VWSCoronaDashboard1: tried it, got into a fight with the compiler, lost my patience

packages/app/src/pages/landelijk/vaccinaties.tsx
VWSCoronaDashboard8: Moving some of the complexity elsewhere can not be avoided, but with a data transforming function that gets applied before hitting the chart API I think it would have saved you a lot of complexity in the API and further down the chart.

packages/app/src/domain/vaccines/use-vaccine-names.ts
```diff
@@ -1,9 +1,12 @@
...
+import { useMemo } from 'react';
 import { colors } from '~/style/theme';
 
 export function useVaccineNames(
   lastValue: NlVaccineAdministeredValue
 ): string[] {
-  const productNames = Object.keys(colors.data.vaccines);
-  return productNames.filter((x) => (lastValue as any).hasOwnProperty(x));
+  return useMemo(() => {
+    const productNames = Object.keys(colors.data.vaccines);
```
VWSCoronaDashboard10, line 9: this looks weird/unsafe, using theme properties as "data input". Feels like it's better to have a separate list of product names somewhere?

VWSCoronaDashboard1: This data isn't available anywhere else, so for now this is the way to go.

VWSCoronaDashboard1: I also beg to differ when it comes to this being unsafe. The theme colors for the different products have to exist, the product names on the actual data item are optional (because not all products are available yet), if a color hasn't been defined for the specific product then it shouldn't be in the graph. So, it's either not showing the data point or, showing in in black, both of which are undesirable.

VWSCoronaDashboard10: Isn't it as simple as this:

```tsx
const vaccines = [
  'pfizer',
  'moderna',
  'astra_zeneca',
  'cure_vac',
  'janssen',
  'sanofi',
] as const;

vaccines.forEach((x) =>
  assert(colors.data.vaccines[x], `missing vaccine color for vaccine ${x}`)
);

export function useVaccineNames(
  lastValue: NlVaccineAdministeredValue
): string[] {
  return useMemo(() => {
    const productNames = Object.keys(colors.data.vaccines);
    return productNames.filter((x) => (lastValue as any)[x] !== undefined);
  }, [lastValue]);
}

VWSCoronaDashboard1: So, then you keep the list in two places. I'm not sure if this is really adding so much robustness to be honest...

VWSCoronaDashboard10: I think it is because it's explicit instead of magical. It's just weird that all possible products are defined in our theme. That file should only define colors and such, it shouldn't be used as "data", right?

PR #1941, opened 2021-02-16
Implement sewer chart built with visx

VWSCoronaDashboard10:
image

Notes:

- The formatDate `short`-type was not captured, instead the fallback was used to render the `short` type. I've renamed the `short` type to `day-month`, because that's the formatter used behind the scenes.


packages/app/src/components-styled/sewer-chart/logic.ts
```diff
@@ -0,0 +1,331 @@
...
+    const distance = 25;
+
+    const closest =
+      voronoiLayout.find(
+        inputPoint.x - dimensions.padding.left,
+        inputPoint.y - dimensions.padding.top,
+        distance
+      ) || undefined;
+
+    datum = closest?.data;
```
VWSCoronaDashboard4, line 233: does the rest of the method (including returning) work well if there is no closest? perhaps `if (!closest) { return }`

packages/app/src/utils/formatDate.ts
```diff
@@ -108,23 +108,69 @@ export function formatDateFromMilliseconds(
...
+  dateOrTimestamp: Date | number,
+  style: formatStyle = 'day-month'
+) {
+  const date =
+    dateOrTimestamp instanceof Date
+      ? dateOrTimestamp
+      : new Date(dateOrTimestamp as number);
+
+  const cacheKey = `${date.getTime()}-${style}`;
+  const dateCached = formatCache[cacheKey];
```
VWSCoronaDashboard4, line 131: :100: nice!

packages/app/src/utils/formatDate.ts
```diff
@@ -108,23 +108,69 @@ export function formatDateFromMilliseconds(
...
+
+  if (dateCached) return dateCached;
+
+  let dateFormatted: string;
+
+  switch (style) {
```
VWSCoronaDashboard4, line 137: Perhaps split this method into two parts: the caching and the formatting part?

I'm not usually in favour of switches, but compared to the original if/else it seems neater for most of the code. In a separate method you can get rid of the breaks in favour of returns. This would allow for neater formatting of the Relative date formatting as well (outside the switch)

packages/app/src/components-styled/sewer-chart/components/toggle-outlier-button.tsx
```diff
@@ -0,0 +1,16 @@
+import css from '@styled-system/css';
+import styled from 'styled-components';
+
+export const ToggleOutlierButton = styled.button(
+  css({
+    bg: 'rgba(218, 218, 218, 0.2)',
```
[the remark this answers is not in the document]
VWSCoronaDashboard10: Coud be but they look rather specific to this single button..

VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/sewer-chart/logic.ts
```diff
@@ -0,0 +1,331 @@
...
+}
+
+/**
+ * Create scales and datum=>coordinate-mappers (getX, getY)
+ */
+export function useSewerChartScales(
+  values: SewerChartValue[],
+  bounds: { width: number; height: number }
+) {
+  const xScale = useMemo(() => {
```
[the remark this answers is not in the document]
VWSCoronaDashboard10: It might have some impact on performance but I'll take the risk (since it already was a premature optimisation)

packages/app/src/components-styled/sewer-chart/logic.ts
VWSCoronaDashboard10: both `datum` and `point` should be returned as `undefined` if there's no closest point.

VWSCoronaDashboard4 (approved):

packages/app/src/components-styled/sewer-chart/logic.ts
```diff
@@ -0,0 +1,320 @@
...
+function dedupe<T>(key: keyof T) {
+  return (x: T, index: number, list: T[]) =>
+    list.findIndex((x2) => x[key] === x2[key]) === index;
+}
+
+/**
+ * get max value of items mapped by a getter
+ */
+function getMax<T>(items: T[], getValue: (item: T) => number | undefined) {
+  return Math.max(...items.map((x) => getValue(x) || 0));
```
VWSCoronaDashboard8, line 285: nit-nit; -Infinity might be better than 0 :)

packages/app/src/components-styled/sewer-chart/logic.ts
```diff
@@ -0,0 +1,320 @@
...
+}
+
+/**
+ * utility hook in order to calculate pan-distance based on given points
```
VWSCoronaDashboard8, line 289: What's a pan distance? 🤔 I think a bit more explanation would be useful

packages/app/src/utils/formatDate.ts
VWSCoronaDashboard8: nit; Great idea. But I think a generic memoize helper would have my preference. No need to implement/expose caching logic ourselves even though it is very basic.

packages/common/src/utils/formatNumber.ts
```diff
@@ -5,7 +5,7 @@ const locale = getLocale();
 const NumberFormat = new Intl.NumberFormat(locale);
 
 export function formatNumber(
-  value: number | string | undefined | null
+  value: number | string | undefined | null | { valueOf(): number }
```
VWSCoronaDashboard8, line 8: I don't get this one...

VWSCoronaDashboard8 (approved):

packages/app/src/components-styled/sewer-chart/logic.ts
VWSCoronaDashboard10: ehm, the distance your pointer travelled, how would you call that..?

VWSCoronaDashboard10: I'll rename the function to `usePointDistance` and explain the use-case in the comments.

packages/common/src/utils/formatNumber.ts
VWSCoronaDashboard10: leftover from an older implementation .. will clean it up

PR #1950, opened 2021-02-17
Bring lighthouse score back to 💯

VWSCoronaDashboard10:
## Summary
Fix some html errors like lists with `div`s instead of `li`s.
Also fixed an a11y issue where `aria-labelledby` wasn't correctly used. The attribute `aria-labelledby` can be set on stuff like an svg and should point to the element with an id which holds the description. The bug was that the element with the description also had an `aria-labelledby`-property, instead of the `id` property.


VWSCoronaDashboard4 (approved):
:100:

PR #1962, opened 2021-02-17
Add unit tests for data sorting

VWSCoronaDashboard13:
## Summary

Added unit tests for data-sorting in commons

NOTE: last test is currently failing and these tests were created based on reading the data sorting methods with no background knowledge of how this should work 

## Detailed design

Created some arrays for the time series data and am reusing them by creating new arrays with the same values. 

#### Tests
- The first test is just validating that the original time series data is not getting mutated by sorting of the same referenced array (I can remove this if preferred)
- Sorting works correctly
- Throws an error if `date_unix` is missing on all data points in a date series
- Throws an error if `date_end_unix` is missing on all data points in a date span series
- Throws an error if `date_start_unix` is missing on all data points in a date span series
- [Failing] Throws an error if `date_end_unix` is missing on the second data point in a date span series. I noticed only the first item in a series was getting checked to determine if a series was valid. I wasn't sure if this was intentional so I wasn't sure how to move forward with this. 

## Unresolved questions

Should I remove the last test and are there any other tests that we would like to add?

### DIFF
--- a/packages/common/src/__tests__/data-sorting.ts
+++ b/packages/common/src/__tests__/data-sorting.ts
@@ -1,0 +1,237 @@
+import { sortTimeSeriesInDataInPlace } from '../data-sorting';
+
+// Items that should be ignored
+const filler = {
+  filler1: {},
+  filler2: {
+    last_value: true,
+  },
+};
+
+const dateSeriesValues = [
+  {
+    date_unix: 1582848000,
+    admissions_on_date_of_admission: 0,
+    admissions_on_date_of_reporting: 0,
+    date_of_insertion_unix: 1612966742,
+  },
+  {
+    date_unix: 1582934400,
+    admissions_on_date_of_admission: 0,
+    admissions_on_date_of_reporting: 0,
+    date_of_insertion_unix: 1612966742,
+  },
+  {
+    date_unix: 1582761600,
+    admissions_on_date_of_admission: 0,
+    admissions_on_date_of_reporting: 0,
+    date_of_insertion_unix: 1612966742,
+  },
+  {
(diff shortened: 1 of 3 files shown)


packages/common/src/__tests__/data-sorting.ts
```diff
@@ -0,0 +1,237 @@
...
+    // Date span series values should be different
+    expect(dateSpanSeriesValues).not.toStrictEqual(
+      invalidTestData.dateSpanSeries.values
+    );
+
+    // Sorting in place
+    expect(() => sortTimeSeriesInDataInPlace(invalidTestData)).toThrow(Error);
+  });
+
+  it('Should throw an error when one date_end_unix is missin in a date span series', () => {
```
VWSCoronaDashboard8, line 211: Our json schema validation already enforces the existence of properties across all values, so we should not have to test for this.

packages/common/src/__tests__/data-sorting.ts
```diff
@@ -0,0 +1,237 @@
...
+    last_value: true,
+  },
+};
+
+const dateSeriesValues = [
+  {
+    date_unix: 1582848000,
+    admissions_on_date_of_admission: 0,
+    admissions_on_date_of_reporting: 0,
+    date_of_insertion_unix: 1612966742,
```
VWSCoronaDashboard8, line 16: date_of_insertion_unix has no part in the sorting algorithms, so you could leave this out. This timestamp should ideally not even be part of the values, but one level higher, because it only indicates when the data as a whole has been updated. But this is legacy...

VWSCoronaDashboard8 (approved):

PR #1963, opened 2021-02-17
Fix the prop errors from ArrowIcon

VWSCoronaDashboard:
## Summary

Applying the CSS prop on an SVG icon would result in console warnings. By wrapping it in styled first, the errors dissapear.

## Motivation

- ✨ _A clean console is a safe console._ ✨ 
- Having a single place to export rotated arrow icons is easier to maintain than remembering to update the css={css({})} props everywhere

### DIFF
--- a/packages/app/src/components-styled/article-teaser.tsx
+++ b/packages/app/src/components-styled/article-teaser.tsx
@@ -1,6 +1,6 @@
 import css from '@styled-system/css';
 import styled from 'styled-components';
-import { StyledArrowIcon } from '~/domain/topical/mini-trend-tile';
+import { ArrowIconRight } from '~/components-styled/arrow-icon';
 import { getImageSrc } from '~/lib/sanity';
 import siteText from '~/locale';
 import { Article, Block, ImageBlock } from '~/types/cms';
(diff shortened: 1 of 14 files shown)


packages/app/src/components-styled/article-teaser.tsx
```diff
@@ -1,6 +1,6 @@
 import css from '@styled-system/css';
 import styled from 'styled-components';
-import ArrowIcon from '~/assets/arrow.svg';
+import { StyledArrowIcon } from '~/domain/topical/mini-trend-tile';
```
VWSCoronaDashboard4: Maybe name it `ArrowRightIcon`? I think there is also a place where there is an ArrowLeftIcon needed (back button in `article-detail`, used on Article page, and `editorial-detail` on Editorial page).

VWSCoronaDashboard4 (approved):

packages/app/src/components-styled/article-teaser.tsx
VWSCoronaDashboard: I forgot to push some changes, but it's named `ArrowIconRight` now.

VWSCoronaDashboard: I've added ArrowIconLeft too, good catch!

PR #1996, opened 2021-02-18
Fix webp support for background images

VWSCoronaDashboard10:
## Summary

Use an early feature detection in order to have the browser fetch webp resources instead of their old-fashioned counterpart.

### DIFF
--- a/packages/app/src/components-styled/background-image.tsx
+++ b/packages/app/src/components-styled/background-image.tsx
@@ -1,4 +1,3 @@
-import { maybeTransformImageUrlToWebp } from '~/lib/sanity';
 import styled, { css } from 'styled-components';
 import {
   backgroundPosition,
@@ -12,6 +11,7 @@ import {
   position,
   PositionProps,
 } from 'styled-system';
+import { maybeCreateWebpUrl } from '~/lib/sanity';
 import { styledShouldForwardProp } from '~/utils/styled-should-forward-prop';
 
 export type BackgroundImageLocalProps = {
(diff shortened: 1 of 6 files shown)


VWSCoronaDashboard1 (approved):

VWSCoronaDashboard:
An alternative to feature detection is to use `<picture>` for background images. We can use webp with png/jpg fallbacks just like we use in the articles. You can't use CSS for background images anymore, but I'm sure you can get good results with `object-fit: cover`

[Example of using a normal `` as a cover image.](https://codepen.io/robinrendle/pen/BywNVX)

VWSCoronaDashboard (approved):

VWSCoronaDashboard14 (approved):

packages/app/src/components-styled/background-image.tsx
```diff
@@ -26,10 +28,25 @@ export const BackgroundImage = styled.div.withConfig({
...
   backgroundRepeat,
   layout,
-  position
+  position,
+  (x) => {
+    const webpUrl = maybeCreateWebpUrl(x.backgroundImageUrl);
+    const prefix = x.backgroundImagePrefix ? `${x.backgroundImagePrefix},` : '';
+    const suffix = x.backgroundImageSuffix ? `,${x.backgroundImageSuffix}` : '';
+
+    return css`
```
VWSCoronaDashboard4, line 41: I think passing an object here, instead of using the templates string, could be less confusing than backtick-ascii-art.

The webp part could also be a separate variable.

VWSCoronaDashboard4 (approved):

PR #2020, opened 2021-02-19
Use TSDX to build the common package

VWSCoronaDashboard:
## Summary

Bundle the common package more efficiently with an out-of-the-box batteries included bundler for libraries.
Replaces this PR: https://github.com/minvws/nl-covid19-data-dashboard/pull/1900

## Motivation

- Out of the box good defaults for bundling
- Efficient bundling of lodash/lodash-es
- Prettier and linting built-in
- Watch mode
- React components are supported
- TypeScript support
- Limits the common package to 10KB max, enforceable though CI/CD in GitHub
- PostInstall script automatically builds a new dist whenever you install a new package

## Detailed design

Technically the change was rather simple.
- I've copied the old common folder to `common-old`, started a new one with `npx tsdx create common` and moved the src folder into the new common package.
- I had to remove absolute imports in common, but it might work again since I tweaked tsconfig after initial testing. Either way it should be easy to add to tsconfig again if we need it.
- I changed `yarn dev` to run `common` and `app` in watch mode instead of building common first before running Next in dev mode. I'm not sure if this will work well, but we can test it out?


VWSCoronaDashboard1 (approved):

VWSCoronaDashboard8:
I added a task to wrap this up to Trello. Closing it for now to keep the list clean

PR #2049, opened 2021-02-23
Update accessible link

VWSCoronaDashboard14:
There was some confusion on how to use the accessible from the Lokale. By adding a `{{source}}` in the Lokale they have more control and overview of what is happening. Also removed the visibility text and used an `aria-label` what gave me a little more control but with Voiceover the workings are the same.

### DIFF
--- a/packages/app/src/components-styled/content-header/metadata.tsx
+++ b/packages/app/src/components-styled/content-header/metadata.tsx
@@ -62,7 +62,7 @@ export function Metadata(props: MetadataProps) {
         icon={<DatabaseIcon aria-hidden />}
         items={dataSources}
         label={text.source}
-        accessibilityString={siteText.accessibility.link_source}
+        accessibilityText={siteText.accessibility.text_source}
         accessibilitySubject={accessibilitySubject}
       />
 
@@ -77,7 +77,7 @@ export function Metadata(props: MetadataProps) {
         }
         items={dataDownloads}
         label={text.download}
-        accessibilityString={siteText.accessibility.link_download}
+        accessibilityText={siteText.accessibility.text_download}
         accessibilitySubject={accessibilitySubject}
       />
     </Box>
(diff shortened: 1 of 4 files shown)


packages/app/src/locale/en.json
```diff
@@ -1954,8 +1954,8 @@
       "vaccin_levering_en_prikken": "The left-hand side of the graph shows how many vaccine doses have been delivered to the Netherlands and how many doses have been administered. The right-hand side of the graph shows how many vaccine doses we expect will be delivered and how many doses are to be administered in the coming weeks.",
       "vaccinatie_draagvlak": "This graph shows the percentage of the population that wishes to be vaccinated against COVID-19."
     },
-    "link_source": "View date source for \"{{subject}}\" at",
-    "link_download": "Download date source for \"{{subject}}\" at"
+    "link_source": "View date source for \"{{subject}}\" at {{source}}",
```
VWSCoronaDashboard4, line 1957: This needs to be adjusted in lokalise as well; existing keys won't re-import.

packages/app/src/components-styled/content-header/metadata.tsx
```diff
@@ -63,10 +62,8 @@ export function Metadata(props: MetadataProps) {
         icon={<DatabaseIcon aria-hidden />}
         items={dataSources}
         label={text.source}
-        accessibilityText={replaceVariablesInText(
-          siteText.accessibility.link_source,
-          { subject: accessibilitySubject }
-        )}
+        accessibilityString={siteText.accessibility.link_source}
```
VWSCoronaDashboard4, line 65: I'd keep `accessibilityText`; or perhaps `accessibilityLabel`

packages/app/src/components-styled/external-link.tsx
```diff
@@ -4,15 +4,22 @@ interface ExternalLinkProps {
...
+  ariaLabel?: string;
 }
 
-export function ExternalLink({ href, children, className }: ExternalLinkProps) {
+export function ExternalLink({
+  href,
+  children,
+  className,
+  ariaLabel,
+}: ExternalLinkProps) {
```
VWSCoronaDashboard4, line 15: Maybe an `assert` here, to check -when ariaLabel is provided- that it its length is greater than zero. With empty strings the links can completely disappear in certain software.

VWSCoronaDashboard1 (approved):

PR #2060, opened 2021-02-23
Add sewer message bar

VWSCoronaDashboard4:
## Summary

Add sewer message bar

## Detailed design

I have repurposed the `WarningTile` component to work with two variants. One `neutral` with white background and normal text. And `emphasis` that uses the yellow background with bold, large text.


VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/warning-tile.tsx
```diff
@@ -29,60 +46,72 @@ export function WarningTile({ message }: WarningMessageProps) {
...
+      borderBottomRightRadius: 1,
+      borderTopRightRadius: 1,
+      pr: 4,
+      '> *': {
+        mt: 0,
+        mb: 3,
+        ':last-child': {
+          mb: 0,
+        },
       },
```
VWSCoronaDashboard10, line 114: You might be able to extend the Box component (`const Children = styled(Box)`) and use the `spacing` prop instead

VWSCoronaDashboard10 (approved):

VWSCoronaDashboard8 (approved):

packages/app/src/components-styled/warning-tile.tsx
VWSCoronaDashboard4: I've given this a try, but I think this screams for a separate task/PR to clean up the markdown rendering and how we handle the styling of it.

PR #2066, opened 2021-02-23
Remove margin-top and on paragraph elements

VWSCoronaDashboard14:
Globally remove all the margin-top from the paragraph tag, went through **all** the pages to spot some difference in the layout compared to the live version. Numerous small changes that were relying on the margin-top from the paragraph tag.

### DIFF
--- a/packages/app/src/components-styled/kpi-tile.tsx
+++ b/packages/app/src/components-styled/kpi-tile.tsx
@@ -26,11 +26,11 @@ export function KpiTile({
       <Box>{children}</Box>
       {description && (
         <Box
-          as="p"
+          as="div"
           maxWidth="400px"
           fontSize={2}
           lineHeight={2}
-          mb={0}
+          mb={3}
           dangerouslySetInnerHTML={{
             __html: description,
           }}
--- a/packages/app/src/components-styled/kpi-tile.tsx
+++ b/packages/app/src/components-styled/kpi-tile.tsx
@@ -26,11 +26,11 @@ export function KpiTile({
       <Box>{children}</Box>
       {description && (
         <Box
-          as="div"
+          as="p"
           maxWidth="400px"
-          mt={3}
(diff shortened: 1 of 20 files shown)


packages/app/src/components-styled/kpi-tile.tsx
```diff
@@ -26,11 +26,11 @@ export function KpiTile({
       <Box>{children}</Box>
       {description && (
         <Box
-          as="div"
+          as="p"
```
VWSCoronaDashboard4, line 29: I don't think this is right. Markdown will output `<p>` 's as well, so you'll get p's in p's.

VWSCoronaDashboard4 (changes requested):

packages/app/src/style/global-style.tsx
```diff
@@ -115,4 +115,8 @@ button::-moz-focus-inner {
 figure {
   margin: 0
 }
+
+p {
+  margin-top: 0;
+}
```
VWSCoronaDashboard10, line 121: I'd say that it can be applied to all elements (`*`), though not sure what others think of that..

VWSCoronaDashboard8 (approved):
You are setting a margin-bottom on a lof of components internally. While using margin-bottom is better then using margin-top, it is still bad practice to include an outside margin setting on a reusable component. 

The parent context should take care of spacing different elements. 

I'm going to approve this, but I feel in some instances you have moved the problem to a different location instead of making a structural fix.

packages/app/src/style/global-style.tsx
VWSCoronaDashboard8: I would be ok with that but let's not blow up the scope of this PR. I already feel that some instances here are just moving the problem, because of setting an outside margin on what should be a reusable component.

VWSCoronaDashboard4 (approved):

PR #2075, opened 2021-02-24
Remove double anchors in article strip

VWSCoronaDashboard14:
In the article strip, there were nested anchors, moved to a span for the "Read more" and let the whole block still be clickable.

### DIFF
--- a/packages/app/src/components-styled/article-strip.tsx
+++ b/packages/app/src/components-styled/article-strip.tsx
@@ -1,6 +1,5 @@
 import css from '@styled-system/css';
 import styled from 'styled-components';
-import ArrowIcon from '~/assets/arrow.svg';
 import { Box } from '~/components-styled/base';
 import { Tile } from '~/components-styled/tile';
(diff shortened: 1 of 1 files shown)


packages/app/src/components-styled/article-strip.tsx
```diff
@@ -73,13 +73,10 @@ function ArticleStripItem(props: ArticleStripItemProps) {
           <Text mt={0} mb={2}>
             {title}
           </Text>
-          <LinkWithIcon
```
VWSCoronaDashboard10, line 76: what would you think of adjusting the LinkWithIcon API and make the `href` optional so we can render a span instead of an anchor when there's no `href`?

VWSCoronaDashboard8: If there is no href would you still consider it a link? LinkWithIcon with an optional link sounds confusing to me.

VWSCoronaDashboard8 (approved):

PR #2090, opened 2021-02-25
Add milestone vaccine page

VWSCoronaDashboard14:
Add milestones to the vaccine page.

Added a new schema
Added an object milestone 
Added the vaccine element


packages/app/src/domain/vaccine/milestones.tsx
```diff
@@ -0,0 +1,295 @@
...
+import { Heading, InlineText, Text } from '~/components-styled/typography';
+import { colors } from '~/style/theme';
+import { RichContentBlock } from '~/types/cms';
+import { formatDate } from '~/utils/formatDate';
+import siteText from '~/locale/index';
+
+const MAX_ITEMS_VISIBLE = 5;
+const CIRCLE_SIZE = 26;
+
+interface MileProp {
```
VWSCoronaDashboard4, line 17:
```suggestion
interface MileProps {
```

packages/app/src/domain/vaccine/milestones.tsx
```diff
@@ -0,0 +1,295 @@
...
+  title: string;
+  date: string;
+}
+
+interface ExpectProp {
```
VWSCoronaDashboard4, line 22:
```suggestion
interface ExpectProps {
```

packages/app/src/domain/vaccine/milestones.tsx
```diff
@@ -0,0 +1,295 @@
...
+    position: 'relative',
+    padding: 0,
+    border: 'none',
+    background: 'none',
+    font: 'inherit',
+    color: 'blue',
+    outline: 'inherit',
+    cursor: 'pointer',
+
+    '&:hover': {
```
VWSCoronaDashboard4, line 212: also add a `:focus` style

packages/app/src/domain/vaccine/milestones.tsx
```diff
@@ -0,0 +1,295 @@
...
+  css({
+    position: 'relative',
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+    width: CIRCLE_SIZE,
+    height: CIRCLE_SIZE,
+    minWidth: CIRCLE_SIZE,
+    borderRadius: '100%',
+    backgroundColor: x.isLast ? '#fff' : 'header',
```
VWSCoronaDashboard4, line 242: nit: there is a named color `white` in the theme too

packages/app/src/pages/landelijk/vaccinaties.tsx
```diff
@@ -33,14 +33,24 @@ import {
...
+import { MileStones } from '~/domain/vaccine/milestones';
 
 export const getStaticProps = createGetStaticProps(
   getLastGeneratedDate,
   getNlData,
   getText,
   createGetContent<{
-    articles?: ArticleSummary[];
-  }>(createPageArticlesQuery('vaccinationsPage'))
+    milestones: any;
```
VWSCoronaDashboard4, line 44: `any`

packages/app/src/queries/vaccine-page-query.ts
```diff
@@ -0,0 +1,15 @@
+import { targetLanguage } from '~/locale';
+
+export const vaccinePageQuery = `
+  *[_type=='vaccinationsPage']{
+    "title": title.${targetLanguage},
+      "description": description.${targetLanguage},
```
VWSCoronaDashboard4, line 6: nit: the indenting is a bit wonky here

packages/app/src/domain/vaccine/milestones.tsx
```diff
@@ -0,0 +1,295 @@
...
+  title: string;
+  description: RichContentBlock[];
+  miles: MileProp[];
+  expected: ExpectProp[];
+}
+
+export function MileStones(props: MileStoneProps) {
+  const { title, miles, description, expected } = props;
+
+  const [isExpanded, setIsExpanded] = useState(false);
```
VWSCoronaDashboard4, line 35: perhaps have it expanded by default, and then with `useEffect` set it to false? that way with JS turned off all content is accessibile

VWSCoronaDashboard4 (commented):
Some nitpicks :-)

Maybe some of the logic could be tidied up (in a different PR), but for now it's fine

packages/app/src/domain/vaccine/milestones.tsx
VWSCoronaDashboard8: MileProp?

packages/app/src/domain/vaccine/milestones.tsx
VWSCoronaDashboard8: "Conversation marked as resolved" maar ik zie nog geen code wijziging

packages/app/src/domain/vaccine/milestones.tsx
```diff
@@ -0,0 +1,295 @@
...
+
+interface MileProp {
+  title: string;
+  date: string;
+}
+
+interface ExpectProp {
+  item: string;
+}
+interface MileStoneProps {
```
VWSCoronaDashboard8: Milestone is 1 woord in het engels. Ik zou het component `MilestonesView` ofzo noemen, zodat je een `milestones` data field kunt hebben, want `mile` vind ik niet geschikt als naam.

packages/app/src/domain/vaccine/milestones.tsx
```diff
@@ -0,0 +1,295 @@
...
+  title: string;
+  description: RichContentBlock[];
+  miles: MileProp[];
+  expected: ExpectProp[];
```
VWSCoronaDashboard8: Dit zijn geen props naar types die onderdeel zijn van je component props. Ze moeten dus geen Prop suffix hebben.  `mile` en `expect` vind ik geen geschikte namen. Een mile is iets anders, en expected is te vaag.

packages/app/src/domain/vaccine/milestones.tsx
```diff
@@ -0,0 +1,295 @@
...
+}
+
+export function MileStones(props: MileStoneProps) {
+  const { title, miles, description, expected } = props;
+
+  const [isExpanded, setIsExpanded] = useState(false);
+
+  function expandedHandler() {
+    setIsExpanded(true);
+  }
```
VWSCoronaDashboard8, line 41: This wrapping function doesn't really serve a purpose and only obfuscates the code. You can use `() => setIsExpanded(true)` directly.

packages/app/src/domain/vaccine/milestones.tsx
```diff
@@ -0,0 +1,295 @@
...
+
+  return (
+    <Tile>
+      <Heading level={2} m={0}>
+        {title}
+      </Heading>
+      <Box maxWidth="maxWidthText" mb={2}>
```
VWSCoronaDashboard8, line 48: maxWidth is set to a string here

packages/app/src/domain/vaccine/milestones.tsx
```diff
@@ -0,0 +1,295 @@
...
+        <RichContent blocks={description} />
+      </Box>
+
+      <UnorderedList>
```
VWSCoronaDashboard8, line 52: No need to call this `UnorderedList` unless you have an `OrderedList` component in the same context and it has a meaning to the code. You can just call this `List` I think.

packages/app/src/domain/vaccine/milestones.tsx
```diff
@@ -0,0 +1,295 @@
...
+        <ListItem isFirst={true}>
```
VWSCoronaDashboard8, line 53: I think you should be able to target the first list item through normal CSS of the UnorderedList parent component.

packages/app/src/domain/vaccine/milestones.tsx
```diff
@@ -0,0 +1,295 @@
...
+            css={css({ transform: 'translateY(-6px)' })}
+          >
+            <Text fontWeight="bold" m={0}>
+              2021
+            </Text>
+          </Box>
+        </ListItem>
+
+        {miles.length > MAX_ITEMS_VISIBLE && !isExpanded && (
+          <ListItemButton>
```
VWSCoronaDashboard8, line 66: ListItemButton doesn't seem to be a button. It doesn't have a click handler and its children contain an actual button, so the name is confusing.

packages/app/src/domain/vaccine/milestones.tsx
```diff
@@ -0,0 +1,295 @@
...
+    </Tile>
+  );
+}
+
+const UnorderedList = styled.ul(
+  css({
+    listStyleType: 'none',
+    p: 0,
+    m: 0,
+    mb: 3,
```
VWSCoronaDashboard10, line 148: You can remove this bottom margin and instead wrap all `UnorderedLists` in a Box component with the spacing prop.

The advantage is that there won't be any outside margin (currently the case due to your last list with bottom margin) and that it's not necessary to manually set margins on other components like the `Text` inbetween the lists.

packages/app/src/domain/vaccine/milestones.tsx
VWSCoronaDashboard10: nit on the nit: there's no color named `white` in our theme. Instead it will simply use the color-value 'white', just like it would work with `hotpink` ;-)

packages/app/src/domain/vaccine/milestones.tsx
```diff
@@ -0,0 +1,295 @@
...
+                </CircleIcon>
+
+                <Box
+                  pl={CIRCLE_SIZE / 2}
+                  position="relative"
+                  maxWidth="maxWidthText"
+                  width="100%"
+                  zIndex={2}
+                >
+                  {index !== miles.length - 1 ? (
```
VWSCoronaDashboard8, line 93: You are repeating the same check 4 times here `index !== miles.length - 1 `.

Create a variable isLast so you only need to do it once. 

Also the property isLast on the individual components doesn't tell anything about what it does. Maybe you can name it after the effect it has on each of the components

packages/app/src/domain/vaccine/milestones.tsx
```diff
@@ -0,0 +1,295 @@
...
+                    <>
+                      <InlineText
+                        color={'gray'}
+                        css={css({ position: 'absolute', top: '-1.3rem' })}
+                      >
+                        {formatDate(new Date(mile.date))}
```
VWSCoronaDashboard8, line 99: if mile.date is not a Date don't call it that. If it is a unix timestamp in milliseconds I'd prefer `date_ms`. If it comes directly from the backend it should be `date_unix` and be in seconds, but that would mean there is also a conversion bug here since `new Date()` expects milliseconds.

packages/cms/src/schemas/objects/milestone.ts
```diff
@@ -0,0 +1,30 @@
...
+    },
+    { name: 'title', type: 'localeString', title: 'Titel' },
+  ],
+  preview: {
+    select: {
+      title: 'title.nl',
+      date: 'date'
+    },
+    prepare(selection: any) {
+      const {title, date} = selection
```
VWSCoronaDashboard10, line 23: Do you happen to know the types of `title` and `date`? If so, can you please type them?
```suggestion
    prepare(selection: { title: string; date: ??? }) {
      const {title, date} = selection
```

packages/app/src/domain/vaccine/milestones.tsx
```diff
@@ -0,0 +1,295 @@
...
+const UnorderedList = styled.ul(
+  css({
+    listStyleType: 'none',
+    p: 0,
+    m: 0,
+    mb: 3,
+  })
+);
+
+const ListItem = styled.li<{ isLast?: boolean; isFirst?: boolean }>((x) =>
```
VWSCoronaDashboard8, line 152: This looks messy with the isFirst boolean.

I think you better create two separate components if they are that far apart in styling. A `FirstItem` and a regular `Item` and switch between those in the parent context.

packages/app/src/pages/landelijk/vaccinaties.tsx
```diff
@@ -43,14 +43,24 @@ import { colors } from '~/style/theme';
...
-    articles?: ArticleSummary[];
-  }>(createPageArticlesQuery('vaccinationsPage'))
+    milestones: MileStoneProps;
+    highlight: {
+      articles?: ArticleSummary[];
+    };
+  }>(
+    `{
+      "milestones": ${vaccinePageQuery},
+      "highlight": ${createPageArticlesQuery('vaccinationsPage')}
```
VWSCoronaDashboard10, line 61: nice!

VWSCoronaDashboard10 (approved):

packages/app/src/domain/vaccine/milestones.tsx
```diff
@@ -0,0 +1,295 @@
...
+  css({
+    position: 'relative',
+    display: 'flex',
+    alignItems: x.isLast ? 'center' : 'flex-start',
+    paddingBottom: x.isFirst ? '2.5rem' : 4,
+    paddingTop: x.isLast ? `${1.3 + 2}rem` : '',
```
VWSCoronaDashboard8, line 158: I think this is the same as 3.3 rem? :)

packages/app/src/domain/vaccine/milestones.tsx
```diff
@@ -0,0 +1,295 @@
...
+      width: x.isLast ? '0' : '2px',
+      height: '100%',
+      backgroundColor: 'header',
+    },
+
+    '&::after': {
+      content: '""',
+      position: 'absolute',
+      bottom: '50%',
+      left: CIRCLE_SIZE / 2 - 1,
```
VWSCoronaDashboard8, line 176: You keep repeating this `CIRCLE_SIZE / 2 - 1`. Better create a variable from it, so you only need to do this calculation once

packages/app/src/domain/vaccine/milestones.tsx
```diff
@@ -0,0 +1,295 @@
...
+
+const CircleIcon = styled.div<{ isLast: boolean }>((x) =>
+  css({
+    position: 'relative',
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+    width: CIRCLE_SIZE,
+    height: CIRCLE_SIZE,
+    minWidth: CIRCLE_SIZE,
```
VWSCoronaDashboard8, line 242: I think if you set a width and height minwidth doesn't really have a function

packages/app/src/domain/vaccine/milestones.tsx
```diff
@@ -0,0 +1,295 @@
...
+    borderRadius: '100%',
+    backgroundColor: 'header',
+  })
+);
+
+const Background = styled.div(
+  css({
+    position: 'absolute',
+    top: 0,
+    left: '-2rem',
```
VWSCoronaDashboard8, line 272: 2rem and 4rem spacings are part of the theme. Try to use more theme spacing sizes because the CSS in these components seems to be pretty much completely detached from the theme, which increases technical dept.

packages/cms/src/schemas/documents/pages/vaccinations-page.ts
```diff
@@ -4,5 +4,37 @@ export default {
...
+      title: 'Mijlpalen',
+      description:
+        'Je kan hier mijlpalen toevoegen, de laatste zal uitgelicht worden. Ze worden op het dashboard op datum geordend van oud naar nieuw.',
+      of: [{ type: 'milestone', direction: 'asc' }],
+      options: {
+        sortable: false
+      }
+    },
+    {
+      name: 'expected',
```
VWSCoronaDashboard8, line 30: Expected is too vague. Since it is really 'expected milestones' why not call it that?  `expected_milestones`

VWSCoronaDashboard8 (changes requested):

packages/app/src/domain/vaccine/milestones.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard14: Sanity is giving me a string, I could also accept strings in the formatDate function or parse it to a new Date like here.

packages/app/src/domain/vaccine/milestones.tsx
VWSCoronaDashboard14: I've set the min-width because when the screen gets smaller the circles got squashed by the flexbox, this prevents it.

packages/app/src/domain/vaccine/milestones.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard8: OK then it's fine if it's an ISO date string. I was assuming this was a number.

packages/app/src/domain/vaccine/milestones.tsx
VWSCoronaDashboard8: I suspect you can also get the same result by setting flex shrink to 0, but this is fine then.

packages/app/src/domain/vaccine/milestones.tsx
VWSCoronaDashboard8: Later on I concluded that looking at the CSS it would be better to create separate components for first and other.

packages/app/src/domain/vaccine/milestones.tsx
VWSCoronaDashboard8: correction; three different components. A fist, a last and a regular in-between styled version.

VWSCoronaDashboard14: @VWSCoronaDashboard8 Totally agree, gonna do that in a separate pull request. This code kinda grew along the way I was working on it.

packages/app/src/domain/vaccine/milestones-view.tsx
```diff
@@ -13,32 +13,31 @@ import siteText from '~/locale/index';
 
 const MAX_ITEMS_VISIBLE = 5;
 const CIRCLE_SIZE = 26;
+const CIRCLE_HALF = CIRCLE_SIZE / 2;
 
-interface MileProps {
+type MilestonesType = {
```
VWSCoronaDashboard8, line 18: No need to place a Type suffix on a type. The fact that it is declared as a type and starts with a capitol latter should be enough.

packages/app/src/domain/vaccine/milestones-view.tsx
```diff
@@ -13,32 +13,31 @@ import siteText from '~/locale/index';
...
-}
-export interface MileStoneProps {
+};
+export interface MilestoneViewProps {
   title: string;
   description: RichContentBlock[];
-  miles: MileProps[];
-  expected: ExpectProps[];
+  milestones: MilestonesType[];
+  expectedMilestones: ExpectedMilestonesType[];
```
VWSCoronaDashboard8, line 30:
```suggestion
  milestones: Milestones[];
  expectedMilestones: ExpectedMilestones[];
```

packages/app/src/domain/vaccine/milestones-view.tsx
```diff
@@ -13,32 +13,31 @@ import siteText from '~/locale/index';
...
+  const { title, milestones, description, expectedMilestones } = props;
 
   const [isExpanded, setIsExpanded] = useState(true);
 
   useEffect(() => setIsExpanded(false), []);
 
-  function expandedHandler() {
-    setIsExpanded(true);
-  }
+  const expandedHandler = () => setIsExpanded(true);
```
VWSCoronaDashboard8, line 40: If you're not removing the wrapper function then maybe give it a name like a handler usually has with handle + an event type like `handleExpand`. Function names in general should start with a verb.

packages/app/src/domain/vaccine/milestones-view.tsx
```diff
@@ -116,54 +113,50 @@ export function MileStones(props: MileStoneProps) {
...
-        {expected.map((item, index) => (
-          <ExpectedListItem key={index}>{item.item}</ExpectedListItem>
-        ))}
-      </UnorderedList>
+      </Box>
+      <Box pl={CIRCLE_SIZE * 1.5}>
+        <Text color="gray" m={0} pl={3} mb={2}>
+          {siteText.milestones.verwacht}
+        </Text>
+        <Box as="ul" p={0} m={0} css={css({ listStyleType: 'none' })}>
```
VWSCoronaDashboard8, line 131: Actually what you are rendering is more like an ordered list, since milestones are ordered in time. So to be semantically correct I think this should be `ol`

VWSCoronaDashboard8 (approved):

PR #2096, opened 2021-02-26
Fix preview deployments by exporting pages as `/{pagename}/index.html`

VWSCoronaDashboard10:
## Summary
Our static preview environment cannot handle requests like `/{pagename}` which should resolve to `{pagename}.html`.
The `trailingSlash: true` will export the pages as `/{pagename}/index.html`.

### DIFF
--- a/packages/app/next.config.js
+++ b/packages/app/next.config.js
@@ -20,7 +20,21 @@ const nextConfig = {
   env: {
     COMMIT_ID: commitHash,
   },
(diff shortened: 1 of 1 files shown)


VWSCoronaDashboard (changes requested):
I don't want this merged without being conditional for deploy previews only. That's why I'm requesting changes to block this PR :).

janhoogeveen (approved):

PR #2101, opened 2021-02-26
Implement sewer chart animations

VWSCoronaDashboard10:
## Summary

Implement an animating sewer chart. Animations are disabled for IE11 and people who prefer reduced motion.


packages/app/src/components-styled/sewer-chart/sewer-chart.tsx
```diff
@@ -181,45 +188,62 @@ export function SewerChart(props: SewerChartProps) {
...
-                getX={scales.getX}
-                getY={() => 26}
-                color="rgba(89, 89, 89, 0.8)"
-                radius={4}
-                dottedOutline
-              />
-            </Group>
-          )}
+          <ScatterPlot
+            data={uniqWith(outlierValues, (val1, val2) => {
```
VWSCoronaDashboard8, line 215: I'm surprised you didn't wrap this in a useMemo 🙂 Is that because there are never many outliers anyway?

packages/app/src/utils/use-is-ie11.ts
```diff
@@ -0,0 +1,14 @@
+import { useEffect, useState } from 'react';
+
+/**
+ * Detect Trident layout engine used by IE11
+ */
+export function useIsIE11(initialValue = false) {
```
VWSCoronaDashboard8, line 6: Nice 👍  But I believe we are also camel-casing acronyms like they are words, so I would suggest to write this as `useIsIe11` but I must admit that looks butt-ugly, and the worst match for camel-casing I came across so far. 
An alternative would be `useIsInternetExplorer11`. Eeeh 😕  well fuk it. Do whatever you feel is right.

packages/app/src/utils/use-is-mounted.ts
```diff
@@ -1,7 +1,23 @@
 import { useEffect, useState } from 'react';
+import { useIsMountedRef } from './use-is-mounted-ref';
 
-export function useIsMounted() {
+/**
+ * @param delayMs optionally mutate state after a delay
+ */
+export function useIsMounted({ delay }: { delay?: number } = {}) {
```
VWSCoronaDashboard8, line 7: I think I liked delayMs better :) I assume you had that before because it's still in the comment above.

packages/app/src/utils/use-path-morph.tsx
```diff
@@ -0,0 +1,32 @@
...
+): MotionValue<string> {
+  const value = useMotionValue<string>(d);
+  const configRef = useRef(config);
+  configRef.current = config;
+
+  useEffect(() => {
+    const interpolator = interpolatePath(value.get(), d);
+
+    animate(0, 1, {
+      ...configRef.current,
```
VWSCoronaDashboard8, line 26: I don't quite follow why you are handling config in this way. Maybe add a comment.

VWSCoronaDashboard8 (approved):

packages/app/src/components-styled/sewer-chart/sewer-chart.tsx
```diff
@@ -181,45 +188,62 @@ export function SewerChart(props: SewerChartProps) {
...
+              /**
+               * filter circles which are close to each other, no need to
+               * render them on top of each other.
+               */
+              const threshold = 4; // 10px
```
VWSCoronaDashboard10, line 220:
```suggestion
              const threshold = 4; // 4px
```

packages/app/src/components-styled/sewer-chart/sewer-chart.tsx
VWSCoronaDashboard10: Ah yes, I'll do that. I think it won't make a lot of difference though, but better be safe than sorry.

packages/app/src/utils/use-is-mounted.ts
VWSCoronaDashboard10: Ah you got me there, sure 👍

packages/app/src/utils/use-path-morph.tsx
VWSCoronaDashboard10: Hmm, Now I realise it's also buggy, the effect won't be triggered on config updates.

Initially I wanted to prevent triggering the effect on every render when someone calls the hook with:
```tsx
usePathMorph(path, { ... })
```
because that would pass a new object on every render, thus calling the effect on every render.

I don't think I can safely serialize that config object, so we'll just have to be careful and use a `useMemo` hook in the calling context.

packages/app/src/utils/use-is-ie11.ts
0x80: Good find 😄

VWSCoronaDashboard8 (approved):

packages/app/src/utils/use-is-ie11.ts
VWSCoronaDashboard1: Or rename it to `isBadBrowser`?

VWSCoronaDashboard1 (approved):

PR #2114, opened 2021-02-28
Update milestones and simplify it

VWSCoronaDashboard14:
Wasn't happy with the milestones and all the checks in the markup tried to simplify it.
- Removed overlapping CSS
- Removed complicated inline logic (except for the checks if it falls between the max items and if is it expanded)
- Fixed the safari visual bug of the expander dots on the left
- Now in the last item, the Icon is aligned with the height of the first sentence

Also the `date` key is not a unix timestamp or something it is just a string like `08-02-2021`.

**Still to do:**
Update the schema names, but since it is live I don't wanna risk it to lose content.

### DIFF
--- a/packages/app/src/domain/vaccine/milestones-view.tsx
+++ b/packages/app/src/domain/vaccine/milestones-view.tsx
@@ -10,24 +10,25 @@ import { colors } from '~/style/theme';
 import { RichContentBlock } from '~/types/cms';
 import { formatDate } from '~/utils/formatDate';
 import siteText from '~/locale/index';
+import { asResponsiveArray } from '~/style/utils';
 
 const MAX_ITEMS_VISIBLE = 5;
 const CIRCLE_SIZE = 26;
 const CIRCLE_HALF = CIRCLE_SIZE / 2;
 
-type MilestonesType = {
+type Milestones = {
   title: string;
   date: string;
 };
 
-type ExpectedMilestonesType = {
+type ExpectedMilestones = {
   item: string;
 };
 export interface MilestoneViewProps {
   title: string;
   description: RichContentBlock[];
-  milestones: MilestonesType[];
-  expectedMilestones: ExpectedMilestonesType[];
+  milestones: Milestones[];
+  expectedMilestones: ExpectedMilestones[];
 }
 
 export function MilestonesView(props: MilestoneViewProps) {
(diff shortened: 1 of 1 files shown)


packages/app/src/domain/vaccine/milestones-view.tsx
```diff
@@ -115,127 +121,101 @@ export function MilestonesView(props: MilestoneViewProps) {
...
-          ))}
+
+      {expectedMilestones && (
+        <Box pl={`calc(1rem + ${CIRCLE_SIZE}px)`}>
+          <Text color="gray" m={0} pl={3} mb={2}>
+            {siteText.milestones.verwacht}
+          </Text>
+          <Box as="ul" p={0} m={0} css={css({ listStyleType: 'none' })}>
+            {expectedMilestones.map((item, index) => (
+              <ExpectedListItem key={index}>{item.item}</ExpectedListItem>
```
VWSCoronaDashboard8, line 140: Nit; this is usually a good indication that the variable should have a different name. You could name it `milestone` or just `x` as we often times do with map and filter.

packages/app/src/domain/vaccine/milestones-view.tsx
```diff
@@ -115,127 +121,101 @@ export function MilestonesView(props: MilestoneViewProps) {
...
-    '&::after': {
-      content: '""',
-      position: 'absolute',
-      bottom: '50%',
-      left: CIRCLE_HALF - 1,
-      width: x.isLast ? '2px' : '0',
-      height: x.isLast ? '50%' : '100%',
-      backgroundColor: 'white',
-    },
+const ListItem = styled.li(css({ ...ListItemStyles }));
```
VWSCoronaDashboard8, line 165:
```suggestion
const ListItem = styled.li(css(ListItemStyles));
```

VWSCoronaDashboard8: Also `ListItemStyles` doesn't seem to be correct, as the capital letter would indicate that it's either a component or a type. I would suggest calling it `commonListItemStyles`, but anything with a lower-case letter should do

VWSCoronaDashboard8 (changes requested):
Nice cleanup!

VWSCoronaDashboard8 (approved):

PR #2123, opened 2021-03-01
Yet another chart in Visx

VWSCoronaDashboard8:
## Summary

This work is combining lessons from previous charts with ideas that were still floating around. Its main goal was simplify and and improve abstractions as much as possible so that we can have a clear common ground to build other charts. In addition it is extending flexibility so that multiple types of similar charts could be merged into one.

This is a big refactor of MultiLineChart, which was by itself a fork of LineChart. 

In its current state: 

- It is already a drop-in replacement of MultiLineChart.
- It is a long way towards being a replacement for LineChart (mainly benchmark line and date span annotations still need to be added)
- It is a long way towards being a replacement for "highcharts" charts that use range data (like sterftemonitor)
- It hopefully lays the groundwork for more complex charts too, like the vaccine delivery chart.

## Main takeaways

- The `seriesConfig` defines what data to render and how that data should be used. It can handle both single value data (line and area trend) and double value (range trend)
- The tooltip formatting function gets its data in a standardized way that gives it maximum flexibility. The original data, plus the seriesConfig plus the metric property that is nearest. This allows us to define default tooltip layout that is usable for more than simple 1 line charts. (See screenshot below)
- Sub components like markers and trends are smaller and more decoupled.
- Most logic is tucked away a few re-usable hooks. The tooltip logic is decoupled from the hover state calculation.
- It uses linear scales for both x and y. The unix timestamps from the data are used as-is without conversion to/from Date objects.

This is the chart as a replacement of the current MultiLineChart used for vaccine support:

image

Here's what a botched version of the "vaccine support" looks like when configured differently in `seriesConfig` to show line, area and range trends. Also the tooltip here is the default / data-driven based on seriesConfig.

image

The botched version of vaccine_support is still in the code for illustration purposes. The idea is to take it out before this gets merged in develop. That's why I placed a "hold" label on this PR.


packages/app/src/components-styled/time-series-chart/time-series-chart.tsx
```diff
@@ -0,0 +1,325 @@
...
+ *
+ * Known Issues:
+ * - Nearest point / tooltip valueKey calculation seems to be off by some
+ *   margin. This is currently not used in any charts but would be nice to solve
+ *   at some point.
+ */
+export type TimeSeriesChartProps<T extends TimestampedValue> = {
+  title: string; // Used for default tooltip formatting
+  values: T[];
+  seriesConfig: SeriesConfig<T>;
```
VWSCoronaDashboard1, line 89: This does assume that all of the series originates from the same values list where each value has multiple properties that can be referenced by the series config.
What if it needs to display trends that come from multiple values collections? We could require the data to be shaped into a unified value object, but that'll get us in trouble when property names overlap.
In the area-chart I tried to solve this by merging the values and series config into one object:

`type Series<T> = {
  values: T[];
  seriesConfig: SeriesConfig<T>;
}`

That was you can pass in multiple Series instances.

VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/time-series-chart/time-series-chart.tsx
VWSCoronaDashboard8: Yes, I have assumed in this chart that those series can be merged into one type as if it was supplied from the same schema. Setting `null` for properties that do not span the whole series. 

I don't think we should make the chart deal with different kinds of data in multiple formats, because that get's very complex very quickly, and 98% of use cases wouldn't need that functionality.

I think it's safer to manually pick and choose what needs to be displayed and format the data so that the chart can handle it in a simplified way. We can deal with that complexity before hitting the chart I think.

packages/app/src/components-styled/time-series-chart/components/area-trend.tsx
```diff
@@ -0,0 +1,67 @@
...
+}: AreaTrendProps) {
+  const [isHovered, setIsHovered] = useState(false);
+
+  const handleHover = (
+    event: TouchEvent<SVGElement> | MouseEvent<SVGElement>
+  ) => {
+    const isLeave = event.type === 'mouseleave';
+    setIsHovered(!isLeave);
+    onHover(event);
+  };
```
VWSCoronaDashboard10, line 36: I'm not sure what magic is happening behind the scenes of AreaClosed and LinePath, but caching `(useCallback) the handleHover makes sure no unneeded (and potential expensive) re-renders happen.

packages/app/src/components-styled/time-series-chart/components/chart-container.tsx
```diff
@@ -0,0 +1,43 @@
...
+  padding,
+  ariaLabelledBy,
+  children,
+}: ChartContainerProps) {
+  return (
+    <svg
+      width={width}
+      height={height}
+      role="img"
+      aria-labelledby={ariaLabelledBy}
```
VWSCoronaDashboard10, line 36: adding `style={{ touchAction: 'pan-y' }}` allows the user to "move" their finger horizontally over the chart, without vertically scrolling the page.

packages/app/src/components-styled/time-series-chart/components/date-line-marker.tsx
```diff
@@ -0,0 +1,86 @@
...
+  border-left-style: dashed;
+  border-left-color: ${(props) => props.color};
+`;
+
+const Container = styled.div`
+  position: absolute;
+  pointer-events: none;
+  top: 0;
+  bottom: 0;
+`;
```
VWSCoronaDashboard10, line 45: Why aren't we using styled-system's css() helper in this file?

packages/app/src/components-styled/time-series-chart/components/date-span-marker.tsx
```diff
@@ -0,0 +1,38 @@
...
+  width: 100%;
+  background-color: rgba(0, 0, 0, 0.03);
+`;
+
+const Container = styled.div`
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  transform: translate(-50%, 0);
+`;
```
VWSCoronaDashboard10, line 16: Same here, I thought we didn't use the styled-components ``` styled.x`` ``` syntax.

packages/app/src/components-styled/time-series-chart/components/line-trend.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const [isHovered, setIsHovered] = useState(false);
+
+  const handleHover = (
+    event: TouchEvent<SVGElement> | MouseEvent<SVGElement>
+  ) => {
+    const isLeave = event.type === 'mouseleave';
+    setIsHovered(!isLeave);
+    onHover(event);
+  };
+
```
VWSCoronaDashboard10, line 35: same here about caching the callback

packages/app/src/components-styled/time-series-chart/components/line-trend.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const strokeDasharray = style === 'dashed' ? 4 : undefined;
+
+  return (
+    <LinePath
+      style={{ pointerEvents: 'all' }}
```
VWSCoronaDashboard10, line 40: hmm but in that case we should also cache objects like these .. the potentially premature optimisation is possibly not worth the hassle.

packages/app/src/components-styled/time-series-chart/components/series.tsx
```diff
@@ -0,0 +1,103 @@
...
+  onHover: HoverHandler;
+  seriesConfig: SeriesConfig<T>;
+  seriesList: SeriesList;
+  getX: GetX;
+  /**
+   * @TODO it's maybe not worth it creating the getY functions in the hook and
+   * passing them along. Since we also need the yScale here anyway, we might as
+   * well let each component make its own y getters based on yScale. GetX is
+   * used in more places and is more universal so that might still be worth
+   * passing along instead of xScale.
```
VWSCoronaDashboard10, line 28: Have you looked into creating a context with all these getters instead?

packages/app/src/components-styled/time-series-chart/logic/dimensions.ts
```diff
@@ -0,0 +1,26 @@
...
+    ...defaultPadding,
+    left: paddingLeft ?? defaultPadding.left,
+  };
+
+  const bounds: Bounds = {
+    width: width - padding.left - padding.right,
+    height: height - padding.top - padding.bottom,
+  };
+
+  return { padding, bounds };
```
VWSCoronaDashboard10, line 25: I think you'd want to memoize the bounds (and possibly also the padding), because you're passing these objects on meomoized components (for example Axis). This will cost you performance instead of winning it (it will compare all props on every render, though we know upfront they are differnet on every render).

packages/app/src/components-styled/time-series-chart/logic/legend.ts
```diff
@@ -0,0 +1,67 @@
...
+            ? transparentize(0.7, annotation.color)
+            : colors.data.emphasis,
+          label: annotation.label,
+          shape: 'square',
+        } as LegendItem);
+      }
+    }
+
+    /**
+     * In current charts we only show a legend if there are more then 1 items. So
```
VWSCoronaDashboard10, line 56:
```suggestion
     * In current charts we only show a legend if there are more than 1 items. So
```

packages/app/src/domain/deceased/deceased-monitor-section.tsx
```diff
@@ -39,40 +35,40 @@ export function DeceasedMonitorSection({
         title={text.deceased_monitor_chart_title}
         description={text.deceased_monitor_chart_description}
       >
-        <DeceasedMonitor
+        {/* <DeceasedMonitor
```
VWSCoronaDashboard10, line 38: comment intentional?

packages/app/src/pages/landelijk/vaccinaties.tsx
```diff
@@ -66,11 +65,11 @@ export const getStaticProps = createGetStaticProps(
 );
 
 const VaccinationPage: FCWithLayout<typeof getStaticProps> = ({
-  text: siteText,
+  text: locale,
```
VWSCoronaDashboard10, line 68: we use `siteText` everywhere we're using the "global" text shape, why going for `locale` here?

VWSCoronaDashboard10 (approved):
nice work!

packages/app/src/components-styled/time-series-chart/components/date-line-marker.tsx
VWSCoronaDashboard8: I inherited some of this from line chart components, but to be honest I wasn't so aware anymore about if and when we use this type of syntax. I'm ok trying to stick with one approach. I am personally still tempted to use plain SC syntax when I create components that have no theming dependencies, but it's probably better to use SS anyway in case you want to add in some theme related variables later.

packages/app/src/components-styled/time-series-chart/components/line-trend.tsx
VWSCoronaDashboard8: This was also inherited. I have now removed the style objects as it didn't seem to make any difference...

packages/app/src/components-styled/time-series-chart/components/series.tsx
VWSCoronaDashboard8: Might be an interesting idea especially once we start using more of these "trend" components. I'll leave that for a future exercise. I'm sure we'll find more of these type of improvements  once we start expanding these components.

packages/app/src/components-styled/time-series-chart/logic/dimensions.ts
VWSCoronaDashboard8: Good catch 👍

packages/app/src/pages/landelijk/vaccinaties.tsx
VWSCoronaDashboard8: You're right I should keep it consistent. I am not fond of the word siteText, especially since there's usually another variable named "text" that comes from it. I would prefer locale but that's a kind of rename that should be done app-wide.

PR #2128, opened 2021-03-02
Create vaccine progress schema's

VWSCoronaDashboard10:
## Summary

Schema's for the vaccination progress.

### DIFF
--- a/packages/app/schema/nl/vaccine_administered_daily.json
+++ b/packages/app/schema/nl/vaccine_administered_daily.json
@@ -1,0 +1,53 @@
+{
+  "definitions": {
+    "value": {
+      "title": "nl_vaccine_administered_daily_value",
+      "type": "object",
+      "required": [
+        "administrations_per_second",
+        "seconds_per_administration",
+        "moving_average",
+        "date_start_unix",
+        "date_end_unix",
+        "date_of_insertion_unix"
+      ],
+      "additionalProperties": false,
+      "properties": {
+        "administrations_per_second": {
+          "type": "number"
+        },
+        "seconds_per_administration": {
+          "type": "number"
+        },
+        "moving_average": {
+          "type": "integer"
+        },
+        "date_start_unix": {
+          "type": "integer"
+        },
+        "date_end_unix": {
+          "type": "integer"
+        },
+        "date_of_insertion_unix": {
+          "type": "integer"
+        }
+      }
+    }
+  },
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "nl_vaccine_administered_daily",
+  "type": "object",
+  "required": ["values", "last_value"],
+  "additionalProperties": false,
+  "properties": {
+    "values": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/value"
+      }
+    },
+    "last_value": {
+      "$ref": "#/definitions/value"
+    }
+  }
+}
--- a/packages/app/schema/nl/vaccine_administered_daily.json
+++ b/packages/app/schema/nl/vaccine_administered_daily.json
@@ -19,7 +19,7 @@
         "seconds_per_administration": {
           "type": "number"
         },
-        "moving_average": {
+        "administrations_per_week_moving_average": {
           "type": "integer"
         },
         "date_start_unix": {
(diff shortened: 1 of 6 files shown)


packages/app/schema/nl/vaccine_administered_daily.json
```diff
@@ -0,0 +1,53 @@
...
+      ],
+      "additionalProperties": false,
+      "properties": {
+        "administrations_per_second": {
+          "type": "number"
+        },
+        "seconds_per_administration": {
+          "type": "number"
+        },
+        "moving_average": {
```
VWSCoronaDashboard4, line 22: there is `administrations_per_second`, and `administrations_per_week`; I'd then also name this `administrations_moving_average`

VWSCoronaDashboard4 (approved):
1 opinion added

packages/app/schema/nl/vaccine_administered_daily.json
VWSCoronaDashboard10: Makes sense, though I named it `moving_average` because the subject should already be clear from the schema name (and I could not apply that logic to the `x_per_y` and `y_per_x` properties). But I don't have a strong opinion about it, so I'll change it :-)

VWSCoronaDashboard1 (approved):

packages/app/schema/nl/vaccine_administered_planned.json
```diff
@@ -0,0 +1,45 @@
...
+      "required": [
+        "administrations_per_week",
+        "date_start_unix",
+        "date_end_unix",
+        "date_of_insertion_unix"
+      ],
+      "additionalProperties": false,
+      "properties": {
+        "administrations_per_week": {
+          "type": "number"
```
VWSCoronaDashboard8, line 15: If this is not an average, I think it should be an integer.

packages/app/schema/nl/vaccine_administered_daily.json
```diff
@@ -0,0 +1,53 @@
...
+          "type": "integer"
```
VWSCoronaDashboard8, line 23: This is an average, so I would expect this to be type number

packages/app/schema/nl/vaccine_administered_daily.json
VWSCoronaDashboard10: I've made it more clear by naming it `administrations_per_week_moving_average`, even better imho.

VWSCoronaDashboard8 (changes requested):

packages/app/schema/nl/vaccine_administered_daily.json
```diff
@@ -0,0 +1,53 @@
+{
+  "definitions": {
+    "value": {
+      "title": "nl_vaccine_administered_daily_value",
+      "type": "object",
+      "required": [
+        "administrations_per_second",
+        "seconds_per_administration",
```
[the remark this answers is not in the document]
VWSCoronaDashboard10: > Isn't this redundant / two versions of the same number? Or are we not allowed to do y = (1 / x) in the FE 😄

yep, we want as few client-side calculations  as possible.

packages/app/schema/nl/vaccine_administered_rate_moving_average.json
```diff
@@ -0,0 +1,53 @@
+{
+  "definitions": {
+    "value": {
+      "title": "nl_vaccine_administered_rate_value",
+      "type": "object",
+      "required": [
+        "doses_per_second",
+        "seconds_per_dose",
+        "doses_sum",
```
0x80, line 9: This is not the sum for the total time period (date_start/end), but the daily average for that period. So I'd remove the word `sum` and just use `doses`

packages/app/schema/nl/vaccine_administered_rate_moving_average.json
```diff
@@ -0,0 +1,53 @@
...
+      "additionalProperties": false,
+      "properties": {
+        "doses_per_second": {
+          "type": "number"
+        },
+        "seconds_per_dose": {
+          "type": "number"
+        },
+        "doses_sum": {
+          "type": "integer"
```
0x80, line 23: All averages should be floating point, even if we display them as integer in the interface.

0x80 (changes requested):

packages/app/schema/nl/vaccine_administered_rate_moving_average.json
VWSCoronaDashboard10: in relation to `doses_per_second`, shouldn't we name it `doses_per_day` instead? Or do we regularly have `x` and `x_per_y`?

PR #2132, opened 2021-03-02
Fix webp background images resulting in duplicate requests

VWSCoronaDashboard10:
## Summary

The current webp implementation results in duplicate requests for a single background image due to webp-detection kicking in after the css has been parsed.

~~We'll disable it for now and dive into another approach.~~

I've reimplemented the background image using a combination of the picture element with the css `object-fit` property. I've also added a quality parameter to the sanity cnd requests, 65% seems to look good enough to me.


VWSCoronaDashboard4 (approved):

packages/app/public/init.js
```diff
@@ -8,20 +8,6 @@
...
-    var webp = new Image();
-    webp.onload = function () {
-      window.document.documentElement.classList.add('has-webp-support');
-    };
-    webp.src =
-      'data:image/webp;base64,UklGRjIAAABXRUJQVlA4ICYAAACyAgCdASoBAAEALmk0mk0iIiIiIgBoSygABc6zbAAA/v56QAAAAA==';
-  } catch (err) {
-    /** */
-  }
-
```
VWSCoronaDashboard8, line 24: This wasn't useful for when we later want to support webp again? It doesn't seem to hurt performance.

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard1 (approved):

packages/app/public/init.js
VWSCoronaDashboard10: We still have it in source control when we would like to use it in the future :-)

packages/app/src/components-styled/highlight-teaser.tsx
```diff
@@ -20,13 +18,25 @@ export type HighlightTeaserProps = {
   };
 };
 
+const articleTeaserImageSizes = [
+  [0, 733],
+  [768, 445],
+  [960, 406],
+];
+
```
VWSCoronaDashboard13, line 26: for my understanding, how are you choosing these image sizes?

VWSCoronaDashboard13 (approved):

packages/app/src/components-styled/highlight-teaser.tsx
VWSCoronaDashboard10: It's indeed a bit vague, but it's map of viewport-width => image-width. In this case the following will happen:

- a viewport width of `0px` to `768px` will use an image width a width of `733px`
- a viewport width of `768px` to `960px` will use an image width a width of `445px`
- a viewport width of `960px` and up will use an image width of `406px`

The image widths have been determined by resizing the window and see how the images behave on certain breakpoints. There's probably a smarter way to achieve this, but for now it does the trick.

VWSCoronaDashboard10:
I've put this PR on hold because I think there are still a couple of bugs. Will pick it up when there's more time.

packages/app/src/components-styled/highlight-teaser.tsx
```diff
@@ -48,6 +55,24 @@ export function HighlightTeaser(props: HighlightTeaserProps) {
   );
 }
 
+const ZoomContainer = styled(ZoomContainerUnstyled)``;
```
VWSCoronaDashboard4, line 58: This could be a component, and then imported on the various places it is used

VWSCoronaDashboard4 (approved):

VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/highlight-teaser.tsx
VWSCoronaDashboard10: Hmm, I don't think it would make sense to have this as a separate component in our component directory.. I guess it's a bit too specific for the situation its currently used.

packages/app/src/components-styled/article-teaser.tsx
```diff
@@ -49,6 +52,24 @@ export function ArticleTeaser(props: ArticleTeaserProps) {
   );
 }
 
+const ZoomContainer = styled(ZoomContainerUnstyled)``;
```
VWSCoronaDashboard8, line 55: Is this a left-over?

packages/app/src/components-styled/cms/sanity-image.tsx
```diff
@@ -2,22 +2,28 @@ import { Image } from '~/components-styled/image';
 
 type SanityImageProps = {
   src: string;
-  srcSet?: string;
   extension: string;
+  srcSet?: string;
+  sizes?: string;
+  className?: string;
```
VWSCoronaDashboard8, line 8: Since this class name has a very specific target (the underlying picture element) I think it would be good to have that in the name. For example `pictureElementClassName` or just `pictureClassName`. Because from the outside/public API a few levels higher you have no idea where the class will end up.

packages/app/src/components-styled/highlight-teaser.tsx
```diff
@@ -84,29 +109,3 @@ function Arrow() {
...
-        height={height}
-        backgroundImageUrl={url}
-        backgroundPosition={bgPosition}
-        backgroundRepeat="no-repeat"
-        backgroundSize="cover"
-        aria-label={image.alt}
-      />
-    </Box>
-  );
-}
```
VWSCoronaDashboard8, line 112: I feel like I'm reading a lot of code for the second time. Maybe add a @TODO to clean things up by sharing some code.

VWSCoronaDashboard8 (approved):

packages/app/src/components-styled/article-teaser.tsx
VWSCoronaDashboard10: Nope, it's used. What made you think it's a left-over?

packages/app/src/components-styled/cms/sanity-image.tsx
VWSCoronaDashboard10: I understand the rationale, but in this case I'll stick to the className prop .. it was previously already handling className props, it only wasn't properly typed (className was used as `styled(SanityImage)` which passes on a className to the wrapped component, TS wouldn't complain either).

packages/app/src/components-styled/article-teaser.tsx
VWSCoronaDashboard8: Does the styled wrapper do anything functionally? It seemed to be just passing along the component as-is.

VWSCoronaDashboard10: Ah, yes it does. it makes it possible to target the ZoomContainer as a nested selector:

```tsx
const X = styled(() => ...)

const Y = styled.div({
  // here we'll reference X which is a styled component
  [X]: { color: 'hotpink' }
})
```

PR #2135, opened 2021-03-02
Some visual tweaks to the actueel header components

VWSCoronaDashboard1:
## Summary

- Change the order (first the header, then the search bar, then the warningbox)
- Header text is now bold
- Tighten the margins on mobile
- Add a link to the 'landelijk' page to the landelijk actueel header
- Change the colors of the warningbox
- Warningbox fits the size of its contents instead of full width
- Warningbox doesn't show icon on mobile
- Reduce the length of the search bar (no longer full size)

## Motivation

Design requests

### Governance

- [ ] Documentation is added

### DIFF
--- a/packages/app/src/domain/topical/topical-page-header.tsx
+++ b/packages/app/src/domain/topical/topical-page-header.tsx
@@ -1,6 +1,5 @@
 import { ReactNode } from 'react';
 import { ArrowIconLeft } from '~/components-styled/arrow-icon';
-
 import { Box } from '~/components-styled/base';
 import { LinkWithIcon } from '~/components-styled/link-with-icon';
 import { RelativeDate } from '~/components-styled/relative-date';
@@ -13,12 +12,14 @@ interface TopicalPageHeaderProps {
   title: ReactNode;
   lastGenerated: number;
   showBackLink?: boolean;
+  link?: ReactNode;
 }
 
 export function TopicalPageHeader({
   title,
   lastGenerated,
   showBackLink,
+  link,
 }: TopicalPageHeaderProps) {
   return (
     <Box spacing={3}>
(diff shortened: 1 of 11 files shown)


packages/app/src/domain/topical/topical-page-header.tsx
```diff
@@ -13,12 +12,14 @@ interface TopicalPageHeaderProps {
   title: ReactNode;
   lastGenerated: number;
   showBackLink?: boolean;
+  link?: ReactNode;
```
VWSCoronaDashboard10, line 15: We can make it more clear this is a render prop:
```suggestion
  renderLink?: ReactNode;
```

packages/app/src/domain/topical/topical-page-header.tsx
```diff
@@ -31,16 +32,27 @@ export function TopicalPageHeader({
...
+          alignItems="baseline"
         >
-          {title}
-        </Heading>
+          <Heading
+            level={1}
+            m={0}
+            mb={{ _: 2, lg: 0 }}
+            lineHeight={0}
+            fontSize={{ _: '2rem', lg: '2.75rem' }}
```
VWSCoronaDashboard10, line 50:
```suggestion
            fontSize={{ _: 4, lg: '2.75rem' }}
```

packages/app/src/pages/index.tsx
```diff
@@ -90,18 +92,30 @@ const Home: FCWithLayout<typeof getStaticProps> = (props) => {
...
+                  icon={<ArrowIconRight />}
+                  iconPlacement="right"
+                  fontWeight="bold"
+                >
+                  {text.title_link}
+                </LinkWithIcon>
+              }
+            />
+
+            <Box width={{ _: '100%', lg: '75%' }}>
```
VWSCoronaDashboard10, line 112: width is 100% by default I think

```suggestion
            <Box width={{ lg: '75%' }}>
```

VWSCoronaDashboard10 (approved):
Looking good 👍

VWSCoronaDashboard8 (approved):

packages/app/src/domain/topical/topical-page-header.tsx
VWSCoronaDashboard10: Orrr .. wait, this doesn't make sense because it isn't a render prop at all.

Im only confused by the name of the `prop` being `link` which could be a lot things..

VWSCoronaDashboard1: Would `linkComponent` be more appropriate?

VWSCoronaDashboard8: Yes that would be a good name I think, but I'm about to replace it with text/href after this so don't worry 😅 I believe the Link layout is standard for all section headings and therefor the header should instantiate the link component. I am also adding the text and href for each section to lokalize.

PR #2138, opened 2021-03-02
Create the "tile"-like content header

VWSCoronaDashboard10:
## Summary

The vaccine progress content header will be rendered with a Tile layout, hence these changes.

I'd suggest to review it with the ignore whitespace option (`&w=1` query param) to reduce the amount of changed LOC.

image

### DIFF
--- a/packages/app/src/components-styled/content-header/index.tsx
+++ b/packages/app/src/components-styled/content-header/index.tsx
@@ -1,20 +1,22 @@
+import { assert } from '@corona-dashboard/common';
 import css from '@styled-system/css';
 import { ReactNode } from 'react';
 import styled from 'styled-components';
+import { Box } from '~/components-styled/base';
 import {
   Metadata,
   MetadataProps,
 } from '~/components-styled/content-header/metadata';
 import { HeadingWithIcon } from '~/components-styled/heading-with-icon';
+import { Tile } from '~/components-styled/tile';
 import {
   Heading,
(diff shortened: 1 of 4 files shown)


packages/app/src/components-styled/content-header/index.tsx
```diff
@@ -128,69 +130,101 @@ export function ContentHeader(props: ContentHeaderProps) {
...
-            flexDirection="row"
-            flexWrap="nowrap"
-            alignItems="center"
-            mb={-2}
-          >
-            <Box>
-              <Heading level={headingLevel} mb={0}>
-                {title}
-              </Heading>
+      <ContainerComponent>
```
VWSCoronaDashboard4, line 171: This component becomes very generic / multi use with this change. Why not create a new component for the tile, and have a `h1` -less version of this component underneath?

The resulting code becomes harder for me to develop on without being afraid of breaking changes.

VWSCoronaDashboard4 (changes requested):

packages/app/src/components-styled/content-header/index.tsx
VWSCoronaDashboard10: Ah yes, good idea. I'll pick that up tomorrow - thanks :-)

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard8 (approved):

PR #2142, opened 2021-03-02
Update line-chart chart axes with overflow visible

VWSCoronaDashboard14:
We could increase the padding to make more room for the chart-axes to avoid clipping, but I rather add an overflow visible.

Screenshot 2021-03-02 at 17 26 56


VWSCoronaDashboard1 (approved):

VWSCoronaDashboard8 (changes requested):
I don't think we should fix this by allowing the label to overflow. Was that requested specifically? We have padding props for the chart to allow for the labels on the y-axis to have more space. These paddings should be set so that the largest numbers align nicely with the content. If you allow overflow than the label it doesn't align anymore.

VWSCoronaDashboard8:
I'm going to approve it anyway, since we could do both. Allow overflow for when paddings are not set correctly, and then adjust paddings if we notice the label overflowing.

VWSCoronaDashboard8 (approved):

PR #2143, opened 2021-03-02
Implement actueel section headings

VWSCoronaDashboard8:
## Summary

This implements section headings for the Actueel page. In order to accomplish this I had to make a few alterations:

- Rename component from TopicalPageHeader TopicalSectionHeader
- Make all sections h2. I've added a hidden h1 for semantic correctness. Not sure if that is the best solution.
- Remove choropleth title from tile.
- Remove title and "subtitle" and link from article-list component. Since that is now the responsibility of the section header. I've also disabled the locale text for those items and moved them to live with the other section heading definitions.
- Move some lokalize header titles to a new "sections" key. And have this section key on common / nl / vr / gm.
- Move sitemap links below the articles, as the last section on the page, because that's what the design looked like and it seems to make sense to have all sections come before the sitemap.


packages/app/src/domain/topical/topical-section-header.tsx
```diff
@@ -0,0 +1,94 @@
...
+          borderBottomColor="border"
+          pb={{ _: 2, lg: 3 }}
+          mb={{ _: 2, lg: 3 }}
+          display="flex"
+          flexDirection={{ _: 'column', lg: 'row' }}
+          alignItems="baseline"
+        >
+          <Heading
+            level={2}
+            fontWeight={700}
```
VWSCoronaDashboard4, line 57: why not `bold` ?

packages/app/src/locale/nl.json
```diff
@@ -2230,18 +2230,51 @@
...
+      }
+    },
+
     "metadata": {
       "title": "Coronadashboard | COVID-19 | Rijksoverheid.nl",
       "description": "Informatie over de ontwikkeling van het coronavirus in Nederland."
     },
     "risiconiveaus": {
-      "selecteer_titel": "De risicokaart van Nederland",
+      "selecteer_titel": "__Wordt niet meer gebruikt",
```
VWSCoronaDashboard4, line 2277: this won't work, lokalize will ignore these new values and fill the old text back in

packages/app/src/locale/nl.json
```diff
@@ -2230,18 +2230,51 @@
...
-    "title_link": "Bekijk alle cijfers van Nederland",
+    "title": "Actueel",
     "the_netherlands": "Nederland",
+
+    "secties": {
+      "actuele_situatie": {
+        "titel": "De actuele situatie in {{the_netherlands}}",
+        "link": {
+          "text": "Bekijk alle cijfers van Nederland",
+          "href": "/landelijk/vaccinaties"
```
VWSCoronaDashboard4, line 2264: Not sure why the href should be part of lokalize?

packages/app/src/pages/actueel/gemeente/[code].tsx
```diff
@@ -92,14 +94,14 @@ const TopicalMunicipality: FCWithLayout<typeof getStaticProps> = (props) => {
...
         <MaxWidth>
           <TileList>
-            <TopicalPageHeader
+            <TopicalSectionHeader
               showBackLink
               lastGenerated={Number(props.lastGenerated)}
               title={replaceComponentsInText(text.title, {
-                municipalityName: <strong>{municipalityName}</strong>,
+                municipalityName: municipalityName,
+                gmCode,
```
Comment (VWSCoronaDashboard4) [inline packages/app/src/pages/actueel/gemeente/[code].tsx:102]: what is the `gmCode` doing here?

VWSCoronaDashboard1 (approved):

packages/app/src/pages/actueel/veiligheidsregio/[code].tsx
```diff
@@ -75,15 +76,46 @@ const TopicalSafetyRegion: FCWithLayout<typeof getStaticProps> = (props) => {
...
       />
+
       <Box bg="white" pb={4}>
+        {/**
+         * Since now the sections have a H2 heading I think we need to include
+         * a hidden H1 here.
+         *
+         * @TODO figure out what the title should be
+         */}
+        <VisuallyHidden>
```
Comment (VWSCoronaDashboard4) [inline packages/app/src/pages/actueel/veiligheidsregio/[code].tsx:87]: This does not make much sense to me. Can't we have a boolean toggle to add to the first component to toggle the h2 to a h3?

Visually hidden texts are nice for small texts, but not for page titles. Also, it would return in bad structure here: `<p><h1/></p>`. Besides that, search engines might penalise <h1>s with styling to make it invisible.

packages/app/src/pages/actueel/veiligheidsregio/[code].tsx
```diff
@@ -75,15 +76,46 @@ const TopicalSafetyRegion: FCWithLayout<typeof getStaticProps> = (props) => {
...
               showBackLink
               lastGenerated={Number(props.lastGenerated)}
-              title={replaceComponentsInText(text.title, {
-                safetyRegionName: <strong>{props.safetyRegionName}</strong>,
-              })}
+              title={replaceComponentsInText(
+                text.secties.actuele_situatie.titel,
+                {
+                  safetyRegionName: props.safetyRegionName,
+                  vrCode,
```
Comment (VWSCoronaDashboard4) [inline packages/app/src/pages/actueel/veiligheidsregio/[code].tsx:104]: What is the `vrCode` doing here? It's not part of the lokalize string?

VWSCoronaDashboard4 (changes requested):
I don't think we should take the shortcut with the visually hidden H1.

Apart from that only small comments.

packages/app/src/locale/nl.json
VWSCoronaDashboard8: I thought that's only the case for EN? In any case I will pass on these changes to whoever is updating lokalize

packages/app/src/locale/nl.json
VWSCoronaDashboard8: I think the two belong together. You can change the text to something else but without knowing what the related link is you don't know if what you're changing is correct. Also I think it makes sense to be able to change the link to somewhere else if you give them control over the link text.

packages/app/src/pages/actueel/gemeente/[code].tsx
Comment (VWSCoronaDashboard8) [reply packages/app/src/pages/actueel/gemeente/[code].tsx:102]: It being used in as a replacement in the link header text.

packages/app/src/pages/actueel/veiligheidsregio/[code].tsx
Comment (VWSCoronaDashboard8) [reply packages/app/src/pages/actueel/veiligheidsregio/[code].tsx:87]: Why do you think the h1 will be inside a paragraph? 

I could toggle the first section to be heading 1, but that is not correct either, because not all sections can be considered being part of the first section.

Comment (VWSCoronaDashboard8) [reply packages/app/src/pages/actueel/veiligheidsregio/[code].tsx:87]: Ah I see now VisuallyHidden is using a p. That is odd.

packages/app/src/pages/actueel/gemeente/[code].tsx
Comment (VWSCoronaDashboard8) [reply packages/app/src/pages/actueel/gemeente/[code].tsx:102]: Ah you're right sorry doesn't seem to be correct

VWSCoronaDashboard10 (approved):

packages/app/src/domain/topical/topical-section-header.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+export function TopicalSectionHeader({
+  title,
+  lastGenerated,
+  showBackLink,
+  link,
+  description,
+}: TopicalSectionHeaderProps) {
+  return (
+    <Box spacing={3} mt={{ _: 2, md: 4 }}>
```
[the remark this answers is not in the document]
VWSCoronaDashboard8: I'm not sure this code was introduced by me, but I wouldn't know an easy fix without refactoring the page layout in general.

packages/app/src/domain/topical/topical-section-header.tsx
```diff
@@ -0,0 +1,94 @@
...
+          <InlineText color="bodyLight" fontSize={2}>
+            {replaceComponentsInText(text.common_actueel.laatst_bijgewerkt, {
+              date: <RelativeDate dateInSeconds={lastGenerated} />,
+              time: formatDateFromSeconds(lastGenerated, 'time'),
+            })}
+          </InlineText>
+        )}
+
+        {description && (
+          <Box mb={4} maxWidth={'30em'}>
```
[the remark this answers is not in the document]
VWSCoronaDashboard8: I'll change it to padding. Although I'm not sure what the advantage of that is.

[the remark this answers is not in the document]
VWSCoronaDashboard8: On second thought I am keeping it margin, because some components (like article list) still have a margin top, and in those cases I think you want the margins to collapse to only have the biggest one of the two.

PR #2154, opened 2021-03-03
Fix long labels being rendered out of the viewbox on the horizontal bar chart

VWSCoronaDashboard1:
## Summary

As the title states

## Motivation

Bug report

## Detailed design

The left padding of the barchart is now calculated on the actual computed text length of the longest label.
This is done by first determining which label text is the longest and the creating and adding a temporary svg text element to the dom. This element's `getComputedTextLength` method is then used to determine the actual rendered length of the text.

Also, when doing this fix I moved the label font size logic in to the `useBarchartCoordinates` hook, since all the calculations are done there anyway.

image

### Governance

- [ ] Documentation is added


packages/app/src/components-styled/bar-chart/bar-chart-coordinates.ts
```diff
@@ -31,28 +34,67 @@ export interface BarChartCoordinates {
   getBarOffset: (x: BarChartValue) => number | undefined;
   getLabel: (x: BarChartValue) => string;
   getTooltipCoordinates: GetTooltipCoordinates<BarChartValue>;
+  labelFontSize: string;
+}
+
+function calculateMaximumLabelLength(labels: string[], fontSize: string) {
+  const longestLabel = labels.reduce(
```
VWSCoronaDashboard4, line 41: What if one label is `ll` and another is `m`? Render width does not relate linearly to the amount fo characters.

packages/app/src/components-styled/bar-chart/bar-chart-coordinates.ts
```diff
@@ -31,28 +34,67 @@ export interface BarChartCoordinates {
...
+    (longest, label) => (label.length > longest.length ? label : longest),
+    ''
+  );
+
+  if (typeof window !== 'undefined' && longestLabel.length > 0) {
+    const textElement = window.document.createElementNS(SVG_NS, 'text');
+    textElement.textContent = longestLabel;
```
VWSCoronaDashboard4, line 48: Instead of checking only the label with the most characters; maybe put all labels in here, separated by newlines. Then read out the offsetwidth?

No need for the `createElementNS` magic I think, it can be done with a `div` if the fontfamily/fontsize is right.

packages/app/src/components-styled/bar-chart/bar-chart-coordinates.ts
```diff
@@ -31,28 +34,67 @@ export interface BarChartCoordinates {
...
+    textElement.setAttributeNS(null, 'font-size', fontSize);
```
VWSCoronaDashboard4, line 49: Be sure to also set the font family, as it night default to a different font

VWSCoronaDashboard1: don't we set a default body fontfamily? I can still set the family explicitly, but I think this probably will work correctly with out as well.

packages/app/src/components-styled/bar-chart/bar-chart-coordinates.ts
VWSCoronaDashboard1: True, this is a bit of a best guess

VWSCoronaDashboard1: True, this is a bit of a best guess scenario

packages/app/src/components-styled/bar-chart/bar-chart-coordinates.ts
VWSCoronaDashboard1: This solution actually requires a lot more elements to be created. I'll have to create a div, then loop over the labels and add them as spans and create BR elements between them.
I've been trying different approaches, but I have to admit that the original solution yielded the most consistent results...

VWSCoronaDashboard1: In fact, I've clicked through all of the safety regions and couldn't really find an instance where this messes up

packages/app/src/components-styled/bar-chart/bar-chart-coordinates.ts
VWSCoronaDashboard1: Done

VWSCoronaDashboard8 (approved):

PR #2158, opened 2021-03-03
Rendering fewer ticks on bar chart when width is less than 400px

VWSCoronaDashboard13:
## Summary

Rendering fewer ticks on the bar chart on mobile so there is no tick label overlap

## Detailed design

Conditionally setting the number of ticks in `generateBarChartCoordinates` based on the width

### DIFF
--- a/packages/app/src/components-styled/bar-chart/bar-chart-coordinates.ts
+++ b/packages/app/src/components-styled/bar-chart/bar-chart-coordinates.ts
@@ -61,7 +61,7 @@ function generateBarChartCoordinates(
   const barsHeight = values.length * 35;
   const height = barsHeight + spacing.top + spacing.bottom;
 
-  const numTicks = 10;
+  const numTicks = width > 400 ? 10 : 6;
 
   const getValue = (value: BarChartValue): number => value.value;
   const getLabel = (value: BarChartValue): string => value.label;


VWSCoronaDashboard4 (approved):

packages/app/src/components-styled/bar-chart/bar-chart-coordinates.ts
```diff
@@ -61,7 +61,7 @@ function generateBarChartCoordinates(
   const barsHeight = values.length * 35;
   const height = barsHeight + spacing.top + spacing.bottom;
 
-  const numTicks = 10;
+  const numTicks = width > 400 ? 10 : 6;
```
VWSCoronaDashboard1, line 64: In this PR: https://github.com/minvws/nl-covid19-data-dashboard/pull/2154/files#diff-4af1f3aab2c05a3d786e2f4e6277c9acd28014b03cea47364c1fd97489f705ca
I've added the `useBreakpoints()` hook, its probably better to determine this tick number based on one of the breakpoints there in. In this case `breakpoints.xs`.

VWSCoronaDashboard1 (changes requested):

packages/app/src/components-styled/bar-chart/bar-chart-coordinates.ts
VWSCoronaDashboard13: Alright, I'll wait for that PR to get merged in before implementing this update

VWSCoronaDashboard13: Actually I just read another thread on slack discussing this issue where the consensus was to keep the existing implementation!

VWSCoronaDashboard10: Nope, your implementation is the correnct one indeed, we can merge this 👍

VWSCoronaDashboard10 (approved):

PR #2164, opened 2021-03-03
Vaccine progress header

VWSCoronaDashboard10:
## Summary

- Default `<ContentHeader/>`'s title has become optional in order to allow rendering that component without title: the vaccination page renders the title inside a separate Tile.
- `<TwoKpiSection />` can receive a spacing prop to define the space between the sections. Also all Box props have been removed, they were not used.
- There are still a couple of static texts inside the vaccine header component. I'll fix these in another PR.

### DIFF
--- a/packages/app/src/components-styled/content-header/index.tsx
+++ b/packages/app/src/components-styled/content-header/index.tsx
@@ -152,13 +152,7 @@ export function ContentHeader(props: ContentHeaderProps) {
                 headingLevel={headingLevel}
               />
             ) : (
-              <Box
-                display="flex"
-                flexDirection="row"
-                flexWrap="nowrap"
-                alignItems="center"
-                mb={-2}
-              >
+              <Box display="flex" flexWrap="nowrap" alignItems="center" mb={-2}>
                 <Box>
                   <Heading level={headingLevel} mb={0}>
                     {title}
--- a/packages/app/src/components-styled/content-header/index.tsx
+++ b/packages/app/src/components-styled/content-header/index.tsx
@@ -143,30 +143,34 @@ export function ContentHeader(props: ContentHeaderProps) {
             )}
           </CategoryHeading>
         )}
-        {icon ? (
-          <HeadingWithIcon
-            icon={icon}
-            title={title}
-            headingLevel={headingLevel}
-          />
-        ) : (
-          <Box
-            display="flex"
-            flexDirection="row"
-            flexWrap="nowrap"
-            alignItems="center"
-            mb={-2}
-          >
-            <Box>
-              <Heading level={headingLevel} mb={0}>
-                {title}
-              </Heading>
-            </Box>
-          </Box>
+        {title && (
+          <>
+            {icon ? (
+              <HeadingWithIcon
+                icon={icon}
+                title={title}
+                headingLevel={headingLevel}
+              />
+            ) : (
+              <Box
+                display="flex"
+                flexDirection="row"
+                flexWrap="nowrap"
+                alignItems="center"
+                mb={-2}
+              >
+                <Box>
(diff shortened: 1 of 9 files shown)


packages/app/src/components-styled/content-header/index.tsx
```diff
@@ -143,30 +143,34 @@ export function ContentHeader(props: ContentHeaderProps) {
...
-            mb={-2}
-          >
-            <Box>
-              <Heading level={headingLevel} mb={0}>
-                {title}
-              </Heading>
-            </Box>
-          </Box>
+        {title && (
+          <>
```
VWSCoronaDashboard8, line 147: Fragment seems redundant

packages/app/src/components-styled/content-header/index.tsx
```diff
@@ -143,30 +143,34 @@ export function ContentHeader(props: ContentHeaderProps) {
...
+            {icon ? (
+              <HeadingWithIcon
+                icon={icon}
+                title={title}
+                headingLevel={headingLevel}
+              />
+            ) : (
+              <Box
+                display="flex"
+                flexDirection="row"
```
VWSCoronaDashboard8, line 157: row is default

packages/app/src/domain/vaccine/vaccine-content-header.tsx
```diff
@@ -0,0 +1,224 @@
...
+
+interface VaccinesContentHeaderProps {
+  text: Locale['vaccinaties'];
+  data: National;
+}
+
+export function VaccineContentHeader({
+  text,
+  data,
+}: VaccinesContentHeaderProps) {
```
VWSCoronaDashboard8, line 31: This component seems to be much more than what you would normally expect to be in a header. Also it contains its own ContentHeader. So I think the name is maybe a little misleading but I don't have a suggestion either.

packages/app/src/domain/vaccine/vaccine-content-header.tsx
```diff
@@ -0,0 +1,224 @@
...
+            dateOfInsertionUnix:
+              data.vaccine_administered_total.last_value.date_of_insertion_unix,
+            dataSources: [],
+          }}
+        />
+      </Box>
+    </Box>
+  );
+
+  function componentCallback(callbackInfo: ComponentCallbackInfo) {
```
VWSCoronaDashboard8, line 130: Hopefully we can capture all this in the new chart components, because this is getting really specific and verbose.

VWSCoronaDashboard8 (approved):

packages/app/src/components-styled/content-header/index.tsx
VWSCoronaDashboard10: I thought using a fragment here is easier for the eyes, compared to:
```tsx
{title && 
  (icon ? (
    ...
  ) : (
    ...
  )}

packages/app/src/components-styled/content-header/index.tsx
VWSCoronaDashboard10: I've only changed indentation on these lines, but I'll clean it up :)

packages/app/src/domain/vaccine/vaccine-content-header.tsx
VWSCoronaDashboard10: Ah, I could rename it to `VaccinePageHeader` if that would make more sense?

VWSCoronaDashboard8: Yes I'm not too worried about this. Maybe `VaccinePageIntroduction` or `VaccinePageLeadingContent` ?

VWSCoronaDashboard1 (approved):

PR #2178, opened 2021-03-04
Create vaccine progress ticker

VWSCoronaDashboard10:
## Summary

This PR adds the vaccination ticker component.

image


packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
```diff
@@ -0,0 +1,244 @@
...
+  data: NlVaccineAdministeredRateMovingAverageValue;
+}
+
+export function VaccineTicker({ data }: VaccineTickerProps) {
+  const tickDuration = data.seconds_per_dose * 1000;
+  const tickCount = Math.floor(data.doses_per_second * 60);
+  const [counter, setCounter] = useState(0);
+
+  useEffect(() => {
+    const id = window.setInterval(() => setCounter((x) => x + 1), tickDuration);
```
VWSCoronaDashboard4, line 26: Nit; for the next time we work on thsi code perhaps:

for these sorts of animations it is handier to let the timing be done by `requestAnimationFrame`. `setInterval` can be handled differently when you switch tabs. Eg if the tab with our App is in the background, upon returning it would go through a bunch of callback in a short period of time to 'catch up' after the browser has paused them.

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
```diff
@@ -0,0 +1,244 @@
...
+        width: '100%',
+        height: '100%',
+        minWidth: CONTAINER_WIDTH,
+      })}
+    >
+      {Array(numTicks)
+        .fill(null)
+        .map((_, index) => (
+          <Tick
+            deg={index * stepDegree + 360 / numTicks}
```
VWSCoronaDashboard4, line 118: What ias a `deg`? Short for degree? That would be the unit, not the property, I guess. Maybe rename it to `rotation` or something similar?

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
```diff
@@ -0,0 +1,244 @@
...
+  index: number;
+  angle: number;
+  deg: number;
+  isSelected: boolean;
+  isLast: boolean;
+  tickDuration: number;
+  tickWidth: number;
+  tickLength: number;
+}) {
+  const translateX =
```
VWSCoronaDashboard4, line 156: You can get rid of this calculation by adding a second `rotate` in the transform; it would then transform like:

1. rotate the tick to the desired angle that this rectangle needs to be
2. translate it to the radius
3. rotate it "minute-hand on the clock" style

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
```diff
@@ -0,0 +1,244 @@
...
+}
+
+function Shapes({
+  counter,
+  tickDuration,
+}: {
+  counter: number;
+  tickDuration: number;
+}) {
+  const shapes = useShapes();
```
VWSCoronaDashboard4, line 189: pretty neat solution!

VWSCoronaDashboard4 (commented):
Looking good. Some comments to simplify the calculations a bit, and improve the interval-handling

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
VWSCoronaDashboard10: good idea, will implement it.

VWSCoronaDashboard1 (approved):

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
```diff
@@ -18,14 +19,26 @@ interface VaccineTickerProps {
...
+      if (!isMountedRef.current) return;
+
+      setCounter((x) => x + 1);
+
+      timeoutId = window.setTimeout(
+        () => requestAnimationFrame(tick),
+        tickDuration
+      );
+    }
+    tick();
```
VWSCoronaDashboard1, line 39: AFAIK the RAF handler receives a current timestamp. So instead of using the timeout you could use something like this:
```
function tick(now: number) {
  if ((now - _last) >= 1000) {
     _last = now;
    doStuff();
  }
}
```

VWSCoronaDashboard1 (changes requested):

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
VWSCoronaDashboard1: Oops, I clicked on request changes, I meant to just click on approve. So don't feel like this needs to be changes.

VWSCoronaDashboard10: Ok, I also don't see a lot of added value of this approach compared to a setTimeout inside the RAF :-)

PR #2209, opened 2021-03-08
POC messages

VWSCoronaDashboard4:
For inspiration

Noteworthy to looks at:

In `cli`:
* `yarn run messages:interfaces` creates a Typescript interface and an array of keys it has found, based on uses of `formatString(...)` and `formatBlock(...)`
* `yarn run messages:schemas` creates Sanity schema's from the interfaces created by `yarn run messages:interfaces`
* Extracting interfaces/keys now happens using rudimentary regex work, but can work well by walking through the AST using certain typescript functions.
* `yarn run messages` combines both

In `app`:
* `createGetMessages` generates a query for the provided interface / name of 'bucket' of messages.
* `formatMessages` transforms the returned messages into strings/components (for the blocks). This could be expanded later into a bit more powerful formatting, such as string replacements (for GM/VR names)
* with adjustments/naming conventions we could opt to make the use of the name `messages` solemnly for this; we could then extract replace `formatString('foo')` for `messages.foo`. (Important when extracting keys.)

In `cms`:
* Generated blocks of field configs are in `schemas/messages` and are then imported where needed.

## Adding new files / bucket

1. Include the needed code (formatString, formatBlock, createGetMessages, formatMessages) using a new name
2. `yarn run messages` in cli
3. include the generated schema on the desired page

## Adding new strings  files / bucket

1. Include the needed code (formatString, formatBlock) using an existing bucket name
2. `yarn run messages` in cli

(Also needed with this hacky code, but can be fixed to eliminate: fix interfaces import, fix empty sanity response)

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -4,6 +4,7 @@
   "version": "0.0.0",
   "license": "EUROPEAN UNION PUBLIC LICENCE v. 1.2",
   "dependencies": {
+    "@formatjs/cli": "^3.1.6",
     "@formatjs/intl-datetimeformat": "^3.2.7",
     "@formatjs/intl-getcanonicallocales": "^1.4.5",
     "@formatjs/intl-locale": "^2.4.14",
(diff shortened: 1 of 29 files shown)


VWSCoronaDashboard4:
No need to keep this POC PR alive.

PR #2214, opened 2021-03-08
Sterftemonitor using new chart

VWSCoronaDashboard8:
## Summary

This is the sterftemonitor chart using the new TimeSeriesChart component instead of Highcharts. 

image

### DIFF
--- a/packages/app/src/domain/deceased/deceased-monitor-section.tsx
+++ b/packages/app/src/domain/deceased/deceased-monitor-section.tsx
@@ -2,16 +2,17 @@ import {
   NationalDeceasedCbs,
   RegionalDeceasedCbs,
 } from '@corona-dashboard/common';
+import { ParentSize } from '@visx/responsive';
 import { AnchorTile } from '~/components-styled/anchor-tile';
-import { Box } from '~/components-styled/base';
 import { ChartTile } from '~/components-styled/chart-tile';
-import { Legend } from '~/components-styled/legend';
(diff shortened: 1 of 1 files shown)


VWSCoronaDashboard1 (approved):

packages/app/src/domain/deceased/deceased-monitor-section.tsx
```diff
@@ -35,45 +36,38 @@ export function DeceasedMonitorSection({
...
-            ]}
-          />
-        </Box>
+        <ParentSize>
+          {({ width }) => (
+            <TimeSeriesChart
+              title={text.deceased_monitor_chart_title}
+              width={width}
+              values={data.values}
+              ariaLabelledBy=""
```
VWSCoronaDashboard4, line 45: If we don't need it, leave it out?

VWSCoronaDashboard4: (I guess it's a required prop, maybe make it optional?)

VWSCoronaDashboard4 (approved):

packages/app/src/domain/deceased/deceased-monitor-section.tsx
VWSCoronaDashboard8: It's a required prop and I think it should be required, but this PR is just meant as a starting point for whoever is picking up the task of implementing this chart.

PR #2217, opened 2021-03-08
Add two column choropleth layout and the risk level map also put the escalation map in the new layout

VWSCoronaDashboard14:
Added a `topical-choropleth-layout` component that can receive a choropleth and content for the right side. Later this would also be nice to use the same for the new risk levels layout. 

On the manuplicity, safety region and the nation should now show a choropleth.

Screenshot 2021-03-08 at 13 41 33


VWSCoronaDashboard4 (approved):

VWSCoronaDashboard1 (approved):

packages/app/src/domain/topical/topical-choropleth-layout.tsx
```diff
@@ -0,0 +1,43 @@
+import { ReactNode } from 'react';
+import { Box } from '~/components-styled/base';
+import { TopicalSectionHeader } from '~/domain/topical/topical-section-header';
+import { TopicalTile } from '~/domain/topical/topical-tile';
+
+interface TopicalChoroplethTileProps {
+  title: string;
+  choropleth: ReactNode;
```
VWSCoronaDashboard8, line 8: That's a bit of an odd prop to have next to children. In any case I would name it `choroplethComponent` since we use that pattern in other place. But I would expect the main thing of the tile to be the choropleth and to have that as children and maybe the other component on props. 

You could also iterate over multiple children, but I don't have a clear idea about this abstraction yet.

VWSCoronaDashboard8: Yeah looking at the code below you have two parts. The map and the "other stuff". There's different ways of tackling this, but maybe just leave it as you did now and only rename the prop. 

My preference would be to not have this component. Because you could do this:

Place
``` 
<TopicalTile>
   <TopicalSectionHeader title={title} />

</TopicalTile>
```
directly in the page. And create "two-column" layout component where you pass two children, one is the choropleth and the other child is everything else wrapped in a div.  So it would be:

```

<TopicalTile>
   <TopicalSectionHeader title={title} />
   <TwoColumnLayout>
     <Choropleth/>
<div>
   ... the other stuff
</div>
</TwoColumnLayout>
</TopicalTile>
```

The two column layout could do something similar as TwoKPITile where you enforce 2 children and than do the styling based on the wrapping those children.

Then you use composition better and have more generic components and less prop drilling

packages/app/src/pages/actueel/gemeente/[code].tsx
```diff
@@ -276,9 +298,73 @@ const TopicalMunicipality: FCWithLayout<typeof getStaticProps> = (props) => {
               </TopicalTile>
             </Box>
 
+            <TopicalChoroplethTile
+              title={
+                siteText.common_actueel.secties.positief_getest_kaart.titel
+              }
+              choropleth={
+                <>
```
Comment (VWSCoronaDashboard8) [inline packages/app/src/pages/actueel/gemeente/[code].tsx:306]: This fragment has no function

packages/app/src/pages/actueel/veiligheidsregio/[code].tsx
```diff
@@ -255,6 +276,70 @@ const TopicalSafetyRegion: FCWithLayout<typeof getStaticProps> = (props) => {
               </TopicalTile>
             </Box>
 
+            <TopicalChoroplethTile
+              title={
+                siteText.common_actueel.secties.positief_getest_kaart.titel
+              }
+              choropleth={
+                <>
```
Comment (VWSCoronaDashboard8) [inline packages/app/src/pages/actueel/veiligheidsregio/[code].tsx:284]: same

packages/app/src/pages/index.tsx
```diff
@@ -250,9 +267,73 @@ const Home: FCWithLayout<typeof getStaticProps> = (props) => {
               </TopicalTile>
             </Box>
 
+            <TopicalChoroplethTile
+              title={
+                siteText.common_actueel.secties.positief_getest_kaart.titel
+              }
+              choropleth={
+                <>
```
VWSCoronaDashboard8, line 275: etc

VWSCoronaDashboard1 (approved):

packages/app/src/pages/actueel/gemeente/[code].tsx
```diff
@@ -252,27 +251,24 @@ const TopicalMunicipality: FCWithLayout<typeof getStaticProps> = (props) => {
...
-                  legendComponent={
-                    <EscalationMapLegenda
-                      data={choropleth.vr}
-                      metricName="escalation_levels"
-                      metricProperty="level"
-                    />
-                  }
-                >
+              <ChoroplethTwoColumnLayout
+                legenda={
```
Comment (VWSCoronaDashboard8) [inline packages/app/src/pages/actueel/gemeente/[code].tsx:260]: I wonder if it's possible to have the legenda as part of the children composition, but let's leave that as an exercise for later.

Comment (VWSCoronaDashboard8) [reply packages/app/src/pages/actueel/gemeente/[code].tsx:260]: Maybe name the prop `legendComponent` to make it clearer. We usually do that with components on props. Also "legenda" is a Dutch word AFAIK in English it's legend.

packages/app/src/pages/actueel/gemeente/[code].tsx
```diff
@@ -252,27 +251,24 @@ const TopicalMunicipality: FCWithLayout<typeof getStaticProps> = (props) => {
...
+                  <EscalationMapLegenda
```
Comment (VWSCoronaDashboard8) [inline packages/app/src/pages/actueel/gemeente/[code].tsx:261]: Maybe also rename "legenda" here. Or leave it for later if that was already part of the code. I recently renamed the Legend component from Legenda so I suspect there are more cases in the codebase.

packages/app/src/pages/actueel/gemeente/[code].tsx
```diff
@@ -342,25 +361,23 @@ const TopicalMunicipality: FCWithLayout<typeof getStaticProps> = (props) => {
...
-              />
-            </TopicalChoroplethTile>
+                </Box>
+                <Box>
+                  <Text>
+                    {siteText.positief_geteste_personen.map_toelichting}
+                  </Text>
+                  <Box
+                    mb={4}
+                    css={css({ '> div': { justifyContent: 'flex-start' } })}
```
Comment (VWSCoronaDashboard8) [inline packages/app/src/pages/actueel/gemeente/[code].tsx:371]: If ChartRegionControls is a styled component you could target it with `{ChartRegionControls}` in the css function. It makes it a little clearer.

packages/app/src/pages/actueel/veiligheidsregio/[code].tsx
```diff
@@ -230,58 +230,78 @@ const TopicalSafetyRegion: FCWithLayout<typeof getStaticProps> = (props) => {
...
+                      regionThresholds.tested_overall.infected_per_100k
+                    }
+                    title={
+                      siteText.positief_geteste_personen.chloropleth_legenda
+                        .titel
+                    }
+                  />
+                }
+              >
+                <Box>
```
Comment (VWSCoronaDashboard8) [inline packages/app/src/pages/actueel/veiligheidsregio/[code].tsx:304]: nit; I don't think you need this extra box

packages/app/src/pages/index.tsx
```diff
@@ -217,36 +215,24 @@ const Home: FCWithLayout<typeof getStaticProps> = (props) => {
...
+                    data={choropleth.vr}
+                    metricName="escalation_levels"
+                    metricProperty="level"
+                    lastDetermined={
+                      choropleth.vr.escalation_levels[0].last_determined_unix
+                    }
+                  />
+                }
+              >
+                <Box>
```
VWSCoronaDashboard8, line 235: Same for this one

packages/app/src/pages/index.tsx
```diff
@@ -256,27 +242,51 @@ const Home: FCWithLayout<typeof getStaticProps> = (props) => {
...
-                borderTopColor={colors.silver}
-                mx={{ _: -3, md: 0 }}
-              />
-              <TopicalTile py={0}>
-                <Box mx={-3}>
-                  <EscalationLevelExplanations />
                 </Box>
-              </TopicalTile>
-            </Box>
+                <Box>
```
VWSCoronaDashboard8, line 246: And this one

packages/app/src/pages/index.tsx
```diff
@@ -256,27 +242,51 @@ const Home: FCWithLayout<typeof getStaticProps> = (props) => {
...
+                      regionThresholds.tested_overall.infected_per_100k
+                    }
+                    title={
+                      siteText.positief_geteste_personen.chloropleth_legenda
+                        .titel
+                    }
+                  />
+                }
+              >
+                <Box>
```
VWSCoronaDashboard8, line 289: etc

VWSCoronaDashboard8 (approved):
Looks like a big chunk of work! 💯

PR #2223, opened 2021-03-08
Fix sewer date rendering

VWSCoronaDashboard10:
## Summary

- Render the date range in the sewer chart x-axis tooltip.
- Render the vertical "hover-bar" using the start- and end date.
- Visually display tooltip after tiny delay to process dimensions and positioning of the tooltip (prevents the initial "jump" of the tooltip)
- Use number formatter when displaying numbers

### DIFF
--- a/packages/app/src/components-styled/sewer-chart/logic.ts
+++ b/packages/app/src/components-styled/sewer-chart/logic.ts
@@ -13,12 +13,15 @@ export interface Dimensions {
   padding: Record<'top' | 'right' | 'bottom' | 'left', number>;
 }
 
-export interface SewerChartValue {
+export type SewerChartValue = {
   id: string;
   name: string;
   value: number;
   dateMs: number;
-}
+} & (
+  | { dateStartMs: number; dateEndMs: number }
+  | { dateStartMs?: never; dateEndMs?: never }
+);
 
 export function useSewerChartValues(
   data: Regionaal | Municipal,
@@ -38,7 +41,12 @@ export function useSewerChartValues(
             .map((x) => ({
               name: '__average',
               value: x.average,
-              dateMs: x.date_end_unix * 1000,
+              dateMs:
+                (x.date_start_unix +
+                  (x.date_end_unix - x.date_start_unix) / 2) *
+                1000,
+              dateStartMs: x.date_start_unix * 1000,
+              dateEndMs: x.date_end_unix * 1000,
             }))
             .map((x) => ({
               ...x,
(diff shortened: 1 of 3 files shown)


packages/app/src/components-styled/sewer-chart/logic.ts
```diff
@@ -269,7 +277,13 @@ export function useLineTooltip({
     });
   }
 
-  return { datum, point, findClosest: setInputPoint, clear } as const;
+  return { datum, point, findClosest: setInputPoint, clear } as {
+    findClosest: typeof setInputPoint;
+    clear: typeof clear;
+  } & (
+    | { datum: undefined; point: undefined }
```
VWSCoronaDashboard4, line 284: Why a dutch name for this?

VWSCoronaDashboard4 (approved):

VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/sewer-chart/logic.ts
VWSCoronaDashboard8: I think what is meant here is the English singular form of data.

packages/app/src/components-styled/sewer-chart/sewer-chart.tsx
```diff
@@ -283,8 +283,17 @@ export function SewerChart(props: SewerChartProps) {
...
-                x={lineTooltip.point.x - 4}
-                width={8}
+                x={
+                  lineTooltip.datum.dateStartMs
+                    ? scales.xScale(lineTooltip.datum.dateStartMs)
+                    : lineTooltip.point.x - 4
+                }
+                width={
+                  lineTooltip.datum.dateStartMs
+                    ? scales.xScale(lineTooltip.datum.dateEndMs) -
```
VWSCoronaDashboard8, line 293: You could move this to a place where it's only calculated once on one value of the data. Because all samples in the values array will always span the same time. Maybe move it to the scales hook. I think I did something similar in the TimeSeriesChart

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard14 (approved):

packages/app/src/components-styled/sewer-chart/logic.ts
```diff
@@ -13,12 +13,15 @@ export interface Dimensions {
...
-export interface SewerChartValue {
+export type SewerChartValue = {
   id: string;
   name: string;
   value: number;
   dateMs: number;
-}
+} & (
+  | { dateStartMs: number; dateEndMs: number }
+  | { dateStartMs?: never; dateEndMs?: never }
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: it just means that both dateStartMs and dateEndMs have to be set, or none at all

packages/app/src/components-styled/sewer-chart/sewer-chart.tsx
VWSCoronaDashboard10: We're talking about one thousandth of a millisecond here, 0.001ms - don't think it's really necessary to calculate it only once :-)

PR #2224, opened 2021-03-08
Improve sewer chart animations

VWSCoronaDashboard10:
## Summary

Improve the sewer chart scatter plot animation by plotting the outlier-data inside the chart but with limited Y-values. This allows us to tween the outlier circle-elements to their real position when outlier visibility is toggled.


VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/sewer-chart/components/toggle-outlier-button.tsx
```diff
@@ -1,21 +1,75 @@
...
       cursor: 'default',
     },
   })
 );
+
+/**
+ * transform rgba with transparency to its hex alternative on a white background.
+ * https://stackoverflow.com/questions/15898740/how-to-convert-rgba-to-a-transparency-adjusted-hex
+ */
+function hexify(color: string) {
```
VWSCoronaDashboard8, line 58: Doesn't polished already include something like this? I've used `transparantize` before in the code

packages/app/src/components-styled/sewer-chart/sewer-chart.tsx
```diff
@@ -71,26 +70,65 @@ export function SewerChart(props: SewerChartProps) {
   /**
    * filter and shape data for our line-chart and scatter plot
    */
-  const { averageValues, stationValues } = useSewerChartValues(data, timeframe);
+  const {
+    averageValues: averageLineData,
+    stationValues: stationScatterData,
+  } = useSewerChartValues(data, timeframe);
```
VWSCoronaDashboard8, line 76: nit; If a hook only returns two things, an array is more useful for this reason I think

VWSCoronaDashboard8 (approved):

packages/app/src/components-styled/sewer-chart/components/toggle-outlier-button.tsx
VWSCoronaDashboard10: I cannot find function capable of doing such a transformation..

PR #2226, opened 2021-03-08
Add schema for risk levels for safety regions

VWSCoronaDashboard14:
Add the schema for the dynamic text in the safety regions' risk levels.
3 new keys added:


packages/app/schema/vr/escalation_level.json
```diff
@@ -17,8 +20,23 @@
...
-    "valid_from_unix": { "type": "integer" },
-    "last_determined_unix": { "type": "integer" },
-    "date_of_insertion_unix": { "type": "integer" }
+    "date_start_measurement": {
+      "type": "integer"
+    },
+    "date_end_measurement": {
+      "type": "integer"
+    },
+    "date_next_estimate": {
```
VWSCoronaDashboard1, line 29: If the dates are unix timestamps, I believe the convention is to suffix them with `_unix`.

VWSCoronaDashboard1 (changes requested):

packages/app/schema/vr/escalation_level.json
```diff
@@ -5,6 +5,9 @@
   "required": [
     "date_unix",
     "level",
+    "date_start_measurement_unix",
```
VWSCoronaDashboard4, line 8: what's a measurement?

VWSCoronaDashboard8 (changes requested):
I think we need to have a meeting to clear some things up and come up with good names, because this will get confusing

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard1 (approved):

PR #2243, opened 2021-03-09
Test build times with all getStaticPaths emptied

VWSCoronaDashboard:
## Summary

Reduce build times by loading municipality and safety region pages on first demand instead of on build-time.

With these changes, a build takes 5 minutes to complete.
We can further reduce this by lazy loading images though an image CDN instead of resizing them all on build-time.

## Motivation

Faster build times leads to a shorter feedback loop.
Building less pages on every build potentially means lower bandwidth/CDN usage.


VWSCoronaDashboard8:
Added to Trello tech dept

PR #2251, opened 2021-03-10
Allow y-axis to start from a non-zero value

VWSCoronaDashboard10:
## Summary

This PR allows the Y-axis to start from an automatically determined nice and round value instead of always using `0`.

old:
image

new:
image


VWSCoronaDashboard10:
Closed because we always want to start by `0`.

PR #2261, opened 2021-03-10
Create dynamic trend icons

VWSCoronaDashboard10:
## Summary

Dynamically create the trend icons in the tooltip and the legend based on the series config

image

### DIFF
--- a/packages/app/src/components-styled/time-series-chart/components/tooltip.tsx
+++ b/packages/app/src/components-styled/time-series-chart/components/tooltip.tsx
@@ -20,14 +20,7 @@ import { VisuallyHidden } from '~/components-styled/visually-hidden';
 import { formatDateFromSeconds } from '~/utils/formatDate';
 import { formatNumber, formatPercentage } from '~/utils/formatNumber';
 import { useBreakpoints } from '~/utils/useBreakpoints';
-import {
-  DataOptions,
-  isAreaSeriesDefinition,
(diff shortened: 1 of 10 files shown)


VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/time-series-chart/components/tooltip.tsx
```diff
@@ -194,25 +206,48 @@ const TooltipList = styled.ol`
...
   }
-`;
+
+  if (isRangeSeriesDefinition(config)) {
+    return <RangeTrendIcon config={config} />;
+  }
+
+  if (isAreaSeriesDefinition(config)) {
+    return <AreaTrendIcon config={config} />;
+  }
```
VWSCoronaDashboard8, line 247: You can replace this with a single switch statement on config.type. The type property with its string literal makes it a discriminating union, and TS will know based on that property what the type is. Then you don't need these type guard functions.

VWSCoronaDashboard8 (approved):

packages/app/src/components-styled/time-series-chart/components/tooltip.tsx
VWSCoronaDashboard10: ah yes! I saw that pattern and told myself to implement it but totally forgot it :-)

PR #2262, opened 2021-03-11
Update Actueel quick links and data sitemap layout

VWSCoronaDashboard13:
## Summary

Moved the quick links and data sitemap to an expandable button on the actueel collapsible button. Also includes "filtering" of the links depending on which page it is on.

## Motivation

Redesigned

## Detailed design

Created a `LinkBlock` component that styles the quicklinks and data sitemap. It accepts `quickLinksHeader`, `quickLinks`, `dataSitemapHeader` and `dataSitemap`. 

Talked to Aryan about filtering and we decided to just define the links for each of the pages rather than creating shared logic for these links and the side bars. To keep the links separate from the rendering code, I moved the links to a `useDataSitemap` hook. It accepts the base (page), an optional code (the selected region), and data which is needed to determine if the sewer link is disabled for the gemeente page.

The "disabled" link is just rendering a span in the its place


packages/app/src/domain/topical/link-block/data-sitemap/data-sitemap.tsx
```diff
@@ -0,0 +1,45 @@
...
+  code?: string | string[];
+};
+
+export function DataSitemap(props: DataSitemapProps) {
+  const { base, code } = props;
+  const linkGroups = useDataSitemapLinks(base, code);
+
+  return (
+    <Box>
+      <Heading level={3} fontSize={3}>
```
VWSCoronaDashboard8, line 20: The heading has a level and "as" prop. That should be enough normally be enough to render a certain style and make it show up in the markup as a specific element. Unless 3 is not part of the predefined header styles...

packages/app/src/domain/topical/link-block/data-sitemap/data-sitemap.tsx
```diff
@@ -0,0 +1,45 @@
...
+        {siteText.nationaal_actueel.data_sitemap_titel}
+      </Heading>
+
+      <Box
+        display="flex"
+        flexWrap="wrap"
+        css={css({
+          '> div': {
```
VWSCoronaDashboard8, line 28: Are you targeting the LinkGroup component here? You could do that explicitly via `${LinkGroup}` if LinkGroup is a styled-component.

If I understand the code correctly another (I think cleaner) solution could be to just wrap the LinkGroup in a Box where you apply these styles

packages/app/src/domain/topical/link-block/data-sitemap/use-data-sitemap-links.tsx
```diff
@@ -0,0 +1,194 @@
+import siteText from '~/locale/index';
+import { LinkGroupProps } from './../link-group';
+
+export function useDataSitemapLinks(
```
VWSCoronaDashboard8, line 4: This hook is returning more than links, so I think this could be named `useDataSitemap`

packages/app/src/domain/topical/link-block/link-block.tsx
```diff
@@ -0,0 +1,71 @@
...
+}
+
+const Container = styled(Box)(
+  css({
+    flexDirection: asResponsiveArray({ _: 'column', md: 'row' }),
+    display: 'flex',
+    px: 4,
+    py: 4,
+
+    '> div:not(:last-child)::after': {
```
VWSCoronaDashboard8, line 44: Looking at the markup above wouldn't `'> div:not(:last-child)` target the Sitemap component always? Would it be possible to wrap that in a Box and apply these styles there?

packages/app/src/domain/topical/link-block/link-block.tsx
```diff
@@ -0,0 +1,71 @@
...
+  })
+);
+
+const Sitemap = styled(Box)(
+  css({
+    flexBasis: asResponsiveArray({ _: '66%', lg: '75%' }),
+    pt: asResponsiveArray({ _: 4, md: 0 }),
+    pl: asResponsiveArray({ _: 0, md: 4 }),
+  })
+);
```
VWSCoronaDashboard8, line 71: I don't see an advantage in creating these components separately. If you use Box directly in the markup, then you can apply these styles more easily because then you don't need the asResponsiveArray helper functions.

The same goes for Container if you manage to remove the child selector

packages/app/src/domain/topical/link-block/link-group.tsx
```diff
@@ -0,0 +1,73 @@
...
+        iconPlacement="right"
+        fontWeight="bold"
+      >
+        {text}
+      </LinkWithIcon>
+    </Item>
+  );
+}
+
+const StyledHeader = styled(InlineText)(
```
VWSCoronaDashboard8, line 53: Inline text renders as a span by default. That seems an odd choice for a header. Maybe you can render the container box as "section" and apply a real heading component here to make it more semantically correct?

packages/app/src/domain/topical/link-block/data-sitemap/data-sitemap.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { Heading } from '~/components-styled/typography';
+import { Box } from '~/components-styled/base';
+import siteText from '~/locale/index';
+import { asResponsiveArray } from '~/style/utils';
+import { LinkGroup } from './../link-group';
+import { useDataSitemapLinks } from './use-data-sitemap-links';
+
+export type DataSitemapProps = {
+  base: 'landelijk' | 'veiligheidsregio' | 'gemeente';
+  code?: string | string[];
```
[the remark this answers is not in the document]
VWSCoronaDashboard13: yeah, I was confused about a type so I didn't want to change it, but I got clarification and updated this and casted the code as a string before passing it in 👍

packages/app/src/domain/topical/link-block/data-sitemap/data-sitemap.tsx
VWSCoronaDashboard13: good to know there are predefined styles! There is a defined style for h3

packages/app/src/domain/topical/link-block/link-block.tsx
VWSCoronaDashboard13: I made these separate components because I felt that it looked really cluttered in the jsx and made it more difficult to read and I had seen this done in other files within the project. What is the agreed upon approach on deciding when to create a separate component vs. directly in the markup?

packages/app/src/domain/topical/link-block/link-group.tsx
VWSCoronaDashboard13: This was copied from the existing DataSitemap component so I'm not sure if that was done intentionally? This component refers to this so I don't think it makes sense to render these as individual sections 
image
image
I think a span still makes sense in this situation. What do you think?

packages/app/src/domain/topical/link-block/link-block.tsx
VWSCoronaDashboard8: I don't think we have a rule for that. But in general I think it makes sense to use props all props are supported. For example these three props all all part of the Box interface so you wouldn't need the css function to cover any of them. 

Having these components named and removing the implementation from the root markup is kind of nice, but the same time you loose overview of that those components are doing. In this case I would prefer to see these props directly but maybe some colleagues have different opinion. I'll share this question in Slack...

packages/app/src/domain/topical/link-block/link-group.tsx
VWSCoronaDashboard8: To me all the black things look like they should be some level of heading element

VWSCoronaDashboard1: @VWSCoronaDashboard8 Those headers can't really be headings, we just render them as such. But semantically we didn't want them to render as H* elements because of screenreaders.

VWSCoronaDashboard8: I don't get why a screenreader would be bothered by knowing that verpleeghuiszorg falls under kwetsbare groepen. If its relevant information for us, why is it irrelevant for visually impaired?

VWSCoronaDashboard8 (approved):

packages/app/src/domain/topical/link-block/link-group.tsx
VWSCoronaDashboard1: Mainly because it would clutter up the main navigation of the page. These link would need to end up somewhere in the hierarchy, but they'd be the same after each page navigation.

VWSCoronaDashboard1: But it has been stated before that we'd like to see a follow-up to the original accessibility analysis, to see if this approach is correct or not. It's quite confusing, to be honest, what exactly is best practice here.

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard1 (approved):

packages/app/src/domain/topical/link-block/use-data-sitemap.tsx
```diff
@@ -0,0 +1,195 @@
+import siteText from '~/locale/index';
+import { LinkGroupProps } from './link-group';
+
+export function useDataSitemap(
+  base: 'landelijk' | 'veiligheidsregio' | 'gemeente',
+  code?: string,
+  data?: any
```
VWSCoronaDashboard4, line 7: This can be typed as

```suggestion
  data?: National | Regionaal | Municipal
```

I guess?

packages/app/src/pages/actueel/gemeente/[code].tsx
```diff
@@ -187,36 +190,43 @@ const TopicalMunicipality: FCWithLayout<typeof getStaticProps> = (props) => {
...
-                      municipalityName: municipalityName,
-                    }
-                  ),
-                },
-              ]}
-            />
+            <CollapsibleButton
+              label={siteText.common_actueel.overview_links_header}
+            >
+              <LinkBlock
```
Comment (VWSCoronaDashboard4) [inline packages/app/src/pages/actueel/gemeente/[code].tsx:196]: Maybe this component could be called the `SiteMap` component? It currently has a fairly generic name but very specific properties.

packages/app/src/pages/actueel/veiligheidsregio/[code].tsx
```diff
@@ -179,23 +182,32 @@ const TopicalSafetyRegion: FCWithLayout<typeof getStaticProps> = (props) => {
...
-                  ),
-                },
-                { href: '/gemeente', text: text.quick_links.links.gemeente },
-              ]}
-            />
+            <CollapsibleButton
+              label={siteText.common_actueel.overview_links_header}
+            >
+              <LinkBlock
+                quickLinksHeader={text.quick_links.header}
```
Comment (VWSCoronaDashboard4) [inline packages/app/src/pages/actueel/veiligheidsregio/[code].tsx:189]: Not sure if the following would improve this code per se, or not:

Maybe this logic can also be moved to its own utility function, like done with the `useDataSitemap`? When done, you can pass props to `LinkBlock` for the code and data etc and `useDataSitemap` can be called from inside this component.

packages/app/src/domain/topical/link-block/use-data-sitemap.tsx
```diff
@@ -0,0 +1,195 @@
+import siteText from '~/locale/index';
+import { LinkGroupProps } from './link-group';
+
+export function useDataSitemap(
```
VWSCoronaDashboard4, line 4: This utility is named to look like a hook, but I don't think it's actually a hook.

packages/app/src/domain/topical/link-block/use-data-sitemap.tsx
VWSCoronaDashboard13: 🤩

packages/app/src/domain/topical/link-block/use-data-sitemap.tsx
VWSCoronaDashboard13: ah you are right. I will change this to a util file and update the name to `getDataSitemap`

packages/app/src/pages/actueel/veiligheidsregio/[code].tsx
Comment (VWSCoronaDashboard13) [reply packages/app/src/pages/actueel/veiligheidsregio/[code].tsx:189]: Some of the logic seems very specific, so I wanted to leave this as is since I'm not sure exactly how all of it works. Since they're quick links I thought it made sense to keep them configurable anyways. Are you ok with leaving this as is?

PR #2267, opened 2021-03-11
Add build script and update Docker file

VWSCoronaDashboard:
## Summary

The builds in Docker are slow. We're changing the build process so the static build of the application gets built outside of the Docker context. We then copy the results of the build in an Nginx container.

## Motivation

It's an easy step we can take to reduce build-times.

## Detailed design

- We've added a `build.sh` bash script that needs to be run with the correct environment variables
- Docker copies the results from the `exports` folder containing the Dutch and English static build of the site into the Docker image
- Docker retrieves the Nginx configuration files

## Instructions for pipeline
`bash build.sh //make sure to provide the right env vars here`
`docker build -t cdb:latest . //no env vars needed`


build.sh
```diff
@@ -0,0 +1,41 @@
...
+# Prepare the export folders
+mkdir -p exports/nl
+mkdir -p exports/en
+
+# Install dependencies
+yarn
+yarn workspace @corona-dashboard/common build
+
+# Validate data
+yarn workspace @corona-dashboard/cli validate-json
```
VWSCoronaDashboard10, line 17: don't we also want to validate the last values?

VWSCoronaDashboard: I've re-added the line and enabled it by default. We're not using Docker anywhere but production anymore. There's no reason to make it conditional :).

PR #2282, opened 2021-03-12
Keep icon from wrapping onto it's own line in LinkWithIcon

VWSCoronaDashboard13:
## Summary

Keep icon from wrapping onto it's own line in LinkWithIcon

## Detailed design

Used existing fix of wrapping the last word and the icon in a span and applied it to all links returned by LinkWithIcon

### DIFF
--- a/packages/app/src/components-styled/link-with-icon.tsx
+++ b/packages/app/src/components-styled/link-with-icon.tsx
@@ -3,6 +3,7 @@ import { ReactNode } from 'react';
 import { UrlObject } from 'url';
 import { Link } from '~/utils/link';
 import { Box } from './base';
+import styled from 'styled-components';
 
 interface LinkWithIconProps {
   href: UrlObject | string;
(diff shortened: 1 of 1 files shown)


VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/link-with-icon.tsx
```diff
@@ -49,28 +50,34 @@ export function LinkWithIcon({
...
+              <IconSmall icon={icon} width={11} height={10} />
+              {words[words.length - 1]}
+            </IconWrapper>
           </>
         )}
         {headingLink && (
           <Box paddingRight={isSingleWord ? `calc(0.5rem + 18px)` : ''}>
             {!words.length ? children : firstWords}
-            <span css={css({ display: 'inline-block' })}>
+            <IconWrapper>
```
VWSCoronaDashboard8, line 72: You could use `<InlineText>` (which renders as span) with the display and textDecoration properties directly I think.

VWSCoronaDashboard8 (approved):

packages/app/src/components-styled/link-with-icon.tsx
VWSCoronaDashboard13: yeah, that also works, but I pulled it out to a separate component since it's used 3 times so I figured it would be easier to have one place to edit the styles

packages/app/src/components-styled/link-with-icon.tsx
```diff
@@ -101,3 +108,10 @@ export function LinkWithIcon({
     );
   }
 }
+
+const IconWrapper = styled.span(
+  css({
+    display: 'inline-block',
+    textDecoration: 'inherit',
```
VWSCoronaDashboard, line 115: `white-space: nowrap` might also work. Not sure if one is better than the other :).

VWSCoronaDashboard (approved):

PR #2284, opened 2021-03-12
Get rid of highcharts 🥳

VWSCoronaDashboard10:
### changes to the `<TimeSeriesChart />`:

- Fix a broken tooltip on incomplete time series
- Improve mouse event listening by capturing mouse events on the container. Previously a rect was drawn which didn't caught all mouse movements (and thus not updating tooltip position)
- Create option (`DataOptions.isNearestPointOnly`) to interact with a single marker instead of all markers on the matching Y-axis
- Allow tooltips to show a short label, configured with `SeriesConfig.shortLabel`
- Improve hover state UI rendering by only storing the point in state
- create tooltip click handler prop (`onTooltipClick`)

#### after PR feedback

- Fix server side rendering, this also removed the need of the `<ParentSize />` component
- Refactor series click handler by passing it the selected serie (and value, if needed)
- Rename `DataOptions.isNearestPointOnly` option to `markNearestPointOnly`
- The SeriesConfig is now a generic type, this results in a smarter click handler which knows exactly what types it could return.
- The DateMarker is now always rendered, removed the configurability.

### Highcharts
- Port the behavior line chart to visx
- Port the "besmettelijke mensen" line chart to visx
- buhbye highcharts! 👋 

### Todo
- implement new tooltip layout
- improve tooltip positioning, especially on narrow screens


packages/app/src/components-styled/time-series-chart/components/range-trend.tsx
```diff
@@ -25,14 +27,20 @@ export function RangeTrend({
 }: RangeTrendProps) {
   const id = useUniqueId();
 
+  const nonNullSeries = useMemo(
+    () =>
+      series.filter((x) => isPresent(x.__value_a) && isPresent(x.__value_b)),
+    [series]
+  );
```
VWSCoronaDashboard8, line 34: nit; I think a hook useNonNullSeries would be nice. Then hook can then also include the logic/switch between single and double values.

packages/app/src/components-styled/time-series-chart/components/tooltip.tsx
```diff
@@ -138,6 +138,32 @@ export function DefaultTooltip<T extends TimestampedValue>({
...
+        if (x.type === 'range') {
+          return (
+            x.metricPropertyLow === __valueKey ||
+            x.metricPropertyHigh === __valueKey
+          );
+        }
+
+        if (x.type === 'line' || x.type === 'area') {
+          return x.metricProperty === __valueKey;
+        }
```
VWSCoronaDashboard8, line 157: Looking at this logic I now think that it would be better to get the seriesConfigIndex for the nearest point passed to the tooltip instead of the valueKey.

Since valueKey === metricProperty and that's something you can get from config + config index. Then you don't have to make this distinction between types here and it would be trivial to just pick the config out of seriesConfig straight away, without filtering the config list.

Could you add a @todo maybe for future improvement? Also the only reason I need it __valueKey is to shut up the linter while keeping the variable visible. So now you can just name it valueKey

packages/app/src/components-styled/time-series-chart/logic/common.ts
```diff
@@ -2,6 +2,7 @@ export interface DataOptions {
   valueAnnotation?: string;
   forcedMaximumValue?: number;
   isPercentage?: boolean;
+  isNearestPointOnly?: boolean;
```
VWSCoronaDashboard8, line 5: The "is" part of this throws me off.

Some suggestions showNearestPointOnly markNearestPointOnly highlightNearestPointOnly

packages/app/src/components-styled/time-series-chart/logic/scales.ts
```diff
@@ -73,9 +73,9 @@ export function useScales<T extends TimestampedValue>(args: {
       xScale,
       yScale,
       getX: (x: SeriesItem) => xScale(x.__date_unix),
-      getY: (x: SeriesSingleValue) => yScale(x.__value),
-      getY0: (x: SeriesDoubleValue) => yScale(x.__value_a),
-      getY1: (x: SeriesDoubleValue) => yScale(x.__value_b),
+      getY: (x: SeriesSingleValue) => yScale(x.__value || NaN),
+      getY0: (x: SeriesDoubleValue) => yScale(x.__value_a || NaN),
+      getY1: (x: SeriesDoubleValue) => yScale(x.__value_b || NaN),
```
VWSCoronaDashboard8, line 78: What's the advantage of using NaN here over undefined?

packages/app/src/components-styled/time-series-chart/logic/series.ts
```diff
@@ -106,11 +109,11 @@ export type SeriesItem = {
...
 };
 export interface SeriesSingleValue extends SeriesItem {
-  __value: number;
+  __value: number | null;
 }
 export interface SeriesDoubleValue extends SeriesItem {
-  __value_a: number;
-  __value_b: number;
+  __value_a: number | null;
+  __value_b: number | null;
```
VWSCoronaDashboard8, line 116: I guess we could also just make these optional properties with some small adaptations, which would be a little more inline with our effort to avoid null. Maybe later...

packages/app/src/components-styled/time-series-chart/time-series-chart.tsx
```diff
@@ -111,6 +110,8 @@ export type TimeSeriesChartProps<T extends TimestampedValue> = {
    * contain things that are essential to rendering a full tooltip layout
    */
   dataOptions?: DataOptions;
+  disableLegend?: boolean;
+  onTooltipClick?: (x: TooltipData<T>) => void;
```
VWSCoronaDashboard8, line 114: Not sure that this means... The event that opens the tooltip? A click inside an already open tooltip?

What is it for? It seems like you are opening up the component for external tooltip logic, but I wonder in what situation we would need that. Something to do with mobile?

VWSCoronaDashboard10: Yes I was also not really fond of the name .. Its a click handler which should receive some information about the currently "active" serie, hence the tooltipData. 
The click handler is used by the behavior line chart: a click "on a line" activates that line.

What would you suggest?

packages/app/src/components-styled/time-series-chart/logic/scales.ts
VWSCoronaDashboard10: xScale/yScale doesn't accept undefined

VWSCoronaDashboard:
Amazing work. 🙌

packages/app/src/components-styled/time-series-chart/logic/scales.ts
```diff
@@ -73,9 +73,9 @@ export function useScales<T extends TimestampedValue>(args: {
       xScale,
       yScale,
       getX: (x: SeriesItem) => xScale(x.__date_unix),
-      getY: (x: SeriesSingleValue) => yScale(x.__value),
-      getY0: (x: SeriesDoubleValue) => yScale(x.__value_a),
-      getY1: (x: SeriesDoubleValue) => yScale(x.__value_b),
+      getY: (x: SeriesSingleValue) => yScale(x.__value || NaN),
```
VWSCoronaDashboard4, line 76: shouldn't it be `?? NaN` in case `__value === 0` ?

packages/app/src/components-styled/time-series-chart/logic/hover-state.ts
```diff
@@ -125,102 +128,141 @@ export function useHoverState<T extends TimestampedValue>({
...
-          switch (config.type) {
-            case 'line':
-            case 'area':
-              return {
+    /**
+     * Point markers on range data are rendered differently, so we split them
+     * out here, so we avoid having to create a union type and complicate
+     * things.
+     */
+    const rangePoints: HoveredPoint<T>[] = seriesConfig
```
VWSCoronaDashboard4, line 188: Maybe this (and other calculated variables here) could also use the `useMemo`?

packages/app/src/pages/landelijk/besmettelijke-mensen.tsx
```diff
@@ -77,33 +76,35 @@ const InfectiousPeople: FCWithLayout<typeof getStaticProps> = (props) => {
...
-                  },
-                  {
-                    label: text.legenda_marge,
-                    color: 'data.margin',
-                    shape: 'square',
-                  },
-                ]}
-              />
-            </>
+            <ParentSize>
```
VWSCoronaDashboard4, line 79: :+1:

packages/app/src/domain/behavior/behavior-line-chart-tile.tsx
```diff
@@ -92,30 +95,40 @@ export function BehaviorLineChartTile({
...
-                      end: value.date_end_unix,
-                    },
-                  } as Value)
-                : undefined
-            )
-            .filter(isPresent)
+      <ParentSize>
+        {({ width }) => (
+          <TimeSeriesChart
+            title={'Gedragsregels'}
```
VWSCoronaDashboard4, line 101: Hardcoded copy?

packages/app/setupTests.js
```diff
@@ -1,6 +1,6 @@
 import '@testing-library/jest-dom/extend-expect';
 
-// Disable dynamic imports for now (Sorry HighCharts)
+// Disable dynamic imports for now
```
VWSCoronaDashboard4, line 3: `next/dynamic` is used in `packages/app/src/components/charts/index.tsx`. That file and this mock can be removed, I guess?

VWSCoronaDashboard4 (approved):
Nice! Some small comments. I trust other devs with regard to more in depth knowledge about the line charts

packages/app/src/components-styled/time-series-chart/time-series-chart.tsx
VWSCoronaDashboard8: onElementClick or onSeriesClick? Also if possible pass only what is needed by the outside world, because I can imagine that valueKey (and maybe value itself) for example would be enough. 

I think it's confusing to pass the whole tooltip state, because then I would expect some external tooltip to use it.

packages/app/src/components-styled/time-series-chart/logic/hover-state.ts
VWSCoronaDashboard8: I think that would only have a benefit if the seriesConfig changes more frequently than the input values. I believe currently the values and chart config are all pretty much static, and since this part is wrapped in a useCallback it should then only be executed once.

VWSCoronaDashboard8: On second thought the useCallback wouldn't help with that and things would still be executed in runtime, but it looks like this is part of hoverState calculation which has a useMemo wrapper.

packages/app/src/components-styled/time-series-chart/time-series-chart.tsx
VWSCoronaDashboard8: .... onSeriesSelect?

packages/app/src/domain/behavior/behavior-line-chart-tile.tsx
```diff
@@ -92,30 +95,40 @@ export function BehaviorLineChartTile({
...
+              isPercentage: true,
+              showNearestPointOnly: true,
+            }}
+            tickValues={[0, 25, 50, 75, 100]}
+            showDateMarker
+            onTooltipClick={(x) =>
+              setCurrentId(
+                x.valueKey.replace(`_${type}`, '') as BehaviorIdentifier
+              )
+            }
```
VWSCoronaDashboard8, line 128: Yes, it looks like all you need is valueKey. So `onSeriesSelect(valueKey: keyof<T>)` would be a clean and minimal API

packages/app/src/domain/behavior/behavior-line-chart-tile.tsx
```diff
@@ -92,30 +95,40 @@ export function BehaviorLineChartTile({
...
+                strokeWidth: x.valueKey === selectedValueKey ? 3 : 2,
+                color:
+                  x.valueKey === selectedValueKey
+                    ? colors.data.primary
+                    : '#E7E7E7',
+              }))}
+            disableLegend
+            dataOptions={{
+              isPercentage: true,
+              showNearestPointOnly: true,
```
VWSCoronaDashboard8, line 120: I doubt if this is an option for the data. I feel it's more an option for the chart itself. Just like disableLegend and showDateMarker 

The latter should I think be removed all together since all of our charts will now show dates on the x-axis I think. At least it should be the default. But that can be picked up outside this PR of course.

packages/app/src/components-styled/time-series-chart/logic/scales.ts
VWSCoronaDashboard10: good catch!

packages/app/src/components-styled/time-series-chart/components/range-trend.tsx
VWSCoronaDashboard10: I looked into it but couldn't get the types to work nicely, I'll leave it as it for now.

packages/app/src/components-styled/time-series-chart/components/point-markers.tsx
```diff
@@ -55,7 +56,9 @@ export function PointMarkers<T extends TimestampedValue>(
 ) {
   const { points, size = MARKER_POINT_SIZE } = props;
 
-  if (!points.length) return null;
+  const nonNullPoints = points.filter((x) => isPresent(x.y));
+
+  if (!nonNullPoints.length) return null;
```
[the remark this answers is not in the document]
VWSCoronaDashboard10: I see that I already filter the points in the hover-state, this check here is obsolete 👍

packages/app/src/domain/behavior/behavior-line-chart-tile.tsx
VWSCoronaDashboard10: I've refactored it to return the selected `seriesConfig`, I assumed that would make more sense when for example working with a `range`: the returned key could be both the `metricPropertyLow` and `metricPropertyHigh` for a single serie. Now it will simply yield the serie config itself.

packages/app/src/components-styled/time-series-chart/time-series-chart.tsx
```diff
@@ -111,26 +113,35 @@ export type TimeSeriesChartProps<T extends TimestampedValue> = {
    */
   dataOptions?: DataOptions;
   disableLegend?: boolean;
-  onTooltipClick?: (x: TooltipData<T>) => void;
+  onSeriesClick?: (serieConfig: C[number], value: T) => void;
```
VWSCoronaDashboard8, line 116: I think serie is a Dutch word. A single series of numbers would still be called a series I believe in EN. Maybe just call this `config` or even `seriesConfig` is ok with me since we still have TS help out.

packages/app/src/components-styled/time-series-chart/time-series-chart.tsx
```diff
@@ -111,26 +113,35 @@ export type TimeSeriesChartProps<T extends TimestampedValue> = {
...
+
+  /**
+   * By default markers for all series are displayed on hover, also the tooltip
+   * will display all series of the selected X-value. The `markNearestPointOnly`
+   * will result in a user interacting with the single nearest point only.
+   */
+  markNearestPointOnly?: boolean;
```
VWSCoronaDashboard8, line 123: 👍 Yay for comments

packages/app/src/components-styled/time-series-chart/time-series-chart.tsx
```diff
@@ -140,10 +151,11 @@ export function TimeSeriesChart<T extends TimestampedValue>({
     tooltipOpen,
   } = useTooltip<TooltipData<T>>();
 
+  const [sizeRef, { width }] = useElementSize<HTMLDivElement>(840);
```
VWSCoronaDashboard8, line 154: What's the 840 for? Maybe place it in a constant with a descriptive name. So you choose to make every chart responsible for figuring out its width automatically? Seems like a good idea, since we can always control outside width by resizing the container I guess 👍  

Maybe later we can define an aspect ratio per chart, but that's up to design I guess.

VWSCoronaDashboard8 (approved):

packages/app/src/components-styled/time-series-chart/time-series-chart.tsx
VWSCoronaDashboard10: Yep, I think it perfectly makes sense to have width be defined by the parent container. I'll move it to a constant 👍

PR #2290, opened 2021-03-12
Add CoronaMelder warnings over time chart

VWSCoronaDashboard14:
Used the new `TimeSeriesChart`


packages/app/src/locale/en.json
```diff
@@ -2715,6 +2715,21 @@
         "text": "Most recent factsheet CoronaMelder app",
         "href": "https://www.coronamelder.nl/media/Factsheet_Corona_latest.pdf"
       }
+    },
+    "linechart": {
```
VWSCoronaDashboard8, line 2719: Line chart is probably not the best identifier here. Better pick a Dutch name for the chart that is also about the subject.

packages/app/src/pages/landelijk/gedrag.tsx
```diff
@@ -75,6 +80,22 @@ const BehaviorPage: FCWithLayout<typeof getStaticProps> = ({
...
+          values={data.corona_melder_app.values}
+          linesConfig={[
+            {
+              metricProperty: 'warned_daily',
+            },
+          ]}
+          metadata={{
+            source: corona_melder_app.linechart.bronnen.coronamelder,
+          }}
+        />
```
VWSCoronaDashboard8: Why did you choose to use the normal LineChart here?

packages/app/src/pages/landelijk/gedrag.tsx
```diff
@@ -75,6 +80,22 @@ const BehaviorPage: FCWithLayout<typeof getStaticProps> = ({
           reference={nl_gedrag.reference}
         />
 
+        <LineChartTile
+          timeframeOptions={['all', '5weeks', 'week']}
+          title={corona_melder_app.linechart.title}
+          ariaDescription={''}
```
VWSCoronaDashboard8: It's required for a reason, so maybe fill it with some locale text. 

We should maybe have a central discussion about the need for this property, since I'm not 100% sure it adding much value at the moment.

packages/app/src/pages/landelijk/gedrag.tsx
```diff
@@ -177,6 +198,38 @@ const BehaviorPage: FCWithLayout<typeof getStaticProps> = ({
...
+          {(timeframe) => (
+            <ParentSize>
+              {({ width }) => (
+                <TimeSeriesChart
+                  title={corona_melder_app.linechart.tooltip.title}
+                  timeframe={timeframe}
+                  width={width}
+                  values={data.corona_melder_app.values}
+                  showDateMarker
+                  ariaLabelledBy=""
```
VWSCoronaDashboard8: Would be good to fill this with something too

packages/app/src/pages/landelijk/gedrag.tsx
```diff
@@ -177,6 +198,38 @@ const BehaviorPage: FCWithLayout<typeof getStaticProps> = ({
...
+            source: corona_melder_app.linechart.bronnen.coronamelder,
+          }}
+          title={corona_melder_app.linechart.title}
+          description={corona_melder_app.linechart.description}
+        >
+          {(timeframe) => (
+            <ParentSize>
+              {({ width }) => (
+                <TimeSeriesChart
+                  title={corona_melder_app.linechart.tooltip.title}
```
VWSCoronaDashboard8, line 196: By default the tooltip often has the same title as the chart. If you want to give the tooltip a different title I would suggest to call that property `tooltip_title` in lokalize, because the other tooltip thing is not specifically for the tooltip (see comment below)

packages/app/src/pages/landelijk/gedrag.tsx
```diff
@@ -177,6 +198,38 @@ const BehaviorPage: FCWithLayout<typeof getStaticProps> = ({
...
+                  width={width}
+                  values={data.corona_melder_app.values}
+                  showDateMarker
+                  ariaLabelledBy=""
+                  paddingLeft={40}
+                  seriesConfig={[
+                    {
+                      type: 'area',
+                      metricProperty: 'warned_daily',
+                      label: corona_melder_app.linechart.tooltip.warnings,
```
VWSCoronaDashboard8, line 207: This label is not just for the tooltip. It is normally also for the legend, but it so happens that this chart is rendering one trend and therefor doesn't have a legend enabled. So a better / more general name would be `labels.warnings` or something. If we start calling this by tooltip it will confuse people later maybe or people start copying this pattern.

packages/app/src/pages/landelijk/gedrag.tsx
VWSCoronaDashboard8: Ah I see you already removed it :)

packages/app/src/locale/en.json
VWSCoronaDashboard8: Suggestion: `waarschuwingen_over_tijd` or since there is only one chart here `grafiek` would also do although that doesn't scale well in case we ever want to show more charts.

packages/app/src/pages/landelijk/gedrag.tsx
VWSCoronaDashboard14: I was comparing it to the new chart but forgot to remove it

VWSCoronaDashboard4 (approved):

PR #2298, opened 2021-03-15
Fix flickering bug menu

VWSCoronaDashboard14:
By using the `has-no-js` and animation the menu is always closed, and if there is no javascript available it expands after a second.


VWSCoronaDashboard4 (approved):

VWSCoronaDashboard10 (commented):
I have the feeling this component became too complex. Can you rewrite/simplify it to use the same logic as we use in the collapsible-section? (measuring the content-height and animating the height of a container element, instead of the maxheight-thing).

VWSCoronaDashboard14:
@VWSCoronaDashboard10 you are right, seems like my implementation is not suitable for a hosted static build. This can be merged yet and I'll have to take another look. I'll put it on hold and take another look later today.

VWSCoronaDashboard14:
@VWSCoronaDashboard10 this should be the right implementation that we've made. Sadly animating the height to auto doesn't work with a transition, hence why I had to choose the `maxHeight` solution (works identical).

PR #2304, opened 2021-03-16
Prevent the background rectangle from rendering outside the x-axis

VWSCoronaDashboard10:
## Summary

This PR fixes an issue where the background rectangle of a chart could get rendered outside of the x-axis bounds. It will now be clipped to the x-axis-domain of the values.

image


VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/line-chart/logic/background-rectangle.tsx
```diff
@@ -20,8 +20,30 @@ export function addBackgroundRectangleCallback(
...
+         */
+        const [rectangleStartMs, rectangleEndMs] = dateRange.map((x) =>
+          x.getTime()
+        );
+
+        /**
+         * Create clipped backround rectangle range to fit within displayed range
+         */
+        const dateRangeClipped = [
+          new Date(Math.max(rectangleStartMs, displayedStartMs)),
```
VWSCoronaDashboard4, line 41: (Not sure if I understand the code correctly, so take this with a grain of salt:)

Shouldn't one of these be a `Math.min` to prevent clipping the other way?

VWSCoronaDashboard4 (approved):

packages/app/src/components-styled/line-chart/logic/background-rectangle.tsx
VWSCoronaDashboard10: nope, this is the minimum value if the range, and the minimum value should be the rectangle min value as long as it isn't lower than the displayed min value.

PR #2319, opened 2021-03-17
Schema for Vaccinated per age group

VWSCoronaDashboard4:
## Summary

Schema for Vaccinated per age group

The approach is equal to other age group schema's. Per age group there's the `age_group_range` with a total and percentage of the population.

There can be an `age_group_range` with value of eg `18+` to cover the total.


packages/app/schema/nl/__index.json
```diff
@@ -164,6 +164,9 @@
     },
     "vaccine_administered_planned": {
       "$ref": "vaccine_administered_planned.json"
+    },
+    "vaccinated_per_age_group": {
```
VWSCoronaDashboard10, line 168: Don't we use "vaccine_administered" instead of "vaccinated"? In that case I think this should be called `vaccine_administered_per_age_group`

packages/app/schema/nl/vaccinated_per_age_group.json
```diff
@@ -0,0 +1,58 @@
...
+        },
+        "age_group_total": {
+          "type": "number"
+        },
+        "fully_vaccinated": {
+          "type": "number"
+        },
+        "partially_vaccinated": {
+          "type": "number"
+        },
```
VWSCoronaDashboard10, line 45: can these be of type integer?

packages/app/schema/nl/__index.json
VWSCoronaDashboard8: Yes, please use the same subject prefix as the others.

packages/app/schema/nl/vaccinated_per_age_group.json
```diff
@@ -0,0 +1,58 @@
...
+  "definitions": {
+    "value": {
+      "title": "nl_vaccinated_per_age_group_value",
+      "additionalProperties": false,
+      "type": "object",
+      "required": [
+        "age_group_percentage",
+        "age_group_range",
+        "age_group_total",
+        "coverage",
```
VWSCoronaDashboard8, line 24: I think coverage is a little vague. I am assuming it's a percentage and it's the percentage fully vaccinated. So I would suggest to call it `fully_vaccinated_percentage`

packages/app/schema/nl/vaccinated_per_age_group.json
```diff
@@ -0,0 +1,58 @@
...
+      ],
+      "properties": {
+        "age_group_range": {
+          "type": "string"
+        },
+        "age_group_percentage": {
+          "type": "number"
+        },
+        "age_group_total": {
+          "type": "number"
```
VWSCoronaDashboard8, line 38: This should be integer I think

VWSCoronaDashboard8 (changes requested):

packages/app/schema/nl/__index.json
VWSCoronaDashboard4: This is specifically about (partially) vaccinated people, not about administered doses.

VWSCoronaDashboard10: But doesn't 1 administered vaccine === 1 dose applied === 1 (partially) vaccinated human being?

packages/app/schema/nl/vaccine_coverage_per_age_group.json
```diff
@@ -0,0 +1,58 @@
...
+      "title": "nl_vaccine_coverage_per_age_group_value",
+      "additionalProperties": false,
+      "type": "object",
+      "required": [
+        "age_group_percentage",
+        "age_group_range",
+        "age_group_total",
+        "fully_vaccinated_percentage",
+        "date_of_insertion_unix",
+        "date_unix",
```
VWSCoronaDashboard8, line 26: You might want to add `date_of_report_unix` as well. It's not a requirement maybe but we are going to retroactively add it to all existing schema's AFAIK, so we might as well add it to all new ones to save some time.

VWSCoronaDashboard8 (approved):

packages/app/schema/nl/__index.json
```diff
@@ -41,7 +41,8 @@
     "vaccine_administered_lnaz",
     "vaccine_administered_total",
     "vaccine_administered_rate_moving_average",
-    "vaccine_administered_planned"
+    "vaccine_administered_planned",
+    "vaccine_coverage_per_age_group"
```
VWSCoronaDashboard8, line 45: I suggest to take it out of the required list so it can be merged before the data is there. Then also don't forget to re-run generate typescript

VWSCoronaDashboard8 (approved):

packages/common/src/types/data.ts
```diff
@@ -680,6 +681,20 @@ export interface NlVaccineAdministeredPlannedValue {
...
+export interface NlVaccineCoveragePerAgeGroupValue {
+  age_group_range: string;
+  age_group_percentage: number;
+  age_group_total: number;
+  fully_vaccinated: number;
+  partially_vaccinated: number;
+  fully_vaccinated_percentage: number;
+  date_unix: number;
+  date_of_insertion_unix: number;
+  date_of_report_unix?: number;
```
VWSCoronaDashboard8, line 696: You forgot to add date of report to the required list

VWSCoronaDashboard8 (changes requested):

PR #2341, opened 2021-03-18
Add source to positive tested choropleth on topical page

VWSCoronaDashboard14:
Initially, I wanted to use the `metadata` component but I think it is a bit too tied to the tiles.

Screenshot 2021-03-18 at 09 56 12


VWSCoronaDashboard1 (approved):

packages/app/src/pages/actueel/gemeente/[code].tsx
```diff
@@ -387,6 +388,26 @@ const TopicalMunicipality: FCWithLayout<typeof getStaticProps> = (props) => {
...
+                        siteText.positief_geteste_personen.bronnen.rivm
+                          .aria_text
+                      }
+                      href={
+                        siteText.positief_geteste_personen.bronnen.rivm.href
+                      }
+                    >
+                      {siteText.positief_geteste_personen.bronnen.rivm.text}
+                    </ExternalLink>
+                  </Text>
```
Comment (VWSCoronaDashboard8) [inline packages/app/src/pages/actueel/gemeente/[code].tsx:410]: It looks like this part ideally should be captured in a reusable component. It's a lot of details to place directly on the root of a page layout.

packages/app/src/pages/actueel/veiligheidsregio/[code].tsx
```diff
@@ -366,6 +367,26 @@ const TopicalSafetyRegion: FCWithLayout<typeof getStaticProps> = (props) => {
...
+                        siteText.positief_geteste_personen.bronnen.rivm
+                          .aria_text
+                      }
+                      href={
+                        siteText.positief_geteste_personen.bronnen.rivm.href
+                      }
+                    >
+                      {siteText.positief_geteste_personen.bronnen.rivm.text}
+                    </ExternalLink>
+                  </Text>
```
Comment (VWSCoronaDashboard8) [inline packages/app/src/pages/actueel/veiligheidsregio/[code].tsx:389]: And if you're repeating the same thing again, then even more so.

packages/app/src/pages/index.tsx
```diff
@@ -342,6 +343,27 @@ const Home: FCWithLayout<typeof getStaticProps> = (props) => {
...
+                          .aria_text
+                      }
+                      href={
+                        siteText.positief_geteste_personen.bronnen.rivm.href
+                      }
+                    >
+                      {siteText.positief_geteste_personen.bronnen.rivm.text}
+                    </ExternalLink>
+                  </Text>
+
```
VWSCoronaDashboard8, line 366: So yes, if you're repeating yourself 3 times I think that's the threshold. Then it should definitely be refactored into a component.

VWSCoronaDashboard8 (changes requested):

VWSCoronaDashboard14:
@VWSCoronaDashboard8 I've integrated this in the `metadata` component now

packages/app/src/components-styled/metadata.tsx
```diff
@@ -28,16 +36,31 @@ export function Metadata({ date, source, obtained }: MetadataProps) {
...
+              ` ${replaceVariablesInText(siteText.common.metadata.obtained, {
+                date: formatDateFromSeconds(obtained, 'weekday-medium'),
+              })}`}
+            {dateString && source ? ' · ' : null}
+            {source
+              ? `${siteText.common.metadata.source}: ${source.text}`
+              : null}
+          </Text>
+        </Box>
+      )}
```
VWSCoronaDashboard8, line 63: There doesn't seem to be a clear reason why these two variations of metadata are combined into a single component. I don't mind this for now, but I feel that ideally it should be two separate components. Metadata (or ContentHeaderMetadata) and TileMetadata for example.

VWSCoronaDashboard8 (approved):
Much better 👍

PR #2342, opened 2021-03-18
Fix SSR charts and choropleths

VWSCoronaDashboard10:
## Summary

By adding a viewbox prop to the svg the chart will scale down instead of crop. Additionally we'll set a bigger initial padding.left on the chart in order to display large y-axis labels by default.

Open issues:

- In safari when the last svg path loses focus by tabbing out of it, probably the tooltip receives focus for a split second until it gets removed from the dom a split second later. This might be fixable by rendering the tooltip in the dom before the chart, instead of after it, or render the tooltip inside a portal.


VWSCoronaDashboard8 (approved):

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard (approved):

packages/app/src/components/choropleth/safety-region-choropleth.tsx
```diff
@@ -132,36 +130,46 @@ export function SafetyRegionChoropleth<T, K extends RegionsMetricName>(
     [highlightCode]
   );
 
-  const hasSelectHander = !!onSelect;
+  const {
+    isTabInteractive,
+    tabInteractiveButton,
+    anchorEventHandlers,
+  } = useTabInteractiveButton(
+    replaceVariablesInText(siteText.choropleth.a11y.tab_navigatie_button, {
```
VWSCoronaDashboard4, line 138: perhaps writing out `accessibility` in full is clearer for these names

VWSCoronaDashboard8 (approved):

packages/app/src/utils/reverse-router.ts
```diff
@@ -0,0 +1,79 @@
+import { MunicipalityPageName } from '~/components/choropleth/select-handlers/types';
+
+// prettier-ignore
+export const reverseRouter = {
```
VWSCoronaDashboard4, line 4: :100:

VWSCoronaDashboard4 (approved):

packages/app/src/components/choropleth/hooks/use-tab-interactive-button.tsx
```diff
@@ -0,0 +1,73 @@
...
+import styled from 'styled-components';
+import { useHotkey } from '~/utils/hotkey/use-hotkey';
+
+export function useTabInteractiveButton(label: string) {
+  const [isTabInteractive, setIsTabInteractive] = useState(false);
+  const toggleButtonRef = useRef<HTMLButtonElement>(null);
+  const timeoutRef = useRef<number>();
+
+  useHotkey('escape', () => toggleButtonRef.current?.focus(), {
+    disabled: !isTabInteractive,
```
VWSCoronaDashboard8, line 12:
```suggestion
    isDisabled: !isTabInteractive,
```

nit; Since it's not a prop I think this is more correct

packages/app/src/components/choropleth/safety-region-choropleth.tsx
VWSCoronaDashboard8: I think we'll get used to `a11y` fast enough. Usually I'm all for using whole names, but this one is just long and annoying enough to warrant using a11y IMO.

packages/app/src/locale/nl.json
```diff
@@ -1899,6 +1899,9 @@
     "VR25": "http://www.veiligheidsregioflevoland.nl/"
   },
   "choropleth": {
+    "a11y": {
```
VWSCoronaDashboard8, line 1902: Actually I quite like the acronym (or whatever it's called) since it clearly sets to content apart from normal strings. If it was just the word "accessibility" it wouldn't be so clear that this content is special / non-visible.

packages/app/src/utils/reverse-router.ts
```diff
@@ -0,0 +1,79 @@
...
+      return reverseRouter.actueel.gm(code);
+    case 'positief-geteste-mensen':
+      return reverseRouter.gm.positiefGetesteMensen(code);
+    case 'sterfte':
+      return reverseRouter.gm.sterfte(code);
+    case 'ziekenhuis-opnames':
+      return reverseRouter.gm.ziekenhuisopnames(code);
+    case 'rioolwater':
+      return reverseRouter.gm.rioolwater(code);
+  }
```
VWSCoronaDashboard8, line 76: I think you forgot to use vr here?

VWSCoronaDashboard1 (approved):

packages/app/src/utils/reverse-router.ts
VWSCoronaDashboard8: You might as wel use Vr and Gm in the name of the functions as wel IMO. I would use it everywhere in stead of Municipality and SafetyRegion

packages/app/src/utils/reverse-router.ts
VWSCoronaDashboard8: Yes this was very much needed :)

VWSCoronaDashboard8 (approved):
I think I've found one bug, but otherwise 💯

packages/app/src/utils/reverse-router.ts
VWSCoronaDashboard10: Good catch, removed these functions because they are unused :-)

PR #2343, opened 2021-03-18
Fix relative date regression

VWSCoronaDashboard10:

### DIFF
--- a/packages/app/src/utils/formatDate.ts
+++ b/packages/app/src/utils/formatDate.ts
@@ -1,4 +1,5 @@
 import { isSameDay, isToday, isYesterday, subDays } from 'date-fns';
+import { isDefined } from 'ts-is-present';
 import siteText from '~/locale/index';
 import { getLocale } from '~/utils/getLocale';
 import { assert } from './assert';
@@ -22,10 +23,6 @@ if ('__setDefaultTimeZone' in Intl.DateTimeFormat) {
   (Intl.DateTimeFormat as any).__setDefaultTimeZone('Europe/Amsterdam');
 }
 
-function isDayBeforeYesterday(date: number | Date): boolean {
-  return isSameDay(date, subDays(Date.now(), 2));
-}
-
 const locale = getLocale();
 
 // TypeScript is missing some types for `Intl.DateTimeFormat`.
@@ -39,7 +36,6 @@ type formatStyle =
   | 'time'
   | 'long'
   | 'medium'
-  | 'relative'
   | 'iso'
   | 'axis'
   | 'axis-with-year'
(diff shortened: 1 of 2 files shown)


VWSCoronaDashboard1 (approved):

packages/app/src/utils/formatDate.ts
```diff
@@ -165,19 +161,53 @@ function getFormattedDate(date: Date, style: formatStyle) {
...
+/**
+ * Returns one of `vandaag` | `gisteren` | `eergisteren`.
+ * If given date is more than 2 days ago it will return `undefined`.
+ */
+export function formatRelativeDate(input: DateInput) {
+  const date = getDate(input);
+
+  if (typeof window === 'undefined') {
+    throw new Error('formatRelativeDate cannot be called server-side');
+  }
```
VWSCoronaDashboard8, line 179:
```suggestion
  assert (window, 'formatRelativeDate cannot be called server-side') 
```

I think Next might have a function or variable to tell you if you're server-side, so maybe that's more elegant than checking window

packages/app/src/utils/formatDate.ts
```diff
@@ -165,19 +161,53 @@ function getFormattedDate(date: Date, style: formatStyle) {
...
+    ? siteText.utils.date_today
+    : isYesterday(date)
+    ? siteText.utils.date_yesterday
+    : isDayBeforeYesterday(date)
+    ? siteText.utils.date_day_before_yesterday
+    : undefined;
+}
+
+function isDayBeforeYesterday(date: Date): boolean {
+  return isSameDay(date, subDays(Date.now(), 2));
```
VWSCoronaDashboard8, line 191: I was going to complain about the name subDays but it's not in our hands :)

packages/app/src/utils/formatDate.ts
```diff
@@ -165,19 +161,53 @@ function getFormattedDate(date: Date, style: formatStyle) {
...
+  return isToday(date)
+    ? siteText.utils.date_today
+    : isYesterday(date)
+    ? siteText.utils.date_yesterday
+    : isDayBeforeYesterday(date)
+    ? siteText.utils.date_day_before_yesterday
+    : undefined;
+}
+
+function isDayBeforeYesterday(date: Date): boolean {
```
VWSCoronaDashboard8, line 190: nit; I think you can remove the return type annotation

VWSCoronaDashboard8 (approved):

packages/app/src/utils/formatDate.ts
VWSCoronaDashboard10: nope: https://github.com/vercel/next.js/pull/7651

PR #2355, opened 2021-03-19
Convert charts positief geteste mensen

VWSCoronaDashboard8:
## Summary

Converts all line charts of positive tested people pages to `TimeSeriesChart`. In order to have all information in the tooltip a new trend type "invisible" has been added, to render additional property values in the tooltip.

With the addition of this invisible type, we are also able to replace the vaccine support multi line tooltip with a default one, so I have also done this as part of this PR.

The tooltip labels used for the differerent properties are stored in a central shared place in Lokalize between NL, VR and GM. I think this is a pattern we can use more. It allows us to have exactly the same code for each charts between the different contexts. Typically these labels will not differ. 

I think it would be good to do the same for the legend labels, but I've left this as is, since no additional text was required.

image
image
image
image


VWSCoronaDashboard1 (approved):

packages/app/src/locale/en.json
```diff
@@ -635,7 +641,7 @@
       "obtained": "obtained on {{date}}"
     },
     "barScale": {
-      "signaalwaarde": "Alert value"
+      "signaalwaarde": "__niet meer gebruikt"
```
VWSCoronaDashboard4, line 644: this won't get in lokalize, it will be ignored when importing

packages/app/src/components-styled/time-series-chart/logic/series.ts
```diff
@@ -46,6 +47,38 @@ export type AreaSeriesDefinition<T extends TimestampedValue> = {
...
+ * An invisible series config does not render any trend but the value shows up
+ * in the tooltip, in order with the rest of the items. This allows us to place
+ * any value from any metric property under or in between the others, with its
+ * own label.
+ *
+ * This can be used for example to show a total count at the bottom, or the
+ * percentage counterpart of an absolute value.
+ */
+export type InvisibleSeriesDefinition<T extends TimestampedValue> = {
+  type: 'invisible';
```
VWSCoronaDashboard4, line 60: what about naming this type `tooltip`? It's only used in the tooltip, where it will be all but invisible.

VWSCoronaDashboard4 (approved):

packages/app/src/components-styled/time-series-chart/logic/series.ts
VWSCoronaDashboard8: I have considered it, but I liked this better. I think it is a little clearer since the other trends also show up in the tooltip, and the filter in code on "isVisible" makes sense to me more then "isNotTypeTooltip" or something.

packages/app/src/locale/en.json
VWSCoronaDashboard8: Yeah it doesn't matter much. Nothing I can do about it from here.

packages/app/src/components-styled/time-series-chart/logic/series.ts
VWSCoronaDashboard8: `tooltipOnly` would be a candidate in my opinion, but still it doesn't read as nicely I think.

packages/app/src/components-styled/time-series-chart/components/tooltip/tooltip-series-list.tsx
```diff
@@ -69,48 +69,73 @@ export function TooltipSeriesList<T extends TimestampedValue>({
 
   const seriesConfig: SeriesConfig<T> = markNearestPointOnly
     ? [config[configIndex]]
-    : [...config].reverse();
+    : [...config];
```
VWSCoronaDashboard10, line 72: is the tooltip still in the right order?

packages/app/src/components-styled/time-series-chart/logic/legend.tsx
```diff
@@ -3,14 +3,14 @@ import { useMemo } from 'react';
...
-import { SeriesConfig } from './series';
+import { SeriesConfig, isVisible } from './series';
 
 export function useLegendItems<T extends TimestampedValue>(
   config: SeriesConfig<T>,
   dataOptions?: DataOptions
 ) {
   const legendItems = useMemo(() => {
-    const items = [...config].reverse().map<LegendItem>((x) => ({
+    const items = [...config].filter(isVisible).map<LegendItem>((x) => ({
```
VWSCoronaDashboard10, line 13: if you remove the `reverse` you could also drop the `[...config]` spread

VWSCoronaDashboard10 (approved):
💯

packages/app/src/components-styled/time-series-chart/components/tooltip/tooltip-series-list.tsx
VWSCoronaDashboard8: It's now in the order of the config, and the legend from left to right, which I think is the only order that makes sense . I've reversed the config of the vaccine support chart because of this, because I suspect the reverse was once put in as a quick fix to flip the content on that chart.

PR #2360, opened 2021-03-22
Add a new markdown component 

VWSCoronaDashboard14:
With the new way of getting text in the dashboard, all the markdown was broken.  I've created the `Markdown` component that translates it to HTML. Therefore I also removed all the old code that was parsing the markdown in the lokale. 

Right now it returns just a div with the parsed markdown, I wanted that div for styling purposes (not a lot but still). But I could also see that just returning only the parsed markdown is good enough. Or we could make it a more flexible component that accepts styling etc. but please let know.


packages/app/src/pages/landelijk/positief-geteste-mensen.tsx
```diff
@@ -393,27 +392,23 @@ const PositivelyTestedPeople = (props: StaticProps<typeof getStaticProps>) => {
                 }
               />
               <Text>{ggdText.positief_getest_week_uitleg}</Text>
-              <Text>
-                <strong
-                  css={css({ '& > span': { color: 'data.primary' } })}
-                  dangerouslySetInnerHTML={{
-                    __html: replaceKpisInText(
```
VWSCoronaDashboard4, line 400: With these changes, is `replaceKpisInText` still used anywhere? Otherwise this helper can be removed from the code base.

packages/app/src/components-styled/markdown.tsx
```diff
@@ -0,0 +1,28 @@
...
+    .use(externalLinks, {
+      target: '_blank',
+      rel: ['noopener', 'noreferrer'],
+    })
+    .use(remark2react)
+    .processSync(content).result;
+
+  return <MarkdownStyles>{parsedMarkdown as ReactNode}</MarkdownStyles>;
+}
+const MarkdownStyles = styled.div({
```
VWSCoronaDashboard4, line 24: Afaik `remark2react` allows for other control over components / styling; but that would maybe a good one for a future PR.

VWSCoronaDashboard4 (approved):
Nice improvement!

packages/app/src/pages/landelijk/positief-geteste-mensen.tsx
VWSCoronaDashboard14: I'll give it a look 👍

packages/app/src/components-styled/markdown.tsx
VWSCoronaDashboard1: So, this means that the markdown conversion happens on the client. I see unified, amongst others, being imported here. How does this change affect the bundle size?

VWSCoronaDashboard1 (commented):
This implementation is cleaner than what we had before. But, in the previous incarnation, the markdown conversion happened during SSR. Now it happens client-side (if I read the code correctly of course), I already asked in a code comment. But let me state here again, how does this change affect the bundle size?
Has this approach been discussed beforehand? If so, then ignore my comment ;)

VWSCoronaDashboard:
My 2 cents:

Short term:
- The getText helper with a whitelist of keys was difficult to maintain
- Changing a text to support markdown was a "complex" task

Using a component to wrap around text should solve both of these annoyances. That why I think it's acceptable to have a slightly higher bundle size now. I get that UX > DX, but in this case I believe that DX is worth the tradeoff, especially looking at the long term.

Long term:
- Markdown is a hack. We should never have supported from content from Lokalise.
- We need to identify the places where we use markdown. With this PR we get it for free. Do a find in your IDE for `<Markdown />` and you know where those texts are placed in the codebase.
- We need to put all rich formatted text into Sanity as block content.

VWSCoronaDashboard8 (approved):

PR #2362, opened 2021-03-22
Add nursing home moving average schema and charts

VWSCoronaDashboard8:
## Summary

This adds properties for 7-day running average to the nursing home schemas. All other 7-day averages can be added in the same way: Copy the original property, add a `_running_average` suffix, and then make it's type `number`.

It now also includes #2438


VWSCoronaDashboard1 (approved):

packages/app/schema/nl/nursing_home.json
```diff
@@ -23,9 +23,15 @@
         "newly_infected_people": {
           "type": "integer"
         },
+        "newly_infected_people_running_average": {
```
VWSCoronaDashboard4, line 26: in other schema's the suffix `_moving_average` is used; maybe use that here as well?

VWSCoronaDashboard8: Nice catch you're right 👍

VWSCoronaDashboard4 (approved):

PR #2363, opened 2021-03-22
Add markdown with react-markdown

VWSCoronaDashboard14:
After some investigation, it was indeed true that handling markdown on the client increased the bundle size by almost 40kb on certain pages where it was used. After shopping around it seems that react-markdown is 23kb and almost half as small. In the end, we need to remove this altogether, but this is a nice middle ground how it works.

Also since this library doesn't support external links I introduced my own small helper. It is not perfect but it uses the same technique as the one from remark. 

Similar to https://github.com/minvws/nl-covid19-data-dashboard/pull/2360 gonna close that one.


packages/app/src/components-styled/markdown.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const isExternalURL = (url: string) => /^https?:\/\//.test(url);
+
+  return (
+    <ReactMarkdown
+      source={content}
+      renderers={{
+        link: (props) => (
+          <a
+            href={props.href}
+            rel="noreferrer"
```
VWSCoronaDashboard4, line 17:
```suggestion
            rel="noreferrer noopener"
```

packages/app/src/components-styled/markdown.tsx
```diff
@@ -0,0 +1,26 @@
...
+            target={isExternalURL(props.href) ? '_blank' : ''}
+          >
+            {props.children}
+          </a>
```
VWSCoronaDashboard1, line 21: I think this instance can be turned into a `const` outside of the component. Currently this is recreated for each instance of this component while the renderer is completely static.

VWSCoronaDashboard1 (changes requested):

VWSCoronaDashboard4 (approved):

VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/markdown.tsx
```diff
@@ -0,0 +1,29 @@
+import { ReactNode } from 'react';
+import ReactMarkdown from 'react-markdown';
+
+interface MarkdownProps {
+  content: string;
+}
+
+interface LinkProp {
+  children: ReactNode[];
```
VWSCoronaDashboard8, line 9: Why is this an array of nodes? Usually children are defined as ReactNode. I think enforcing an array is probably a mistake

VWSCoronaDashboard8 (changes requested):

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard (approved):

PR #2364, opened 2021-03-22
Fix missing article titles

VWSCoronaDashboard4:
## Summary

Fix missing article titles

Also straightened out a related todo comment in a different file


VWSCoronaDashboard8 (approved):

VWSCoronaDashboard (approved):
This is a good example of why having a fallback language is a good idea with the `coalesce` function. 

```
// Default values when missing or null in document
*[_type == 'movie']{..., 'rating': coalesce(rating, 'unknown')}
```

PR #2365, opened 2021-03-22
Add back the sidebar on safety regions and municipality level

VWSCoronaDashboard14:
No data props were given to the sidebar hence why it wasn't showing.


VWSCoronaDashboard4 (approved):

VWSCoronaDashboard:
Should data be a required prop for MunicipalityLayout? We could prevent this error with TS.

VWSCoronaDashboard8:
@VWSCoronaDashboard4 @VWSCoronaDashboard yes lets make data a required property for all layouts as part of this PR

VWSCoronaDashboard8 (approved):
Looks good, but data should be made required to avoid this in the future

VWSCoronaDashboard10:
I've reverted the change which made the data required. Both the `/veiligheidsregio` and `/gemeente` routes don't need data because they don't render a sidebar.
To prevent future errors I've updated the layout props to the following:
```tsx
type MunicipalityLayoutProps = {
  lastGenerated: string;
  children?: React.ReactNode;
} & (
  | {
      data: Municipal;
      municipalityName: string;
    }
  | {
      isLandingPage: true;
      data?: undefined;
      municipalityName?: undefined;
    }
);
```
Which means the component can receive one of these:
1. `data`-prop in combination with `municipalityName`-prop, OR:
2. the `isLandingPage`-prop with value `true` in combination with an undefined `data`- and `municipalityName`-prop.

As you can see I've included the `municipalityName` because that was also left empty, resulting in no municipality name rendered in the sidebar.

All the above also applies for the safety region layout.

PR #2367, opened 2021-03-22
Add schema safety region risk level indicators

VWSCoronaDashboard14:
Schema for adding 2 new keys on the safety regions risk level page
Positive tested per 100k and hospital admissions per million.


packages/common/src/types/data.ts
```diff
@@ -947,6 +947,8 @@ export interface RegionalElderlyAtHomeValue {
 }
 export interface VrEscalationLevel {
   level: number;
+  positive_tested_per_100k?: number;
+  hospital_admissions_per_million?: number;
```
VWSCoronaDashboard8, line 951: Seems like you forgot to add them to the required list. Or maybe you did this on purpose to not break develop? I think I would only advise to use "optional" on the root of the schema, otherwise it quickly becomes messy to keep track of toggling things on and off.

I would suggest to add these properties to the required list and then don't merge it into develop. So use the "old" workflow for mutations of existing schema's.

VWSCoronaDashboard8 (changes requested):

VWSCoronaDashboard14:
@VWSCoronaDashboard8 I've integrated this in the Metadata component now.

VWSCoronaDashboard8 (approved):

packages/app/schema/vr/escalation_level.json
```diff
@@ -18,6 +20,12 @@
       "minimum": 1,
       "maximum": 4
     },
+    "positive_tested_per_100k": {
+      "type": "integer"
+    },
+    "hospital_admissions_per_million": {
+      "type": "integer"
+    },
```
VWSCoronaDashboard8, line 28: Are you sure these need to be integer?

VWSCoronaDashboard8 (changes requested):
If I look at other schemas per_100k the definitions are type number. So it would be logical to expect decimal here as well.

PR #2372, opened 2021-03-23
Load restriction icons as images

VWSCoronaDashboard:
## Summary

This PR changes icons to be images instead of inline components.
It saves ~100BK from the restrictions table bundle.

## Motivation

Roughly 100KB of inlined components is unacceptable. Images can load much more efficiently if not inlined, browsers don't need to parse the JS to load it.

## Technical details

You can't easily style an SVG that's loaded as an img element. That's why I resorted to applying different SVG filters to change colors of the icons. I used an [online tool to replicate the existing colors](https://codepen.io/sosuke/pen/Pjoqqp) as good as possible.

Filters are well supported by browsers, and can be considered progressive enhancement. If a filter doesn't work, you still see black and white icon.

https://caniuse.com/svg-filters

I think the only downside of this approach is that you need to manually convert hex colors to filters. Whenever the hex colors change in the codebase, we need to change the filters manually. While it adds to maintenance, it's an easy chore and also easy to spot if we forget to update.


packages/app/src/domain/restrictions/lockdown-table.tsx
```diff
@@ -17,22 +16,54 @@ export function LockdownTable(props: LockdownTableProps) {
...
+  // #A11050
+  if (escalationLevel === 3) {
+    filter =
+      'invert(15%) sepia(48%) saturate(4967%) hue-rotate(317deg) brightness(92%) contrast(101%)';
+  }
+  // #68032F
+  if (escalationLevel === 4) {
+    filter =
+      'invert(9%) sepia(48%) saturate(6614%) hue-rotate(322deg) brightness(85%) contrast(103%)';
+  }
```
VWSCoronaDashboard1, line 54: nit: Please turn this into a switch statement ;)

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard8 (approved):

packages/app/src/domain/restrictions/lockdown-table.tsx
VWSCoronaDashboard10: and they can all be return statements, no need to mutate a filter variable

packages/app/src/domain/restrictions/restriction-icons.ts
```diff
@@ -1,92 +1,37 @@
...
-  '41_onderwijs_40': OnderwijsKinderopvang,
-  '41_onderwijs_41': OnderwijsNoodopvang,
-  avondklok,
-  bezoek,
-  lopend,
-  eenPersoonDoorgestreept,
-  gedeeltelijkOpenRugzak,
-  geenWedstrijden,
-  sporterMetZweetband,
+export const restrictionIcons: Record<string, string | null> = {
```
VWSCoronaDashboard10, line 1: can this object be moved to common to be used both by the cms as the client app? including resolving the svg's from the client app's domain?

VWSCoronaDashboard10 (approved):

packages/app/src/domain/restrictions/lockdown-table.tsx
```diff
@@ -17,22 +16,54 @@ export function LockdownTable(props: LockdownTableProps) {
...
 
-  const color = useEscalationColor(4);
-
   if (breakpoints.lg) {
-    return <DesktopLockdownTable data={data} color={color} />;
+    return <DesktopLockdownTable data={data} escalationLevel={4} />;
   }
 
-  return <MobileLockdownTable data={data} color={color} />;
+  return <MobileLockdownTable data={data} escalationLevel={4} />;
```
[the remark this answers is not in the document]
VWSCoronaDashboard: Color was also derived from a hardcoded number 4, just one component higher. In that sense we're not changing anything with this PR. It would be smart to use a variable for both instances so we know 100% sure both tables use the same escalation level.

packages/app/src/domain/restrictions/restriction-icons.ts
VWSCoronaDashboard: If we would use a duplicate restictions icons folder copied to the public folder in the CMS? Yes. 

I don't know how we'd be able to share the images between the applications. Any thoughts?

packages/app/src/domain/restrictions/lockdown-table.tsx
VWSCoronaDashboard: And done!

packages/app/src/domain/restrictions/lockdown-table.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard: I've left a TODO in the code that we need to hook this into the CMS or the dataset at some point.

packages/app/src/domain/restrictions/restriction-icons.ts
VWSCoronaDashboard10: isn't it possible to load images in the cms with ``?

VWSCoronaDashboard: You mean from the CMS. Maybe. Not sure if CSP/Akamai would interfere with this. You mind if we see this as an enhancement and work in this in the future?

PR #2377, opened 2021-03-23
Add schema for NL intensive case NICE date of admission

VWSCoronaDashboard8:
## Summary

This PR now also includes the key changes in the various components.

### DIFF
--- a/packages/app/infra/download.js
+++ b/packages/app/infra/download.js
@@ -7,6 +7,7 @@ const API_URL =
   process.env.API_URL ||
   'https://coronadashboard.rijksoverheid.nl/json/latest-data.zip';
 
+// eslint-disable-next-line no-console
 console.log(`Downloading json data from this location: ${API_URL}`);
 
 (async () => {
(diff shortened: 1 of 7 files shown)


VWSCoronaDashboard1 (approved):

packages/common/src/types/data.ts
```diff
@@ -261,6 +261,8 @@ export interface NationalIntensiveCareNice {
 }
 export interface NationalIntensiveCareNiceValue {
   admissions_moving_average: number;
+  admissions_on_date_of_admission: number;
```
VWSCoronaDashboard4, line 264: these could be integers?

VWSCoronaDashboard4 (approved):

packages/app/infra/download.js
```diff
@@ -7,6 +7,8 @@ const API_URL =
   process.env.API_URL ||
   'https://coronadashboard.rijksoverheid.nl/json/latest-data.zip';
 
+console.log(`Downloading json data from this location: ${API_URL}`);
```
VWSCoronaDashboard10, line 10: Not sure whether we should log variables which are secret

VWSCoronaDashboard10 (approved):

packages/app/infra/download.js
VWSCoronaDashboard1: the vercel build logs aren't public, are they?

PR #2380, opened 2021-03-23
Add styling for the "Risiconiveau van dit moment" on safety regions

VWSCoronaDashboard14:
A flexible component that renders a new tile on the safety region risk level page that displays with a bit more styling the determining risk levels, hospital admissions, and the positive tested people for that safety region.

Screenshot 2021-03-24 at 10 11 27

This is with dummy data, the correct data still needs to be connected.


packages/app/src/components-styled/escalation-level-icon.tsx
```diff
@@ -10,11 +10,13 @@ const escalationThresholds = regionThresholds.escalation_levels.level;
 interface EscalationLevelIconProps {
   level: EscalationLevel;
   isSmall?: boolean;
+  isBig?: boolean;
```
VWSCoronaDashboard8, line 13: Now something can be small and big at the same time.... If there are three flavors it would be better to implement this as a variant with a string literal. So for example a prop named `variant` or `size` with values `small | medium | large` 

I prefer the word large over big, since it's more in line with our other theme naming.

packages/app/src/components-styled/escalation-level-icon.tsx
```diff
@@ -37,10 +40,12 @@ export function EscalationLevelIcon({
...
   isSmall?: boolean;
-}>(({ color, isSmall }) => {
-  const size = isSmall
-    ? asResponsiveArray({ _: 20, sm: 20 })
-    : asResponsiveArray({ _: 24, sm: 22 });
+  isBig?: boolean;
+}>(({ color, isSmall, isBig }) => {
+  let size = asResponsiveArray({ _: 24, sm: 22 });
+  if (isSmall) size = asResponsiveArray({ _: 20, sm: 20 });
+  if (isBig) size = asResponsiveArray({ _: 45, sm: 45 });
```
VWSCoronaDashboard8, line 47: I find it odd that we mix logic of a size prop with media queries. So we have 3 different sizes of the component with each their own media queries to make them respond differently to screen sizes on top of that. Seems a bit convoluted to me.

packages/app/src/pages/veiligheidsregio/[code]/risiconiveau.tsx
```diff
@@ -228,4 +292,118 @@ const RegionalRestrictions = (props: StaticProps<typeof getStaticProps>) => {
...
+interface ListItemProps {
+  icon: ReactNode;
+  title: string;
+  hasBorderBottom?: boolean;
+  escalationColor?: string;
+  description?: string;
+  date?: number | number[];
+  children?: ReactNode;
+  amount?: number;
+  isAroundDate?: boolean;
```
Comment (VWSCoronaDashboard8) [inline packages/app/src/pages/veiligheidsregio/[code]/risiconiveau.tsx:304]: Should this really be all optional? Seems like a lot. If you leave everything out, is it still a valid / complete list item?

packages/app/src/pages/veiligheidsregio/[code]/risiconiveau.tsx
```diff
@@ -228,4 +292,118 @@ const RegionalRestrictions = (props: StaticProps<typeof getStaticProps>) => {
...
+          <Box
+            height={9}
+            width={9}
+            backgroundColor={escalationColor}
+            borderRadius="50%"
+            mr={2}
+          />
+          <Text m={0}>
+            {replaceComponentsInText(description, {
+              amount: <InlineText fontWeight="bold">{`${amount} `}</InlineText>,
```
Comment (VWSCoronaDashboard8) [inline packages/app/src/pages/veiligheidsregio/[code]/risiconiveau.tsx:372]:
```suggestion
              amount: <InlineText fontWeight="bold">{amount}</InlineText>,
```

VWSCoronaDashboard8 (changes requested):

packages/app/src/pages/veiligheidsregio/[code]/risiconiveau.tsx
Comment (VWSCoronaDashboard14) [reply packages/app/src/pages/veiligheidsregio/[code]/risiconiveau.tsx]: I totally get you, every list item has a small difference in dealing with data or dates. I thought creating new components or rendering everything inline and passing it on wasn't that nice. 

If you only use the standard list item without any optional props it should render an empty item with just a title and an icon. 
Screenshot 2021-03-26 at 11 26 30

packages/app/schema/vr/escalation_level.json
```diff
@@ -18,6 +20,12 @@
       "minimum": 1,
       "maximum": 4
     },
+    "positive_tested_per_100k": {
+      "type": "integer"
+    },
+    "hospital_admissions_per_million": {
+      "type": "integer"
+    },
```
VWSCoronaDashboard8, line 28: These need to be number / decimal.

packages/app/schema/vr_collection/escalation_levels.json
```diff
@@ -23,6 +25,12 @@
       "minimum": 1,
       "maximum": 4
     },
+    "positive_tested_per_100k": {
+      "type": "integer"
+    },
+    "hospital_admissions_per_million": {
+      "type": "integer"
+    },
```
VWSCoronaDashboard8, line 33: And then of course these too

packages/app/src/pages/veiligheidsregio/[code]/risiconiveau.tsx
```diff
@@ -228,4 +290,130 @@ const RegionalRestrictions = (props: StaticProps<typeof getStaticProps>) => {
...
+  children,
+  description,
+  escalationColor,
+  amount,
+  isAroundDate,
+}: ListItemProps) {
+  const { siteText, formatDateFromSeconds } = useIntl();
+
+  return (
+    <StyledList>
```
Comment (VWSCoronaDashboard8) [inline packages/app/src/pages/veiligheidsregio/[code]/risiconiveau.tsx:317]: This component is called ListItem, yet the first thing it renders (as a container) is a list. Seems like a mixup. If this is really about the list item, when the container should be lifted to the calling context using composition. So render the StyledList where you are using the ListItem, and make ListItem its child.

VWSCoronaDashboard8 (changes requested):

packages/app/src/pages/veiligheidsregio/[code]/risiconiveau.tsx
Comment (VWSCoronaDashboard8) [reply packages/app/src/pages/veiligheidsregio/[code]/risiconiveau.tsx]: I think the problem starts because you are trying to treat everything in that list as the same component. You could make a distinction between root level list items and other items.

Root-level list items seem to always have an icon, title and date so you can make those required. Then each of them is able to (optionally) contain a sublist.  That sublist then contains list items that are different components, because their icon is smaller and they do not have dates.

Those sub-lists see to always have childen and those children are rendered using bullet points instead of icons. So that is a third type of list item.

So a clean implementation would be to create three distinct list item types, each with their ons specific strict set of props/behavior. Then you don't need the variant I'm assuming and you can make most properties non-optional.

Comment (VWSCoronaDashboard14) [reply packages/app/src/pages/veiligheidsregio/[code]/risiconiveau.tsx]: Alright, gonna separate it a bit more. Thanks!

VWSCoronaDashboard8 (approved):

PR #2406, opened 2021-03-25
G number chart

VWSCoronaDashboard13:
## Summary

G-number chart 

## Motivation

New chart 🎉

## Detailed design

image

Created a VerticalBarChart component that is modeled after the TimeSeriesChart so it has a lot of the same parameters, accepts multiple series (But only accepts the new BarSeriesDefinition), and uses a lot of the same base components since both are based on Time Series. This was done for the sake of consistency, potentially allowing this chart to accept multiple series for a stacked bar chart, and to make it easier to merge this into the TimeSeriesChart in the future if desired.

The two charts share the following components:
- ChartContainer (onHover is now optional since hovering the entire chart is not applicable for bars)
- Overlay
- Axes

Note: These above components could be used with minimal alterations which is why I opted to use them rather than making new components. I can make simplified versions for the bar chart if that is preferable.

All differing components/logic were recreated in matching logic and component directories within VerticalBarChart
Some notable differences: 
- Series config only accepts the new "bar" type
- VerticalBarChart accepts date spans, but it uses the date_end_unix value to determine where it is placed on the x-axis
- VerticalBarChart shows negative values
- The onHover function is passed an index (from the bar that is hovered) rather than using a bisect function (this seemed simpler code wise, but the bisect function can be re-added)

Visually the graph looks different from the original design, however these have already been discussed with a designer. For example, the bottom axes should be consistent with existing TimeSeriesCharts and the tooltip should be consistent across all graphs if possible.

Note: Most of the comments/TODO were copied from the TimeSeriesChart

## Drawbacks

There is the option to pass in multiple series to this chart component, but it doesn't really handle it properly since there is no use case or design for this yet. This means there is a little bit of code that is not being used, but I do think keeping a consistent interface with the TimeSeriesChart is worth this drawback and we can always build on top of this chart to account for those cases when the design is available in the future.


packages/app/src/components-styled/vertical-bar-chart/components/bar-hover.tsx
```diff
@@ -0,0 +1,27 @@
...
+  barWidth,
+}: BarHoverProps<T>) {
+  return (
+    <Bar
+      fill={'rgba(0, 0, 0, 0.03)'}
+      x={point.x}
+      y={0}
+      width={barWidth}
+      height={bounds.height}
+      transform={`translate(-${barWidth / 2}, 0)`}
```
VWSCoronaDashboard4, line 24:
```suggestion
      transform={'translateX(-50%)'}
```

packages/app/src/components-styled/vertical-bar-chart/logic/series.tsx
```diff
@@ -0,0 +1,123 @@
...
+  values: T[],
+  seriesConfig: BarSeriesConfig<T>
+) {
+  const metricProperties = seriesConfig.flatMap((x) => x.metricProperty);
+
+  const extremeValues = values
+    .map((x) => {
+      const trendValues = Object.values(pick(x, metricProperties)) as (
+        | number
+        | null
```
VWSCoronaDashboard4, line 112: can we use `undefined` here?

packages/app/src/domain/tested/g-number-bar-chart-tile.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+interface GNumberBarChartTileProps {
+  data: NlGNumber;
+  siteText: AllLanguages;
+  timeframeOptions?: TimeframeOption[];
+}
+
+export function GNumberBarChartTile({
+  data: __data,
+  // siteText,
```
VWSCoronaDashboard4, line 38: comments can be removed if not needed

packages/app/src/domain/tested/g-number-bar-chart-tile.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const last_value: NlGNumberValue = values[values.length - 1];
+
+  return (
+    <ChartTileWithTimeframe
+      title={text.title}
+      description={text.description}
+      timeframeOptions={timeframeOptions}
+      metadata={{
+        date: last_value.date_of_insertion_unix,
+        // source: 'source',
```
VWSCoronaDashboard4, line 60: this probably needs to be added

packages/app/src/domain/tested/g-number-bar-chart-tile.tsx
```diff
@@ -0,0 +1,101 @@
...
+                  label: 'G Number',
+                },
+              ]}
+              formatTooltip={({ value }) => {
+                return (
+                  <>
+                    <InlineText fontWeight="bold">
+                      {`${formatPercentage(value.g_number)}% `}
+                    </InlineText>
+                    getaald
```
VWSCoronaDashboard4, line 91: move to locale

packages/app/src/domain/tested/g-number-bar-chart-tile.tsx
```diff
@@ -0,0 +1,101 @@
...
+      metadata={{
+        date: last_value.date_of_insertion_unix,
+        // source: 'source',
+      }}
+    >
+      {(timeframe) => (
+        <ParentSize>
+          {({ width }) => (
+            <VerticalBarChart
+              title="G Number"
```
VWSCoronaDashboard4, line 67: move to locale

packages/app/src/domain/tested/g-number-bar-chart-tile.tsx
```diff
@@ -0,0 +1,101 @@
...
+              dataOptions={{
+                isPercentage: true,
+              }}
+              seriesConfig={[
+                {
+                  type: 'bar',
+                  metricProperty: 'g_number',
+                  color: colors.data.primary,
+                  secondaryColor: colors.red,
+                  label: 'G Number',
```
VWSCoronaDashboard4, line 82: move to locale

packages/app/src/pages/index.tsx
```diff
@@ -146,6 +147,9 @@ const Home = (props: StaticProps<typeof getStaticProps>) => {
               <Search />
             </Box>
 
+            {/* @ts-expect-error @TODO Pass in data and sitetext */}
```
VWSCoronaDashboard4, line 150: siteText / locale can be hooked with `useIntl`

packages/app/src/components-styled/vertical-bar-chart/logic/series.tsx
```diff
@@ -0,0 +1,123 @@
...
+    return [];
+  }
+
+  if (isDateSeries(values)) {
+    return values.map((x) => ({
+      /**
+       * This is messy and could be improved.
+       */
+      __value: (x[metricProperty] ?? undefined) as number | undefined,
+      // @ts-expect-error @TODO figure out why the type guard doesn't work
```
VWSCoronaDashboard4, line 67: there are various todos like this in this PR, perhaps figure out a way to deal with them?

packages/app/src/components-styled/vertical-bar-chart/components/bar-hover.tsx
VWSCoronaDashboard13: This is what I tried originally, but it didn't work :/

packages/app/src/components-styled/vertical-bar-chart/logic/series.tsx
VWSCoronaDashboard13: This was pulled from the TimeSeriesChart so keeping this null is more consistent. I believe this needs to be null since that's what is provided in the data?

packages/app/src/components-styled/vertical-bar-chart/components/bar-hover.tsx
VWSCoronaDashboard10: Probably because this isn't a css transform but an svg transform

VWSCoronaDashboard10 (commented):
actually looking pretty good 👍

packages/app/src/components-styled/vertical-bar-chart/components/bar-trend.tsx
```diff
@@ -0,0 +1,104 @@
...
+        const fillColor = value.__value
+          ? value.__value > 0
+            ? color
+            : secondaryColor
+          : 'transparent';
+
+        const { x, y, height } = getRectPosition(value);
+        return (
+          <Group
+            id={barId}
```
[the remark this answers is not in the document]
VWSCoronaDashboard13: it's not necessary. I can remove it 👍

packages/app/src/domain/tested/g-number-bar-chart-tile.tsx
```diff
@@ -0,0 +1,101 @@
...
+      description={text.description}
+      timeframeOptions={timeframeOptions}
+      metadata={{
+        date: last_value.date_of_insertion_unix,
+        // source: 'source',
+      }}
+    >
+      {(timeframe) => (
+        <ParentSize>
+          {({ width }) => (
```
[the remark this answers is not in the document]
VWSCoronaDashboard13: done 👍

packages/app/src/components-styled/vertical-bar-chart/logic/series.tsx
```diff
@@ -0,0 +1,112 @@
...
+  isDateSeries,
+  isDateSpanSeries,
+  TimestampedValue,
+} from '@corona-dashboard/common';
+import { useMemo } from 'react';
+import { pick } from 'lodash';
+import { isPresent } from 'ts-is-present';
+import { SeriesSingleValue } from '~/components-styled/time-series-chart/logic/series';
+
+export type BarSeriesConfig<
```
VWSCoronaDashboard8, line 11: Since this file is local to your chart, I suggest you call it SeriesConfig, to keep the code more in-line with time series chart. It makes it easier to merge things later if we want to.

packages/app/src/components-styled/vertical-bar-chart/logic/series.tsx
```diff
@@ -0,0 +1,112 @@
...
+  values: T[],
+  seriesConfig: BarSeriesConfig<T>
+) {
+  return useMemo(() => getSeriesList(values, seriesConfig), [
+    values,
+    seriesConfig,
+  ]);
+}
+
+export type BarSeriesList = SeriesSingleValue[][];
```
VWSCoronaDashboard8, line 33: Same here, this can just be called SeriesList like in the other code

packages/app/src/components-styled/vertical-bar-chart/components/bar-trend.tsx
```diff
@@ -0,0 +1,104 @@
...
+) {
+  return function getRectPosition(value: SeriesSingleValue) {
+    const valuePosition = getY(value);
+
+    return {
+      x: getX(value),
+      y: Math.min(zeroPosition, valuePosition),
+      height: Math.abs(zeroPosition - valuePosition),
+    };
+  };
```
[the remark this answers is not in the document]
VWSCoronaDashboard13: ahh yeah, that is cleaner 👍

packages/app/src/components-styled/time-series-chart/components/axes.tsx
```diff
@@ -17,7 +17,7 @@ import { useIntl } from '~/intl';
 
 type AxesProps = {
   bounds: Bounds;
-  xScale: ScaleLinear<number, number>;
+  xScale: ScaleLinear<number, number> | ScaleBand<number>;
```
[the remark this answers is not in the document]
VWSCoronaDashboard13: Do you think this warrants a duplicate component specifically for the bar chart for just this accepting a different type of scale? I feel like that might not be necessary until we need them to look different since they currently have the exact same logic

packages/app/src/components-styled/time-series-chart/components/tooltip/tooltip-series-list.tsx
```diff
@@ -66,7 +65,7 @@ export function TooltipSeriesList<T extends TimestampedValue>({
 
   const dateString = getDateStringFromValue(value);
 
-  const seriesConfig: SeriesConfig<T> = markNearestPointOnly
+  const seriesConfig: TooltipSeriesConfig<T> = markNearestPointOnly
```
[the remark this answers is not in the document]
VWSCoronaDashboard13: I created a tooltip directory for the VerticalBarChart and moved a lot of the tooltip positioning and visual code to a `TooltipWrapper` component that both tooltips use 👍

packages/app/src/components-styled/time-series-chart/components/tooltip/types.ts
```diff
@@ -42,3 +51,12 @@ export type TooltipData<T extends TimestampedValue> = {
...
 ) => React.ReactNode;
+
+export type TooltipSeriesConfig<T extends TimestampedValue> = (
+  | LineSeriesDefinition<T>
+  | AreaSeriesDefinition<T>
+  | RangeSeriesDefinition<T>
+  | InvisibleSeriesDefinition<T>
+  | StackedAreaSeriesDefinition<T>
+  | BarSeriesDefinition<T>
+)[];
```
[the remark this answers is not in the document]
VWSCoronaDashboard13: This type was removed 👍

packages/app/src/components-styled/time-series-chart/components/axes.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard8: No you're right if it's only a minor change like this we should keep the component shared 👍

packages/app/src/components-styled/vertical-bar-chart/logic/series.tsx
```diff
@@ -0,0 +1,110 @@
...
+  values: T[],
+  metricProperty: keyof T
+): SeriesSingleValue[] {
+  if (values.length === 0) {
+    return [];
+  }
+
+  if (isDateSeries(values)) {
+    return values.map((x) => ({
+      __value: (x[metricProperty] ?? undefined) as number | undefined,
```
VWSCoronaDashboard8, line 56:
```suggestion
      __value: x[metricProperty] as number | undefined,
```
Maybe not your code, but I think this does exactly the same

packages/app/src/components-styled/vertical-bar-chart/logic/series.tsx
```diff
@@ -0,0 +1,110 @@
...
+      // @ts-expect-error @TODO figure out why the type guard doesn't work
+      __date_unix: x.date_unix,
+    }));
+  }
+
+  if (isDateSpanSeries(values)) {
+    return values.map((x) => ({
+      __value: (x[metricProperty] ?? undefined) as number | undefined,
```
VWSCoronaDashboard8, line 64:
```suggestion
      __value: x[metricProperty]  as number | undefined,
```

VWSCoronaDashboard8 (approved):

packages/app/src/components-styled/vertical-bar-chart/logic/series.tsx
VWSCoronaDashboard13: It looks like this is necessary to fix a typescript error. It is also present on the TimeSeriesChart series.js file

PR #2421, opened 2021-03-26
Improve article and weekly/custom blocks topical page

VWSCoronaDashboard14:
- Update the styling of blocks on the topical page
- Remove the editorial tile
- Add an option to have 2 custom highlighted content blocks in the cms
- Add a category input field in the CMS
- Display on the article detail, overview, and on the topical page the category names

In the future, the category has predefined tags.


packages/app/src/components-styled/article-detail.tsx
```diff
@@ -28,6 +29,14 @@ export function ArticleDetail({ article }: ArticleDetailProps) {
         </LinkWithIcon>
 
         <Box spacing={2}>
+          <InlineText
+            css={css({ textTransform: 'uppercase' })}
```
VWSCoronaDashboard10, line 33: `textTransform` sounds like a property which can be part of the `Text` and `InlineText` component. This can be implemented the same way is we did for the Box's `spacing` prop (by creating a `/style/functions/text-transform.ts` file)

packages/app/src/components-styled/highlight-teaser.tsx
```diff
@@ -73,17 +98,17 @@ function ZoomContainerUnstyled({
...
-    minHeight: '26rem',
     overflow: 'hidden',
     textDecoration: 'none',
-    color: 'body',
+    backgroundColor: x.isWeekly ? 'blue' : undefined,
+    color: x.isWeekly ? 'white' : 'body',
+
+    'span, time': {
+      color: x.isWeekly ? 'white' : undefined,
+    },
```
VWSCoronaDashboard10, line 111: please don't use strong selectors like these, in the future they will become a pain in the ass and they are also an indication that something is wrong with the codebase.
For example: the actual mistake here is that there's a color set inside the `PublicationDate` component. This color should be set by the container component (in your case on line 54). 

I'd suggest you to:
1.  remove the color from the PublicationDate component
2. apply the `isWeekly ? 'white' : 'annotation'` style on line 54
3. set correct color of all PublicationDate instance's parent components

packages/app/src/domain/topical/article-list.tsx
```diff
@@ -44,14 +45,14 @@ export function ArticleList({ articleSummaries }: ArticleListProps) {
...
+export const ArticleBox = styled.div(
   css({
     marginBottom: 4,
     width: asResponsiveArray({
       _: '100%',
       sm: 'calc(50% - 16px)',
-      md: 'calc(33% - 32px)',
-      lg: 'calc(33% - 32px)',
+      md: 'calc(33.33% - 32px)',
+      lg: 'calc(33.33% - 32px)',
```
VWSCoronaDashboard10, line 55: How about `calc(1 / 3 * 100% - 32px)`

packages/app/src/domain/topical/highlights-tile.tsx
```diff
@@ -0,0 +1,59 @@
...
+        <HighlightTeaser
+          cover={weeklyHighlight.cover}
+          href={weeklyHighlight.slug.current}
+          title={weeklyHighlight.title}
+          category={'Weekbericht'}
+          publicationDate={weeklyHighlight.publicationDate}
+          isWeekly
+        />
+      </ArticleBox>
+      {highlights.map((item: HighlightTeaserProps, index: number) => (
```
VWSCoronaDashboard10, line 46: types can be inferred
```suggestion
      {highlights.map((item, index) => (
```

packages/app/src/pages/actueel/gemeente/[code].tsx
```diff
@@ -258,16 +260,17 @@ const TopicalMunicipality = (props: StaticProps<typeof getStaticProps>) => {
               />
             </CollapsibleButton>
 
-            {content.editorial && content.highlight && (
-              <>
+            {content.weeklyHighlight && content.highlights && (
```
Comment (VWSCoronaDashboard10) [inline packages/app/src/pages/actueel/gemeente/[code].tsx:263]: do you want to render when `content.highlights` is an empty array? If not: check also the length

packages/app/src/pages/index.tsx
```diff
@@ -216,15 +218,15 @@ const Home = (props: StaticProps<typeof getStaticProps>) => {
               />
             </CollapsibleButton>
 
-            {content.editorial && content.highlight && (
+            {content.weeklyHighlight && content.highlights && (
```
VWSCoronaDashboard10, line 221: same here

packages/app/src/queries/topical-page-query.ts
```diff
@@ -11,31 +11,33 @@ export function getTopicalPageQuery(_context: GetStaticPropsContext) {
...
+        "category": category.${locale},
+        "label":label.nl,
+        href,
+        "cover": {
+          ...cover,
+          "asset": cover.asset->  
+        }
       }
-    }[0]
+    }[0].highlights
```
VWSCoronaDashboard10, line 41: pretty!

packages/cms/src/schemas/documents/fields/article-fields.ts
```diff
@@ -9,6 +9,16 @@ export const ARTICLE_FIELDS = [
         en: (fieldRule: any) => fieldRule.reset().required(),
       }),
   },
+  {
+    title: 'Category',
```
VWSCoronaDashboard10, line 13: shouldn't the title be in dutch?

packages/cms/src/schemas/documents/pages/topical-page.ts
```diff
@@ -4,58 +4,75 @@ export default {
...
+              name: 'title',
+              type: 'localeString',
+              validation: (Rule: any) =>
+                Rule.fields({
+                  nl: (fieldRule: any) => fieldRule.reset().required(),
+                  en: (fieldRule: any) => fieldRule.reset().required(),
+                }),
+            },
+            {
+              title: 'Category',
```
VWSCoronaDashboard10, line 31: same here

packages/cms/src/schemas/documents/pages/topical-page.ts
```diff
@@ -4,58 +4,75 @@ export default {
...
+      title: 'Custom highlight blocks',
+      name: 'highlights',
+      type: 'array',
+      of: [
         {
-          title: 'Alternatieve tekst (toegankelijkheid)',
-          name: 'alt',
-          type: 'localeString',
-        },
+          title: 'Highlighted articles',
```
VWSCoronaDashboard10, line 12: same here

packages/cms/src/schemas/documents/pages/topical-page.ts
```diff
@@ -4,58 +4,75 @@ export default {
...
-    },
-    {
-      title: 'Afbeelding',
-      name: 'cover',
-      type: 'image',
-      options: {
-        hotspot: true,
-      },
-      fields: [
+      title: 'Custom highlight blocks',
```
VWSCoronaDashboard10, line 7: same here

packages/app/src/domain/topical/highlights-tile.tsx
```diff
@@ -0,0 +1,59 @@
...
+      flexDirection={{ _: 'row', lg: 'row' }}
+    >
+      <ArticleBox>
+        <HighlightTeaser
+          cover={weeklyHighlight.cover}
+          href={weeklyHighlight.slug.current}
+          title={weeklyHighlight.title}
+          category={'Weekbericht'}
+          publicationDate={weeklyHighlight.publicationDate}
+          isWeekly
```
VWSCoronaDashboard10, line 43: The `isWeekly` prop on a `HighlightTeaser` doesn't really make sense, context-wise. Maybe we can use a variants prop? `variant='blue'` with a default value of `white`?

packages/app/src/locale/en.json
```diff
@@ -667,6 +667,7 @@
     "gisteren": "Yesterday",
     "vandaag": "Today",
     "read_more": "Read more",
+    "read_weekly": "Read the weekly message",
```
VWSCoronaDashboard10, line 670: This is imho not a common message but very specific one. Please rename the key to something like `read_weekly_message` and move it to a topical- or article-related context of the locale files.

VWSCoronaDashboard10 (changes requested):
Looking good overall, couple of changes requested though :-)

packages/app/src/queries/topical-page-query.ts
VWSCoronaDashboard14: 🤣

packages/app/src/components-styled/highlight-teaser.tsx
```diff
@@ -98,16 +107,16 @@ function ZoomContainerUnstyled({
   );
 }
 
-const StyledHightlightTeaser = styled.a<{ isWeekly?: boolean }>((x) =>
+const StyledHightlightTeaser = styled.a<{ variant?: string }>((x) =>
```
VWSCoronaDashboard10, line 110: shouldn't variant either be one if `blue` or `default`/`white`?

packages/app/src/components-styled/highlight-teaser.tsx
```diff
@@ -18,6 +18,7 @@ export type HighlightTeaserProps = {
   category: string;
   publicationDate?: string;
   isWeekly?: boolean;
+  variant?: 'blue';
```
VWSCoronaDashboard10, line 21: The default variant is missing, name it `white` or `default` or something like that .. you could create a separate type (`variant?: HightlightTeaserType;`) and use this type in the `StyledHightlightTeaser` component too.

packages/app/src/components-styled/typography/text.tsx
```diff
@@ -12,24 +12,29 @@ import {
...
+  TextTransformProps;
 
 /**
  * A generic text component that can be used for any paragraph or other piece of
  * text, only for headers we have a dedicated Title component.
  */
 export const Text = styled.p.withConfig({
   shouldForwardProp: styledShouldForwardProp,
-})<TextProps>(compose(margin, padding, typography, color));
+})<TextProps>(compose(margin, padding, typography, color), textTransform);
```
VWSCoronaDashboard10, line 33: it won't work as argument of the `compose`?

packages/app/src/domain/topical/highlights-tile.tsx
```diff
@@ -30,20 +31,21 @@ export function HighlightsTile(props: HighlightsTileProps) {
...
       spacingHorizontal={breakpoints.lg}
-      spacing={3}
       flexDirection={{ _: 'row', lg: 'row' }}
+      mr={0}
     >
       <ArticleBox>
         <HighlightTeaser
           cover={weeklyHighlight.cover}
-          href={weeklyHighlight.slug.current}
+          href={`weekberichten/${weeklyHighlight.slug.current}`}
```
VWSCoronaDashboard10, line 40: you're missing a root slash in this href

packages/app/src/pages/actueel/gemeente/[code].tsx
```diff
@@ -260,7 +260,7 @@ const TopicalMunicipality = (props: StaticProps<typeof getStaticProps>) => {
               />
             </CollapsibleButton>
 
-            {content.weeklyHighlight && content.highlights && (
+            {content.weeklyHighlight && content.highlights.length && (
```
Comment (VWSCoronaDashboard10) [inline packages/app/src/pages/actueel/gemeente/[code].tsx:263]: Here's classic mistake being made :-) see:
```tsx
// this is the current situation:
{content.weeklyHighlight && content.highlights.length && <> ... </>}

// when the highlights array is empty, the length will be `0`
{content.weeklyHighlight && 0 && <> ... </>}

// it will stop evaluating when there's a falsy value

{content.weeklyHighlight && 0}

// and as you might know
const str = somethingTruthy && 'yeah!'
str === 'yeah!'

// so, this:
{content.weeklyHighlight && 0}

// basically becomes this:
{0}

// and voilia, suddenly a `0` is rendered by react :-)

File: packages/app/src/style/functions/text-transform.ts [hunk 19]
```diff
@@ -0,0 +1,15 @@
+import css from "@styled-system/css"
+import { isDefined } from "ts-is-present"
+import { styleFn } from 'styled-system';
+
+export interface TextTransformProps {
+  textTransform?: 'none' | 'capitalize' | 'uppercase' | 'lowercase' | 'initial' | 'inherit';
+}
+
+export const textTransform: styleFn = (x: TextTransformProps) => {
+  if (isDefined(x.textTransform)) {
```
VWSCoronaDashboard10, line 10: I think we can make the textTransform responsive by using the `asResponsiveArray` as you can see in the spacing implementation (together with the `ResponsiveValue` type)

VWSCoronaDashboard10: This allows us to use the textTransform as any other styled system prop `textTransform={{ _: 'uppercase', md: 'lowercase' }}`

VWSCoronaDashboard10: but I can image this won't ever be used .. so it's also fine to leave it as it is :-)

VWSCoronaDashboard10 (changes requested):
getting there!

packages/app/src/domain/topical/highlights-tile.tsx
```diff
@@ -0,0 +1,62 @@
+import { Box } from '~/components-styled/base';
+import {
+  HighlightTeaser,
+  HighlightTeaserProps,
+} from '~/components-styled/highlight-teaser';
+import { useBreakpoints } from '~/utils/useBreakpoints';
+import { ArticleBox } from './article-list';
+import { Block, ImageBlock } from '~/types/cms';
+
+export interface weeklyHighlightProps {
```
VWSCoronaDashboard8, line 10:
```suggestion
export interface WeeklyHighlightProps {
```

packages/app/src/domain/topical/highlights-tile.tsx
```diff
@@ -0,0 +1,62 @@
...
+  title: string;
+  slug: {
+    current: string;
+  };
```
VWSCoronaDashboard8, line 14: Why is it shaped like this? What's the point in having a nested property if there is only one? Also I wonder what "current" means in the context of a slug.

VWSCoronaDashboard14: @VWSCoronaDashboard8 the way sanity gives back a generated slug based on the title is `slug.current`. It wasn't my choice using the format this way.

PR #2424, opened 2021-03-26
Create formatDateFromTo formatter

VWSCoronaDashboard10:


packages/app/src/intl/hooks/use-intl.ts
```diff
@@ -214,6 +214,34 @@ export function useIntl() {
...
+   *
+   *     formatDateFromTo(1 maart, 7 maart)
+   *     output: { dateFrom: '1', dateTo: '7 maart' }
+   *
+   * or:
+   *
+   *    formatDateFromTo(29 maart, 4 april)
+   *    output: { dateFrom: '29 maart', dateTo: '4 april' }
+   */
+  function formatDateFromTo(dateFromInput: DateInput, dateToInput: DateInput) {
```
VWSCoronaDashboard1, line 231: nit: is this really a formatting function though? Isn't it more of a `createDateFromTo`? I kind of expect a `format*` function to return a string...

VWSCoronaDashboard1 (approved):

packages/app/src/intl/hooks/use-intl.ts
VWSCoronaDashboard10: it is indeed formatting, but the formatted values are in an object. In hindsight I think it's cleaner to return a tuple instead of an object.

VWSCoronaDashboard8: Yes I agree on the tuple. Also, i think the suffix "input" is not really helping. Arguments to a function are inherently always considered input.  It doesn't tell me anything about that it is, and confusingly the input date is then transformed to a date with getDate(). So what is what? I'm assuming before getDate it is not actually a Date object then. Is it a timestamp? Could the arguments be called `fromDateUnix` and `toDateUnix`?

VWSCoronaDashboard8: I would similarly suggest to use variable names `fromDate`, `toDate`, `fromDateText` and `toDateText` below. I think it reads better.

VWSCoronaDashboard8 (approved):

packages/app/src/intl/hooks/use-intl.ts
```diff
@@ -239,7 +239,7 @@ export function useIntl() {
       : formatDate(dateFrom);
     const dateToText = formatDate(dateTo);
 
-    return { dateFrom: dateFromText, dateTo: dateToText };
+    return [dateFromText, dateToText] as [dateFrom: string, dateTo: string];
```
VWSCoronaDashboard8, line 242: I think you can simply say `as const` here

VWSCoronaDashboard10: yep, but this syntax allows us to name values in tuples :-)

VWSCoronaDashboard8:
Another naming suggestion: Call the function formatDateSpanString and name the arguments start and end instead of from/to. It might make it easier to write clear variable names.

packages/app/src/intl/hooks/use-intl.ts
VWSCoronaDashboard10: the DateInput type is this:
```tsx
type DateInput = { seconds: number } | { milliseconds: number };
```
Which forces us to explicitly define the date in the calling context being either seconds or milliseconds.

I agree on the naming things.

packages/app/src/intl/hooks/use-intl.ts
VWSCoronaDashboard8: I don't get it. Don't you always have to name the tuple items at the receiving end anyway?

VWSCoronaDashboard10: Yep, but now TS can display the name of each value of tuple. 
image

In this case it doesn't really add a lot of value because it's probably clear enough which is which, but yeah. I thought it couldn't harm.

packages/app/src/intl/hooks/use-intl.ts
VWSCoronaDashboard10: Renamed `DateInput` to `DateDefinition` by lack of inspiration.

PR #2438, opened 2021-03-28
Nursing home 7 day avg chart

VWSCoronaDashboard8:
## Summary

Add a new trend type "bar" to the TimeSeriesChart allowing us to show absolute reported values together with a moving average trend line.

In addition this PR fixes:
- Misalignment of the labels on the x-axis for shorter timeframes
- Tooltip not displaying 0 values.

The moving average trend still needs to be connected to real data, but I think that shouldn't stop us from review/QA, since the line and tooltip will just show different values when connected.

image
image
image

### DIFF
--- a/packages/app/src/components-styled/time-series-chart/components/axes.tsx
+++ b/packages/app/src/components-styled/time-series-chart/components/axes.tsx
@@ -81,6 +81,8 @@ export const Axes = memo(function Axes({
     [startUnix, endUnix, formatDateFromSeconds]
   );
 
+  const hasYearLabel = formatXAxis(startUnix).length > 6;
+
   return (
     <g css={css({ pointerEvents: 'none' })}>
       <GridRows
@@ -115,10 +117,21 @@ export const Axes = memo(function Axes({
           fontSize: 12,
           /**
            * Using anchor middle the line marker label will fall nicely on top
-           * of the axis label
+           * of the axis label.
+           *
+           * The only times at which we can not use middle is if we are
+           * rendering a year in the label, because it becomes too long.
            */
           textAnchor:
-            x === startUnix ? 'start' : x === endUnix ? 'end' : 'middle',
+            x === startUnix
+              ? hasYearLabel
+                ? 'start'
+                : 'middle'
+              : x === endUnix
+              ? hasYearLabel
+                ? 'end'
+                : 'middle'
+              : 'middle',
         })}
         hideTicks
       />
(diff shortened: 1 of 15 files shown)


VWSCoronaDashboard13 (approved):

VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/time-series-chart/components/bar-trend.tsx
```diff
@@ -0,0 +1,111 @@
...
+    <>
+      {nonNullSeries.map((item, index) => {
+        return (
+          /**
+           * We could use Visx shape Bar here, but it is almost the same as rect
+           * and rect gives a bit more flexibility for example with regards to
+           * hover state if we choose to implement that.
+           */
+          <rect
+            id={`bar-${index}`}
```
VWSCoronaDashboard10, line 61: why the id?

packages/app/src/pages/landelijk/verpleeghuiszorg.tsx
```diff
@@ -56,7 +54,7 @@ const NursingHomeCare = (props: StaticProps<typeof getStaticProps>) => {
   const infectedLocationsText = siteText.verpleeghuis_besmette_locaties;
   const positiveTestedPeopleText =
     siteText.verpleeghuis_positief_geteste_personen;
-  const locationDeaths = siteText.verpleeghuis_oversterfte;
+  const deceased = siteText.verpleeghuis_oversterfte;
```
VWSCoronaDashboard10, line 57: isn't `deceased` is a bit too vague?
```suggestion
  const deceasedText = siteText.verpleeghuis_oversterfte;
```

VWSCoronaDashboard10 (approved):

packages/app/src/components-styled/time-series-chart/components/axes.tsx
```diff
@@ -81,6 +81,8 @@ export const Axes = memo(function Axes({
     [startUnix, endUnix, formatDateFromSeconds]
   );
 
+  const hasYearLabel = formatXAxis(startUnix).length > 6;
```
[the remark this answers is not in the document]
VWSCoronaDashboard8: I don't think that would be a big improvement. Getting the year span also doesn't necessarily mean that the label was formatted with a year included. I think I can rename the variable "isLongLabel" then the purpose is clear. It doesn't have to be about the year, because if the full month was written out then we also would have to use the same logic for positioning the label probably. The check is more about "is the label short enough to center it".

packages/app/src/components-styled/time-series-chart/components/bar-trend.tsx
VWSCoronaDashboard8: Good question, I carried it over from my older bar chart.

PR #2444, opened 2021-03-29
Update sidebar with a button for Measures and Risk level items

VWSCoronaDashboard14:
The sidebar can now render a button styling for an item.
It introduces the `isButton` and `buttonVariant` properties to display them nicely on the screen. 

Default
Screenshot 2021-03-29 at 14 39 21

Top and bottom stacked on top of each other
Screenshot 2021-03-29 at 14 39 31

### DIFF
--- a/packages/app/src/components-styled/aside/menu.tsx
+++ b/packages/app/src/components-styled/aside/menu.tsx
@@ -11,6 +11,7 @@ import { Link } from '~/utils/link';
 import { Box } from '../base';
 import { Category } from './category';
 import { Title } from './title';
+import { asResponsiveArray } from '~/style/utils';
 
 type Url = UrlObject | string;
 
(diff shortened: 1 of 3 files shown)


packages/app/src/components-styled/aside/menu.tsx
```diff
@@ -87,11 +82,23 @@ export function MetricMenuItemLink({
   const isActive = isActivePath(router, href);
 
   return (
-    <MetricMenuItem>
-      <Link href={href} passHref>
-        <StyledLink isActive={isActive}>{content}</StyledLink>
-      </Link>
-    </MetricMenuItem>
+    <>
```
VWSCoronaDashboard4, line 85: Why not use a completely separate component `MetricMenuButton` and use it in the menu? Now this component get very crowded and hard to read.

When you edit the sidebar, it's much easier to read `MetricMenuButton` than `MetricMenuItemLink` and have to debug what combination of props make it behave like it does.

packages/app/src/components-styled/aside/menu.tsx
```diff
@@ -105,6 +112,41 @@ function isActivePath(router: NextRouter, href: Url) {
...
+const MetricMenuButton = styled.li<{
+  buttonVariant: buttonVariantType;
+  isActive?: boolean;
+}>((x) =>
+  css({
+    borderRadius: x.buttonVariant === 'default' ? 5 : undefined,
+    border: '1px solid',
+    borderColor: 'border',
+
+    borderTopLeftRadius: x.buttonVariant === 'top' ? 5 : undefined,
```
VWSCoronaDashboard4, line 136: If you'd wrap these components, you could use `first-child`/`last-child` ?

VWSCoronaDashboard4 (changes requested):

packages/app/src/components-styled/aside/menu.tsx
VWSCoronaDashboard14: I can't quite figure out an elegant way to do that, I've tried some different approaches.

PR #2445, opened 2021-03-29
Chart for infected per age group per 100k over time

VWSCoronaDashboard4:
## Summary

Chart and interactive legend for infected per age group per 100k over time


packages/app/src/domain/tested/infected-per-age-group/components/age-group-legend.tsx
```diff
@@ -0,0 +1,182 @@
...
+    requestAnimationFrame(() => {
+      const item: HTMLElement | null = document.querySelector(
+        `button[data-metric-property=${metricProperty}]`
+      );
+      if (item) {
+        item.focus();
+      }
+    });
+  }
+
```
VWSCoronaDashboard1, line 42: Shouldn't this be in a useEffect() block? I'm not sure how this affects SSR otherwise.

VWSCoronaDashboard1 (approved):

packages/app/src/domain/tested/infected-per-age-group/components/age-group-legend.tsx
VWSCoronaDashboard4: it's in an event listener; SSR does not trigger the event listeners

VWSCoronaDashboard1:
I'm QA-ing by accident, but this caught my eye when clicking through the chart on Vercel:

image

And here's when I select all the legend items:
image

I would expect the charts to be exactly the same, or am I misunderstanding something?

packages/app/src/domain/tested/infected-per-age-group/infected-per-age-group.tsx
```diff
@@ -0,0 +1,187 @@
...
+  const { siteText } = useIntl();
+  const text = siteText.infected_per_agroup;
+
+  const ageGroupLegendConfig: (LineSeriesDefinition<NlTestedPerAgeGroupValue> & {
+    isSelected: boolean;
+  })[] = useMemo(() => {
+    return SERIES_CONFIG.map((baseAgeGroup) => {
+      const isSelected =
+        ageGroupSelection.includes(baseAgeGroup.metricProperty) ||
+        ageGroupSelection.length === 0;
```
VWSCoronaDashboard8, line 143: I don't get why you're injecting isSelected here in the config when you're already passing the full list of selected items to the legend. Seems redundant and adding complexity to the types.

VWSCoronaDashboard8 (commented):
I think everything here related to "age group" interactivity could be quite easily made generic as it has nothing to do with age groups specifically and we might want to use this pattern in the future. Maybe put it on the backlog / tech depth.

packages/app/src/domain/tested/infected-per-age-group/components/age-group-legend.tsx
VWSCoronaDashboard10: It's weird focus is lost on rerender, focus can only be lost during unmount/remount of a component

packages/app/src/domain/tested/infected-per-age-group/infected-per-age-group.tsx
```diff
@@ -0,0 +1,185 @@
...
+  /* Conditionally wrap tooltip over two columns due to amount of items */
+  const tooltipColumns =
+    ageGroupSelection.length === 0 || ageGroupSelection.length > 4 ? 2 : 1;
+  const InfectedPerAgeGroupWrapper = styled.div(
+    css({
+      [`${TooltipChildren}`]: {
+        columns: tooltipColumns,
+      },
+    })
+  );
```
VWSCoronaDashboard10, line 167: You're generating a new component on every render, this is why you needed the requestAnimationFrame hack.

VWSCoronaDashboard10: so move this component outside of the function body and then you should be able to clean up the requestAnimationFrame :-)

packages/app/src/domain/tested/infected-per-age-group/components/age-group-legend.tsx
```diff
@@ -0,0 +1,183 @@
...
+                </ItemButton>
+              </Item>
+            );
+          })}
+        </List>
+        <ResetButton
+          onClick={onReset}
+          css={css({
+            visibility: hasNoneSelected ? 'hidden' : 'visible',
+          })}
```
VWSCoronaDashboard10, line 67: the ResetButton has a fontWeight prop defined which is not used from the outside and you're applying dynamic styles here with a local variable which would be prettier as a prop. I suggest to rename the `fontWeight`-prop to `isHidden` and set visibility-styles in that component itself.

```tsx
<ResetButton
  onClick={onReset}
  isHidden={hasNoneSelected}

File: packages/app/src/domain/tested/infected-per-age-group/infected-per-age-group.tsx [hunk 5]
```diff
@@ -0,0 +1,185 @@
...
+        ageGroupSelection.length === 0
+    );
+  }, [ageGroupLegendConfig, ageGroupSelection]);
+
+  /* Conditionally wrap tooltip over two columns due to amount of items */
+  const tooltipColumns =
+    ageGroupSelection.length === 0 || ageGroupSelection.length > 4 ? 2 : 1;
+  const InfectedPerAgeGroupWrapper = styled.div(
+    css({
+      [`${TooltipChildren}`]: {
```
VWSCoronaDashboard10, line 163: can't we fix this with configuration instead of using a strong css selector? It looks a bit like spaghetti when we start importing sub-components of the time-series-chart and apply these kind of transformations on it from the outside. TS also won't warn us when we change the styles of TooltipChildren, which might conflict with this implementation. Imho it would be more transparent to have it set with configuration, translating into a prop.


Also the value of the left columns is rendered closer to the label of the second column, this might be confusing though it might something for design to decide on:
image

packages/app/src/domain/tested/infected-per-age-group/infected-per-age-group.tsx
```diff
@@ -0,0 +1,185 @@
...
+  return (
+    <InfectedPerAgeGroupWrapper>
+      <TimeSeriesChart
+        values={values}
+        timeframe={timeframe}
+        seriesConfig={[...ageGroupChartConfig]}
+        disableLegend
+      />
+      <AgeGroupLegend
+        seriesConfig={[...ageGroupLegendConfig]}
```
VWSCoronaDashboard10, line 178: why do you clone the array?

packages/app/src/domain/tested/infected-per-age-group/components/age-group-legend.tsx
```diff
@@ -0,0 +1,183 @@
...
+  seriesConfig,
+  ageGroupSelection,
+  onToggleAgeGroup,
+  onReset,
+}: AgeGroupLegendProps) {
+  const { siteText } = useIntl();
+  const text = siteText.infected_per_agroup;
+
+  const hasNoneSelected = ageGroupSelection.length === 0;
+  const hasAllSelected = ageGroupSelection.length === 0;
```
VWSCoronaDashboard10, line 25: these two will always hold the same value?

VWSCoronaDashboard8 (approved):

packages/app/src/domain/tested/infected-per-age-group/logic/use-age-group-selection.ts
```diff
@@ -0,0 +1,35 @@
+import { useState } from 'react';
+
+export function useAgeGroupSelection() {
```
VWSCoronaDashboard10, line 3: This hook is named rather specific, but the logic is very generic. I think we could create a shared hook, something like this:
```tsx
function useList<T>(initialValue: T[] = []) {
  const [list, setList] = useState<T[]>(initialValue)

  const add = useCallback((x: T) => {
    setList(items => items.concat(x))
  }, [])

  const remove = useCallback((x: T) => {
    setList(items => items.filter((x) => x !== item))
  }, [])

  const toggle = useCallback((item: T) => setList(items =>
    items.includes(item)
      ? remove(item)
      : add(remove)
  ), [add, remove])

  const reset = useCallback(() => setList([]), [])

  return {
    list,
    add,
    remove,
    toggle,
    clear
  }
}

**Review by VWSCoronaDashboard10 — APPROVED** [review]

File: packages/app/src/domain/tested/infected-per-age-group/infected-per-age-group.tsx [hunk 9]
```diff
@@ -149,37 +144,32 @@ export function InfectedPerAgeGroup({ timeframe }: InfectedPerAgeGroup) {
...
+          </div>
+        )}
       />
       <AgeGroupLegend
-        seriesConfig={[...ageGroupLegendConfig]}
-        ageGroupSelection={ageGroupSelection}
-        onToggleAgeGroup={toggleAgeGroup}
-        onReset={resetAgeGroupSelection}
+        seriesConfig={ageGroupLegendConfig}
+        ageGroupSelection={list}
```
VWSCoronaDashboard8, line 169: You were almost there.... This could be just "selection" and the component named "SelectableLegend" or something. and the other prop onToggleSelection.

PR #2459, opened 2021-03-30
Update chart-tile

VWSCoronaDashboard14:
Update of the chart-tile where it is a bit more flexible instead of using a normal `ChartTile` and a `TimeSeriesTile`. They are now both combined.

You could set a `timeframeOptions` prop for the Tile and injects the timeframe directly into the `TimeSeriesChat` (or any other that uses the `timeseries` prop.

The only thing to do is waiting what to do with the AriaLabels, I currently left them out and waiting for we have a final decision on that.


packages/app/src/components-styled/chart-tile.tsx
```diff
@@ -140,3 +140,65 @@ function ChartTileHeader({
...
+    <ChartTileContainer metadata={metadata}>
+      <NewChartTileHeader
+        title={title}
+        description={description}
+        timeframe={timeframe}
+        timeframeOptions={timeframeOptions}
+        onTimeframeChange={setTimeframe}
+      />
+      {timeframeOptions && React.cloneElement(children, { timeframe })}
+      {!timeframeOptions && children}
```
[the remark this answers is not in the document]
VWSCoronaDashboard8: I think we need to support two variations of children here. 

1. A normal react component is expected when you don't set a timeframe option.
2. A "function as a child" is expected when you set a timeframe option. In that case the function is called with the timeframe setting. (like the ChartTileWithTimeframe component does right now)

If you want you could check the type of the children based on the switch between timeframe or  no timeframe, but I don't think it's necessary.

packages/app/src/components-styled/chart-tile.tsx
```diff
@@ -1,72 +1,72 @@
...
-      <ChartTileHeader
-        title={title}
-        description={description}
-        ariaDescription={ariaDescription}
-        ariaLabelledBy={ariaLabelledBy}
-      />
-      {children}
-    </ChartTileContainer>
+    <Box display="flex" flexDirection={{ _: 'column', lg: 'row' }}>
+      <Box maxWidth="maxWidthText" mr="auto" pr={{ _: 0, lg: 3 }}>
```
VWSCoronaDashboard8, line 28:
```suggestion
      <Box maxWidth={maxWidthText} mr="auto" pr={{ _: 0, lg: 3 }}>
```

packages/app/src/components-styled/chart-tile.tsx
```diff
@@ -1,72 +1,72 @@
...
+      {children && (
+        <Box
+          display="inline-table"
+          alignSelf={{ _: 'flex-start', lg: 'flex-end' }}
+          mb={hasExtraToggle ? asResponsiveArray({ _: 3, lg: 0 }) : 3}
+          css={css({
+            transform: hasExtraToggle
+              ? asResponsiveArray({ lg: 'translateY(100%)' })
+              : undefined,
+            zIndex: 3,
```
VWSCoronaDashboard8, line 41: This looks like a funky way to solve a layout problem. Maybe css grid is an option now we don't support IE11 anymore?

packages/app/src/components-styled/chart-tile.tsx
```diff
@@ -1,72 +1,72 @@
...
+    </Box>
   );
 }
 
-export function ChartTileWithTimeframe({
+interface ChartTileProps {
+  title: string;
+  metadata: MetadataProps;
+  description?: string;
+  children?: React.ReactNode;
```
VWSCoronaDashboard8, line 55: I don't think children should be optional in a chart tile. We would never want to render a chart tile without a chart right?

packages/app/src/components-styled/chart-tile.tsx
```diff
@@ -1,72 +1,72 @@
...
+}
+
+export function ChartTile({
   title,
   description,
+  children,
   metadata,
-  timeframeOptions = ['all', '5weeks', 'week'],
+  timeframeOptions,
   timeframeInitialValue = 'all',
```
VWSCoronaDashboard8, line 67: nit; initial value is dependent on the options, so I think it would be more robust to take the first option as the default. Technically you can't know that "all" will be part of the options.

packages/app/src/components-styled/chart-tile.tsx
```diff
@@ -76,67 +76,160 @@ export function ChartTileWithTimeframe({
...
+      </ChartTileHeader>
+      {/* Clone element and assign a timeframe to it if there are timeframeOptions */}
+      {timeframeOptions
+        ? Array.isArray(children)
+          ? children.map((item, index) =>
+              React.cloneElement(item as React.ReactElement, {
+                timeframe,
+                key: index,
+              })
+            )
```
VWSCoronaDashboard8, line 97: What's the use case for an array of children where you set timeframe on each of the children? If the chart is expected to be the children then I don't think we should support and array of them.

packages/app/src/components-styled/chart-tile.tsx
```diff
@@ -76,67 +76,160 @@ export function ChartTileWithTimeframe({
...
+//           <ChartTimeControls
+//             timeframeOptions={timeframeOptions}
+//             timeframe={timeframe}
+//             onChange={onTimeframeChange}
+//           />
+//         </div>
+//       )}
+//     </Box>
+//   );
+// }
```
VWSCoronaDashboard8, line 235: This can all go of course

packages/app/src/domain/deceased/deceased-monitor-section.tsx
```diff
@@ -31,7 +31,7 @@ export function DeceasedMonitorSection({
         </AnchorTile>
       )}
 
-      <ChartTile
+      <NewChaChartTilertTile
```
VWSCoronaDashboard8, line 34: Typo

packages/app/src/pages/gemeente/[code]/rioolwater.tsx
```diff
@@ -159,29 +156,28 @@ const SewerWater = (props: StaticProps<typeof getStaticProps>) => {
             </KpiTile>
           </TwoKpiSection>
 
-          <ChartTileWithTimeframe
+          <ChartTile
             title={text.linechart_titel}
             metadata={{ source: text.bronnen.rivm }}
             timeframeOptions={['all', '5weeks']}
+            hasExtraToggle
```
Comment (VWSCoronaDashboard8) [inline packages/app/src/pages/gemeente/[code]/rioolwater.tsx:163]: I don't really like the name of this prop, since a dropdown is not a toggle. hasAdditionalNavigation?

packages/app/src/pages/gemeente/[code]/rioolwater.tsx
```diff
@@ -159,29 +156,28 @@ const SewerWater = (props: StaticProps<typeof getStaticProps>) => {
...
-                  range_description: text.graph_range_description,
-                  display_outliers: text.display_outliers,
-                  hide_outliers: text.hide_outliers,
-                }}
-              />
-            )}
-          </ChartTileWithTimeframe>
+            <SewerChart
+              data={data}
+              timeframe={'all'}
```
Comment (VWSCoronaDashboard8) [inline packages/app/src/pages/gemeente/[code]/rioolwater.tsx:167]: I think this needs to be removed

packages/app/src/pages/landelijk/positief-geteste-mensen.tsx
```diff
@@ -227,56 +227,52 @@ const PositivelyTestedPeople = (props: StaticProps<typeof getStaticProps>) => {
...
 
-          <ChartTileWithTimeframe
+          <ChartTile
             title={text.linechart_titel}
             description={text.linechart_toelichting}
             metadata={{
               source: text.bronnen.rivm,
             }}
+            timeframeOptions={['all', '5weeks', 'week']}
+            timeframeInitialValue="all"
```
VWSCoronaDashboard8, line 237: nit; This was the default already

packages/app/src/pages/landelijk/positief-geteste-mensen.tsx
```diff
@@ -227,56 +227,52 @@ const PositivelyTestedPeople = (props: StaticProps<typeof getStaticProps>) => {
...
 
-          <ChartTileWithTimeframe
+          <ChartTile
             title={siteText.infected_per_age_group.title}
             description={siteText.infected_per_age_group.description}
             metadata={{
               source: text.bronnen.rivm,
             }}
+            timeframeOptions={['all', '5weeks', 'week']}
+            timeframeInitialValue="all"
```
VWSCoronaDashboard8, line 272: same for this one

packages/app/src/pages/veiligheidsregio/[code]/rioolwater.tsx
```diff
@@ -151,37 +148,35 @@ const SewerWater = (props: StaticProps<typeof getStaticProps>) => {
...
-                  range_description: text.graph_range_description,
-                  display_outliers: text.display_outliers,
-                  hide_outliers: text.hide_outliers,
-                }}
-              />
-            )}
-          </ChartTileWithTimeframe>
+            <SewerChart
+              data={data}
+              timeframe={'5weeks'}
```
Comment (VWSCoronaDashboard8) [inline packages/app/src/pages/veiligheidsregio/[code]/rioolwater.tsx:160]: Looks like this needs to be removed, and set as initial value on the parent tile

PR #2468, opened 2021-03-30
Update styling sidebar button

VWSCoronaDashboard14:
Small padding update on the buttons in the sidebar

### DIFF
--- a/packages/app/src/components-styled/aside/menu.tsx
+++ b/packages/app/src/components-styled/aside/menu.tsx
@@ -26,12 +26,14 @@ export function Menu({ children }: { children: ReactNode }) {
 export function CategoryMenu({
   title,
   children,
+  isSmallPaddingTop,
 }: {
   title: string;
   children: ReactNode;
+  isSmallPaddingTop?: boolean;
 }) {
   return (
-    <Box as="li" spacing={3} pt={5}>
+    <Box as="li" spacing={3} pt={isSmallPaddingTop ? 4 : 5}>
       <Category>{title}</Category>
       <Menu>{children}</Menu>
     </Box>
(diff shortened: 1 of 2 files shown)


packages/app/src/components-styled/aside/menu.tsx
```diff
@@ -26,12 +26,14 @@ export function Menu({ children }: { children: ReactNode }) {
 export function CategoryMenu({
   title,
   children,
+  isSmallPaddingTop,
```
VWSCoronaDashboard1, line 29: nit: call this `hasSmallTopPadding`

VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/aside/menu.tsx
VWSCoronaDashboard8: I suspect that the small padding is a result of something else? Like isFirstChild or something... Maybe you can come up with a name that indicates _why_ it is different (regardless of padding)

VWSCoronaDashboard8 (approved):

packages/app/src/components-styled/aside/menu.tsx
VWSCoronaDashboard14: @VWSCoronaDashboard8 there isn't much special going on with first-child. The default is 5 for that component but was asked to reduce it to 4 on the national layout for the first item to match with the new buttons a bit better.

VWSCoronaDashboard8: First child was just a guess. If it only applies to the first item I would still call it something like isFirstItem, because that tells you more about the context than hasSmallPaddingTop

PR #2495, opened 2021-04-01
Convert all CMS javascript code to typescript

VWSCoronaDashboard1:
## Summary

The title says it all

## Motivation

TS > JS

## Detailed design

There are  a lot of `any` typings in this code. Largely because a lot of Sanity stuff isn't typed very neatly yet. Also because probably we're using some undocumented features here and there.
The preview components aren't used yet, so I just did a quick port of those. They'll get some more loving once the live previews are being implemented.
Also, I ran `sanity upgrade`.
Addendum: This PR also fixes the problem where the language switcher isn't shown for some documents where it IS needed. The fix is now to just always show the language picker.

### Governance

- [ ] Documentation is added


VWSCoronaDashboard8 (approved):
I wonder if we still have TS benefits given the amount of "any" types used, but I guess it's still an improvement over plain JS.

VWSCoronaDashboard10 (approved):

PR #2498, opened 2021-04-01
Convert R-number chart to use TimeSeriesChart

VWSCoronaDashboard13:
## Summary

Converted the R-number chart to use TimeSeriesChart and fixed one hover bug in the TimeSeriesChart when there are no points being hovered.

### DIFF
--- a/packages/app/src/components-styled/time-series-chart/logic/hover-state.ts
+++ b/packages/app/src/components-styled/time-series-chart/logic/hover-state.ts
@@ -185,6 +185,13 @@ export function useHoverState<T extends TimestampedValue>({
       .filter(isDefined);
 
     /**
+     * If there are no points getting hovered, hover state should be undefined
+     */
+    if (!linePoints.length) {
+      return undefined;
+    }
+
+    /**
      * Point markers on range data are rendered differently, so we split them
      * out here, so we avoid having to create a union type and complicate
      * things.
--- a/packages/app/src/components-styled/time-series-chart/logic/hover-state.ts
+++ b/packages/app/src/components-styled/time-series-chart/logic/hover-state.ts
@@ -188,7 +188,7 @@ export function useHoverState<T extends TimestampedValue>({
      * If there are no points getting hovered, hover state should be undefined
      */
     if (!linePoints.length) {
-      return undefined;
+      return;
     }
 
     /**
--- a/packages/app/src/domain/tested/reproduction-chart-tile.tsx
+++ b/packages/app/src/domain/tested/reproduction-chart-tile.tsx
@@ -29,7 +29,10 @@ export function ReproductionChartTile({
    * of all the values before the first datapoint with a null value to
    * display in the chart
    */
-  const values = data.values.filter((x) => isPresent(x.index_average));
+  const values = data.values.slice(
+    0,
+    data.values.findIndex((x) => !isPresent(x.index_average))
+  );
   const last_value = last(values) as NationalReproductionValue;
 
   return (
@@ -51,7 +54,7 @@ export function ReproductionChartTile({
             {
               type: 'line',
               metricProperty: 'index_average',
-              label: 'R',
+              label: text.lineLegendLabel,
               color: colors.data.primary,
             },
           ]}
--- a/packages/app/src/domain/tested/reproduction-chart-tile.tsx
+++ b/packages/app/src/domain/tested/reproduction-chart-tile.tsx
@@ -1,6 +1,11 @@
 import { useIntl } from '~/intl';
 import { colors } from '~/style/theme';
-import { NationalReproduction } from '@corona-dashboard/common';
+import { isPresent } from 'ts-is-present';
+import { last } from 'lodash';
+import {
+  NationalReproduction,
+  NationalReproductionValue,
+} from '@corona-dashboard/common';
 import { ChartTileWithTimeframe } from '~/components-styled/chart-tile';
 import { TimeSeriesChart } from '~/components-styled/time-series-chart';
 import { TimeframeOption } from '~/utils/timeframe';
@@ -12,7 +17,7 @@ interface ReproductionChartTileProps {
 }
 
 export function ReproductionChartTile({
-  data: __data,
+  data,
   timeframeOptions = ['all', '5weeks'],
   timeframeInitialValue = 'all',
 }: ReproductionChartTileProps) {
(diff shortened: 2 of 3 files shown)


packages/app/src/components-styled/time-series-chart/logic/hover-state.ts
```diff
@@ -184,6 +184,13 @@ export function useHoverState<T extends TimestampedValue>({
       })
       .filter(isDefined);
 
+    /**
+     * If there are no points getting hovered, hover state should be undefined
+     */
+    if (!linePoints.length) {
+      return undefined;
+    }
```
VWSCoronaDashboard8, line 192: nit; 
`return undefined` would be the same as just `return` I think

packages/app/src/domain/tested/reproduction-chart-tile.tsx
```diff
@@ -0,0 +1,66 @@
+import { useIntl } from '~/intl';
```
VWSCoronaDashboard8, line 1: I'm usually not a fan of creating a wrapper component for a specific chart, because I don't think it adds much value and it makes it more difficult to see on the page what data is being used.

Sometimes I do create a wrapper to easily co-locate some mock data without placing it in the page, but in general I think a chart should be rendered directly in a ChartTile on the page itself.

packages/app/src/domain/tested/reproduction-chart-tile.tsx
```diff
@@ -0,0 +1,66 @@
...
+  /**
+   * There is no data for the last 2 weeks so we are getting a slice
+   * of all the values before the first datapoint with a null value to
+   * display in the chart
+   */
+  const values = __data.values.slice(
+    0,
+    __data.values.map((x) => x.index_average).indexOf(null)
+  );
+  const last_value = __data.last_value;
```
VWSCoronaDashboard8, line 31: This logic would be an argument in favor of using a wrapper component, so it's fine if you leave the wrapper 👍 

Are you sure you want to pick the last value with nulls included?

packages/app/src/domain/tested/reproduction-chart-tile.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { TimeframeOption } from '~/utils/timeframe';
+
+interface ReproductionChartTileProps {
+  data: NationalReproduction;
+  timeframeOptions?: TimeframeOption[];
+  timeframeInitialValue?: TimeframeOption;
+}
+
+export function ReproductionChartTile({
+  data: __data,
```
VWSCoronaDashboard8, line 15: Why'd you name it like that? I only do this when I'm not using the variable, to shut up eslint. Or I use __ to mark a "private" property which isn't meant for public consumption.

packages/app/src/domain/tested/reproduction-chart-tile.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  /**
+   * There is no data for the last 2 weeks so we are getting a slice
+   * of all the values before the first datapoint with a null value to
+   * display in the chart
+   */
+  const values = __data.values.slice(
+    0,
+    __data.values.map((x) => x.index_average).indexOf(null)
+  );
```
VWSCoronaDashboard8, line 30: I think there is an easier way. How about this? (using isPresent from ts-is-present)

```suggestion
  const values = __data.values.filter((x) => isPresent(x.index_average))
  );
```

packages/app/src/pages/landelijk/reproductiegetal.tsx
```diff
@@ -100,33 +97,7 @@ const ReproductionIndex = (props: StaticProps<typeof getStaticProps>) => {
           </TwoKpiSection>
 
           {data.reproduction.values && (
```
VWSCoronaDashboard8, line 99: I think this check might be obsolete. The data should not be optional AFAIK

VWSCoronaDashboard8 (approved):

packages/app/src/domain/tested/reproduction-chart-tile.tsx
VWSCoronaDashboard13: good catch! i will use the last value from values

packages/app/src/domain/tested/reproduction-chart-tile.tsx
VWSCoronaDashboard13: I just did that because I saw it being done elsewhere. I can also just use the original param name 👍

packages/app/src/domain/tested/reproduction-chart-tile.tsx
VWSCoronaDashboard13: Hmm I think I did it that way to avoid potentially removing values in the middle of the data set, but since this uses a linear scale it shouldn't matter. I'll use what you proposed 👍

packages/app/src/domain/tested/reproduction-chart-tile.tsx
```diff
@@ -0,0 +1,68 @@
...
+}: ReproductionChartTileProps) {
+  const { siteText } = useIntl();
+  const text = siteText.reproductiegetal;
+
+  /**
+   * There is no data for the last 2 weeks so we are getting a slice
+   * of all the values before the first datapoint with a null value to
+   * display in the chart
+   */
+  const values = data.values.filter((x) => isPresent(x.index_average));
```
VWSCoronaDashboard10, line 32: >we are getting a slice of all the values before the first datapoint with a null value

The filter isn't exactly doing the same as the comment, it could in theory skip null values and display values after those null values.
If this is not an issue we can keep it like this, but otherwise you should look up the index of the first null value and slice on that index
```tsx
  const values = data.values.slice(0, data.values.findIndex(x => !isPresent(x.index_average)));
```

packages/app/src/domain/tested/reproduction-chart-tile.tsx
```diff
@@ -0,0 +1,68 @@
...
+    >
+      {(timeframe) => (
+        <TimeSeriesChart
+          values={values}
+          timeframe={timeframe}
+          seriesConfig={[
+            {
+              type: 'line',
+              metricProperty: 'index_average',
+              label: 'R',
```
VWSCoronaDashboard10, line 54: I think you can use the `text.lineLegendLabel` text for this label

VWSCoronaDashboard10 (approved):

packages/app/src/domain/tested/reproduction-chart-tile.tsx
VWSCoronaDashboard8: @VWSCoronaDashboard10 I suggested to simplify the filter to this. @VWSCoronaDashboard13 had implemented something with findIndex before which I didn't think was necessary. Your suggestion is a bit more elegant than the original implementation.

VWSCoronaDashboard10 (approved):

PR #2499, opened 2021-04-01
Adjustments tested per age group

VWSCoronaDashboard4:
## Summary

Adjustments tested per age group

* Re-enable tile
* Re-order legend labels
* Split legend into dynamic and static
* Add incomplete range to graph + legend
* remove "week" from time toggles
* "All age groups" is always visible

### DIFF
--- a/packages/app/src/domain/tested/infected-per-age-group/components/age-group-legend.tsx
+++ b/packages/app/src/domain/tested/infected-per-age-group/components/age-group-legend.tsx
@@ -4,6 +4,7 @@ import styled from 'styled-components';
 import { LineSeriesDefinition } from '~/components-styled/time-series-chart/logic';
 import { Text } from '~/components-styled/typography';
 import { useIntl } from '~/intl';
+import { colors } from '~/style/theme';
 import { asResponsiveArray } from '~/style/utils';
 
 interface AgeGroupLegendProps {
@@ -28,7 +29,9 @@ export function AgeGroupLegend({
 
   return (
     <>
-      <Text fontSize={1}>{text.legend_help_text}</Text>
+      <Text fontSize={1} fontWeight="bold" mb={0} mt={2}>
+        {text.legend_help_text}
+      </Text>
       <Legend>
         <List>
           {seriesConfig.map((item) => {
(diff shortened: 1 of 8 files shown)


packages/app/src/domain/tested/infected-per-age-group/components/age-group-legend.tsx
```diff
@@ -4,17 +4,20 @@ import styled from 'styled-components';
 import { LineSeriesDefinition } from '~/components-styled/time-series-chart/logic';
 import { Text } from '~/components-styled/typography';
 import { useIntl } from '~/intl';
+import { colors } from '~/style/theme';
 import { asResponsiveArray } from '~/style/utils';
 
 interface AgeGroupLegendProps {
   seriesConfig: LineSeriesDefinition<NlTestedPerAgeGroupValue>[];
+  alwaysEnabledConfig: LineSeriesDefinition<NlTestedPerAgeGroupValue>[];
```
VWSCoronaDashboard8, line 12: You should only have to pass in an array of metric properties / strings here. It doesn't make much sense to me to pass in complete config objects when you are just to select a few of the complete config that is already passed in on the other prop.

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard1 (approved):

packages/app/src/domain/tested/infected-per-age-group/components/age-group-legend.tsx
VWSCoronaDashboard4: though you need at least 4 properties of the config to render the legend properly; but it could probably be optimised a bit

packages/app/src/domain/tested/infected-per-age-group/infected-per-age-group.tsx
```diff
@@ -57,13 +85,24 @@ export function InfectedPerAgeGroup({
         height={breakpoints.md ? 300 : 250}
         disableLegend
         formatTooltip={(data) => (
-          <div css={css({ columns: tooltipColumns })}>
+          <div css={css({ [`${TooltipList}`]: { columns: tooltipColumns } })}>
```
VWSCoronaDashboard10, line 88: Is this nested css targeting necessary? I thought I saw it working previously without this css?

PR #2504, opened 2021-04-02
Kill lokalize part 1

VWSCoronaDashboard8:
## Summary

The basis for converting and handling Lokalize data in Sanity.


packages/cms/src/plugins/translate/select-language.tsx
```diff
@@ -50,20 +48,12 @@ export default function SelectLanguage(props: SelectLanguageProps) {
...
-                    code={lang.id === 'en' ? 'gb' : lang.id}
-                    width="24"
-                    height="12"
-                  />
-                  {validationErrors[lang.id] ? (
-                    <MdErrorOutline
-                      color={validationErrors[lang.id] ? 'red' : undefined}
-                    />
-                  ) : null}
-                </>
```
VWSCoronaDashboard1, line 64: I think we concluded that the validation error indicator would be kept next to the flag because it indicates in which language the error is contained in.

packages/cms/src/schemas/documents/lokalize-subject.ts
```diff
@@ -0,0 +1,67 @@
...
+export const lokalizeSubject = {
+  name: 'lokalizeSubject',
+  type: 'document',
+  title: 'Lokalize',
+  icon: FaLanguage,
+  /**
+   * We could limit what actions are allowed on these documents. It doesn't seem
+   * there is a way to limit this based on user role.
+   */
+  __experimental_actions: ['create', 'update', 'delete', 'publish'],
```
VWSCoronaDashboard1, line 28: We don't use these experimental features anymore. They are deprecated actually.
The new way of defining actions per document type can be added to `resolve-document-actions.ts`. I think the code there speaks for itself.

VWSCoronaDashboard1 (changes requested):

packages/cms/src/plugins/translate/select-language.tsx
VWSCoronaDashboard8: @VWSCoronaDashboard1 I brought back the red outline for the flag, as I assumed that would be enough of an indication. Right next to the flags you then have the icon with the exclamation mark already, so I feel the combination of both is sufficient.

packages/cms/src/lokalize/import.ts
```diff
@@ -0,0 +1,136 @@
...
+       * until we kill Lokalize those files are the source of truth. However,
+       * newly created text entries in Sanity will remain untouched, so we could
+       * already start to extend the Lokalize string set using Sanity if we want
+       * to.
+       */
+      .createOrReplace(document)
+      .catch((err) => {
+        console.error(
+          `Failed to create document for key ${key}: ${err.message}`
+        );
```
VWSCoronaDashboard10, line 131: Do we want the import/export scripts to fail silently or should it actually throw an error to block build scripts?

VWSCoronaDashboard10: or would console.error result in the same behavior as a throw?

packages/cms/src/schemas/documents/lokalize-subject.ts
```diff
@@ -0,0 +1,62 @@
...
+      name: 'key',
+      type: 'string',
+      /**
+       * This key is filled with the root-level JSON key at import, but could be
+       * changed later after we kill Lokalize. It allows us to rename the keys
+       * we have, which should make it easier to keep things clean.
+       */
+      description:
+        'A unique snake-cased key that will form the root-level JSON key.',
+      // hidden: true,
```
VWSCoronaDashboard10, line 36: comment necessary?

packages/cms/src/schemas/locale/locale-text.ts
```diff
@@ -8,5 +8,12 @@ export default {
     title: lang.title,
     name: lang.id,
     type: 'text',
+    rows: 4,
+    /**
+     * Only NL is required. For EN we use NL as a fallback when exporting. Both
+     * can be 1000 chars long (default is 120)
+     */
+    validation: (Rule: any) =>
+      lang.id === 'nl' ? Rule.required().max(1000) : Rule.max(1000),
```
VWSCoronaDashboard10, line 17: I checked out the branch and saw the following error pop up:
http://localhost:3333/desk/lokalize;jsonKey__ic_opnames_per_dag
image

is this configuration not working or not related?

packages/cms/src/schemas/documents/lokalize-text.ts
```diff
@@ -0,0 +1,54 @@
...
+      const shortLength = shortSubtitle.length;
+      return {
+        title,
+        subtitle:
+          shortLength === SHORTEN_LENGTH
+            ? `${shortSubtitle}...`
+            : shortSubtitle,
+      };
+    },
+  },
```
VWSCoronaDashboard10, line 53: is shortening the subtitle necessary? It looks like the sanity UI will automatically cut off long values together with an ellipsis.

packages/cms/src/lokalize/export.ts
```diff
@@ -0,0 +1,48 @@
...
+      }
+    );
+    fs.writeFileSync(
+      path.join(localeDirectory, 'en_export.json'),
+      prettier.format(JSON.stringify(unflatten(en)), { parser: 'json' }),
+      {
+        encoding: 'utf8',
+      }
+    );
+  });
```
VWSCoronaDashboard10, line 48: you forgot to catch errors on this promise

VWSCoronaDashboard10 (approved):
couple of comments/questions.

packages/cms/src/desk-structure.ts
```diff
@@ -62,139 +30,104 @@ const hiddenDocTypes = [
...
 ];
 
 export default () =>
-  StructureBuilder.list()
+  S.list()
     .title('Content')
     .items([
-      StructureBuilder.listItem()
+      S.listItem()
+        .title('Lokalize')
```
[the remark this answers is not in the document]
VWSCoronaDashboard10: or maybe leave it like this, it's probably more clear for the comm users to name it Lokalize.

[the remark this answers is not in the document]
VWSCoronaDashboard8: In my mind once we have this working we should move to a different structure in a next iteration and then call it something else. With this name it is clear that this feature is there to replace Lokalize at first. But the title we can always change at any time.

packages/cms/src/lokalize/delete.ts
```diff
@@ -0,0 +1,22 @@
+import Queue from 'p-queue';
+import { client } from '../client';
+
+const queue = new Queue({
+  concurrency: 4,
+  interval: 1000 / 25,
```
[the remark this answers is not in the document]
VWSCoronaDashboard8: This came from @VWSCoronaDashboard initial PoC, I am not sure how these limits have been chosen. Maybe it's just the Sanity API rate limit.

packages/cms/src/lokalize/import.ts
VWSCoronaDashboard8: Import is something we do manually and only a few times until we hit the killswitch. So I think error messages in the console are fine.

packages/cms/src/schemas/locale/locale-text.ts
VWSCoronaDashboard8: Yes the 120 limit error needs to be sorted out still. I'll add it to the trello ticket of step 2. Not sure why it's not picking up the 1000 limit

packages/cms/src/schemas/documents/lokalize-subject.ts
VWSCoronaDashboard8: I'm putting them back in to wrap up this PR. I'll add a todo. Using document actions it is not immediately clear to me what the equivalent of and API document update action is and I don't have time to dive into these things now. The feature is also not deprecated yet.

packages/cms/src/client.ts
```diff
@@ -0,0 +1,19 @@
...
+import dotenv from 'dotenv';
+import path from 'path';
+import sanityConfig from '../sanity.json';
+
+/**
+ * @TODO load dotenv elsewhere maybe and take dataset from env
+ */
+dotenv.config({
+  path: path.resolve(process.cwd(), '.env.local'),
+});
```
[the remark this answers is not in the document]
VWSCoronaDashboard8: Simply taking out this bit breaks the code (permissions errors), so I'm leaving this for another day.

packages/cms/src/schemas/documents/lokalize-text.ts
VWSCoronaDashboard8: I must have added it because I was seeing strings that were too long, but I'll add a todo for the follow up work. I don't want to spent more time on this PR now.

VWSCoronaDashboard10 (approved):

VWSCoronaDashboard1 (approved):

PR #2510, opened 2021-04-02
Update chart-tile 

VWSCoronaDashboard14:
Update of the chart tile that can handle the callback `timeframe` better. You don't need a different chart tile when there is a timeframe involved simply set the props on the tile.

The only thing I left out is the aria labels, but there should be a solution coming for that.

Closed the old one due the fact that I kinda took a different approach there and it was easier to branch again from develop 
https://github.com/minvws/nl-covid19-data-dashboard/pull/2459.


packages/app/src/components-styled/chart-tile.tsx
```diff
@@ -1,142 +1,105 @@
...
-            onChange={onTimeframeChange}
+            onChange={setTimeframe}
           />
-        </div>
-      )}
-    </Box>
+        )}
+      </ChartTileHeader>
+      {timeframeOptions && typeof children === 'function'
+        ? children(timeframe as TimeframeOption)
```
VWSCoronaDashboard8, line 101: ... and you then also don't need to cast it to TimeframeOption here I suspect

VWSCoronaDashboard8 (approved):
Looking good! 🥳  I had a few small remarks

packages/app/src/components-styled/chart-tile.tsx
VWSCoronaDashboard14: It is a little weird if I include the initial value when creating the state and later maybe update it with the effect it won't update the active state of the toggle. That's why I need those checks since I'm assigning the state in the useEffect.

packages/app/src/components-styled/chart-tile.tsx
```diff
@@ -1,142 +1,108 @@
...
 }
 
-export function ChartTileWithTimeframe({
+type ChartTileProps = {
+  title: string;
+  metadata: MetadataProps;
+  description?: string;
+  timeframeInitialValue?: TimeframeOption;
+  hasAdditionalNavigation?: boolean;
+  ariaLabelledBy?: string;
```
VWSCoronaDashboard4, line 55: I think you can remove the aria properties for now, so also remove them from where this component is used.

VWSCoronaDashboard4 (approved):

VWSCoronaDashboard8 (approved):

PR #2513, opened 2021-04-02
Code improvements confirmed cases per age group chart

VWSCoronaDashboard4:
## Summary

Code improvements confirmed cases per age group chart

* Moved the static legend items to a regular `legend` component
* Extended the regular `legend` component to accept `dashed` lines
* Above changes allowed for some cleanup in the legend
* Adjusted `tooltip-series-list` to accept an optional `hasMultipleColumns` prop
* Cleaned up the odd code to pass columns from `infected-per-age-group` to the tooltip by using above prop
* Cleanup of `useMemo`, adjusted some names and added comments
* `age-group-legend` is made re-usable as `interactive-legend` component

### DIFF
--- a/packages/app/src/components-styled/interactive-legend.tsx
+++ b/packages/app/src/components-styled/interactive-legend.tsx
@@ -14,7 +14,7 @@ interface InteractiveLegendProps {
   helpText: string;
   seriesConfig: SimplifiedSeriesConfig[];
   selection: string[];
-  onToggleAgeGroup: (ageGroupRange: string) => void;
+  onToggleItem: (item: string) => void;
   onReset: () => void;
 }
 
@@ -22,7 +22,7 @@ export function InteractiveLegend({
   helpText,
   seriesConfig,
   selection,
-  onToggleAgeGroup,
+  onToggleItem,
   onReset,
 }: InteractiveLegendProps) {
   const { siteText } = useIntl();
@@ -41,7 +41,7 @@ export function InteractiveLegend({
             return (
               <Item key={item.label}>
                 <ItemButton
-                  onClick={() => onToggleAgeGroup(item.metricProperty)}
+                  onClick={() => onToggleItem(item.metricProperty)}
                   isActive={hasSelection && isSelected}
                   color={item.color}
                   data-text={item.label}
--- a/packages/app/src/components-styled/interactive-legend.tsx
+++ b/packages/app/src/components-styled/interactive-legend.tsx
@@ -1,38 +1,43 @@
-import { NlTestedPerAgeGroupValue } from '@corona-dashboard/common';
 import css from '@styled-system/css';
 import styled from 'styled-components';
-import { LineSeriesDefinition } from '~/components-styled/time-series-chart/logic';
 import { Text } from '~/components-styled/typography';
 import { useIntl } from '~/intl';
 import { asResponsiveArray } from '~/style/utils';
 
-interface AgeGroupLegendProps {
-  seriesConfig: LineSeriesDefinition<NlTestedPerAgeGroupValue>[];
-  ageGroupSelection: string[];
+interface SimplifiedSeriesConfig {
+  metricProperty: string;
+  label: string;
+  color: string;
+}
+
+interface InteractiveLegendProps {
+  helpText: string;
+  seriesConfig: SimplifiedSeriesConfig[];
+  selection: string[];
   onToggleAgeGroup: (ageGroupRange: string) => void;
   onReset: () => void;
 }
 
-export function AgeGroupLegend({
+export function InteractiveLegend({
+  helpText,
   seriesConfig,
-  ageGroupSelection,
+  selection,
   onToggleAgeGroup,
   onReset,
-}: AgeGroupLegendProps) {
+}: InteractiveLegendProps) {
   const { siteText } = useIntl();
-  const text = siteText.infected_per_age_group;
 
-  const hasSelection = ageGroupSelection.length !== 0;
+  const hasSelection = selection.length !== 0;
 
   return (
     <>
       <Text fontSize={1} fontWeight="bold" mb={0} mt={2}>
-        {text.legend_help_text}
+        {helpText}
       </Text>
       <Legend>
         <List>
           {seriesConfig.map((item) => {
-            const isSelected = ageGroupSelection.includes(item.metricProperty);
+            const isSelected = selection.includes(item.metricProperty);
             return (
               <Item key={item.label}>
                 <ItemButton
@@ -49,7 +54,7 @@ export function AgeGroupLegend({
           })}
           <Item>
             <ResetButton onClick={onReset} isVisible={hasSelection}>
-              {text.reset_button_label}
+              {siteText.common.interactive_legend.reset_button_label}
             </ResetButton>
           </Item>
         </List>
--- a/packages/app/src/components-styled/interactive-legend.tsx
+++ b/packages/app/src/components-styled/interactive-legend.tsx
@@ -146,6 +146,7 @@ const ItemButton = styled.button<{
 const ResetButton = styled.button<{ isVisible: boolean }>(({ isVisible }) =>
   css({
     backgroundColor: 'transparent',
+    cursor: 'pointer',
     color: 'blue',
     py: '6px',
     border: 'none',
--- a/packages/app/src/components-styled/time-series-chart/components/tooltip/tooltip-series-list.tsx
+++ b/packages/app/src/components-styled/time-series-chart/components/tooltip/tooltip-series-list.tsx
@@ -18,10 +18,10 @@ import { colors } from '~/style/theme';
 
 export function TooltipSeriesList<T extends TimestampedValue>({
   data: tooltipData,
-  columns,
+  hasMultipleColumns,
 }: {
   data: TooltipData<T>;
-  columns: 1 | 2;
+  hasMultipleColumns?: boolean;
 }) {
   const {
     value,
@@ -83,7 +83,7 @@ export function TooltipSeriesList<T extends TimestampedValue>({
           {timespanAnnotation.shortLabel || timespanAnnotation.label}
         </Text>
       )}
-      <TooltipList columns={columns ?? 1}>
+      <TooltipList hasMultipleColumns={hasMultipleColumns}>
         {seriesConfig.map((x, index) => {
           switch (x.type) {
             case 'stacked-area':
@@ -162,13 +162,14 @@ export function TooltipSeriesList<T extends TimestampedValue>({
   );
 }
 
-export const TooltipList = styled.ol<{ columns: 1 | 2 }>(({ columns }) =>
-  css({
-    columns,
-    m: 0,
-    p: 0,
-    listStyle: 'none',
-  })
+export const TooltipList = styled.ol<{ hasMultipleColumns: boolean }>(
+  ({ hasMultipleColumns }) =>
+    css({
+      columns: hasMultipleColumns ? 2 : 1,
+      m: 0,
+      p: 0,
+      listStyle: 'none',
+    })
 );
 
 interface TooltipListItemProps {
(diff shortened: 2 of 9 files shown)


packages/app/src/components-styled/time-series-chart/components/tooltip/tooltip-series-list.tsx
```diff
@@ -18,8 +18,10 @@ import { colors } from '~/style/theme';
 
 export function TooltipSeriesList<T extends TimestampedValue>({
   data: tooltipData,
+  columns,
```
VWSCoronaDashboard10, line 21: nit; you could define the default over here instead of on line 86
```suggestion
  columns = 1,
```

packages/app/src/domain/tested/infected-per-age-group/infected-per-age-group.tsx
```diff
@@ -51,43 +48,67 @@ export function InfectedPerAgeGroup({
...
-  }, [ageGroupBaseConfig, alwayEnabled]);
+  }, [seriesConfig, alwayEnabled]);
 
-  const alwaysEnabledConfig = useMemo(() => {
-    return ageGroupBaseConfig.filter((item) =>
-      alwayEnabled.includes(item.metricProperty)
-    );
-  }, [ageGroupBaseConfig, alwayEnabled]);
+  /* Static legend contains always enabled items and the under reported item */
+  const staticLegendItems: LegendItem[] = useMemo(() => {
```
VWSCoronaDashboard10, line 82: nit: caching this one and the previous one (dynamicLegendConfig) is not really necessary performance-wise, the legend component isn't doing a lot of magic with this data..

VWSCoronaDashboard1 (approved):

packages/app/src/domain/tested/infected-per-age-group/infected-per-age-group.tsx
VWSCoronaDashboard10: actually the same applies for the seriesConfig (no realneed to wrap that in a useMemo), because this `InfectedPerAgeGroup` is not re-rendered that often. Only on click of a legend item it will rerender, but that will also trigger a new seriesConfig array ..

So I guess you can get rid of all the useMemo hooks and experience even better performance than before.

packages/app/src/components-styled/time-series-chart/components/tooltip/tooltip-series-list.tsx
```diff
@@ -18,8 +18,10 @@ import { colors } from '~/style/theme';
...
 }: {
   data: TooltipData<T>;
+  columns: 1 | 2;
```
VWSCoronaDashboard8, line 24: If these are the only options that we'll ever use (which makes sense to me), I think an optional boolean `twoColumns` or `twoColumnLayout` would have been a little nicer API wise

packages/app/src/components-styled/time-series-chart/components/tooltip/tooltip-series-list.tsx
```diff
@@ -81,7 +83,7 @@ export function TooltipSeriesList<T extends TimestampedValue>({
           {timespanAnnotation.shortLabel || timespanAnnotation.label}
         </Text>
       )}
-      <TooltipList>
+      <TooltipList columns={columns ?? 1}>
```
VWSCoronaDashboard8, line 86: ... then you also don't need this default/fallback things.

packages/app/src/domain/tested/infected-per-age-group/infected-per-age-group.tsx
```diff
@@ -51,43 +48,67 @@ export function InfectedPerAgeGroup({
...
+      )
+      .concat([underReported]);
+  }, [seriesConfig, alwayEnabled, underReported]);
 
   /* Conditionally wrap tooltip over two columns due to amount of items */
-  const tooltipColumns = list.length === 0 || list.length > 4 ? 2 : 1;
+  const tooltipColumns: 1 | 2 = useMemo(
+    () => (list.length === 0 || list.length > 4 ? 2 : 1),
+    [list.length]
+  );
```
VWSCoronaDashboard10, line 100: The `useMemo` hook has AFAIK two main purposes:

1. cache expensive calculations
2. prevent expensive re-renders of a child-component which receives a non-primitive prop

both use-cases indicate that wrapping `list.length === 0 || list.length > 4 ? 2 : 1` in a useMemo hook is not necessary, and is even hurting performance :-)

packages/app/src/domain/tested/infected-per-age-group/infected-per-age-group.tsx
```diff
@@ -51,43 +48,67 @@ export function InfectedPerAgeGroup({
     });
   }, [text.legend]);
 
+  const underReported = useMemo(
```
VWSCoronaDashboard8, line 51:
```suggestion
  const underReportedConfigItem = useMemo(
```

It wasn't immediately clear to me what this variable contained.

packages/app/src/domain/tested/infected-per-age-group/series-config.ts
```diff
@@ -55,3 +56,8 @@ export const SERIES_CONFIG: {
     style: 'dashed',
   },
 ];
+
+export const UNDER_REPORTED_CONFIG: { shape: LegendShape; color: string } = {
+  shape: 'square',
+  color: colors.data.underReported,
+};
```
VWSCoronaDashboard10, line 63: It doesn't feel it's really necessary to create a const for this config

packages/app/src/domain/tested/infected-per-age-group/infected-per-age-group.tsx
VWSCoronaDashboard8, line 100: I would use a boolean as suggested earlier, but if not, it helps if you name it clearly as a number, because otherwise it reads as if it contains the actual columns component. I'd suggest `numTooltipColumns`

VWSCoronaDashboard10 (changes requested):
there are a couple of unnecessary useMemo hooks to clean up, besides that approved 👍

packages/app/src/domain/tested/infected-per-age-group/series-config.ts
```diff
@@ -1,7 +1,8 @@
 import { NlTestedPerAgeGroupValue } from '@corona-dashboard/common';
+import { LegendShape } from '~/components-styled/legend';
 import { colors } from '~/style/theme';
 
-export const SERIES_CONFIG: {
+export const BASE_SERIES_CONFIG: {
   metricProperty: keyof Omit<
     NlTestedPerAgeGroupValue,
```
VWSCoronaDashboard8, line 7: I you use `<T>` you don't need to tie this to the specific schema for NlTestedPerAgeGroupValue. Also the Omit is not necessary I think. The series config metricProperty is defined as `keyof T` which is fine IMO. I'd rather keep it simple.

packages/app/src/domain/tested/infected-per-age-group/series-config.ts
```diff
@@ -55,3 +56,8 @@ export const SERIES_CONFIG: {
     style: 'dashed',
   },
 ];
+
+export const UNDER_REPORTED_CONFIG: { shape: LegendShape; color: string } = {
```
VWSCoronaDashboard8, line 60: It might read better if you just placed these things in the file where they are use (if one file) and gave them a normal non-capitalized name. I don't see much reason for isolating this in its own file.

packages/app/src/domain/tested/infected-per-age-group/components/age-group-legend.tsx
```diff
@@ -1,23 +1,20 @@
...
-import css, { SystemStyleObject } from '@styled-system/css';
+import css from '@styled-system/css';
 import styled from 'styled-components';
 import { LineSeriesDefinition } from '~/components-styled/time-series-chart/logic';
 import { Text } from '~/components-styled/typography';
 import { useIntl } from '~/intl';
-import { colors } from '~/style/theme';
 import { asResponsiveArray } from '~/style/utils';
 
 interface AgeGroupLegendProps {
```
VWSCoronaDashboard8, line 9: This component is still called AgeGroupLegend (and the name is used in its props), but there is nothing specific to age groups in this whole thing. It could be called `InteractiveLegend` for example and live outside of the age group chart. But I'll leave it up to you if you want to take that on.

VWSCoronaDashboard8 (approved):

packages/app/src/components-styled/time-series-chart/components/tooltip/tooltip-series-list.tsx
VWSCoronaDashboard4: I went with `hasMultipleColumns` as boolean

packages/app/src/domain/tested/infected-per-age-group/series-config.ts
VWSCoronaDashboard4: Where this config is used, the `metricProperty` is used to look for the proper label in `text`. Without omit here there, some other fix is needed there.

packages/app/src/domain/tested/infected-per-age-group/infected-per-age-group.tsx
```diff
@@ -48,46 +40,62 @@ export function InfectedPerAgeGroup({
...
+        color: item.color,
+        style: item.style,
+      })
+    )
+    .concat([underReportedLegendItem]);
 
-  /* Conditionally wrap tooltip over two columns due to amount of items */
-  const tooltipColumns = list.length === 0 || list.length > 4 ? 2 : 1;
+  /* Conditionally let tooltip span over multiple columns */
+  const hasMultipleColumns = list.length === 0 || list.length > 4;
```
VWSCoronaDashboard10, line 84: It almost feels like this logic might become part of the TooltipSeriesList component itself, but maybe we should wait with that until we run into more tall tooltips with lots of items.

packages/app/src/domain/tested/infected-per-age-group/series-config.ts
VWSCoronaDashboard10: Try this:
```tsx
export const BASE_SERIES_CONFIG = {
  ...,
  ...
} as const;

**Review by VWSCoronaDashboard10 — APPROVED** [review]

File: packages/app/src/domain/tested/infected-per-age-group/infected-per-age-group.tsx (hunk 11 above)
Comment (VWSCoronaDashboard4) [reply packages/app/src/domain/tested/infected-per-age-group/infected-per-age-group.tsx:84]: Exactly, and it shouldn't be too difficult to move/adjust it at a later moment.

**Review by VWSCoronaDashboard8 — APPROVED** [review]

File: packages/app/src/domain/tested/infected-per-age-group/infected-per-age-group.tsx (hunk 11 above)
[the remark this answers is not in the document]
Comment (VWSCoronaDashboard4) [reply packages/app/src/domain/tested/infected-per-age-group/infected-per-age-group.tsx:84]: Then all are selected.

File: packages/app/src/components-styled/interactive-legend.tsx [hunk 12]
```diff
@@ -1,41 +1,43 @@
...
 
-interface AgeGroupLegendProps {
-  seriesConfig: LineSeriesDefinition<NlTestedPerAgeGroupValue>[];
-  alwaysEnabledConfig: LineSeriesDefinition<NlTestedPerAgeGroupValue>[];
-  ageGroupSelection: string[];
+interface SimplifiedSeriesConfig {
+  metricProperty: string;
+  label: string;
+  color: string;
+}
```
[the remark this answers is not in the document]
VWSCoronaDashboard4: that's a trip down the rabbit hole that I reverted; not all variants in seriesconfig have the `metricProperty`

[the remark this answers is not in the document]
VWSCoronaDashboard4: and also not all of them have a `color`; though that could be made optional / with a fallback I guess.

PR #2534, opened 2021-04-07
Vaccine stock per supplier chart

VWSCoronaDashboard4:
## Summary

Vaccine stock per supplier chart

### DIFF
--- a/packages/app/src/domain/vaccine/vaccine-stock-per-supplier-chart.tsx
+++ b/packages/app/src/domain/vaccine/vaccine-stock-per-supplier-chart.tsx
@@ -1,0 +1,26 @@
+import { NlVaccineStockValue } from '@corona-dashboard/common';
+import { TimeSeriesChart } from '~/components-styled/time-series-chart';
+import { useIntl } from '~/intl';
+
+interface VaccineStockPerSupplierChartProps {
+  values: NlVaccineStockValue[];
+}
+
+export function VaccineStockPerSupplierChart({
+  values,
+}: VaccineStockPerSupplierChartProps) {
+  const { siteText } = useIntl();
+  const text = siteText.vaccinaties.stock_per_vaccine_chart;
+
+  return (
+    <TimeSeriesChart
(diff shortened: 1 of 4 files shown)


packages/app/src/domain/vaccine/vaccine-stock-per-supplier-chart.tsx
```diff
@@ -0,0 +1,26 @@
+import { NlVaccineStockValue } from '@corona-dashboard/common';
+import { TimeSeriesChart } from '~/components-styled/time-series-chart';
+import { useIntl } from '~/intl';
+
+interface VaccineStockPerSupplierChartProps {
+  values: NlVaccineStockValue[];
+}
+
+export function VaccineStockPerSupplierChart({
```
0x80, line 9: I'm not a fan of these types of wrapper components. I think they obfuscate things in the page composition and don't add much value. Having a specialized component for every chart instance blows up the amount of components. Things that can not be reused should preferably not be components I think but made using composition in the parent context.

packages/app/src/locale/nl.json
```diff
@@ -2367,7 +2367,12 @@
       "titel": "Leveringen",
       "omschrijving": "Deze grafiek laat zien hoeveel COVID-19-vaccins Nederland per week heeft gekregen en verwacht te krijgen van de verschillende producenten."
     },
-    "bereidheid_datums": "Laatste waardes verkregen op {{dateOfInsertion}}. Wordt elke drie weken bijgewerkt."
+    "bereidheid_datums": "Laatste waardes verkregen op {{dateOfInsertion}}. Wordt elke drie weken bijgewerkt.",
+    "stock_per_vaccine_chart": {
```
0x80, line 2371: nit; The schema is called vaccine_stock_per_supplier so maybe it makes sense to call it stock_per_supplier_chart ?

0x80: I see you did use that name already for the chart too. VaccineStockPerSupplierChartProps

PR #2539, opened 2021-04-07
Fix chart hover regression

VWSCoronaDashboard10:
## Summary

I've fixed the same issue with another PR, but on line 288 by checking the existence of the nearestPoint. The removed lines in this PR enable hover interaction for bar charts.

```tsx
    /**
     * Empty hoverstate when there's no nearest point detected
     */
    if (!nearestPoint) {
      return undefined;
    }


VWSCoronaDashboard13 (approved):
🥂

PR #2540, opened 2021-04-07
Add stock per supplier chart

VWSCoronaDashboard4:
## Summary

Add stock per supplier chart


packages/app/src/domain/vaccine/vaccine-stock-per-supplier-chart.tsx
```diff
@@ -0,0 +1,85 @@
...
+      label: productNames.moderna,
+      shape: 'circle' as const,
+      type: 'line' as const,
+    },
+    {
+      metricProperty: 'astra_zeneca' as const,
+      color: colors.data.vaccines.astra_zeneca,
+      label: productNames.astra_zeneca,
+      shape: 'circle' as const,
+      type: 'line' as const,
```
VWSCoronaDashboard8, line 42: I don't think these as const are required if you type vaccineSelectOptions

PR #2550, opened 2021-04-08
Interactive legend for stock per supplier chart

VWSCoronaDashboard4:
## Summary

Interactive legend for stock per supplier chart


VWSCoronaDashboard8 (approved):
It feels like this component needs a refactor soon for various reasons:

- The interactive legend as it is called now is not going to be serving a purpose as a legend but more a select button group.
- The state management of the select and the logic to switch between radio and multi-select behavior needs to be encapsulated in the object I think.

But I'm going to approve it because it works as a sufficient implementation for now.

VWSCoronaDashboard10:
there are build errors:
image

packages/app/src/components-styled/interactive-legend.tsx
```diff
@@ -1,21 +1,23 @@
...
+export interface SimplifiedSeriesConfig {
   metricProperty: string;
   label: string;
   color: string;
+  shape: 'line' | 'circle';
 }
 
 interface InteractiveLegendProps {
   helpText: string;
   seriesConfig: SimplifiedSeriesConfig[];
```
VWSCoronaDashboard8, line 17: Since this prop is not technically configuring the series in the chart, and the name is also used elsewhere with a different type, I'd suggest to call this something different like "selectOptions" , and then also rename the type SimplifiedSeriesConfig to something like SelectOption

packages/app/src/locale/nl.json
```diff
@@ -2376,7 +2376,12 @@
     "stock_per_supplier_chart": {
       "title": "Verwachtte voorraad per vaccin type",
       "description": "Deze grafiek ....",
-      "tooltip_title": "Verwachtte voorraad per vaccin type"
+      "tooltip_title": "Verwachtte voorraad per vaccin type",
+      "legend": {
+        "available" : "Beschikbare {{vaccineName}} vaccins in voorrraad",
+        "total" : "Totaal aantal {{vaccineName}} vaccins in voorraad"
+      },
+      "legend_help_text": "Selecteer vaccintype om de voorraad te bekijken"
```
VWSCoronaDashboard8, line 2384: I think it would be better to call things here "select" instead of "legend" because if the buttons are shown on the top and the legend on the bottom they really can't be seen as a legend anymore and design mentioned that this is also how they want to start using the selector in the other chart that uses this component.

PR #2568, opened 2021-04-09
Wrap up vaccin release MVP

VWSCoronaDashboard8:
## Summary

This implements the last bits for the upcoming vaccin release. 

- Configure vaccin stock per supplier chart for stepped curve
- Move area chart to different position on the page
- Clean up code
- Add timeframe toggle
- Make y-domain constant across all supplier selectors

The MVP does not include week boundary markers

image
image
image


VWSCoronaDashboard10 (approved):

VWSCoronaDashboard4 (approved):

packages/app/src/domain/vaccine/vaccine-stock-per-supplier-chart.tsx
```diff
@@ -22,6 +25,25 @@ export function VaccineStockPerSupplierChart({
...
+                'bio_n_tech_pfizer_total',
+                'moderna_total',
+                'astra_zeneca_total',
+              ])
+            ).filter(isPresent)
+          ),
+        0
+      ),
+    [values]
+  );
```
VWSCoronaDashboard10: Is the reduce necessary or can we apply a flatmap?
```tsx
const maximumValueForAllProperties = useMemo(
  () =>
    Math.max(
      ...values.flatMap(value => 
        Object.values(
          pick(value, [
            'bio_n_tech_pfizer_total',
            'moderna_total',
            'astra_zeneca_total',
          ])
        ).filter(isPresent)
      )
    ),
  [values]
);

VWSCoronaDashboard1 (approved):

packages/app/src/domain/vaccine/vaccine-stock-per-supplier-chart.tsx
VWSCoronaDashboard8: Good suggestion 👍

VWSCoronaDashboard8: Hmm on second thought it feels less logical to me. You then create an array of all the properties in all the values before calculating the max possibly creating a huge array with `values.length * properties.length` items. 

The reduce version on the other hand is calculating max for only the properties in each value and then combining those values in the final outcome. From a resources standpoint that is clearly a better approach IMO.

PR #2584, opened 2021-04-12
Reverse linepoints to match them with the way data is drawn

VWSCoronaDashboard14:
Reverse the data of the linepoints so they match with the config object. 

Mandatory screenshot as example
Screenshot 2021-04-12 at 11 06 14

### DIFF
--- a/packages/app/src/components-styled/time-series-chart/logic/hover-state.ts
+++ b/packages/app/src/components-styled/time-series-chart/logic/hover-state.ts
@@ -218,7 +218,7 @@ export function useHoverState<T extends TimestampedValue>({
             };
         }
       })
-      .filter(isDefined);
+      .filter(isDefined).reverse();
 
     /**
      * Point markers on range data are rendered differently, so we split them
--- a/packages/app/src/components-styled/time-series-chart/logic/hover-state.ts
+++ b/packages/app/src/components-styled/time-series-chart/logic/hover-state.ts
@@ -218,7 +218,7 @@ export function useHoverState<T extends TimestampedValue>({
             };
         }
       })
-      .filter(isDefined).reverse();
+      .filter(isDefined)
 
     /**
      * Point markers on range data are rendered differently, so we split them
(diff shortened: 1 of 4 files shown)


packages/app/src/components-styled/time-series-chart/logic/hover-state.ts
```diff
@@ -218,7 +218,7 @@ export function useHoverState<T extends TimestampedValue>({
             };
         }
       })
-      .filter(isDefined);
+      .filter(isDefined).reverse();
```
VWSCoronaDashboard4, line 221: Should this be applied to all graphs?

VWSCoronaDashboard8: Yes

VWSCoronaDashboard8 (approved):

packages/app/src/components-styled/time-series-chart/logic/hover-state.ts
VWSCoronaDashboard8: I think it might be a little clearer to reverse them at the location where the points are place in the DOM, because then you would understand the reversal better, but maybe just add a comment here to explain why reverse is required.

VWSCoronaDashboard10: I'd vote for reversing during render

packages/app/src/components-styled/time-series-chart/time-series-chart.tsx
```diff
@@ -383,7 +383,7 @@ export function TimeSeriesChart<
               value={values[hoverState.valuesIndex]}
             />
             <PointMarkers points={hoverState.rangePoints} />
-            <PointMarkers points={hoverState.linePoints} />
+            <PointMarkers points={hoverState.linePoints.reverse()} />
```
VWSCoronaDashboard8, line 386: I wouldn't consider this "where they are rendered", to me that would be inside this component where the points are iterated. But again, I don't mind :)

VWSCoronaDashboard14: Aah, got it. Give me a moment @VWSCoronaDashboard8

VWSCoronaDashboard10: also be aware that `Array.prototype.reverse` mutates the array which means that this implementation would constantly flip the order on every rerender :-) With your new implementation it's not the case because you're first mapping the values to a _new_ array, which in turn is flipped.

VWSCoronaDashboard8 (approved):
🎉

PR #2597, opened 2021-04-12
Add h1's for every page that didn't have it

VWSCoronaDashboard14:
A lot of pages especially on nl/vr/gm didn't have a proper h1 also on certain content pages it has been added.

### DIFF
--- a/packages/app/src/pages/gemeente/[code]/positief-geteste-mensen.tsx
+++ b/packages/app/src/pages/gemeente/[code]/positief-geteste-mensen.tsx
@@ -83,6 +83,7 @@ const PositivelyTestedPeople = (props: StaticProps<typeof getStaticProps>) => {
             title={replaceVariablesInText(text.titel, {
               municipality: municipalityName,
             })}
+            headingLevel={1}
             icon={<Getest />}
             subtitle={text.pagina_toelichting}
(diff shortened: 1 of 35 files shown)


packages/app/src/pages/gemeente/[code]/positief-geteste-mensen.tsx
```diff
@@ -83,6 +83,7 @@ const PositivelyTestedPeople = (props: StaticProps<typeof getStaticProps>) => {
             title={replaceVariablesInText(text.titel, {
               municipality: municipalityName,
             })}
+            headingLevel={1}
```
Comment (VWSCoronaDashboard1) [inline packages/app/src/pages/gemeente/[code]/positief-geteste-mensen.tsx:86]: Wasn't the whole idea of the `category` and `screenReaderCategory` properties to be the way we add an H1 to a ContentHeader? It looks as if the category title is now being downgraded to an h2, I'm not sure if the hierarchy of the headings makes sense anymore this way...

Comment (VWSCoronaDashboard1) [reply packages/app/src/pages/gemeente/[code]/positief-geteste-mensen.tsx:86]: Oh, and I just remembered, simply changing the title from H2 to H1 will lead to problems on pages where there are multiple ContentHeader components, that will lead to multiple H1 instances.

PR #2600, opened 2021-04-13
Create time series mini bar chart

VWSCoronaDashboard10:
## Summary
image

Thanks to the horizontal padding on the x-axis and the benchmark line these charts couldn't be created with the default time series chart.. So here's a new chart, quite specific to its use I'm afraid.


packages/app/src/components-styled/time-series-chart/components/benchmark.tsx
```diff
@@ -1,28 +1,48 @@
...
+        <text fontSize="14px" dy={-4} fill={color}>
+          {label ? `${label}: ${value}` : value}
+        </text>
+      </g>
       <Line
         stroke={color}
         strokeDasharray="4,3"
         from={{ x: 0, y: 0 }}
-        to={{ x: width, y: 0 }}
+        to={{ x: width + xRangePadding + xRangePadding, y: 0 }}
```
VWSCoronaDashboard8, line 45:
```suggestion
        to={{ x: width + xRangePadding * 2, y: 0 }}
```

VWSCoronaDashboard10: I multiplied it previously but changed it to `+` to have it a bit more visually present what's happening.

VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/time-series-chart/components/chart-container.tsx
```diff
@@ -37,7 +37,7 @@ export function ChartContainer({
   return (
     <svg
       width={width}
-      viewBox={`0 0 ${width} ${height}`}
+      viewBox={`-0.5 -0.5 ${width} ${height}`}
```
VWSCoronaDashboard8, line 40: It's not clear to me why this is done, so please add a comment.

packages/app/src/components-styled/time-series-chart/logic/dimensions.ts
```diff
@@ -17,34 +24,65 @@ const defaultPadding: Padding = {
...
-  paddingLeft?: number
-) {
+interface DimensionProps {
+  width: number;
+  height: number;
+  paddingLeft?: number;
+  /**
+   * A symmetrical padding ensures the right padding equals the left padding
+   */
+  isSymmetricalPadding?: boolean;
```
VWSCoronaDashboard8, line 34: nit; Somehow "is" doesn't sound so nice here. Maybe `useSymmetricalPadding` or `applyPaddingSymmetrically`.

Also I wonder if this 'isSymmetricalPadding' could have been enforced in the parent context by passing in a value for `paddingRight` which equals paddingLeft. Then I think the logic in this component could be simplified. It's not important but maybe consider it if you haven't already.

VWSCoronaDashboard8 (approved):

packages/app/src/components-styled/time-series-chart/components/benchmark.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard10: correct, I'll add a comment

VWSCoronaDashboard4 (approved):

packages/app/src/components-styled/time-series-chart/components/chart-container.tsx
```diff
@@ -37,6 +37,18 @@ export function ChartContainer({
...
+       * value of 2, the line itself will be rendered _exactly_ on coordinate
+       * (0,2). Due to the 1px size of that line, it will visually be drawn from
+       * (0,1.5) to (0,2.5), with the center of the line still on (0,2).
+       * Those half pixel values can result in blurry lines on some browsers or
+       * monitors.
+       *
+       * In order to fix the above the viewbox is moved for half a pixel.
+       *
+       * inspired by: https://vecta.io/blog/guide-to-getting-sharp-and-crisp-svg-images
+       */
```
VWSCoronaDashboard1, line 51: Nice!

VWSCoronaDashboard1 (approved):

packages/app/src/components-styled/time-series-chart/logic/dimensions.ts
VWSCoronaDashboard10: I'd rather have the measured left calculation on a single location instead of in several chart root containers..

PR #2603, opened 2021-04-13
Custom validation for choropleth data

VWSCoronaDashboard1:
## Summary

Add a custom validation function that checks whether the last_value properties in GM or VR files
is equal to the choropleth value in the GM_COLLECTION or VR_COLLECTION.

## Motivation

Feature request.

### Governance

- [ ] Documentation is added


packages/cli/src/schema-information.ts
```diff
@@ -14,23 +18,35 @@ export type SchemaItemInfo = {
   optional?: boolean;
 };
 
-export function getSchemaInfo(path: string = jsonDirectory) {
-  assert(fs.existsSync(path), `Path ${path} does not exist`);
+export function getSchemaInfo(jsonPath: string = jsonDirectory) {
```
VWSCoronaDashboard8, line 21: nit; I guess you renamed it to avoid confusion with the `path` import? But this path does not point to a json file, so I'd suggest naming it something like `directoryPath`,`directory`, `jsonDirectory` (and renaming the const to defaultJsonDirectory) ...

packages/cli/src/validate-schema/custom-validations/validate-choropleth-values.ts
```diff
@@ -0,0 +1,92 @@
...
+ * The validator loops over these three properties in the collection file and for each finds the value where the 'gmcode' property
+ * is equal to 'GM0014'.
+ * It then compares the 'last_value' property of tested_overall, hospital_nice and sewer with each corresponding datapoint from the collection file.
+ *
+ */
+export const validateChoroplethValues = (
+  collectionJsonFilename: string,
+  collectionJson: Record<string, any>, // The GM_COLLECTION.sjon or VR_COLLECTION.json
+  codeProperty: string, //the gmcode or vrcode property name
+  input: Record<string, any> //GM***.json or VR***.json
```
VWSCoronaDashboard8, line 36: I think you can probably use the `UnknownObject` type (I've used it in some other cli code) here which would make it a little more strict.

packages/cli/src/validate-schema/custom-validations/validate-choropleth-values.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  return results.length
+    ? [
+        `Data in the last_value properties of ${code}.json was not equal to its counterpart in ${collectionJsonFilename}:`,
+        ...results,
+      ]
+    : undefined;
+};
+
+function haveEqualCommonProperties(
```
VWSCoronaDashboard8, line 67: nit;

```suggestion
function hasEqualCommonProperties(
```

Somehow I don't expect "have" to be used for a boolean, but I can't quite put my finger on it.

VWSCoronaDashboard8: Hmm reading the function it doesn't return a boolean but a list of validation errors?

Then I think it should be named differently all together.

packages/cli/src/validate-schema/custom-validations/validate-choropleth-values.ts
```diff
@@ -0,0 +1,92 @@
...
+  lastValue: any,
+  collectionValue: any,
```
VWSCoronaDashboard8, line 69: I think you can type these with UnknownObject or something similar

packages/cli/src/validate-schema/custom-validations/validate-choropleth-values.ts
```diff
@@ -0,0 +1,92 @@
...
+    .map((key) => {
+      return lastValue[key] !== collectionValue[key]
+        ? `property ${propertyName}.${key} is not equal: (last_value) ${lastValue[key]} !== (choropleth data) ${collectionValue[key]}`
+        : undefined;
+    })
+    .filter(isDefined);
+  return result.length ? result.join(', ') : undefined;
+}
+
+function getCommonProperties(left: any, right: any) {
```
VWSCoronaDashboard8, line 83: Looking at the code you are assuming objects here, so why not type them.

packages/cli/src/validate-schema/custom-validations/validate-choropleth-values.ts
```diff
@@ -0,0 +1,92 @@
...
+  return Object.keys(left).filter((key) => right.hasOwnProperty(key));
+}
+
+function getCommonDataProperties(left: any, right: any) {
```
VWSCoronaDashboard8, line 87: Same here

packages/cli/src/validate-schema/custom-validations/validate-choropleth-values.ts
VWSCoronaDashboard1: Type them as what exactly? Object?

VWSCoronaDashboard8: Anything is better than any :) 

`export type UnknownObject = Record<string, unknown>;`

packages/cli/src/validate-schema/custom-validations/validate-choropleth-values.ts
VWSCoronaDashboard1: This is the only instance where I need `any`, all the other instances `UnknownObject` suffices indeed

VWSCoronaDashboard8 (approved):

packages/cli/src/validate-schema/custom-validations/validate-choropleth-values.ts
```diff
@@ -0,0 +1,93 @@
...
+  const collectionJson = JSON.parse(
+    fs.readFileSync(choroplethCollectionPath, { encoding: 'utf8' })
+  );
+  return validateChoroplethValues.bind(
+    undefined,
+    path.basename(choroplethCollectionPath),
+    collectionJson,
+    codeProperty
+  ) as CustomValidationFunction;
+}
```
VWSCoronaDashboard10, line 20: why the function.bind instead of an arrow function?

VWSCoronaDashboard10 (approved):

packages/cli/src/validate-schema/custom-validations/validate-choropleth-values.ts
VWSCoronaDashboard1: You mean an arrow function as a closure?

packages/cli/src/validate-schema/custom-validations/validate-choropleth-values.ts
```diff
@@ -0,0 +1,93 @@
...
+ * It extracts all of the datapoints that both files have in common, then it compares the last_value property from each of these datapoints
+ * in the datafile with the corresponding value in the choropleth file using the codeProperty to find the correct value.
+ *
+ * For example, GM0014.json has three datapoints that are also present in the GM_COLLECTION.json: tested_overall, hospital_nice and sewer.
+ * The validator loops over these three properties in the collection file and for each finds the value where the 'gmcode' property
+ * is equal to 'GM0014'.
+ * It then compares the 'last_value' property of tested_overall, hospital_nice and sewer with each corresponding datapoint from the collection file.
+ *
+ */
+export const validateChoroplethValues = (
```
VWSCoronaDashboard4, line 33: For these methods I would be in favour of explicitly adding the return type as well; the code is a bit hard to read and this one returns either a string, undefined or the result of calling a different method. It can be safe to leave them out since Typescript is very smart, but it would be easier to read/debug.

VWSCoronaDashboard4 (approved):

PR #2609, opened 2021-04-13
Remove SCSS from the app

VWSCoronaDashboard14:
This PR removes all the SCSS from the whole application.

Some things are being solved right now.
- Removed obviously all the SCSS
- Moved the combobox general styling to the global styling, not sure if this can be removed at all
- Every general page such as: `400`, `500`, `over-risiconiveaus` is now using the same layout and structure
- Replaced all `h1`, `h2` and `p` in those pages to `Headings` and `Text` components
- We were using the `over.module.scss` to use for styling content pages, I've added a `ContentLayout` component that can be used for generic pages. 

But we also have a `cms/ContentBlock` which is kinda similar to my `ContentLayout`, so need your opinion on what to do with that. I'm not totally sold on my draft and naming yet.


VWSCoronaDashboard4 (approved):

packages/app/src/pages/404.tsx
```diff
@@ -16,14 +16,10 @@ const NotFound = (props: StaticProps<typeof getStaticProps>) => {
...
     <Layout {...siteText.notfound_metadata} lastGenerated={lastGenerated}>
-      <div className={styles.container}>
-        <MaxWidth>
-          <div className={styles.maxwidth}>
-            <h2>{siteText.notfound_titel.text}</h2>
-            <p>{siteText.notfound_beschrijving.text}</p>
-          </div>
-        </MaxWidth>
-      </div>
+      <ContentLayout>
```
VWSCoronaDashboard8, line 19: Looking at this I'm thinking this component is better named `Content` or `LayoutContent` because it is nested in a Layout component so _that_ should be about layout.

I see that in some layouts this component comes under Head, so in that sense I think Content would be a good name, otherwise the other should be LayoutHead I guess.

VWSCoronaDashboard8 (approved):
Cleans up nicely 👍

packages/app/src/domain/layout/content.tsx
```diff
@@ -24,5 +24,9 @@ const StyledBox = styled(Box)(
     p: {
       fontSize: '1.125rem',
     },
+
+    li: {
+      fontSize: '1.125rem',
```
VWSCoronaDashboard8, line 29: Seems strange to give a list (and also a paragraph) a fixed font-size. Shouldn't they just inherit it from the parent?

VWSCoronaDashboard14: That is a way better solution, didn't cross my mind

PR #2610, opened 2021-04-13
Fix react rehydration errors

VWSCoronaDashboard10:
Dates shouldn't be created during render because the server-side date can be different compared to the date created client-side (especially with a static site: build today, visit tomorrow).
The difference between dates (or any variable actually) between server and client can result in different html output between server and client. These differences can in turn result in react rehydration errors such as missing DOM-elements or DOM-elements with wrong attributes.

The CurrentDateContext will now hold the "today" date. On the server this is populated with the "last generated" date and on the client the state will be mutated after mount. The explicit state mutation will ensure components are correctly re-rendered where necessary, which means rehydration issues won't pop up anymore.

### DIFF
--- a/packages/app/src/components-styled/area-chart/hooks/use-area-configs.ts
+++ b/packages/app/src/components-styled/area-chart/hooks/use-area-configs.ts
@@ -6,6 +6,7 @@ import {
 import { useMemo } from 'react';
 import { isPresent } from 'ts-is-present';
 import { timestampToDate } from '~/components-styled/stacked-chart/logic';
+import { useCurrentDate } from '~/utils/current-date-context';
 import { TimeframeOption } from '~/utils/timeframe';
 import { createUniqueId } from '~/utils/use-unique-id';
 import { AreaDescriptor } from '../area-chart';
@@ -16,20 +17,22 @@ export function useAreaConfigs<T extends TimestampedValue>(
   areaDescriptors: AreaDescriptor<T>[],
   timeframe: TimeframeOption
 ): AreaConfig<T & TimestampedTrendValue>[] {
+  const today = useCurrentDate();
   const areaConfigs = useMemo(() => {
     return areaDescriptors.map<AreaConfig<T & TimestampedTrendValue>>(
       (descriptor) => ({
         values: getAreaData(
           descriptor.values,
           descriptor.displays.map((x) => x.metricProperty),
-          timeframe
+          timeframe,
+          today
         ),
         displays: [
           ...descriptor.displays.map((x) => ({ id: createUniqueId(), ...x })),
         ],
       })
     );
-  }, [areaDescriptors, timeframe]);
+  }, [areaDescriptors, timeframe, today]);
 
   return areaConfigs;
 }
@@ -37,9 +40,10 @@ export function useAreaConfigs<T extends TimestampedValue>(
 export function getAreaData<T extends TimestampedValue>(
   values: T[],
   metricProperties: (keyof T)[],
-  timeframe: TimeframeOption
+  timeframe: TimeframeOption,
+  today: Date
 ): (T & TimestampedTrendValue)[] {
-  const valuesInFrame = getTimeframeValues(values, timeframe);
+  const valuesInFrame = getTimeframeValues(values, timeframe, today);
 
   if (valuesInFrame.length === 0) {
     /**
--- a/packages/app/src/components-styled/area-chart/logic/index.ts
+++ b/packages/app/src/components-styled/area-chart/logic/index.ts
@@ -55,9 +55,10 @@ export function calculateYMax(values: TrendValue[], signaalwaarde = -Infinity) {
  */
 export function getTimeframeValues(
   values: TimestampedValue[],
-  timeframe: TimeframeOption
+  timeframe: TimeframeOption,
+  today: Date
 ) {
-  const boundary = getTimeframeBoundaryUnix(timeframe);
+  const boundary = getTimeframeBoundaryUnix(timeframe, today);
 
   if (isDateSeries(values)) {
     return values.filter((x) => x.date_unix >= boundary);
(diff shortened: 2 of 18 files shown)


VWSCoronaDashboard4 (approved):

VWSCoronaDashboard14 (approved):

packages/app/src/intl/hooks/use-intl.ts
```diff
@@ -196,7 +196,7 @@ export function useIntlHelperContext(locale: string, siteText: AllLanguages) {
         ? siteText.utils.date_today
         : isYesterday(date)
         ? siteText.utils.date_yesterday
-        : isDayBeforeYesterday(date)
+        : isDayBeforeYesterday(date, new Date())
```
VWSCoronaDashboard1, line 199: So, everywhere the `today` date is used, except here. A mistake, or is this correct?

VWSCoronaDashboard1 (approved):

packages/app/src/intl/hooks/use-intl.ts
VWSCoronaDashboard10: https://github.com/minvws/nl-covid19-data-dashboard/pull/2610

VWSCoronaDashboard10: I was already thinking about adding a comment to this line 😛  but it's on purpose because this function can only be called client-side and will throw an error on the server (to prevent rehydration issues)

VWSCoronaDashboard1: Makes perfect sense then ;)

VWSCoronaDashboard4: A general comment that these relative date formatters only run client side would be helpful

VWSCoronaDashboard10: It will throw an error server-side:
```tsx
if (typeof window === 'undefined') {
  throw new Error('formatRelativeDate cannot be called server-side');
}

File: packages/app/src/components-styled/area-chart/logic/index.ts [hunk 2]
```diff
@@ -72,12 +73,12 @@ export function getTimeframeValues(
...
 const oneDayInSeconds = 24 * 60 * 60;
 
-function getTimeframeBoundaryUnix(timeframe: TimeframeOption) {
+function getTimeframeBoundaryUnix(timeframe: TimeframeOption, today: Date) {
   if (timeframe === 'all') {
     return 0;
   }
   const days = getDaysForTimeframe(timeframe);
-  return Date.now() / 1000 - days * oneDayInSeconds;
+  return today.getTime() / 1000 - days * oneDayInSeconds;
```
VWSCoronaDashboard8, line 81: Since we are using timestamps as epoch in seconds throughout the app, it might have been a little cleaner to pass `today` around in seconds as well, for the sake of consistency.

packages/app/src/components-styled/time-series-chart/logic/scales.ts
```diff
@@ -82,7 +84,7 @@ export function useScales<T extends TimestampedValue>(args: {
     };
 
     return result;
-  }, [values, maximumValue, bounds, numTicks]);
+  }, [values, maximumValue, bounds, numTicks, todayDate]);
```
VWSCoronaDashboard8, line 87: ... then you wouldn't have this naming confusion or conversion here either

packages/app/src/components-styled/vertical-bar-chart/logic/scales.tsx
```diff
@@ -39,7 +41,7 @@ export function useScales<T extends TimestampedValue>(args: {
 
   return useMemo(() => {
     if (isEmpty(values)) {
-      const today = Date.now() / 1000;
+      const today = todayDate.getTime() / 1000;
```
VWSCoronaDashboard8, line 44: Or here..

packages/app/src/intl/hooks/use-intl.ts
```diff
@@ -43,8 +43,8 @@ if ('__setDefaultTimeZone' in Intl.DateTimeFormat) {
 
 // Helper functions
 
-function isDayBeforeYesterday(date: Date) {
-  return isSameDay(date, subDays(Date.now(), 2));
+function isDayBeforeYesterday(date: Date, compareDate: Date) {
```
VWSCoronaDashboard8, line 46: Since both dates are being compared, this name `compareDate` does not help for me.

Maybe you can name one "now" and the other "then"? 🤔

packages/app/src/components-styled/area-chart/hooks/use-area-configs.ts
```diff
@@ -16,30 +17,33 @@ export function useAreaConfigs<T extends TimestampedValue>(
   areaDescriptors: AreaDescriptor<T>[],
   timeframe: TimeframeOption
 ): AreaConfig<T & TimestampedTrendValue>[] {
+  const today = useCurrentDate();
```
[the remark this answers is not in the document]
VWSCoronaDashboard8: Also "now / useDateNow" might be an option, because that also lines up with the Date.now() semantics.

packages/app/src/utils/current-date-context.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export function CurrentDateProvider({
+  dateInSeconds,
+  children,
+}: {
+  dateInSeconds: number;
+  children: ReactNode;
+}) {
+  const [date, setDate] = useState(new Date(dateInSeconds * 1000));
+  useEffect(() => setDate(new Date()), []);
```
VWSCoronaDashboard8, line 24: Here you are converting the timestamp from seconds to Date. I guess this is also not needed once you choose to handle all times in seconds.

VWSCoronaDashboard8 (approved):

PR #2630, opened 2021-04-14
Build-time feature flags

VWSCoronaDashboard8:
## Summary

This PR implements build-time feature flags. Each feature can be defined with metric and route properties. 

- Adds `useFeature` hook
- Adds `validate-feature-flags` CLI script to run at build time.
- The generate-sitemap script filters out any routes that belong to disabled features. **TODO** Figure out a way to also disable the page rendering and/or their data fetching. **TODO** work with routes for VR and GM.

This snowballed a bit for several reasons:

- The file structure in packages/cli felt like it could use a makeover, because scripts and logic were not very well separated yet and importing things from each other.
- In order to import features in generate-sitemap, I had to make that file TS. In order to use that TS file I had to make the next.config file TS as well (which is not supported out of the box yet, hence the extra build/compile step)

### CLI validate-feature-flags

- Make sure that metric data is available for enabled features. It can be a metric as a whole, or specific properties of that metric.
- Make sure that metric data is _not_ available for disabled features.
- Checks data across all configured scopes (nl, vr, gm, vr-collection, gm-collection)

image
image

### API

```ts
const stockFeature = useFeature('vaccineStockPerSupplier');
```

```
 {stockFeature.isEnabled && (
            <VaccineStockPerSupplierChart values={data.vaccine_stock.values} />
          )}
```

```ts
export interface Feature {
  name: string;
  isEnabled: boolean;

  /**
   * If it is not clear what this feature does from the name, you can leave some
   * text here.
   */
  description?: string;

  /**
   * Metric scope defines the files in which we enforce the (non-)existence of
   * metricNames.
   */
  metricScopes?: MetricScope[];

  /**
   * A metricName is the root-level schema property used to hold the data in the
   * NL/VR/GM files. By limiting a feature to only 1 metric name, we can keep
   * the code and API simple. If your feature is using more than one metric
   * name, simply split it into multiple named features.
   */
  metricName?: string;

  /**
   * If the feature was built on new metric properties of an existing metric
   * name then you can use this field to enforce the (non)existence of only
   * those properties. The metric name will still be allowed to exist when this
   * feature is disabled.
   */
  metricProperties?: string[];

  /**
   * The route to filter from the sitemap when this feature is disabled. I don't
   * think we'll need to have more than one route per feature as this is quite
   * rare. And you can always choose to define multiple features if you need to
   * cover multiple routes.
   */
  route?: string;
}


export type MetricScope =
  | 'nl'
  | 'vr'
  | 'gm'
  | 'vr_collection'
  | 'gm_collection';
```

```ts
export const features: Feature[] = [
  {
    name: 'downscaling',
    isEnabled: false,
    route: '/afschaling',
  },
  {
    name: 'vaccineStockPerSupplier',
    isEnabled: false,
    metricScopes: ['nl'],
    metricName: 'vaccine_stock',
    metricProperties: ['astra_zeneca_total', 'astra_zeneca_available'],
  },
];
```


packages/cli/src/scripts/validate-features.ts
```diff
@@ -0,0 +1,244 @@
...
+
+    assert(
+      feature.metricName,
+      'If a feature defines metricProperties it should also have a metricName'
+    );
+
+    const promisedResults = feature.metricScopes.map((scope) =>
+      validateMetricPropertiesForScope(
+        feature.metricName!,
+        feature.metricProperties!,
```
VWSCoronaDashboard10, line 131: I think you can get rid of the `!` when yuo assign the properties to variables
```suggestion
    const { metricScopes, metricName } = feature;
    
    assert(
      metricScopes,
      `Missing metricScopes configuration for feature ${feature.name}`
    );

    assert(
      metricName,
      'If a feature defines metricProperties it should also have a metricName'
    );

    const promisedResults = feature.metricScopes.map((scope) =>
      validateMetricPropertiesForScope(
        metricName,
        metricProperties,
```

packages/cli/src/scripts/validate-features.ts
```diff
@@ -0,0 +1,244 @@
...
+      return messages;
+    }
+  } else if (feature.metricName) {
+    assert(
+      feature.metricScopes,
+      `Missing metricScopes configuration for feature ${feature.name}`
+    );
+    const promisedResults = feature.metricScopes.map((scope) =>
+      validateMetricNameForScope(
+        feature.metricName!,
```
VWSCoronaDashboard10, line 155: same here

packages/cli/src/scripts/validate-features.ts
VWSCoronaDashboard8: Yes, that works 👍

VWSCoronaDashboard4 (approved):
Looking good! Only a nitpick commented.

PR #2635, opened 2021-04-15
Update Tile components to be a functional component

VWSCoronaDashboard14:
This started with the bug on the background of the article strip, where I had to refactor some small things to not rely on CSS set directly on the component.


packages/app/src/components/tile.tsx
```diff
@@ -1,15 +1,38 @@
-import { Box } from './base';
 import { css } from '@styled-system/css';
 import styled from 'styled-components';
 import { asResponsiveArray } from '~/style/utils';
 
-export const Tile = styled(Box).attrs({ as: 'article' })(
+interface TileProps {
+  children: React.ReactNode;
+  hasBlueBackground?: boolean;
+  isContentHeight?: boolean;
```
VWSCoronaDashboard1, line 8: I guess `hasContentHeight` is more correct?

VWSCoronaDashboard1 (changes requested):

packages/app/src/components/tile.tsx
VWSCoronaDashboard14: Why didn't I think of that ❓

packages/app/src/components/kpi-tile.tsx
```diff
@@ -21,7 +21,7 @@ export function KpiTile({
   metadata,
 }: KpiTileProps) {
   return (
-    <Tile height="100%">
+    <Tile hasContentHeight>
```
VWSCoronaDashboard10, line 24: I don't think `hasContentHeight` means `height='100%`. I'd rather see it named something like `isFullHeight` or, maybe better, even just keep the `height` props as is and pass it on to the styled component:
```diff
<StyledTile
      hasBlueBackground={hasBlueBackground}
-      hasContentHeight={hasContentHeight}
+      height={height}
    >
```

And to be honest: I think the height should actually be mutated by styles of the parent component of a tile but that's maybe too much refactoring for this PR.

packages/app/src/components/tile.tsx
```diff
@@ -1,15 +1,38 @@
-import { Box } from './base';
 import { css } from '@styled-system/css';
 import styled from 'styled-components';
 import { asResponsiveArray } from '~/style/utils';
 
-export const Tile = styled(Box).attrs({ as: 'article' })(
+interface TileProps {
+  children: React.ReactNode;
+  hasBlueBackground?: boolean;
```
VWSCoronaDashboard8, line 7: We don't actually use the normal boolean naming pattern for props, because then they align with html elements better (a decision I possibly regret) but in that sense this prop could be `blueBackground`

packages/app/src/components/anchor-tile.tsx
```diff
@@ -51,10 +51,14 @@ export function AnchorTile({
   );
 }
 
-const Container = styled(Tile)<{ shadow: boolean }>((x) =>
+const Container = styled.article<{ shadow: boolean }>((x) =>
   css({
+    // spacing: 2,
```
VWSCoronaDashboard10, line 56: comment

packages/app/src/components/tile.tsx
```diff
@@ -1,15 +1,38 @@
...
+  hasContentHeight?: boolean;
```
VWSCoronaDashboard8, line 8: I think we can find a better name for this. Content always has a height to it, and so I don't know what this means when I read it.

packages/app/src/components/tile.tsx
```diff
@@ -1,15 +1,38 @@
...
+}>((x) =>
   css({
     display: 'flex',
     flexDirection: 'column',
-    bg: 'white',
+    bg: x.hasBlueBackground ? 'lightBlue' : 'white',
     p: asResponsiveArray({ _: 3, sm: 4 }),
     borderRadius: 1,
     boxShadow: 'tile',
+    height: x.hasContentHeight ? '100%' : undefined,
```
VWSCoronaDashboard8, line 36: Maybe `isFullHeight` or `isParentHeight`? The is part you can drop see earlier comment.

packages/app/src/components/anchor-tile.tsx
```diff
@@ -79,23 +83,23 @@ const IconContainer = styled.span(
...
-    pt: [3, 3, 3, 3, null],
-    pl: [null, null, null, null, 4],
-    ml: [null, null, null, null, 4],
+    borderTop: asResponsiveArray({
+      _: `1px solid ${colors.silver}`,
+      lg: 'none',
+    }),
+    borderLeft: asResponsiveArray({ lg: '1px solid' }),
+    borderLeftColor: asResponsiveArray({ lg: colors.silver }),
+    pt: asResponsiveArray({ _: 3, lg: undefined }),
```
VWSCoronaDashboard10, line 100:
```suggestion
    pt: asResponsiveArray({ _: 3, lg: 0 }),
```

packages/app/src/components/anchor-tile.tsx
```diff
@@ -79,23 +83,23 @@ const IconContainer = styled.span(
...
+    pl: asResponsiveArray({ lg: 4 }),
+    ml: asResponsiveArray({ lg: 4 }),
+    mt: asResponsiveArray({ _: 3, md: undefined }),
```
VWSCoronaDashboard10, line 103:
```suggestion
    mt: asResponsiveArray({ _: 3, md: 0 }),
```

VWSCoronaDashboard8 (approved):

packages/app/src/components/tile.tsx
```diff
@@ -1,15 +1,38 @@
...
-export const Tile = styled(Box).attrs({ as: 'article' })(
+interface TileProps {
+  children: React.ReactNode;
+  hasBlueBackground?: boolean;
+  hasContentHeight?: boolean;
+}
+
+export function Tile({
+  children,
+  hasBlueBackground,
```
VWSCoronaDashboard10, line 13: nit; Instead of the `hasBlueBackground` prop we could also think about exporting a `<BlueTile />` component. Do whatever feels best.

VWSCoronaDashboard10 (approved):
Approved, though I'd like to see the `hasContentHeight` prop being renamed.

VWSCoronaDashboard1 (approved):

packages/app/src/components/tile.tsx
VWSCoronaDashboard10: ehhhm, that's a rule we came up with in another project, but AFAIK we do use the `is` pattern for booleans currently? Or am I mistaken?

packages/app/src/components/tile.tsx
```diff
@@ -4,35 +4,32 @@ import { asResponsiveArray } from '~/style/utils';
...
   css({
     display: 'flex',
     flexDirection: 'column',
-    bg: x.hasBlueBackground ? 'lightBlue' : 'white',
+    bg: x.backgroundColor ? x.backgroundColor : 'white',
     p: asResponsiveArray({ _: 3, sm: 4 }),
     borderRadius: 1,
     boxShadow: 'tile',
-    height: x.hasContentHeight ? '100%' : undefined,
+    height: x.height ? x.height : undefined,
```
VWSCoronaDashboard10, line 33: fallback to undefined is unnecessary and would even fail when it equals `0` (`height={0}`)
```suggestion
    height: x.height,
```

VWSCoronaDashboard10 (changes requested):

VWSCoronaDashboard10 (approved):

PR #2640, opened 2021-04-16
Escalation level schema

VWSCoronaDashboard1:
## Summary

Add a top-level schema to National that contains the thresholds for the positive tests and hospital admissions. (as seen on this regional page: https://coronadashboard.rijksoverheid.nl/veiligheidsregio/VR02/risiconiveau)

Because these levels are the same for all regions I added them to the national schema, otherwise, they'd get duplicated.
This does mean, however, that the regional pages need to import some of the national data. So, to prevent a regional page from having to include the entire national data file, I added a few functions to `get-data.ts`: `selectNlData`, `selectVrData`, and `selectGmData`. These functions return a subset of the respective data files.

It can be considered to use this pattern for all pages and only select the exact data that is actually used on the page, which would yield some gains in terms of HTML size. It does add a bit of maintenance cost though, so it should perhaps be determined on a per-case basis.
(I have a done a quick spike, and for example, the regional 'risiconiveau' page is over 200Kb smaller when explicitly selecting only the needed data from the VR**.json file)

This ticket is on hold for now. The backend team will have to confirm that the proposed data shape is feasible for them.

### Governance

- [ ] Documentation is added


packages/app/src/domain/escalation-level/thresholds.ts
```diff
@@ -1,64 +1,40 @@
...
-      name: levels['4'].titel,
-      threshold: 250,
-      color: colors.data.scale.magenta[3],
-    },
-    {
+  return useMemo(() => {
+    const escalationThresholds = {
+      positiveTestedEscalationThresholds: thresholds.tested_overall_sum.infected_per_100k.map<CategoricalBarScaleCategory>(
+        (x, index) => ({
+          name: (levels as any)[x.escalation_level.toString()].titel,
```
VWSCoronaDashboard8, line 15: You are assuming here that level[x] has a title, so I wonder why you cast it to any. Seems like you can be specific about it an cast it to something like `Record<string, { title: string}>`.

packages/common/src/types/data.ts
```diff
@@ -724,6 +725,19 @@ export interface NlVaccineStockValue {
...
   date_unix: number;
 }
+export interface NlEscalationThresholds {
+  tested_overall_sum: {
+    infected_per_100k: NlEscalationThresholdValue[];
+  };
+  hospital_nice_sum: {
+    admissions_per_1m: NlEscalationThresholdValue[];
+  };
+  date_of_insertion_unix?: number;
```
VWSCoronaDashboard8, line 735: Looks like you forgot to add this one to the required list

VWSCoronaDashboard8 (approved):

packages/app/src/domain/escalation-level/thresholds.ts
```diff
@@ -12,14 +12,18 @@ export function useEscalationThresholds(thresholds: NlEscalationThresholds) {
     const escalationThresholds = {
       positiveTestedEscalationThresholds: thresholds.tested_overall_sum.infected_per_100k.map<CategoricalBarScaleCategory>(
         (x, index) => ({
-          name: (levels as any)[x.escalation_level.toString()].titel,
+          name: (levels as Record<string, { titel: string }>)[
```
VWSCoronaDashboard8, line 15: nit; You could have written `const levels = siteText.escalatie_niveau.types as Record<string, { titel: string }>` then you don't have to repeat yourself and the code is a bit more readable.

PR #2643, opened 2021-04-16
Update intl hook to accept en-gb formatting

VWSCoronaDashboard14:
Update the intl hook to change the English language to use an `en-GB` formatting for the dates.


packages/app/src/pages/_app.tsx
```diff
@@ -31,8 +31,8 @@ export default function App(props: AppProps) {
   const { Component, pageProps } = props;
 
   // const { locale = 'nl' } = useRouter(); // if we replace this with process.env.NEXT_PUBLIC_LOCALE, next export should still be possible?
-  const locale = process.env.NEXT_PUBLIC_LOCALE || 'nl';
-  const text = languages[locale as LanguageKey];
+  const locale = (process.env.NEXT_PUBLIC_LOCALE || 'nl') as LanguageKey;
+  const text = languages[locale];
```
VWSCoronaDashboard10, line 35: In hindsight it's better to do an assertion over here instead of inside the `useIntlHelperContext`.

 ```suggestion
  const locale = (process.env.NEXT_PUBLIC_LOCALE || 'nl') as LanguageKey;
  assert(['en', 'nl'].includes(locale), `Encountered unknown language: ${locale}`);
  
  const text = languages[locale];
```

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard10 (approved):

PR #2656, opened 2021-04-19
Add schema for hospital admissions by age group

VWSCoronaDashboard14:
- Took the `tested_per_age_group.json` as a reference to transform it to the `hospital_nice_per_age_group` schema.
- Combined the 0-9 and the 10-19 into one category
- Made it not required in the `__index.json` file

### DIFF
--- a/packages/app/schema/nl/hospital_nice_per_age_group.json
+++ b/packages/app/schema/nl/hospital_nice_per_age_group.json
@@ -21,48 +21,48 @@
       "additionalProperties": false,
       "type": "object",
       "required": [
-        "admissions_age_0_19_per_100k",
-        "admissions_age_20_29_per_100k",
-        "admissions_age_30_39_per_100k",
-        "admissions_age_40_49_per_100k",
-        "admissions_age_50_59_per_100k",
-        "admissions_age_60_69_per_100k",
-        "admissions_age_70_79_per_100k",
-        "admissions_age_80_89_per_100k",
-        "admissions_age_90_plus_per_100k",
-        "admissions_overall_per_100k",
+        "admissions_age_0_19_per_million",
+        "admissions_age_20_29_per_million",
+        "admissions_age_30_39_per_million",
+        "admissions_age_40_49_per_million",
+        "admissions_age_50_59_per_million",
+        "admissions_age_60_69_per_million",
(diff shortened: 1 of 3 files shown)


VWSCoronaDashboard1 (approved):

VWSCoronaDashboard8 (approved):

packages/app/schema/nl/hospital_nice_per_age_group.json
```diff
@@ -0,0 +1,78 @@
...
+  },
+  "required": ["values", "last_value"],
+  "definitions": {
+    "value": {
+      "title": "nl_hospital_nice_per_age_group_value",
+      "additionalProperties": false,
+      "type": "object",
+      "required": [
+        
+        "admissions_age_0_19_per_100k",
```
VWSCoronaDashboard4, line 25: I think this one should also be per 1 million?

packages/app/schema/nl/hospital_nice_per_age_group.json
```diff
@@ -0,0 +1,78 @@
...
+    }
+  },
+  "required": ["values", "last_value"],
+  "definitions": {
+    "value": {
+      "title": "nl_hospital_nice_per_age_group_value",
+      "additionalProperties": false,
+      "type": "object",
+      "required": [
+
```
VWSCoronaDashboard4, line 24: what's up with this random newline?

VWSCoronaDashboard14: Good question, that wasn't intended!

packages/app/schema/nl/hospital_nice_per_age_group.json
VWSCoronaDashboard14: I've set out the question since it isn't quite clear in my opinion. So indeed it could be changed later but I want to be sure.

VWSCoronaDashboard8: If you look at the current charts the numbers for IC and hospital admissions seem to be absolute. So no 100k or per million I think.

VWSCoronaDashboard14: Seems like it is per million.

VWSCoronaDashboard8 (approved):

PR #2657, opened 2021-04-19
Add schema for ic admissions nice per age group

VWSCoronaDashboard14:
- Took the `tested_per_age_group.json` as a reference to transform it to the `intensive_care_nice_per_age_group` schema
- Combined the 0-9 and the 10-19 into one category
- Made it not required in the `__index.json` file

### DIFF
--- a/packages/app/schema/nl/intensive_care_nice_per_age_group.json
+++ b/packages/app/schema/nl/intensive_care_nice_per_age_group.json
@@ -1,0 +1,78 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "nl_intensive_care_nice_per_age_group",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "values": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/value"
+      }
+    },
+    "last_value": {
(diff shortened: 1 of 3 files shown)


packages/app/schema/nl/intensive_care_nice_per_age_group.json
```diff
@@ -0,0 +1,78 @@
...
+  },
+  "required": ["values", "last_value"],
+  "definitions": {
+    "value": {
+      "title": "nl_intensive_care_nice_per_age_group_value",
+      "additionalProperties": false,
+      "type": "object",
+      "required": [
+        
+        "admissions_age_0_19_per_million",
```
VWSCoronaDashboard4, line 25: hmm, in the codebase there already is a `_per_million` suffix, but also a `_per_1m` suffix... which one should we use?

VWSCoronaDashboard4 (approved):

packages/app/schema/nl/intensive_care_nice_per_age_group.json
VWSCoronaDashboard14: My personal preference is `_per_million` but I'm not against `_per_1m` either.

PR #2660, opened 2021-04-19
Link mobile users to sidebar instead of vaccination page

VWSCoronaDashboard13:
## Summary

Link mobile users to sidebar instead of the vaccination page from the link in the header on the actual page

## Motivation

Confused users

## Detailed design

Added nl routes to `user-reverse-router` with the index routing to the vaccination page. Used this link in the `top-navigation` and used the `nl.index` route provided by the reverseRouter to overwrite the existing href.

## Drawbacks

That href will need to be programmatically changed if they want to redirect it to a different page


packages/app/src/utils/use-reverse-router.ts
```diff
@@ -7,34 +7,62 @@ export function useReverseRouter() {
 
   return useMemo(() => {
     // prettier-ignore
```
VWSCoronaDashboard10, line 9: looks like your editor ignored this prettier-ignore comment. Might as well just delete this line because it doesn't work everywhere :-)

VWSCoronaDashboard10 (approved):

packages/app/src/utils/use-reverse-router.ts
VWSCoronaDashboard8, line 9: Looks like prettier formatted this file after all? If that's true I think it's fine, but then I suggest to also remove this comment.

VWSCoronaDashboard8 (approved):

packages/app/src/utils/use-reverse-router.ts
VWSCoronaDashboard13: is there a reason why you wanted this to be formatted this way?

packages/app/src/utils/use-reverse-router.ts
```diff
@@ -7,34 +7,62 @@ export function useReverseRouter() {
...
+    const reverseRouter = {
       actueel: {
         vr: (code: string) => `/actueel/veiligheidsregio/${code}`,
-        gm: (code: string) => `/actueel/gemeente/${code}`
+        gm: (code: string) => `/actueel/gemeente/${code}`,
+      },
+
+      nl: {
+        index: () => reverseRouter.nl.vaccinaties() + openMenuSuffix,
+        vaccinaties: () => `/landelijk/vaccinaties`,
```
VWSCoronaDashboard4, line 17: nit: perhaps these could be strings, instead of functions that return strings? not sure if it will work with wherever this is used.

VWSCoronaDashboard4 (approved):

VWSCoronaDashboard1 (approved):

packages/app/src/utils/use-reverse-router.ts
VWSCoronaDashboard13: good catch! works fine as a string 👍

packages/app/src/utils/use-reverse-router.ts
VWSCoronaDashboard13: although I think for the sake of consistency it would actually make sense to keep in as a function

packages/app/src/utils/use-reverse-router.ts
VWSCoronaDashboard13: although, for the sake of consistency I think it would actually make more sense for them to all be functions

packages/app/src/utils/use-reverse-router.ts
VWSCoronaDashboard10: the `// prettier-ignore` disables the formatting for the complete code block, not only the first function. The comment was there to make that block a bit easier for the eyes to read. But feel free to remove it since 't doesn't work on your (and probably other dev's) environment.

packages/app/src/utils/use-reverse-router.ts
VWSCoronaDashboard10: Yes I'd rather have them as functions to have the API predictable.

packages/app/src/utils/use-reverse-router.ts
VWSCoronaDashboard13: sorry wrote that response on the wrong comment haha. Makes sense 👍

PR #2661, opened 2021-04-19
Remove sewer bar chart from VR and G

VWSCoronaDashboard13:
## Summary

Remove sewer bar chart from VR and G 

## Motivation

Too confusing and data is available elsewhere

## Detailed design

- Removed the chart tiles from the sewer pages
- Removed `municipality-sewer-water.util.ts` and `safety-region-sewer-water.util.ts` since they weren't being used anymore
- Removed `useSelectedStationValues` from `/sewer-chart/logic.ts` since it also wasn't being used

Did not remove any schemas because the schema being used for the bar chart, `sewer_per_installation`, appears to be used for the sewer scatter plot


VWSCoronaDashboard4 (approved):
🔥

VWSCoronaDashboard14 (approved):

VWSCoronaDashboard8 (approved):

PR #2663, opened 2021-04-19
Prevent exporting draft translations on production builds

VWSCoronaDashboard10:
## Summary

Prevent exporting lokalize draft translations during a production build.

The env var I currently use to detect the production environment (process`.env.NEXT_PUBLIC_VERCEL_ENV`) probably isn't set during a production build. Todo: figure out how to detect production build properly..

### DIFF
--- a/packages/cms/src/lokalize/export.ts
+++ b/packages/cms/src/lokalize/export.ts
@@ -18,7 +18,14 @@ client
     const nl: Record<string, string> = {};
     const en: Record<string, string> = {};
 
-    for (const document of result) {
+    /**
+     * Both drafts as published documents are part of the query-result because
+     * the client is authenticated with a sanity token.
+     */
+    const drafts = result.filter((x) => x._id.startsWith('drafts.'));
+    const published = result.filter((x) => !x._id.startsWith('drafts.'));
+
+    for (const document of published) {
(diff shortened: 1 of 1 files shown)


VWSCoronaDashboard7:
How about the other way around? Never export unless its develop? Then you can set an environmental to your choosing for preview/develop environments on Vercel.

VWSCoronaDashboard1 (approved):

packages/cms/src/lokalize/export.ts
```diff
@@ -35,6 +42,28 @@ client
...
+     */
+    if (process.env.NEXT_PUBLIC_VERCEL_ENV !== 'production') {
+      for (const document of drafts) {
+        const key = `${document.subject}.${document.path}`;
+        nl[key] = document.text.nl.trim();
+        en[key] = document.text.en?.trim();
+
+        if (!en[key]) {
+          /**
+           * Here we could make an automatic fallback to Dutch texts if English is missing.
```
VWSCoronaDashboard8, line 56: I think I used "could" when I wasn't sure it was what we wanted. But since it's being applied let's remove the word to avoid confusion :)

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard10:
closed in favor of #2687

PR #2669, opened 2021-04-20
Add explicit data selection to all metric pages

VWSCoronaDashboard1:
## Summary

This PR adds functionality that allows each metric page to explicitly select only the properties from a data file (nl.json, vrXX.json, gmXX.json). This prevents the entire data file to be included in the HTML, which yields quite a significant size reduction.

For example, these are some VR pages before:
regio_html_build_before
and this is after:
regio_html_build_after

And here are the regional actueel pages before:
actueel_html_build_before
and after:
actueel_html_build_after

## Motivation

Optimization.

## Detailed design

Instead of, for example, using the `getNlData` method, a new method called `selectNlPageMetricData` is introduced. This method returns all of the default National data that is required on each national metric page. Because the sidebar needs all of those properties. This method accepts an optional list of additional metric names which will be added to the output. This way a page can select exactly what it needs and no more.

So, the most significant changes are in the `get-data.ts` file, the rest of the changes is all based on implementing the usage of those new methods.

## Drawbacks

It adds a bit of verbosity. Before we just pushed the entire data file (National, Regionaal or Municipal) in some components, but now things have to be defined a little bit more explicitly. The size improvements are worth it though, IMHO.

### Governance

- [ ] Documentation is added


packages/app/src/pages/veiligheidsregio/[code]/verpleeghuiszorg.tsx
```diff
@@ -25,11 +28,11 @@ export { getStaticPaths } from '~/static-paths/vr';
 
 export const getStaticProps = createGetStaticProps(
   getLastGeneratedDate,
-  getVrData
+  selectVrPageMetricData()
```
Comment (VWSCoronaDashboard8) [inline packages/app/src/pages/veiligheidsregio/[code]/verpleeghuiszorg.tsx:31]: No metrics on this page?

packages/app/src/static-props/get-data.ts
```diff
@@ -55,6 +67,37 @@ export function createGetContent<T>(
   };
 }
 
+/**
+ * This method returns all the national data that is required by the sidebar,
+ * optional extra metric property names can be added as separate arguments which will
```
VWSCoronaDashboard8, line 72: Right, now I get the empty function calls

packages/app/src/static-props/get-data.ts
```diff
@@ -55,6 +67,37 @@ export function createGetContent<T>(
...
+ * be added to the output
+ *
+ */
+export function selectNlPageMetricData<
+  T extends keyof National = NationalPageMetricNames
+>(...additionalMetrics: T[]) {
+  return selectNlData(...[...nationalPageMetricNames, ...additionalMetrics]);
```
VWSCoronaDashboard8, line 79: nit; Would this work too?

```suggestion
  return selectNlData(...nationalPageMetricNames, ...additionalMetrics);
```

packages/app/src/static-props/get-data.ts
```diff
@@ -55,6 +67,37 @@ export function createGetContent<T>(
...
+}
+
+/**
+ * This method selects only the specified metric properties from the national data
+ *
+ */
+export function selectNlData<T extends keyof National = never>(
```
VWSCoronaDashboard8, line 86: What is the function of never here?

packages/app/src/static-props/get-data.ts
```diff
@@ -55,6 +67,37 @@ export function createGetContent<T>(
...
+  ...metrics: T[]
+) {
+  return () => {
+    const { data } = getNlData();
+
+    const selectedNlData = {} as Pick<National, T>;
```
VWSCoronaDashboard8, line 92: nit; It's a little strange to declare the type like this when you still need to construct its contents. I would find it more logical to cast the return value, when it's ready.

VWSCoronaDashboard8 (commented):
Nice to have these things typed and filtered 💯

VWSCoronaDashboard8 (approved):

packages/app/src/components/data-driven-text.tsx
```diff
@@ -7,31 +7,38 @@ import {
...
-interface DataDrivenTextProps {
-  data: National | Regionaal | Municipal;
+type DataKeys = keyof National | keyof Regionaal | keyof Municipal;
+type DataFile<T> = T extends keyof National
+  ? Pick<National, 'difference' | T>
+  : T extends keyof Regionaal
+  ? Pick<Regionaal, 'difference' | T>
+  : T extends keyof Municipal
+  ? Pick<Municipal, 'difference' | T>
+  : never;
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: I'll add a comment in the file. In short, this typing makes sure that assign a metricName of `keyof National` that the data property needs to be of type `Pick<National, 'difference' | T>`. If the metricName is `keyof Regionaal`, then the data property needs to be of type `Pick<Regionaal, 'difference' | T>`, etc

packages/app/src/static-props/get-data.ts
VWSCoronaDashboard1: For some reason it doesn't, the compiler complains about it.

packages/app/src/static-props/get-data.ts
VWSCoronaDashboard1: This forces the selectNlData method to be called with at least 1 rest parameter

packages/app/src/static-props/get-data.ts
VWSCoronaDashboard1: Fair enough

packages/app/src/static-props/get-data.ts
VWSCoronaDashboard1: Since T is inferred from the rest parameters. If you default the generic to never, it will infer to never if you don't supply any arguments.

packages/app/src/domain/layout/municipality-layout.tsx
```diff
@@ -29,9 +29,9 @@ export const municipalPageMetricNames = [
   'difference',
 ] as const;
 
-export type MunicipalPageMetricNames = typeof municipalPageMetricNames[number];
+export type GMPageMetricNames = typeof gmPageMetricNames[number];
```
VWSCoronaDashboard8, line 32: Please write acronyms as normal words. We do that in other places too

```suggestion
export type GmPageMetricNames = typeof gmPageMetricNames[number];
```

packages/app/src/domain/layout/national-layout.tsx
```diff
@@ -42,9 +42,9 @@ export const nationalPageMetricNames = [
   'difference',
 ] as const;
 
-export type NationalPageMetricNames = typeof nationalPageMetricNames[number];
+export type NLPageMetricNames = typeof nlPageMetricNames[number];
```
VWSCoronaDashboard8, line 45: Same comment

packages/app/src/domain/layout/safety-region-layout.tsx
```diff
@@ -38,11 +38,11 @@ export const safetyRegionPageMetricNames = [
   'difference',
 ] as const;
 
-export type SafetyRegionPageMetricNames = typeof safetyRegionPageMetricNames[number];
+export type VRRegionPageMetricNames = typeof vrPageMetricNames[number];
```
VWSCoronaDashboard8, line 41: You get my point :)

packages/app/src/static-props/get-data.ts
```diff
@@ -131,12 +129,15 @@ export function selectVrData<T extends keyof Regionaal = never>(
   return (context: GetStaticPropsContext) => {
     const vrData = getVrData(context);
 
-    const selectedVrData = {} as Pick<Regionaal, T>;
+    const selectedVrData: Record<string, any> = {};
```
VWSCoronaDashboard8, line 132: Ah bummer I didn't realize you then also had to type this still. You can probably use string, unknown :)

VWSCoronaDashboard8 (changes requested):

packages/app/src/static-props/get-data.ts
VWSCoronaDashboard8: You could avoid using a temporary variable by using reduce. This is using lodash set:

```suggestion
    const selectedVrData: Pick<Regionaal, T> = metrics.reduce((acc, p) => set(acc, p, vrData.data[p]), {})
```

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard4 (approved):
:100:

packages/app/src/static-props/get-data.ts
```diff
@@ -64,10 +108,43 @@ export function getNlData() {
...
+/**
+ * This method returns all the region data that is required by the sidebar,
+ * optional extra metric property names can be added as separate arguments which will
+ * be added to the output
+ *
+ */
+export function selectVrPageMetricData<
+  T extends keyof Regionaal = VrRegionPageMetricNames
+>(...additionalMetrics: T[]) {
+  return selectVrData(...[...vrPageMetricNames, ...additionalMetrics]);
```
VWSCoronaDashboard10, line 120:
```suggestion
  return selectVrData(...vrPageMetricNames, ...additionalMetrics);
```

packages/app/src/static-props/get-data.ts
```diff
@@ -81,10 +158,43 @@ export function getVrData(context: GetStaticPropsContext) {
...
+/**
+ * This method returns all the municipal data that is required by the sidebar,
+ * optional extra metric property names can be added as separate arguments which will
+ * be added to the output
+ *
+ */
+export function selectGmPageMetricData<
+  T extends keyof Municipal = GmPageMetricNames
+>(...additionalMetrics: T[]) {
+  return selectGmData(...[...gmPageMetricNames, ...additionalMetrics]);
```
VWSCoronaDashboard10, line 170:
```suggestion
  return selectGmData(...gmPageMetricNames, ...additionalMetrics);
```

VWSCoronaDashboard10 (approved):

packages/app/src/static-props/get-data.ts
VWSCoronaDashboard1: Alas, the compiler doesn't agree with this. I don't understand why, but it has something to do with the fact these are rest parameters...

packages/app/src/static-props/get-data.ts
VWSCoronaDashboard1: same as my previous comment

PR #2672, opened 2021-04-20
Hospital Admissions per age group charts

VWSCoronaDashboard13:
## Summary

Adding overall and intensive care admission per age group charts

## Detailed design

Created a `hospital` domain that houses the `admissions-per-age-group` chart (almost a duplicate of the positive tested version) which accepts `NlIntensiveCareNicePerAgeGroupValue` and `NlHospitalNicePerAgeGroupValue` allowing us to use this in both places since both types have the same structure. 

Also added the copy in the following structure
``` 
"hospital_admissions_per_age_group": {
    "overall_chart_title": "...",
    "overall_chart_description": "...",
    "ic_chart_title": "...",
    "ic_chart_description": "...",
    "legend": {
      "admissions_age_0_19_per_million": "0-19",
      "admissions_age_20_29_per_million": "20-29",
      "admissions_age_30_39_per_million": "30-39",
      "admissions_age_40_49_per_million": "40-49",
      "admissions_age_50_59_per_million": "50-59",
      "admissions_age_60_69_per_million": "60-69",
      "admissions_age_70_79_per_million": "70-79",
      "admissions_age_80_89_per_million": "80-89",
      "admissions_age_90_plus_per_million": "90+",
      "admissions_overall_per_million": "Alle leeftijden"
    },
    "legend_help_text": "...",
    "line_chart_legend_inaccurate_label": "...",
    "tooltip_labels": {
      "inaccurate": "..."
    }
  },
```

Dummy data isn't great, but it's better than previous iterations 😬

image

## Alternatives

Make separate charts for ic and overall - I think this isn't necessary until there are notable differences between the two


VWSCoronaDashboard1 (approved):

VWSCoronaDashboard10 (approved):

packages/app/src/locale/en.json
```diff
@@ -2949,6 +2935,29 @@
       "inaccurate": "Incomplete"
     }
   },
+  "hospital_admissions_per_age_group": {
+    "overall_chart_title": "Ziekenhuisopnames per leeftijdsgroep door de tijd heen",
+    "overall_chart_description": "Deze grafiek laat zien hoeveel mensen met het coronavirus in een bepaalde leeftijdsgroep in het ziekenhuis zijn opgenomen. In de grafiek worden weektotalen getoond die zijn omgerekend naar 1.000.000 personen in de leeftijdsgroep.",
+    "ic_chart_title": "Intensive care-opnames per leeftijdsgroep door de tijd heen",
+    "ic_chart_description": "Deze grafiek laat zien hoeveel mensen met het coronavirus in een bepaalde leeftijdsgroep in het ziekenhuis zijn opgenomen. In de grafiek worden weektotalen getoond die zijn omgerekend naar 1.000.000 personen in de leeftijdsgroep.",
```
VWSCoronaDashboard8, line 2942: "hospital" and "ic" are two separate data subjects within our application, and they do not overlap even though ic can be seen as part of a hospital of course. 

So it seems there's a little mixup / confusion here with the keys.

packages/app/src/domain/hospital/admissions-per-age-group/admissions-per-age-group.tsx
```diff
@@ -0,0 +1,183 @@
...
+        helpText={text.legend_help_text}
+        selectOptions={interactiveLegendOptions}
+        selection={list}
+        onToggleItem={toggle}
+        onReset={clear}
+      />
+      <Legend items={staticLegendItems} />
+    </>
+  );
+}
```
VWSCoronaDashboard8, line 183: This chart contains a lot of logic related to config and legends. The interactive legend was an afterthought, so I think we will have to take a closer look at improving abstractions to related components, because this feels too much like a workaround for something that could be solved elsewhere.

Entirely not your fault of course. I think you did a great job here. I'll add this to the tech-dept backlog.

VWSCoronaDashboard8 (changes requested):

packages/app/src/locale/en.json
VWSCoronaDashboard13: When you say `data subjects` are you referring to the different tabs in the side bar and how does that inform how it is structured in the locale files? I just put them together because the graph is shared between the two and they are related in that they are grouped together in the sidebar

Would something like this be better?
```
"hospital_admissions_per_age_group_chart": {
    chart text...
},
"hospital_admissions_per_age_group": {
   "chart_title": "...",
   "chart_description": "...",
},
"ic_admissions_per_age_group": {
   "chart_title": "...",
   "chart_description": "...",
},
```

VWSCoronaDashboard8: Yes that looks good to me 👍  With subjects I mean the first name by which we group data/properties on the root level in the schema files. So for NL.json for example you have different data sets starting with hospital_ and ic_ and sewer_ etc. In that sense hospital and ic are considered two different / independent sources of data.

VWSCoronaDashboard13: makes sense 👍

VWSCoronaDashboard8 (approved):

PR #2677, opened 2021-04-20
Add moving average to deceased charts

VWSCoronaDashboard13:
## Summary

Add moving average to deceased charts + copy keys


VWSCoronaDashboard4 (approved):

VWSCoronaDashboard1 (approved):

packages/app/src/pages/gemeente/[code]/sterfte.tsx
```diff
@@ -149,6 +149,17 @@ const DeceasedMunicipalPage = (props: StaticProps<typeof getStaticProps>) => {
...
+                    type: 'line',
+                    metricProperty: 'covid_daily_moving_average',
+                    label:
+                      text.section_deceased_rivm
+                        .line_chart_covid_daily_legend_trend_label_moving_average,
+                    shortLabel:
+                      text.section_deceased_rivm
+                        .line_chart_covid_daily_legend_trend_short_label_moving_average,
+                    color: colors.data.primary,
+                  },
```
Comment (VWSCoronaDashboard10) [inline packages/app/src/pages/gemeente/[code]/sterfte.tsx:162]: The last series of the array will be drawn with the "lowest z-index". In this case the line will be drawn behind the bars. Is that what we want?

VWSCoronaDashboard10 (approved):
Approved, but do have a question about drawing order. This might also apply for your other PR's.

packages/app/src/pages/gemeente/[code]/sterfte.tsx
Comment (VWSCoronaDashboard8) [reply packages/app/src/pages/gemeente/[code]/sterfte.tsx:162]: I think they are drawn with the last item on top. That's also how I would prefer the API to be. 

This is how I configured the bar/line chart for verpleeghuiszorg NL, so I think this configuration is correct.

VWSCoronaDashboard8 (approved):

packages/app/src/pages/gemeente/[code]/sterfte.tsx
Comment (VWSCoronaDashboard14) [reply packages/app/src/pages/gemeente/[code]/sterfte.tsx:162]: The bars are actually drawn on top of the line on the verpleeghuiszorg but since it is blue on blue it isn't noticable. But with for example the pink color you can see it slightly. So it should be first a line and than bar in the seriesConfig.

Screenshot 2021-04-21 at 12 55 06

Comment (VWSCoronaDashboard14) [reply packages/app/src/pages/gemeente/[code]/sterfte.tsx:162]: When reversing the order:
Screenshot 2021-04-21 at 12 55 26

Comment (VWSCoronaDashboard10) [reply packages/app/src/pages/gemeente/[code]/sterfte.tsx:162]: > I think they are drawn with the last item on top. That's also how I would prefer the API to be.
> 
> This is how I configured the bar/line chart for verpleeghuiszorg NL, so I think this configuration is correct.


Nope, the last item in the list is drawn at the back and the first item is rendered on top, there is a `.reverse()` in the `series.tsx`-component. I think it was implemented like this because the most important series should also appear as the first series in the legend..

Comment (VWSCoronaDashboard8) [reply packages/app/src/pages/gemeente/[code]/sterfte.tsx:162]: Allright, I stand corrected 👍

Comment (VWSCoronaDashboard13) [reply packages/app/src/pages/gemeente/[code]/sterfte.tsx]: Makes sense that the line should come first in the legend and tooltip. I'll update all of the graphs where this applies!

PR #2676, opened 2021-04-20
Add moving average to testing graphs

VWSCoronaDashboard13:
## Summary

Added moving average to testing graphs on NL, VR, and G

Also added new copy 🤼

PR #2680, opened 2021-04-21
add 7 days average disability care and living at home

VWSCoronaDashboard14:
Add moving averages for disability care and elderly at home.
Could have added the 7 days average label text in a more common way in Lokalise, but this gives them the flexibility to change it per graph if needed.


VWSCoronaDashboard1 (approved):

VWSCoronaDashboard8 (approved):

packages/app/src/pages/landelijk/gehandicaptenzorg.tsx
```diff
@@ -111,7 +111,15 @@ const DisabilityCare = (props: StaticProps<typeof getStaticProps>) => {
                 timeframe={timeframe}
                 seriesConfig={[
                   {
-                    type: 'area',
+                    type: 'line',
+                    metricProperty: 'newly_infected_people_moving_average',
+                    color: colors.data.primary,
+                    label:
+                      positiveTestedPeopleText.tooltip_labels
```
VWSCoronaDashboard8, line 118: This is not the place to pass tooltip labels. These show up in the legend + the tooltip _if_ there is no shortLabel set. So tooltip labels should go in `shortLabel`.

I think that means you still need to create `_moving_average` texts for labelling the average trends in the legend. Similar to what is done in #2676

VWSCoronaDashboard8 (changes requested):

packages/app/src/pages/landelijk/gehandicaptenzorg.tsx
VWSCoronaDashboard8: I see that there are some similar mistakes in that PR as well.

VWSCoronaDashboard8: These mistakes date back even further, so I will just add this to tech dept. But if you can still fix this for your texts in this PR that would be appreciated.

VWSCoronaDashboard8 (approved):

PR #2682, opened 2021-04-21
Handle null values in the tooltip for the timeSeriesChart

VWSCoronaDashboard14:
Render `-` when a null value is passed in the tooltip.

Screenshot 2021-04-21 at 13 23 46

### DIFF
--- a/packages/app/src/components/time-series-chart/components/tooltip/tooltip-series-list.tsx
+++ b/packages/app/src/components/time-series-chart/components/tooltip/tooltip-series-list.tsx
@@ -62,13 +62,11 @@ export function TooltipSeriesList<T extends TimestampedValue>({
   ) {
     const numberValue = (value[key] as unknown) as number | null;
 
-    if (numberValue === null) return '-';
-
     return isPresent(numberValue)
       ? isPercentage
         ? `${formatPercentage(numberValue)}%`
         : formatNumber(numberValue)
-      : '';
+      : '-';
   }
 
   const dateString = getDateStringFromValue(value);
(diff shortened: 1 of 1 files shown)


packages/app/src/components/time-series-chart/components/tooltip/tooltip-series-list.tsx
```diff
@@ -62,6 +62,8 @@ export function TooltipSeriesList<T extends TimestampedValue>({
   ) {
     const numberValue = (value[key] as unknown) as number | null;
 
+    if (numberValue === null) return '-';
```
VWSCoronaDashboard1, line 65:
```suggestion
    if (numberValue === null) return '-';

    return isPresent(numberValue)
      ? isPercentage
        ? `${formatPercentage(numberValue)}%`
        : formatNumber(numberValue)
      : '-';
  }```

**Review by VWSCoronaDashboard1 — CHANGES_REQUESTED** [review]

File: packages/app/src/components/time-series-chart/components/tooltip/tooltip-series-list.tsx (hunk 1 above)
Comment (VWSCoronaDashboard8) [reply packages/app/src/components/time-series-chart/components/tooltip/tooltip-series-list.tsx:65]: isPresent is already catching null values. So I think we only need this part. Otherwise the second isPresent is only catching undefined and the checks are partially redundant.

```suggestion
    return isPresent(numberValue)
      ? isPercentage
        ? `${formatPercentage(numberValue)}%`
        : formatNumber(numberValue)
      : '-';
```

VWSCoronaDashboard8 (changes requested):

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard1 (approved):

PR #2686, opened 2021-04-21
Topic/production on node images

VWSCoronaDashboard4:
## Summary

* Loads images and files directly from sanity
* Rewrites URLs from `/cms-files/` and `/cms-images`
* Adjusted helper functions and components
* Removed the sync assets script


VWSCoronaDashboard1 (approved):

packages/app/src/lib/sanity.tsx
```diff
@@ -109,15 +109,18 @@ export function getImageProps<T extends ImageBlock>(
...
 
   const width = findClosestSize(defaultWidth, imageResizeTargets);
   const height = width / metadata.dimensions.aspectRatio;
 
   const src = getImageSrc(node.asset, defaultWidth);
-  let srcSet = undefined; //we keep this undefined for SVG's, which don't need srcSets
+
+  /* we keep these undefined for SVGs, which don't need srcSets */
+  let srcSet = undefined;
+  let srcSetWebp = undefined;
```
VWSCoronaDashboard8, line 123: I don't think assigning undefined is any different from assigning nothing. But since these can also become strings I think this would be more correct?

```suggestion
  let srcSet: undefined | string;
  let srcSetWebp: undefined | string;
```

packages/cms/scripts/sync-assets.ts
```diff
@@ -1,167 +1,4 @@
-import { imageResizeTargets } from '@corona-dashboard/common';
-import download from 'download';
-import fs from 'fs-extra';
-import { chunk } from 'lodash';
-import fetch from 'node-fetch';
-import { getConfig } from './logic';
+console.log('Sync assets is deprecated');
```
VWSCoronaDashboard8, line 1: I would prefer to place this in an echo statement in the yarn script definition in package.json so you can remove the file completely.

packages/app/next.config.js
```diff
@@ -1,3 +1,7 @@
+const path = require('path');
+const config = { path: path.resolve('.env.local') };
+require('dotenv').config(config);
```
VWSCoronaDashboard8, line 3: Doesn't dotenv-flow automatically pick up any .env.local file? It seems odd to have to configure this explicitly. The cascade of different env files is what the library is created for it seems.

packages/app/next.config.js
```diff
@@ -7,6 +11,8 @@ const withBundleAnalyzer = require('@next/bundle-analyzer')({
 });
 
 const COMMIT_ID = process.env.NEXT_PUBLIC_COMMIT_ID || 'no-version-found';
+const SANITY_CDN_IMAGES = `https://cdn.sanity.io/images/${process.env.NEXT_PUBLIC_SANITY_PROJECT_ID}/${process.env.NEXT_PUBLIC_SANITY_DATASET}/`;
+const SANITY_CDN_FILES = `https://cdn.sanity.io/files/${process.env.NEXT_PUBLIC_SANITY_PROJECT_ID}/${process.env.NEXT_PUBLIC_SANITY_DATASET}/`;
```
VWSCoronaDashboard8, line 15: nit; looks like normal variables to me.

packages/app/src/pages/landelijk/vaccinaties.tsx
```diff
@@ -52,7 +52,7 @@ export const getStaticProps = createGetStaticProps(
   }>((context) => {
     const { locale = 'nl' } = context;
```
VWSCoronaDashboard8, line 53: Just wondering, can locale returned from context be undefined?

VWSCoronaDashboard8 (approved):

packages/app/src/lib/sanity.tsx
```diff
@@ -127,6 +130,9 @@ export function getImageProps<T extends ImageBlock>(
     srcSet = imageResizeTargets
       .map((size) => `${getImageSrc(asset, size)} ${size}w`)
       .join(', ');
+    srcSetWebp = imageResizeTargets
+      .map((size) => `${getImageSrc(asset, size)}&fm=webp ${size}w`)
+      .join(', ');
```
VWSCoronaDashboard10, line 135: We can get rid of the manual webp stuff and `<picture/>` element with `<source />` elements by adding the `&auto=format` parameter to the request url. The Sanity assets api/cdn will extract webp-compatibility from the request headers.

packages/app/src/components/cms/sanity-image.tsx
```diff
@@ -4,25 +4,24 @@ type SanityImageProps = {
...
   srcSet?: string;
+  srcSetWebp?: string;
   sizes?: string;
   className?: string;
   alt?: string;
 };
 
 export function SanityImage(props: SanityImageProps) {
-  const { src, srcSet, sizes, extension, ...imageProps } = props;
+  const { src, srcSet, srcSetWebp, sizes, extension, ...imageProps } = props;
```
VWSCoronaDashboard10, line 14: In theory we can now use the `next/Image` component with a custom loader attribute set which I think defeats the need of manually configuring sizes/srcSets/etc .. https://nextjs.org/docs/api-reference/next/image

packages/app/src/pages/landelijk/vaccinaties.tsx
VWSCoronaDashboard4: According to Typescript it can, though I guess practically it never will. Not sure where else we can make TS change its mind.

packages/app/src/components/cms/sanity-image.tsx
VWSCoronaDashboard4: I leave that for a separate PR, since it will need some tweaks to get it working properly.

PR #2687, opened 2021-04-22
Allow root properties and array-like text-values in sanity lokalize

VWSCoronaDashboard10:


VWSCoronaDashboard1 (approved):

packages/cms/src/lokalize/import.ts
```diff
@@ -77,6 +70,9 @@ for (const [key, dataText] of Object.entries(nl)) {
          * with the Sanity search feature.
          */
         lokalize_path: `${`${key}.${path}`.split('.').join('::')}`,
+
+        displayEmpty: !value,
```
VWSCoronaDashboard8, line 74: Not entirely consciously I have used snake case for database fields. Not sure about other Sanity types, but to keep it inline with the other fields in this object I think this should be named `display_empty`. And since it's a boolean I think `is_empty` might be good alternative too. I'm not sure it matters how the empty state is then used, so tying it to "display" might be unnecessary / less flexible.

packages/cms/src/schemas/documents/lokalize-text.tsx
```diff
@@ -53,9 +53,14 @@ export const lokalizeText = {
          */
         Rule.fields({
           nl: (fieldRule: any) => fieldRule.reset().required(),
-          en: (fieldRule: any) => fieldRule.reset(),
+          en: (fieldRule: any) => fieldRule.reset().required().warning(),
         }),
     },
+    {
+      title: 'Toon als lege tekst',
```
VWSCoronaDashboard8, line 60: If you choose to use is_empty this could maybe be "Gebruik als lege tekst"

VWSCoronaDashboard8 (approved):

packages/cms/src/lokalize/import.ts
VWSCoronaDashboard10: Not sure about the `is_empty`. For me it feels like it's a derived value based on the content of the text, which would result in a conflicting state where text is not empty though `is_empty` says it is.. Could you agree with `display_empty`?

VWSCoronaDashboard8: Sure, sounds reasonable 👍

PR #2689, opened 2021-04-22
Implement sanity lokalize cli and text transfer mechanism

VWSCoronaDashboard8:
## Summary

The CLI scripts give developers the ability to add / remove text entries from the "Lokalize" part of the Sanity dataset. These commands work on the development dataset. Every addition and removal will be logged to a file which is later used to transfer the new texts to the production dataset. Each mutation has a timestamp and if you run DELETE after ADD for the same key the actions will cancel each other out and no text will be transferred to production.

After each release the log can be cleared. Deletions from production are managed via a separate script that can compare the two datasets and delete all keys that are no longer present in development. I think this way we can keep things fairly simple while avoiding the edge-cases we would likely run in to when trying to automate everything.

### Adding a Text

1. Run `yarn lokalize:add`
2. Choose to type a (new) subject name or pick from a list of existing subjects
3. Specify the path within the subject
4. Specify NL and optionally EN text
5. Review and confirm
6. Choice to repeat from step 3 (to add more texts to the same subject) or exit. 

### Deleting a Text

1. Run `yarn lokalize:delete`
2. Type the name of a subject
3. A list is shown of all texts within this subject
4. Use a multi-select to define what texts should be deleted
5. Review and confirm

### Transfer new texts to production

The `lokalize:sync-additions` script should be run automatically by a Github action when a feature branch is merged into develop. It takes the CLI mutations log and figures out what keys to add. The keys are inserted using "createIfNotExists" so the script can be triggered multiple times without problems.

The script is _not_ removing any deleted texts from production, we only add new ones. Later we clean up the production dataset by running `lokalize:sync-deletions` which performs a full LokalizeText sync so that production mirrors the development dataset.

### List newly added texts

Text items that were added via the CLI are flagged with is_newly_added. This flag is cleared using a custom "PublishOrAccept" action in Sanity which replaces the regular publish on LokalizeText documents. This way we can list all new text keys for communications, and as soon as they publish them (or accept them as-is) the flag will be cleared.

### Other things in this PR

- List newly added lokalize texts in the sanity dashboard
- List recently published documents in the sanity dashboard
- Changed `search_key` with `::` separator back to `key` with dot notation. As the former was meant to also search in Lokalize which we don't need.

### TODO

Some things I'm leaving for later to not pile on to this PR, but they are not essential for migration:

- Move all CLI code to the CLI package
- Make the sync-deletions script interactive so that it displays whatever it is going to delete and asks for user confirmation. Then the script can also be added to package.json since triggering it doesn't yet damage anything.
- Add autocomplete to the lokalize:add/delete scripts, to improve the flow of the text add/delete scripts
- Cache subjects on disk to speed up CLI startup time.


packages/cms/src/lokalize/add-text.ts
```diff
@@ -0,0 +1,148 @@
...
+import { appendTextMutation } from './logic';
+import { LokalizeText } from './types';
+
+(async function run() {
+  const client = getClient('development');
+
+  /**
+   * @TODO We could cache the subjects in a temp location on disk and update
+   * them every time this CLI is run. This way we can avoid having to wait for
+   * the query to return.
```
VWSCoronaDashboard10, line 16: Did you measure how much time the query takes? Because I have the feeling most of the time `ts-node` is busy generating javascript.

packages/cms/src/lokalize/add-text.ts
```diff
@@ -0,0 +1,148 @@
...
+   */
+  const allTexts = (await client
+    .fetch(`*[_type == 'lokalizeText']`)
+    .catch((err) => {
+      throw new Error(`Failed to fetch texts: ${err.message}`);
+    })) as LokalizeText[];
+
+  const allSubjects = Object.keys(
+    allTexts.reduce((acc, x) => set(acc, x.subject, true), {})
+  );
```
VWSCoronaDashboard10, line 26: Might read a bit easier ..
```tsx
import uniq from 'lodash/uniq'
// ...
const allSubjects = uniq(allTexts.map(x => x.subject))
```

packages/cms/src/lokalize/add-text.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  const subjectChoices = Object.keys(
+    allTexts.reduce((acc, x) => set(acc, x.subject, true), {})
+  ).map((x) => ({ title: x, value: x }));
```
VWSCoronaDashboard10, line 30: and here you can use the previous variable
```suggestion
  const subjectChoices = allSubjects.map((x) => ({ title: x, value: x }));
```

packages/cms/src/lokalize/add-text.ts
```diff
@@ -0,0 +1,148 @@
...
+    subject,
+    path: response.path,
+    is_newly_added: true,
+    publish_count: 0,
+    text: {
+      _type: 'localeText',
+      nl: response.nl,
+      en: response.en,
+    },
+  };
```
VWSCoronaDashboard10, line 147: Can you add the LokalizeText interface to this object?

packages/cms/src/lokalize/clear-mutations-log.ts
```diff
@@ -0,0 +1,13 @@
...
+ */
+
+import { clearLogFile } from './logic';
+
+(async function run() {
+  clearLogFile();
+})().catch((err) => {
+  console.error(`Clear failed: ${err.message}`);
+  process.exit(1);
+});
```
VWSCoronaDashboard10, line 13: nit: called function isn't a promise so we could in theory wrap it in a try/catch to clean up a lot of noisy characters :-)

```suggestion
try {
  clearLogFile();
} catch (err) {
  console.error(`Clear failed: ${err.message}`);
  process.exit(1);
}
```

packages/cms/src/lokalize/delete-all.ts
```diff
@@ -0,0 +1,16 @@
...
+    const transaction = client.transaction();
+
+    result.forEach((document) => transaction.delete(document._id));
+
+    transaction.commit().catch((err) => {
+      console.error(`Failed to delete documents: ${err.message}`);
+      process.exit(1);
+    });
+  });
+})();
```
VWSCoronaDashboard10, line 16: you forgot to handle some promises here

```diff
- client.fetch(`*[_type == 'lokalizeText']`).then((result: any[]) => {
+ return client.fetch(`*[_type == 'lokalizeText']`).then((result: any[]) => {

// ...

- transaction.commit().catch((err) => {
+ return transaction.commit().catch((err) => {

// ...

- })();
+ })().catch( ... );

File: packages/cms/src/lokalize/delete-text.ts [hunk 7]
```diff
@@ -0,0 +1,98 @@
...
+    const choices = allTexts.map((x) => ({ title: x.path, value: x.path }));
+
+    const response = await prompts([
+      {
+        type: 'multiselect',
+        name: 'paths',
+        message: `What path in ${subjectResponse.subject}?`,
+        choices,
+        validate: (x) => {
+          console.log('validate', x);
```
VWSCoronaDashboard10, line 53: log on purpose?

packages/cms/src/lokalize/delete-text.ts
```diff
@@ -0,0 +1,98 @@
...
+          return x.length > 0;
+        },
+      },
+      {
+        type: (prev) => (prev.length > 0 ? 'confirm' : null),
+        name: 'confirmed',
+        message: (prev, values) => {
+          console.log('values', values, 'prev', prev);
```
VWSCoronaDashboard10, line 61: log on purpose?

packages/cms/src/lokalize/delete-text.ts
```diff
@@ -0,0 +1,98 @@
...
+        },
+      },
+    ]);
+
+    if (response.confirmed) {
+      const documentIdsToDelete = allTexts
+        .filter((x) => response.paths.includes(x.path))
+        .map((x) => x._id);
+
+      console.log(response.paths, documentIdsToDelete);
```
VWSCoronaDashboard10, line 74: log on purpose?

packages/cms/src/lokalize/logic/mutations.ts
```diff
@@ -0,0 +1,90 @@
...
+    return set(acc, mut.key, {
+      weight: weightByAction[mut.action] + prev.weight,
+      timestamp: mut.timestamp,
+    });
+  }, {} as Record<string, { weight: number; timestamp: string }>);
+
+  return (
+    Object.entries(collapsedKeys)
+      // For these keys the actions cancelled each other out
+      .filter(([__key, v]) => v.weight === 0)
```
VWSCoronaDashboard10, line 79:
```suggestion
      .filter(([__key, x]) => x.weight === 0)
```

or like you did below:
```suggestion
      .filter(([__key, { weight }]) => weight === 0)
```

packages/cms/src/lokalize/logic/mutations.ts
```diff
@@ -0,0 +1,90 @@
...
+ * communication team to prepare for release. Deletions are handled differently
+ * via the sync-deletions script.
+ */
+export function collapseTextMutations(mutations: TextMutation[]) {
+  const weightByAction = {
+    add: 1,
+    delete: -1,
+  } as const;
+
+  const collapsedKeys = mutations.reduce((acc, mut) => {
```
VWSCoronaDashboard10, line 67:
```suggestion
  const collapsedKeys = mutations.reduce((acc, mutation) => {
```
or
```suggestion
  const collapsedKeys = mutations.reduce((acc, x) => {
```

packages/cms/src/lokalize/sync-additions.ts
```diff
@@ -0,0 +1,105 @@
...
+  console.log('mutations', mutations);
+
+  const collapsedMutations = collapseTextMutations(mutations);
+
+  const additions = collapsedMutations.filter((x) => x.action === 'add');
+
+  const allDevTexts = (await devClient.fetch(`*[_type == 'lokalizeText'] |
+    order(subject asc)`)) as LokalizeText[];
+
+  const transaction = prdClient.transaction();
```
VWSCoronaDashboard10, line 47: Might want to rename this to `prdTransaction` to make it clear it's for production

packages/cms/src/lokalize/sync-deletions.ts
```diff
@@ -0,0 +1,119 @@
...
+  const prdKeysMissingInDev = difference(allPrdKeys, allDevKeys);
+
+  /**
+   * Maybe there could be an edge-case where a text addition on dev didn't make
+   * it to prd and is also not in the mutation log anymore. We might as well fix
+   * that here now we have all the data.
+   */
+  const devKeysMissingInPrd = difference(allDevKeys, allPrdKeys);
+
+  const transaction = prdClient.transaction();
```
VWSCoronaDashboard10, line 94: same here

packages/cms/src/lokalize/types.ts
```diff
@@ -0,0 +1,15 @@
...
+  _id: string;
+  path: string;
+  subject: string;
+  key: string;
+  should_display_empty: boolean;
+  is_newly_added: boolean;
+  publish_count: number;
+  text: {
+    _type: 'localeText';
+    nl: string;
```
VWSCoronaDashboard10, line 12: `nl` can be undefined too (e.g. in drafts)
```suggestion
    nl?: string;
```

packages/cms/src/actions/publish-or-accept.ts
```diff
@@ -0,0 +1,92 @@
...
+        is_newly_added: false,
+        publish_count: document.publish_count + 1,
+      };
+
+      patch.execute([{ set: patchData }]);
+
+      publish.execute();
+
+      props.onComplete();
+    },
```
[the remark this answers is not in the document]
VWSCoronaDashboard8: This is just a left-over from the original code where something async was done in the handler. I should remove it 👍

packages/cms/src/lokalize/add-text.ts
VWSCoronaDashboard8: No I didn't good point

packages/cms/src/lokalize/add-text.ts
VWSCoronaDashboard8: Definitely :) I think this is even a bug, since it's not producing a list of strings but I am expecting it further down

packages/cms/src/lokalize/add-text.ts
```diff
@@ -0,0 +1,144 @@
...
+ * Add one or multiple texts to the Sanity "Lokalize" dataset.
+ */
+import { set, uniq } from 'lodash';
+import prompts from 'prompts';
+import { getClient } from '../client';
+import { appendTextMutation } from './logic';
+import { LokalizeText } from './types';
+
+(async function run() {
+  const client = getClient('development');
```
VWSCoronaDashboard4, line 11: Will this always be `development`? Also for eg hotfixes?

packages/cms/src/lokalize/logic/mutations.ts
```diff
@@ -0,0 +1,96 @@
...
+      acc[mutation.key] = {
+        weight: weightByAction[mutation.action] + prev.weight,
+        timestamp: mutation.timestamp,
+      };
+      return acc;
+    },
+    {} as Record<string, { weight: number; timestamp: string }>
+  );
+
+  console.log('collapsedKeys', collapsedKeys);
```
VWSCoronaDashboard4, line 80: log on purpose?

packages/cms/src/lokalize/sync-additions.ts
```diff
@@ -0,0 +1,110 @@
...
+
+const devClient = getClient('development');
+const prdClient = getClient('production');
+
+(async function run() {
+  const mutations = await readTextMutations();
+
+  const collapsedMutations = collapseTextMutations(mutations);
+
+  console.log('collapsedMutations', collapsedMutations);
```
VWSCoronaDashboard4, line 40: log on purpose?

VWSCoronaDashboard4 (approved):
Looking good overall! :100: Some nitpicks / questions :-)

packages/cms/src/lokalize/add-text.ts
VWSCoronaDashboard8: Hmm LokalizedText is typed with an _id field, which makes it rather annoying to use in this case. I find that smart about Firestore, where they kept the id out of the document data.

VWSCoronaDashboard8: I'll do this

```
const text: Omit<LokalizeText, '_id'> = {
    _type: 'lokalizeText',
    key: `${subject}.${response.path}`,
    subject,
    path: response.path,
    is_newly_added: true,
    publish_count: 0,
    should_display_empty: false,
    text: {
      _type: 'localeText',
      nl: response.nl,
      en: response.en,
    },
  };

  return text;
```

VWSCoronaDashboard8: It did expose the fact that should_display_empty was missing, so it's still a good idea to type it of course

packages/cms/src/lokalize/clear-mutations-log.ts
VWSCoronaDashboard8: I gave it this format to be consistent with the other cli scripts, and it gives the ability to easily extend it without refactoring.

packages/cms/src/lokalize/add-text.ts
VWSCoronaDashboard8: Yes I don't see a use-case for hotfixes on production. We can inject new keys into production from develop at any moment by running `sync-additions`

packages/cms/src/lokalize/clear-mutations-log.ts
VWSCoronaDashboard8: I think I'd prefer to keep it like this for now.

PR #2692, opened 2021-04-22
Fix console error collapsible bug

VWSCoronaDashboard14:
Move the `wrapperRef` to his own element, I couldn't spot any difference in terms of working. We also use the same technique on the `collapsible-button` to not directly put it directly on the `DisclosurePanel`.

### DIFF
--- a/packages/app/src/components/collapsible/collapsible-section.tsx
+++ b/packages/app/src/components/collapsible/collapsible-section.tsx
@@ -137,23 +137,24 @@ export const CollapsibleSection = ({
         </Summary>
 
         <Panel
-          ref={wrapperRef}
           style={{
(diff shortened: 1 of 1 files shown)


VWSCoronaDashboard1 (approved):
One general comment though, did you also not spot any difference in behaviour when you removed the wrapperRef entirely? In that case it might be useless ;)

VWSCoronaDashboard14:
@VWSCoronaDashboard1 I didn't spot any difference in how it works no. Let me confirm that a second time and might remove it alltogether. But not 100% sure why it is needed anyway.

VWSCoronaDashboard14:
@VWSCoronaDashboard1 it seems that the `wrapperRef` is still needed, it gives some unexpected behavior when removed.

PR #2693, opened 2021-04-22
Update admission per age group schemas

VWSCoronaDashboard13:
## Summary

Updating the admissions per age group schemas to have a date span

## Motivation

The data is representative of a weeks worth of admissions and needs to be visually represented that way as well. The current data structure makes it hard to do that 

## Detailed design

replaced the `date_unix` with `date_start_unix` and `date_end_unix`

### DIFF
--- a/packages/app/schema/nl/intensive_care_nice_per_age_group.json
+++ b/packages/app/schema/nl/intensive_care_nice_per_age_group.json
@@ -21,7 +21,6 @@
       "additionalProperties": false,
       "type": "object",
       "required": [
-        
         "admissions_age_0_19_per_million",
         "admissions_age_20_29_per_million",
         "admissions_age_30_39_per_million",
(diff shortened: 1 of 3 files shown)


packages/app/schema/nl/intensive_care_nice_per_age_group.json
```diff
@@ -32,7 +31,8 @@
         "admissions_age_80_89_per_million",
         "admissions_age_90_plus_per_million",
         "admissions_overall_per_million",
-        "date_unix",
+        "date_start_unix",
+        "dat_end_unix",
```
VWSCoronaDashboard4, line 35: typo ;-) also in the other one

VWSCoronaDashboard4 (approved):

packages/app/schema/nl/intensive_care_nice_per_age_group.json
VWSCoronaDashboard13: 🤭

PR #2695, opened 2021-04-22
Change header of homepage to h1

VWSCoronaDashboard14:
With the work I've done by removing the SCSS from the app, I've changed most of the headers already to a h1. 
The homepage had a `visibilityHidden`  h1 that yielded the same that is in the `topicalHeader` so that is by default a h1 now and removed the unnecessary hidden text, that was not even working.


VWSCoronaDashboard1 (approved):

VWSCoronaDashboard10 (approved):

packages/app/src/domain/topical/topical-section-header.tsx
```diff
@@ -55,7 +55,7 @@ export function TopicalSectionHeader({
           alignItems="baseline"
         >
           <Heading
-            level={2}
+            level={1}
```
VWSCoronaDashboard8, line 58: I don't get this. The component is called a section header. How is it ok to assume that every section will be H1? Seems like it's by definition not a section then.

VWSCoronaDashboard14: In my opinion, all those headings should be an h1 in the way they are used. Following your feedback, I rather rename the component `topical-header` or something similar. I don't expect them they get used more than once per page. @VWSCoronaDashboard8

VWSCoronaDashboard8: There seem to be 3 pages that each use 5 `TopicalSectionHeader` components, so I don't see how that would work?

VWSCoronaDashboard14: My bad, didn't look good 😢

VWSCoronaDashboard8 (approved):

PR #2698, opened 2021-04-22
Create hot reloadable sanity lokalize texts

VWSCoronaDashboard10:
The useLokalizeText hook will return an object which contains all lokalize translations.

Additionally, with `enableHotReload: true` it can also connect with Sanity
to receive real-time (hot) updates of draft translations.

If enabled a tiny button will be rendered at the bottom-right which is only
visible on hover.

This button toggles through three states:

- undefined (gray): app is rendered with translations from filesystem
- live (green): app is listening to real-time sanity updates
- paths (blue): app is displaying the pathnames of every key instead of the value

The live/paths state can also be toggled with shortkey shift+t.

The hot reload feature is only available on vercel deployments. Locally it can be activated with the env var:
```
NEXT_PUBLIC_HOT_RELOAD_LOKALIZE=1
```


packages/app/src/locale/use-lokalize-text.tsx
```diff
@@ -0,0 +1,174 @@
...
+import styled from 'styled-components';
+import { isDefined } from 'ts-is-present';
+import DatabaseIcon from '~/assets/database.svg';
+import { getClient } from '~/lib/sanity';
+import { LanguageKey, languages } from '~/locale';
+import { LokalizeText } from '~/types/cms';
+import { useHotkey } from '~/utils/hotkey/use-hotkey';
+import { useIsMountedRef } from '~/utils/use-is-mounted-ref';
+
+export function useLokalizeText(
```
VWSCoronaDashboard1, line 22: We already know what the hook does exactly due to your demo this morning. But perhaps you can add a general comment that explains what the display modes exactly entail.

packages/app/src/locale/use-lokalize-text.tsx
```diff
@@ -0,0 +1,174 @@
...
+      .then((client) => client.fetch(query))
+      .then((documents: LokalizeText[]) => {
+        const flattenTexts = createFlattenTexts(documents)[locale];
+        setText(() => unflatten(flattenTexts, { object: true }));
+      })
+      .catch((err) => console.error(err));
+
+    let subscription: Subscription | undefined;
+
+    getClient().then(
```
VWSCoronaDashboard1, line 63: why do you need to create the client twice here? Is there a particular reason why simply `const client = await getClient()` doesn't suffice?

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard4 (approved):

packages/app/src/locale/use-lokalize-text.tsx
VWSCoronaDashboard10: good catch, I can clean this up.

VWSCoronaDashboard1: I think this deals with that situation:
https://stackoverflow.com/questions/53332321/react-hook-warnings-for-async-function-in-useeffect-useeffect-function-must-ret
But indeed, if whatever is returned from the await needs cleaning up, then it's difficult to do that in the return function. Perhaps if you return the promise in that same function and perform the clean-up only when the promise has resolved and if not, cancel it, or something? I admit it's a gnarly one...

packages/app/src/locale/use-lokalize-text.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+    getClient().then(async (client) => {
+      await client
+        .fetch(query)
+        .then((documents: LokalizeText[]) => {
+          if (!isMountedRef.current) return;
+          const flattenTexts = createFlattenTexts(documents)[locale];
+          setText(() => unflatten(flattenTexts, { object: true }));
+        })
+        .catch((err) => console.error(err));
```
VWSCoronaDashboard8, line 80: Not sure why you are mixing async/await with promises here. I think it is cleaner to stick with async/await where possible.

```suggestion
      const documents: LokalizeText[] = await client.fetch(query)
      
      if (!isMountedRef.current) return;
          
      const flattenTexts = createFlattenTexts(documents)[locale];
      setText(() => unflatten(flattenTexts, { object: true }));
```

I think the catch can be moved to the getClient promise chain, because there it seems to be missing, and it would still print any error that gets thrown here.

packages/app/src/locale/use-lokalize-text.tsx
```diff
@@ -0,0 +1,193 @@
...
+           * Here we map all locale text object values to their "path"
+           */
+          (unflatten(
+            Object.keys(flatten(text)).reduce((res, key) => {
+              res[key] = key.replace(/\./g, '::');
+              return res;
+            }, {} as Record<string, string>),
+            { object: true }
+          ) as typeof text)
+        : undefined,
```
VWSCoronaDashboard8, line 126: nit; I think you can make this more readable by using one or two temporary variables. This feels a bit much with nesting IMO

packages/app/src/locale/use-lokalize-text.tsx
```diff
@@ -0,0 +1,193 @@
...
+    [displayMode, text]
+  );
+
+  const toggleEl = enableHotReload ? (
```
VWSCoronaDashboard8, line 130: nit; To me this abbreviation is not so common so I wonder if it's worth it.

```suggestion
  const toggleElement = enableHotReload ? (
```

packages/app/src/lib/sanity.tsx
```diff
@@ -38,7 +39,10 @@ const config = {
 export const PortableText = BlockContent;
 
 // Set up the client for fetching data in the getProps page functions
-export const client = sanityClient(config);
+export async function getClient() {
```
VWSCoronaDashboard8, line 42: Do we really need this as a dynamic import? If client instantiation needs to be limited to on-demand, we can maybe solve it by creating the client only once on first call of the function. That could be done without async then.

packages/app/src/types/cms.d.ts
```diff
@@ -154,3 +154,19 @@ export type TitleDescriptionBlock = {
...
+  _type: 'lokalizeText';
+  _id: string;
+  _createdAt: string;
+  _updatedAt: string;
+
+  lokalize_path: string;
+  subject: string;
+  path: string;
+  text: {
+    nl: string | undefined;
```
VWSCoronaDashboard8, line 168: Wasn't nl text required?

packages/app/src/types/cms.d.ts
```diff
@@ -154,3 +154,19 @@ export type TitleDescriptionBlock = {
...
   title: string;
 };
+
+export interface LokalizeText {
+  _type: 'lokalizeText';
+  _id: string;
+  _createdAt: string;
+  _updatedAt: string;
+
+  lokalize_path: string;
```
VWSCoronaDashboard8, line 164: To avoid confusion between path and lokalize_path, I think it might be good to refer to anything that is a combination of subject & path as "key". Then key would be the unique id / object key what defines the text.

Also, the lokalize_key could be named `search_key` or just `key` because it is also used to each within Sanity.  The search prefix might be nice because then it's clear why the key is using :: notation instead of dot. 

But regardless of the format then key is always a combination of subject and path.

packages/cms/src/lokalize/import.ts
```diff
@@ -62,7 +62,11 @@ for (let [key, dataText] of Object.entries(nl)) {
 
       return {
         _type: 'lokalizeText',
-        _id: `jsonKey__${key}.${path}`,
+        /**
+         * ids with a `.` are only accessible for authenticated users. We'll
+         * replace dots with dashes.
+         */
+        _id: `jsonKey__${key}.${path}`.replace(/\./g, '-'),
```
VWSCoronaDashboard8, line 69: Since we decided we'll not use opaque ids after the transition (for dev/prd sync reasons), I suggest to make the prefix `lokalize__` instead of `jsonKey__`

packages/common/src/domain/lokalize/index.ts
```diff
@@ -0,0 +1,64 @@
+import { LokalizeText } from '@corona-dashboard/app/src/types/cms';
+
+export function createFlattenTexts(
```
VWSCoronaDashboard8, line 3:
```suggestion
export function createFlattenedTexts(
```

of 

```suggestion
export function createFlatTexts(
```

VWSCoronaDashboard8 (approved):

packages/app/src/locale/use-lokalize-text.tsx
VWSCoronaDashboard10: ah of course.. leftover from some refactoring

packages/app/src/locale/use-lokalize-text.tsx
VWSCoronaDashboard10: fair point :p

packages/app/src/types/cms.d.ts
VWSCoronaDashboard10: It can be empty inside drafts

PR #2708, opened 2021-04-26
Set the tick values of the reproduction chart

VWSCoronaDashboard14:
A more dynamic approach I've added.
But I could also use the manual approach `tickValues={[0, 0.5, 1.0, 1.5, 2, 2.5]}` but in the end, it is visually the same.

### DIFF
--- a/packages/app/src/domain/tested/reproduction-chart-tile.tsx
+++ b/packages/app/src/domain/tested/reproduction-chart-tile.tsx
@@ -49,7 +49,7 @@ export function ReproductionChartTile({
 
   const tickSteps = Array(highestTick / tickStep + 1)
     .fill(tickStep)
-    .map((v, i) => i * v);
+    .map((step, index) => index * step);
 
   return (
     <ChartTile
@@ -80,7 +80,7 @@ export function ReproductionChartTile({
               label: siteText.common.signaalwaarde,
             },
           }}
-          xtickValues={tickSteps}
+          tickValues={tickSteps}
         />
       )}
     </ChartTile>
--- a/packages/app/src/domain/tested/reproduction-chart-tile.tsx
+++ b/packages/app/src/domain/tested/reproduction-chart-tile.tsx
@@ -34,24 +34,21 @@ export function ReproductionChartTile({
     0,
     data.values.findIndex((x) => !isPresent(x.index_average))
   );
-  const last_value = last(values) as NationalReproductionValue;
 
-  const findMaxAverage = values.reduce((prev, current) =>
-    isPresent(prev.index_average) > isPresent(current.index_average)
-      ? prev
-      : current
-  );
-
+  /**
+   * Calculate the highest tick and round it up on the closet number on the 0.5 scales.
+   * Then create an array with every tickStep till the highestTick
+   * to use for the tickValues in the chart.
+   */
   const tickStep = 0.5;
 
-  assert(
-    findMaxAverage.index_average,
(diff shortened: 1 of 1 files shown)


VWSCoronaDashboard4 (approved):

packages/app/src/domain/tested/reproduction-chart-tile.tsx
```diff
@@ -35,6 +36,25 @@ export function ReproductionChartTile({
   );
   const last_value = last(values) as NationalReproductionValue;
 
+  const findMaxAverage = values.reduce((prev, current) =>
+    isPresent(prev.index_average) > isPresent(current.index_average)
+      ? prev
+      : current
+  );
```
VWSCoronaDashboard10, line 43: this is probably not doing what you'd expect. The isPresent returns a boolean so your'e comparing booleans instead of values, which means the value of `findMaxAverage` is the first value with an `index_average`-value.

I think what you really want is:

```tsx
const max = Math.max(...values.map((x) => x.index_average).filter(isPresent));

**Review by VWSCoronaDashboard10 — CHANGES_REQUESTED** [review]

File: packages/app/src/domain/tested/reproduction-chart-tile.tsx (hunk 1 above)
Comment (VWSCoronaDashboard14) [reply packages/app/src/domain/tested/reproduction-chart-tile.tsx]: @VWSCoronaDashboard10 I didn't catch that one, without the `isPresent` is indeed working but typescript gives me some errors (hence why I added that, but didn't look back apparently). Your solution is working better.

File: packages/app/src/domain/tested/reproduction-chart-tile.tsx [hunk 2]
```diff
@@ -33,7 +33,23 @@ export function ReproductionChartTile({
...
+   * Then create an array with every tickStep till the highestTick
+   * to use for the tickValues in the chart.
+   */
+  const tickStep = 0.5;
+
   const last_value = last(values) as NationalReproductionValue;
+  const maxIndexAverage = Math.max(
+    ...values.map((x) => x.index_average).filter(isPresent)
+  );
+  const highestTick = Math.ceil(maxIndexAverage * 2) / 2;
```
VWSCoronaDashboard8, line 48:
```suggestion
  const highestTick = Math.ceil(maxIndexAverage);
```
Just wondering, this produced different results?

VWSCoronaDashboard8 (approved):

packages/app/src/domain/tested/reproduction-chart-tile.tsx
```diff
@@ -33,7 +33,23 @@ export function ReproductionChartTile({
...
+
+  const tickSteps = Array(highestTick / tickStep + 1)
+    .fill(tickStep)
+    .map((step, index) => index * step);
```
VWSCoronaDashboard8, line 52: Since these 15 lines are all about achieving one specific goal, I think it's nice to place it in a local function and call that. This clutters the code a bit and if you just read `calculateTickValues()` or something and you're not particularly interested in the internals, you can just read over it.

packages/app/src/domain/tested/reproduction-chart-tile.tsx
VWSCoronaDashboard14: It does, without it rounds up to a whole number (3). And otherwise, it rounds to the closest 0.5 value (2.5).

VWSCoronaDashboard1 (approved):

PR #2719, opened 2021-04-26
Add script that strips unused locale keys from nl.json and en.json

VWSCoronaDashboard1:
## Summary

The script can be run by running `yarn strip-locale` in the `packages/cli` directory.
The resulting files are written to the `app/src/locale` directory as `nl_stripped.json` and `en_stripped.json`.

## Motivation

Optimization.

### Governance

- [ ] Documentation is added


packages/app/src/components/sidebar-metric/sidebar-barscale.tsx
```diff
@@ -82,7 +85,9 @@ export function SidebarBarScale<T>({
   );
 
   assert(
-    text.barscale_screenreader_text,
+    typeof text === 'object' &&
+      'barscale_screenreader_text' in text &&
+      text.barscale_screenreader_text,
```
VWSCoronaDashboard8, line 90: Wouldn't it be possible to use simple optional chaining like this?

```suggestion
      text?.barscale_screenreader_text
```

packages/app/src/components/sidebar-metric/sidebar-metric.tsx
```diff
@@ -90,9 +90,10 @@ export function SidebarMetric<T extends { difference: unknown }>({
    * @TODO this should really be called sidebar_metric_description or something
    * as it's not a title at all.
    */
+  const text = siteText[localeTextKey];
   const title =
-    get(siteText, [localeTextKey, 'kpi_titel']) ||
-    get(siteText, [localeTextKey, 'titel_kpi']);
+    (typeof text === 'object' && 'kpi_titel' in text && text.kpi_titel) ||
+    (typeof text === 'object' && 'titel_kpi' in text && text.titel_kpi);
```
VWSCoronaDashboard8, line 96: Similar question here:

```suggestion
    text?.kpi_titel || text?.titel_kpi;
```

.gitignore
```diff
@@ -25,3 +25,5 @@ exports
 
 # TODO rename the Sanity locale exports to en/nl.json and ignore those files instead
 packages/app/src/locale/*_export.json
+/packages/app/src/locale/en_stripped.json
+/packages/app/src/locale/nl_stripped.json
```
VWSCoronaDashboard8, line 29: nit; you could make it the same as the line above

```suggestion
packages/app/src/locale/*_stripped.json
```

packages/cli/src/locale/strip-locale.ts
```diff
@@ -0,0 +1,104 @@
...
+
+const newLocaleObjects = propertyAssignmentNodes
+  .map((x) => createFullPropertyChain(x))
+  .concat(whitelist)
+  // Only keep the deepest paths:
+  .filter(
+    (x, _i, l) =>
+      l.findIndex((y) => y.startsWith(`${x}.`) && y.length > x.length + 1) ===
+      -1
+  )
```
VWSCoronaDashboard8, line 54: nit; I think one letter variable names should be avoided. With the exception of x because we standardized that for short-hand purposes. But in this case x is not so obvious so I'd suggest to name that too, with the benefit of being able to use x to replace y. I'm using `subject` here not sure if that's the best name for it, or I understand the code correctly, but I think it's the same as what subject is in sanity LokalizeText documents...

```suggestion
    (subject, _index, list) =>
      list.findIndex((x) => x.startsWith(`${subject}.`) && x.length > x.length + 1) ===
      -1
  )
```

packages/cli/src/locale/strip-locale.ts
```diff
@@ -0,0 +1,104 @@
...
+  .sort()
+  .reduce(
+    (aggr, chain) => {
```
VWSCoronaDashboard8, line 57: Maybe "path" is a more common name for chain?

packages/cli/src/locale/strip-locale.ts
```diff
@@ -0,0 +1,104 @@
...
+      const NlValue = get(NlOriginal, chain);
+      const EnValue = get(EnOriginal, chain);
+      if (NlValue !== undefined) {
```
VWSCoronaDashboard8, line 60: nit; I think the opposite reads a little better

```suggestion
      if (isDefined(NlValue)) {
```

packages/cli/src/locale/strip-locale.ts
```diff
@@ -0,0 +1,104 @@
...
+        aggr.nl[chain] = NlValue;
+        aggr.en[chain] = EnValue;
+      }
+      return aggr;
+    },
+    { nl: {}, en: {} } as any
```
VWSCoronaDashboard8: Not sure why you'd have to use any here. Seems like it's code you control. Maybe this works?
```suggestion
    { nl: {}, en: {} } as { nl: Record<string, unknown>, en: Record<string, unknown> }
```

packages/cli/src/locale/strip-locale.ts
```diff
@@ -0,0 +1,104 @@
...
+  );
+
+const newJson: any = unflatten(newLocaleObjects, { object: true });
```
VWSCoronaDashboard8: Same here, since you're expecting newJson to have nl and en keys below. But maybe this solves itself when you type the function return value.

packages/app/src/components/sidebar-metric/sidebar-barscale.tsx
VWSCoronaDashboard10: Maybe, but this is mostly done for type narrowing so that has to keep working/be tested

packages/app/src/components/sidebar-metric/sidebar-metric.tsx
VWSCoronaDashboard1: That won't work actually, because `text` is a union of many types. Therefore the compiler will bark at you if you just use the `?` shortcut.

VWSCoronaDashboard8 (approved):

packages/app/src/components/sidebar-metric/sidebar-barscale.tsx
```diff
@@ -30,7 +33,7 @@ export function SidebarBarScale<T>({
 }: SidebarBarScaleProps<T>) {
   const { siteText } = useIntl();
 
-  const text = siteText[localeTextKey] as Record<string, string>;
+  const text = siteText[localeTextKey]; // as Record<string, string>;
```
VWSCoronaDashboard10, line 36: comment

packages/cli/package.json
```diff
@@ -26,7 +29,8 @@
     "validate-last-values": "ts-node src/scripts/validate-last-values.ts",
     "validate-features": "ts-node src/scripts/validate-features.ts",
     "generate-typescript": "ts-node src/scripts/generate-data-types.ts",
-    "generate-sitemap": "ts-node src/scripts/generate-sitemap.ts"
+    "generate-sitemap": "ts-node src/scripts/generate-sitemap.ts",
+    "strip-locale": "ts-node src/locale/strip-locale.ts"
```
VWSCoronaDashboard10, line 33: name of this script/file could be a bit more descriptive, eg `strip-unused-locale-keys`

VWSCoronaDashboard10 (approved):

packages/cli/src/locale/strip-locale.ts
```diff
@@ -0,0 +1,115 @@
...
+      return aggr;
+    },
+    { nl: {}, en: {} } as {
+      nl: Record<string, any | string>;
+      en: Record<string, any | string>;
+    }
+  );
+
+const newJson: {
+  nl: Record<string, any | string>;
```
VWSCoronaDashboard4, line 75: if you name this type, you can use that name to refer to a sub block instead of `any | string`

Or import it from another place, my hunch this is typed stricter somewhere else.

VWSCoronaDashboard4 (approved):

PR #2720, opened 2021-04-26
Fix sitemap.xml for nextjs via node

VWSCoronaDashboard4:
## Summary

Fix sitemap.xml for nextjs via node


packages/cli/package.json
```diff
@@ -26,7 +26,7 @@
     "validate-last-values": "ts-node src/scripts/validate-last-values.ts",
     "validate-features": "ts-node src/scripts/validate-features.ts",
     "generate-typescript": "ts-node src/scripts/generate-data-types.ts",
-    "generate-sitemap": "ts-node src/scripts/generate-sitemap.ts"
+    "generate-sitemap": "echo 'generate-sitemap is deprecated' && exit 1"
```
VWSCoronaDashboard8, line 29: I don't think deprecation is a reason to break the build pipeline, so exit 0 might be more appropriate here.

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard1 (approved):

packages/cli/package.json
VWSCoronaDashboard4: It's already removed from the build script in the `topic/production-on-node` branch

PR #2721, opened 2021-04-26
Update paths sitemap generator

VWSCoronaDashboard14:
Seems like the paths and the environment variables didn't work in the sitemap generator.
This fixes it locally, but can't check on a Vercel deployment if it works with the environment variables.

### DIFF
--- a/packages/cli/src/scripts/generate-sitemap.ts
+++ b/packages/cli/src/scripts/generate-sitemap.ts
@@ -26,8 +26,6 @@ const publicOutputDirectory = path.resolve(
 );
 
 async function main() {
-  console.log(path.resolve('../app/.env.local'))
-
   assert(
     process.env.NEXT_PUBLIC_SANITY_PROJECT_ID,
     'Missing NEXT_PUBLIC_SANITY_PROJECT_ID env var'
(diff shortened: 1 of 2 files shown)


VWSCoronaDashboard1 (approved):

VWSCoronaDashboard4 (approved):

packages/cli/src/scripts/generate-sitemap.ts
```diff
@@ -8,6 +8,7 @@ import { features } from '../../../app/src/config/features';
 import { gmData } from '../../../app/src/data/gm';
 import { vrData } from '../../../app/src/data/vr';
 import { logError } from '../utils';
+require('dotenv').config({ path: '../app/.env.local' })
```
VWSCoronaDashboard8, line 11: We should have a look this dot env business at some point. The mechanism is not designed to work with these explicit imports of specific files I think. The env file should probably live in the root and not app, and be called .env or something that would be picked up automatically by dotenv at least. I'll add this to tech depth.

But this sitemap thing will be replaced once we move to nodejs, so for this specific instance it doesn't matter anyway.

VWSCoronaDashboard8 (approved):

PR #2745, opened 2021-04-29
ci: Backup Sanity data regularly

VWSCoronaDashboard:
## Summary

Backups public sanity data on every new release in GitHub artifacts

## Motivation

- Allows people to set up their own Sanity servers to use with the app
- Backups are _good_


VWSCoronaDashboard10 (changes requested):
see slack

PR #2751, opened 2021-04-30
Move touch bar above the markers for the vaccine area graph

VWSCoronaDashboard14:
Render the dividers under the touch/mouse bar so it's always possible to interact with the graph


VWSCoronaDashboard1 (approved):

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard10 (approved):

VWSCoronaDashboard10 (changes requested):
I don't think this isn't the right solution. It looks like something is going wrong with the nearest point calculations, I guess it would be most efficient to have @VWSCoronaDashboard1 pick this up.

VWSCoronaDashboard10:
closed in favor of #2835

PR #2756, opened 2021-04-30
Rethink text sync strategy

VWSCoronaDashboard8:
## Summary

There were some oversights in the sanity lokalize text sync strategies. Texts that get deleted via the CLI can not actually be deleted straight-away from the development dataset without breaking other branches. This PR aims to solve those issues.

- CLI delete-text now marks the text for deletion, but actual deletion is postponed until the feature branch gets merged.
- The export script filters out any keys that were marked for deletion in the mutations file. So your local feature branch does use a set where the key is already deleted.
- The export script is now called after text add/delete operations.
- The collapse mutations logic now clips at 1 delete action, because without deleting the actual text document you could easily run the delete command on the same key multiple times (which would create edge cases when synced)
- The collapse mutations logic now figures out what texts initially got added and later removed. Because the addition creates a document and deletion doesn't remove it, the `sync-after-feature` script still needs to remove that document from the development set.
- The `sync-additions` script is now named `sync-after-feature` because it now applies deletions to development, and additions to production.
- The `sync-deletions` script is now named `sync-after-release` because it now also includes clearing the mutations log. And then it's clear when the script should be run.
- The `lokalize:import` script is removed from package.json as we should not ever need to run it anymore


packages/cms/src/lokalize/logic/mutations.ts
```diff
@@ -52,43 +53,77 @@ export function readTextMutations() {
...
+      timestamp: mutation.timestamp,
+    };
+    return acc;
+  }, {} as Record<string, { weight: number; timestamp: string }>);
+
+  /**
+   * Because new keys are added immediately, but deletions are only scheduled
+   * for later. We still need to mark the keys that were first newly create, but
+   * later delete as action "delete" otherwise they will not be removed from the
+   * dataset.
```
VWSCoronaDashboard1, line 93:
```suggestion
   * for later. We still need to mark the keys that were first newly created, but
   * later deleted as action "delete" otherwise they will not be removed from the
   * dataset.
```

VWSCoronaDashboard1 (approved):

packages/cms/src/lokalize/sync-after-feature.ts
```diff
@@ -0,0 +1,178 @@
...
+      name: 'isConfirmed',
+      message:
+        'This script should be run (automatically) on develop, right after a feature branch got merged. Are you aware of this?',
+      initial: false,
+    },
+  ]);
+
+  if (!response.isConfirmed) {
+    process.exit(0);
+  }
```
VWSCoronaDashboard10, line 37: How does this prompts thing work when the script is called on gh actions?

packages/cms/src/lokalize/sync-after-feature.ts
```diff
@@ -0,0 +1,178 @@
...
+  if (deletions.length === 0) {
+    console.log('There are no mutations that result in keys to delete');
+    process.exit(0);
+  }
+
+  let successCount = 0;
+  let failureCount = 0;
+
+  const documentIdsToDelete = deletions
+    .map(({ key }) => allTextDocuments.find((x) => x.key === key))
```
VWSCoronaDashboard10, line 145: Here's a bug because the query also includes drafts. Drafts should also be removed, and the client-facing bug is that delete only deletes a draft if that document is found earlier.

I think we should delete both drafts as documents.
```suggestion
    .flatMap(({ key }) => allTextDocuments.filter(x) => x.key === key))
```

packages/cms/src/lokalize/sync-after-feature.ts
```diff
@@ -0,0 +1,178 @@
...
+    .map((x) => x?._id)
+    .filter(isDefined);
+
+  if (documentIdsToDelete.length !== deletions.length) {
```
VWSCoronaDashboard10, line 149: This check will fail when we include drafts

VWSCoronaDashboard10 (changes requested):

packages/cms/src/lokalize/sync-after-feature.ts
VWSCoronaDashboard8: Eeh I guess how it works is it doesn't 😅

packages/cms/src/lokalize/sync-after-feature.ts
VWSCoronaDashboard8: I wasn't aware that drafts are separate documents.

VWSCoronaDashboard10: yep I think they are basically copies of the original document but the `id` is prepended with `drafts.`

VWSCoronaDashboard8: I think the easiest solution is to just not inject drafts from development into production. I think the mental model is a bit easier then. Otherwise the draft and published documents will both show up in the list of new documents on production.

VWSCoronaDashboard10 (approved):

PR #2757, opened 2021-04-30
Improve tooltip jumping on TimeSeriesChart

VWSCoronaDashboard13:
## Summary

Reduce tooltip width changes when hovering on TimeSeriesCharts

## Motivation

Tooltip was changing inside depending on the size of its' content which didn't look very smooth

## Detailed design

Moved the string formatting logic to hooks in the `formatStrings` file in the tooltip directory so it could be used in the time-series-chart to calculate the width of values

Found the largest number of characters that could be shown in the value and set that as the min width for all of the values

## Drawbacks

Doesn't calculate the width for each series individually, but I think this is sufficient and simpler


packages/app/src/components/time-series-chart/components/tooltip/formatStrings.ts
```diff
@@ -0,0 +1,61 @@
+import { useIntl } from '~/intl';
+import { isPresent } from 'ts-is-present';
+import {
+  isDateSpanValue,
+  isDateValue,
+  TimestampedValue,
+} from '@corona-dashboard/common';
+import { useCallback } from 'react';
+
+export function useGetValueString() {
```
VWSCoronaDashboard1, line 10: I'm not sure if it's absolutely necessary to wrap all of these in a useCallback. From what I understand the useCallback hook is specifically meant for functions that are passed down as component props:
https://aheadcreative.co.uk/articles/when-to-use-react-usecallback/

VWSCoronaDashboard1 (approved):

packages/app/src/components/time-series-chart/components/tooltip/formatStrings.ts
```diff
@@ -0,0 +1,61 @@
+import { useIntl } from '~/intl';
```
VWSCoronaDashboard8, line 1: We choose to not camel-case file names

packages/app/src/components/time-series-chart/components/tooltip/formatStrings.ts
VWSCoronaDashboard8: I don't yet see the point in applying useCallback either. Not in the way the functions are use now. Possibly if these are expected to be moved to a more common location and passed down via props it would be useful, but I suggest to keep it simple for now.

Having all these separate hooks is a little verbose. I think they can be a combined in a single hook which exposes a few utility functions. (See later comment also)

packages/app/src/components/time-series-chart/components/tooltip/tooltip-series-list.tsx
```diff
@@ -31,43 +30,12 @@ export function TooltipSeriesList<T extends TimestampedValue>({
...
-
-    return isPresent(numberValue)
-      ? isPercentage
-        ? `${formatPercentage(numberValue)}%`
-        : formatNumber(numberValue)
-      : '-';
-  }
+  const getValueString = useGetValueString();
+  const getRangeValueString = useGetRangeValueString();
+  const getDateStringFromValue = useGetDateStringFromValue();
```
VWSCoronaDashboard8, line 38: I think it would be a little cleaner if you combined these into one hook which returns the three function together like `useStringFormatting `

packages/app/src/components/time-series-chart/components/tooltip/tooltip-series-list.tsx
```diff
@@ -96,9 +64,8 @@ export function TooltipSeriesList<T extends TimestampedValue>({
                   displayTooltipValueOnly={displayTooltipValueOnly}
                 >
                   <b>
-                    {getValueStringForKey(
-                      value,
-                      x.metricProperty,
+                    {getValueString(
```
VWSCoronaDashboard8, line 67: If you typed the function using generics passed to the hook the same way as the original code, you wouldn't need to cast it here. I think that would be cleaner, or is there a specific reason you choose to not do this?

Something like this

`const {getValueString, getRangeValueString, getDateStringFromValue} = useStringFormatting<T>()`

packages/app/src/components/time-series-chart/components/tooltip/types.ts
```diff
@@ -42,6 +42,13 @@ export type TooltipData<T extends TimestampedValue> = {
    * Configuration to display only a value in the tooltip
    */
   displayTooltipValueOnly?: boolean;
+
+  /**
+   * Configure the minimum width of the value element. Example use case is
+   * assigning a width that is large enough to hold the largest value so the
+   * tooltip content does not change in width due to differeing value lengths
```
VWSCoronaDashboard8, line 49: nit; typo

packages/app/src/components/time-series-chart/components/tooltip/formatStrings.ts
VWSCoronaDashboard13: yeah, I can see how it doesn't make sense for this use case. I'll remove them 👍

packages/app/src/components/time-series-chart/time-series-chart.tsx
```diff
@@ -21,6 +21,8 @@ import {
   Tooltip,
   TooltipData,
   TooltipFormatter,
+  useGetRangeValueString,
+  useGetValueString,
```
VWSCoronaDashboard8, line 25: Since these hooks are located in the tooltip subfolder but are used here in the root component I'd suggest to move them to `time-series-chart/logic` or `time-series-chart/hooks` or something similar.

packages/app/src/components/time-series-chart/time-series-chart.tsx
```diff
@@ -210,6 +213,37 @@ export function TimeSeriesChart<
...
+                isPercentage
+              ).length;
+          }
+        })
+      );
+    });
+
+    return `${Math.max(...valueLengths)}ch`;
+  }, [values, seriesConfig, getValueString, getRangeValueString, isPercentage]);
+
```
VWSCoronaDashboard8, line 246: It seems unnecessary to expose all this code in the root component. I think this is a perfect candidate for a `useValueWidth()` hook.

VWSCoronaDashboard8 (changes requested):

packages/app/src/components/time-series-chart/components/tooltip/formatStrings.ts
VWSCoronaDashboard13: oops sorry!

packages/app/src/components/time-series-chart/components/tooltip/tooltip-series-list.tsx
VWSCoronaDashboard13: that would be much cleaner 👍

packages/app/src/components/time-series-chart/components/tooltip/tooltip-series-list.tsx
VWSCoronaDashboard13: I just thought the functions didn't need to be aware of the specific types used for the graph since they are just doing string formatting so it would be a bit cleaner to make them dumber so we could reduce the params as well. I can change it back to how it used to be if that's preferred though

VWSCoronaDashboard1 (approved):

packages/app/src/components/time-series-chart/components/tooltip/tooltip-series-list.tsx
VWSCoronaDashboard8: I agree that the functions should not have to care about the type. Unfortunately the value object is not typed accurately as an escape hatch for other code and therefor the casting is required. With all the casting going on it looks quite messy, but yeah the problem should ideally be fixed elsewhere. 

So I'm fine if you leave it like this. We'll could look into improving the typing of `value` as a separate task

VWSCoronaDashboard8 (approved):

packages/app/src/components/time-series-chart/logic/use-value-width.ts
```diff
@@ -0,0 +1,42 @@
...
+                isPercentage
+              ).length;
+            case 'range':
+              return getRangeValueString(
+                (value[config.metricPropertyLow] as unknown) as number | null,
+                (value[config.metricPropertyHigh] as unknown) as number | null,
+                isPercentage
+              ).length;
+          }
+        })
```
VWSCoronaDashboard10, line 34: I think we could create a formatter function which does exactly this selected range ☝️ .
The formatSeriesValue can be used in this useValueWidth-hook and also in the tooltip series list to prevent all the type castings. Extra advantage is that the output in the DOM and the calculations of the longest value use the same output.

```tsx
  function formatSeriesValue(
    value: T,
    config: SeriesConfig<T>[number],
    isPercentage?: boolean
  ) {
    switch (config.type) {
      case 'line':
      case 'area':
      case 'bar':
      case 'stacked-area':
      case 'invisible':
        return getValueString(
          (value[config.metricProperty] as unknown) as number | null,
          isPercentage
        );
      case 'range':
        return getRangeValueString(
          (value[config.metricPropertyLow] as unknown) as number | null,
          (value[config.metricPropertyHigh] as unknown) as number | null,
          isPercentage
        );
    }
  }
```

packages/app/src/components/time-series-chart/components/tooltip/tooltip-series-list.tsx
VWSCoronaDashboard10: See my comment on `use-value-width.tsx` for a prettier solution, it would boil down to this call:
```tsx
{formatSeriesValue(value, x, options.isPercentage)}

File: packages/app/src/components/time-series-chart/logic/use-value-width.ts (hunk 8 above)
Comment (VWSCoronaDashboard13) [reply packages/app/src/components/time-series-chart/logic/use-value-width.ts]: this is great suggestion 💯

File: packages/app/src/components/time-series-chart/logic/use-string-formatting.ts [hunk 9]
```diff
@@ -0,0 +1,78 @@
...
+          'axis'
+        );
+        const dateEndString = intl.formatDateFromSeconds(
+          value.date_end_unix,
+          'axis'
+        );
+
+        return `${dateStartString} - ${dateEndString}`;
+      }
+    };
```
VWSCoronaDashboard10, line 71: nit; I think this date string formatter is very specific and it's also only called once. I'd suggest moving it back to its original location. The `use-string-formatting` could be renamed to the single formatter function it currently is..

```tsx
// use-format-series-values.tsx

export function useFormatSeriesValue( value, config, isPercentage) { ... }

File: packages/app/src/components/time-series-chart/logic/use-value-width.ts [hunk 10]
```diff
@@ -0,0 +1,26 @@
...
+  const { formatSeriesValue } = useStringFormatting();
+  const valueMaxWidth = useMemo(() => {
+    const valueLengths: number[] = [];
+    seriesConfig.forEach((config) => {
+      return valueLengths.push(
+        ...values.map(
+          (value) => formatSeriesValue(value, config, isPercentage).length
+        )
+      );
+    });
```
VWSCoronaDashboard10, line 20: nit; not sure the impact on performance, but this can create a very long array. We could call Math.max more often;
```suggestion
    const valueLengths: number[] = [];
    seriesConfig.forEach((config) => {
      return valueLengths.push(
        Math.max(...values.map(
          (value) => formatSeriesValue(value, config, isPercentage).length
        ))
      );
    });
```

VWSCoronaDashboard10 (approved):
looking good,  couple of nitpicks left but you have my approval.

PR #2758, opened 2021-05-03
Add actions sync after feature

VWSCoronaDashboard8:
## Summary

This adds a github action to trigger the cms `sync-after-feature` script whenever a feature branch is merged into develop. It triggers on develop push, there is no merge event. Since we do not commit to develop directly, and it also can't hurt to run this script on "non-feature" commits, I think this is fine.


VWSCoronaDashboard1 (approved):

.github/workflows/sync-sanity-lokalize.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Cancel previous workflow
+        uses: styfle/cancel-workflow-action@0.4.0
+        with:
+          access_token: ${{ github.token }}
+      # The bahmutov/npm-install uses caching, so by running it already in the
+      # init task, the other tasks running in parallel should install quickly.
+      - name: Check out repository
+        uses: actions/checkout@v2
+      - name: Install dependencies
+        uses: bahmutov/npm-install@v1
```
VWSCoronaDashboard10, line 21: you can get rid of this part, doesn't do anything useful
```suggestion
```

VWSCoronaDashboard10 (approved):

.github/workflows/sync-sanity-lokalize.yml
VWSCoronaDashboard8: Yeah thought about it, but somehow it felt good to have the init stap in case we want to extend the script later. But maybe that's not reason enough. I'll take it out.

VWSCoronaDashboard10: The `Cancel previous workflow` does make sense though I think .. you removed it all now?

PR #2773, opened 2021-05-03
Add custom validation for moving averages

VWSCoronaDashboard1:
## Summary

This PR adds a custom validation for the moving averages.

## Motivation

Feature request

## Detailed design

The custom validation function `validateMovingAverages` loops through all the metrics in GM and VR and looks for collections that contain objects with property names suffixed by '_moving_average'.
These collections are checked to make sure that the first six entries of these specific metrics have a value of `NULL` _and_ that there are no non-consecutive null values in the rest of the collection.

### Governance

- [ ] Documentation is added

### DIFF
--- a/packages/cli/src/schema/custom-validations/moving-averages.ts
+++ b/packages/cli/src/schema/custom-validations/moving-averages.ts
@@ -1,15 +1,18 @@
 import set from 'lodash/set';
 import { isDefined } from 'ts-is-present';
+import { JSONValue } from './types';
 
 function hasValuesProperty(
-  input: [string, { values?: Record<string, unknown>[] }]
-): input is [string, { values: Record<string, unknown>[] }] {
-  return input[1].hasOwnProperty('values');
+  input: [string, JSONValue]
+): input is [string, { values: Record<string, JSONValue>[] }] {
+  return (
+    input[1] !== null &&
+    typeof input[1] === 'object' &&
+    input[1].hasOwnProperty('values')
+  );
 }
 
-export function validateMovingAverages(
-  input: Record<string, { values?: Record<string, unknown>[] }>
-) {
+export function validateMovingAverages(input: Record<string, JSONValue>) {
   const result = Object.entries(input)
     // first filter out all the non-metric properties (we only want the ones with a values collection)
     .filter(hasValuesProperty)
--- a/packages/cli/src/schema/custom-validations/moving-averages.ts
+++ b/packages/cli/src/schema/custom-validations/moving-averages.ts
@@ -1,3 +1,4 @@
+import set from 'lodash/set';
 import { isDefined } from 'ts-is-present';
 
 export function validateMovingAverages(input: Record<string, any>) {
@@ -10,10 +11,10 @@ export function validateMovingAverages(input: Record<string, any>) {
     .map(([propertyName, value]) => ({
       [propertyName]: value.values.map((x: Record<string, unknown>) => {
         const movingAverageProperties = findMovingAverages(x);
-        return movingAverageProperties.reduce((aggr, p) => {
-          aggr[p] = x[p];
-          return aggr;
-        }, {} as Record<string, unknown>);
+        return movingAverageProperties.reduce(
+          (aggr, p) => set(aggr, p, x[p]),
+          {} as Record<string, unknown>
+        );
       }),
     }))
     // Perform the validations on those moving averages
@@ -27,11 +28,9 @@ export function validateMovingAverages(input: Record<string, any>) {
             ' and '
           )}`;
         }
-        const propsWithNonConsecutiveNullValues = hasNonConsecutiveNullValues(
-          values
-        );
-        if (propsWithNonConsecutiveNullValues.length) {
-          return `${propertyName} has non consecutive NULL values in metric ${propsWithNonConsecutiveNullValues.join(
+        const propsWithNonConsecutiveValues = hasNonConsecutiveValues(values);
+        if (propsWithNonConsecutiveValues.length) {
+          return `${propertyName} has non consecutive values in metric ${propsWithNonConsecutiveValues.join(
             ' and '
           )}`;
         }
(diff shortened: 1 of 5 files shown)


packages/cli/src/schema/custom-validations/moving-averages.ts
```diff
@@ -0,0 +1,97 @@
...
+    if (currentValue === null) {
+      while (currentValue === null) {
+        currentValue = collection[++i]?.[propertyName];
+      }
+      if (isDefined(currentValue) && i < collection.length) {
+        return true;
+      }
+    }
+  }
+  return false;
```
VWSCoronaDashboard8, line 75: This looks overly complicated with nested loops and multiple currentValue assignments.

I suspect all you need one loop and keep a boolean to record if a null value has already been found after a normal value, then the second time that happens (and the boolean was already true) you can exit the loop via a break statement.

But as I mentioned earlier I think you might be testing the wrong thing, because it's null values should be allowed to be non-consecutive if they appear and the beginning and end of the series. Or maybe it's just the wording that is confusing me.

packages/cli/src/schema/custom-validations/moving-averages.ts
```diff
@@ -0,0 +1,97 @@
+import { isDefined } from 'ts-is-present';
+
+export function validateMovingAverages(input: Record<string, any>) {
+  const result = Object.entries(input)
+    // first filter out all the non-metric properties (we only want the ones with a values collection)
+    .filter(([_p, value]) => value.hasOwnProperty('values'))
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: Yea, but tell me how to type this exactly. You'll end up with an incredibly nested record that is no longer readable. Because the any in the parameter record cannot be an unknown, nor is it a record of itself, it is potentially a string or a number or an array. So, while I, sort of, agree with your nitpick here, I went for any out of practical reasons.

packages/cli/src/schema/custom-validations/moving-averages.ts
VWSCoronaDashboard1: fair enough, I misunderstood the original requirements. Since our chat I also used the term 'non-consecutive nulls'.

packages/cli/src/schema/custom-validations/moving-averages.ts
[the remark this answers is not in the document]
VWSCoronaDashboard10: I guess this works:
```
input: Record<string, { values?: Record<string, unknown>[] }
```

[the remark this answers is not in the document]
VWSCoronaDashboard1: well, sort of...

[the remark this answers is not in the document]
VWSCoronaDashboard1: Alright, just had to add a typeguard to further narrow the type

packages/cli/src/schema/custom-validations/moving-averages.ts
```diff
@@ -0,0 +1,94 @@
...
+) {
+  const values = collection.slice(6).map((x) => x[propertyName]);
+  let lastValue = values.pop();
+
+  while (lastValue === null && values.length) {
+    lastValue = values.pop();
+  }
+
+  return values.indexOf(null) > -1;
+}
```
VWSCoronaDashboard8, line 76: It looks like `while (lastValue === null` will not be true if the 7th value is non-null? And then the loop will never be executed.

I think it's unnecessary to mutate and iterate the full array. You can use an early-return statement.

```suggestion
function hasNonConsecutiveValuesInMetric(
  propertyName: string,
  collection: Record<string, unknown>[]
) {
  const values = collection.slice(6).map((x) => x[propertyName]);
  let lastValue;

  for (const value of values) {
     if(lastValue === null && value !== null) {
        return true
     }
  
    lastValue = value
   }
   return false
}
```

VWSCoronaDashboard1: Your assumption is incorrect, but your code suggestion simplifies things even more, so I've made the update

packages/cli/src/schema/custom-validations/moving-averages.ts
```diff
@@ -0,0 +1,101 @@
...
+import { JSONValue } from './types';
+
+function hasValuesProperty(
+  input: [string, JSONValue]
+): input is [string, { values: Record<string, JSONValue>[] }] {
+  return (
+    input[1] !== null &&
+    typeof input[1] === 'object' &&
+    input[1].hasOwnProperty('values')
+  );
```
VWSCoronaDashboard8, line 12: nit; if you're only using one value of the array, you could:
- pass in a single value by doing [1] at the calling context, which would simplify this function.
- take the first value at the top of the function, so that you don't need to write [1] three times.

packages/cli/src/schema/custom-validations/moving-averages.ts
```diff
@@ -0,0 +1,101 @@
...
+        return undefined;
+      })
+    )
+    .flat()
+    .filter(isDefined);
+
+  return result.length ? result : undefined;
+}
+
+function hasNonConsecutiveValues(values: Record<string, unknown>[]) {
```
VWSCoronaDashboard8, line 57: nit; I would expect a function with a name like this to return a boolean. Something like this would be clearer I think:

`getPropertiesWithNonConsecutiveValues`

packages/cli/src/schema/custom-validations/moving-averages.ts
```diff
@@ -0,0 +1,101 @@
...
+    .filter(isDefined);
+}
+
+function hasMovingAverages(input: Record<string, unknown>) {
+  return findMovingAverages(input).length > 0;
+}
+
+function findMovingAverages(input: Record<string, unknown>) {
+  return Object.keys(input).filter((x) => x.endsWith('_moving_average'));
+}
```
VWSCoronaDashboard8, line 101: nit; you could inline this function in `hasMovingAverages ` and then it would still be as readable I think.

VWSCoronaDashboard8 (approved):

packages/cli/src/schema/custom-validations/moving-averages.ts
VWSCoronaDashboard8: Also it's unclear to me why you're taking the second item in the array? Why not [0]?

VWSCoronaDashboard1: Because the first item is the property name and the second one the property value

VWSCoronaDashboard1: The filter is based on `Object.entries(input)`

packages/cli/src/schema/custom-validations/moving-averages.ts
```diff
@@ -0,0 +1,108 @@
...
+  return values
+    .slice(0, 5)
+    .map((x) =>
+      Object.entries(x).map(([propertyName, value]) =>
+        value !== null ? propertyName : undefined
+      )
+    )
+    .flat()
+    .filter((x, index, arr) => index === arr.indexOf(x))
+    .filter(isDefined);
```
VWSCoronaDashboard10, line 99: nit
```suggestion
  return values
    .slice(0, 5)
    .flatMap((x) =>
      Object.entries(x).map(([propertyName, value]) =>
        value !== null ? propertyName : undefined
      )
    )
    .filter((x, index, arr) => isDefined(x) && index === arr.indexOf(x))
```

VWSCoronaDashboard10 (approved):

PR #2767, opened 2021-05-03
Update in schema nursing_home__deceased_daily_moving_average

VWSCoronaDashboard14:

PR #2766, opened 2021-05-03
Update in schema nursing_home__newly_infected_people_moving_average

VWSCoronaDashboard14:

PR #2776, opened 2021-05-03
Update in schema -- nursing home

VWSCoronaDashboard10:
This PR is a combination of #2766 and #2767.


packages/app/schema/nl/__difference.json
```diff
@@ -45,13 +45,13 @@
     "sewer__average": {
       "$ref": "#/definitions/diff_decimal"
     },
-    "nursing_home__newly_infected_people": {
+    "nursing_home__newly_infected_people_moving_average": {
```
VWSCoronaDashboard8, line 48: I think because the source of these differences is a decimal value, the difference type should also be decimal.

VWSCoronaDashboard8 (changes requested):

VWSCoronaDashboard8 (approved):

PR #2781, opened 2021-05-04
Add option to validate a single metric in a data file

VWSCoronaDashboard1:
## Summary

Previously `validate-single-json <schema-name> <json-filename>` could be used to validate one single data file instead of the entire list.
An extra optional command line option has been added that allows to have only ONE metric within the specified datafile to be validated: `validate-single-json <schema-name> <json-filename> <metric-name>`.

## Motivation

Feature request.

## Detailed design

If a metric name is specified the script now first loads the specified schema file, removes all of the required properties and leaves only the specified metric name in that list. It then sets `additionalProperties` to `true` and removes all the property definitions except for the one describing the metric name.

This way all of the other properties are ignored and only the specified metric will be validated.

### Governance

- [ ] Documentation is added

### DIFF
--- a/packages/cli/src/schema/custom-validations/types.ts
+++ b/packages/cli/src/schema/custom-validations/types.ts
@@ -6,6 +6,8 @@ export type JSONValue =
   | JSONValue[]
   | { [key: string]: JSONValue };
 
+export type JSONType = Record<string, JSONValue>;
+
 export type CustomValidationFunction = (
-  input: Record<string, JSONValue>
+  input: JSONType
 ) => string[] | undefined;
--- a/packages/cli/src/schema/custom-validations/types.ts
+++ b/packages/cli/src/schema/custom-validations/types.ts
@@ -6,8 +6,8 @@ export type JSONValue =
   | JSONValue[]
   | { [key: string]: JSONValue };
 
-export type JSONType = Record<string, JSONValue>;
+export type JSONObject = Record<string, JSONValue>;
 
 export type CustomValidationFunction = (
-  input: JSONType
+  input: JSONObject
 ) => string[] | undefined;
--- a/packages/cli/src/scripts/validate-json-single.ts
+++ b/packages/cli/src/scripts/validate-json-single.ts
@@ -1,4 +1,3 @@
-import chalk from 'chalk';
 import fs from 'fs';
 import meow from 'meow';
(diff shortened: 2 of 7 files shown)


packages/cli/src/scripts/validate-json-single.ts
```diff
@@ -2,45 +2,79 @@ import chalk from 'chalk';
...
+ */
+function loadStrippedSchema(metricName: string, basePath: string) {
+  const strippedSchema = loadRootSchema(path.join(basePath, `__index.json`));
+
+  if (!isDefined(strippedSchema.properties[metricName])) {
+    console.error(
+      chalk.bgRed.bold(
+        `  ${metricName} is not a metric in the specified schema '${schemaName}'  \n`
+      )
+    );
```
VWSCoronaDashboard8, line 31: nit; there's a logError helper function in cli utils for this

VWSCoronaDashboard8 (approved):

packages/cli/src/scripts/validate-json-single.ts
```diff
@@ -60,25 +94,29 @@ if (!schemaInformation[schemaName].files.includes(jsonFileName)) {
...
+createValidateFunction(rootSchema, schemaBasePath).then((validateFunction) => {
+  const fileName = path.join(jsonBasePath, jsonFileName);
   const schemaInfo = schemaInformation[schemaName];
+
   const contentAsString = fs.readFileSync(fileName, {
     encoding: 'utf8',
   });
 
-  let data: any = null;
+  let jsonData: any = null;
```
VWSCoronaDashboard8, line 117: I guess you can use the new JSONType for this. Don't know if it'll make any difference

packages/cli/src/schema/custom-validations/types.ts
```diff
@@ -6,6 +6,8 @@ export type JSONValue =
   | JSONValue[]
   | { [key: string]: JSONValue };
 
+export type JSONType = Record<string, JSONValue>;
```
VWSCoronaDashboard8, line 9: You might want to call this `JSONObject` instead. I don't see how JSONType is a clear name for something that must be an object, since a JSON file can also contain an array at the root level.

VWSCoronaDashboard10 (approved):

PR #2786, opened 2021-05-05
Add moving averages in difference components

VWSCoronaDashboard14:
Can't test this properly without the data, have to check again when it's available

- Add new text to the homepage that indicates the moving average
- an option to add a moving average to a KPI bar scale (what would change the text)
- Sidebar differences has been changed for certain metrics
- New keys in Lokalise


packages/app/src/components/difference-indicator.tsx
```diff
@@ -14,13 +14,15 @@ import IconGelijk from '~/assets/gelijk.svg';
...
 import { useIntl } from '~/intl';
+import { InlineText } from '~/components/typography';
 
 interface DifferenceIndicatorProps {
   value: DifferenceDecimal | DifferenceInteger;
   isDecimal?: boolean;
   context?: 'sidebar' | 'tile' | 'inline';
   maximumFractionDigits?: number;
   staticTimespan?: string;
+  absoluteMovingAverage?: number;
```
VWSCoronaDashboard8, line 25: I don't understand this. The difference indicator is able to show the difference based on value. It can be any number and ideally it should be agnostic to what that number means. 

Why do we need an extra property to pass in a moving average? 

And what does it mean that this value is absolute? Isn't a difference indicator by definition relative, because it is displaying a value in relation to another?

packages/app/src/components/difference-indicator.tsx
```diff
@@ -159,13 +163,29 @@ function TileIndicator({
...
+          </IconContainer>
+          <InlineText fontWeight="bold">
+            {formatNumber(value.difference)} {siteText.toe_en_afname.hoger}{' '}
+          </InlineText>
+          <InlineText>
+            {siteText.toe_en_afname.zeven_daags_gemiddelde}
+            {` (${formatNumber(absoluteMovingAverage)})`}
+          </InlineText>
+        </Container>
+      );
```
VWSCoronaDashboard8, line 180: I don't think this is the way to go. If you need a special component for a difference indicator that is substantially different from the original one, I suggest to fork it and name it something like `MovingAverageDifferenceIndicator`.

But the original indicator was just about the graphic, and did not include text. So I doubt if it is even a good idea to create an indicator that has text inlined. Can't we just handle the text outside the indicator like we do with the others?

packages/app/src/pages/landelijk/ziekenhuis-opnames.tsx
```diff
@@ -126,7 +126,7 @@ const IntakeHospital = (props: StaticProps<typeof getStaticProps>) => {
                 metricName="hospital_nice"
                 metricProperty="admissions_on_date_of_reporting"
                 localeTextKey="ziekenhuisopnames_per_dag"
-                differenceKey="hospital_nice__admissions_on_date_of_reporting"
+                differenceKey="hospital_nice__admissions_on_date_of_reporting_moving_average_moving_average"
```
VWSCoronaDashboard8, line 129: This seems like a typo

packages/app/src/pages/veiligheidsregio/[code]/positief-geteste-mensen.tsx
```diff
@@ -285,7 +296,11 @@ const PositivelyTestedPeople = (props: StaticProps<typeof getStaticProps>) => {
               <KpiValue
                 percentage={dataGgdAverageLastValue.infected_percentage}
                 difference={
-                  data.difference.tested_ggd_average__infected_percentage
+                  data.difference
+                    .tested_ggd_average__infected_percentage_moving_average
+                }
+                absoluteMovingAverage={
```
Comment (VWSCoronaDashboard8) [inline packages/app/src/pages/veiligheidsregio/[code]/positief-geteste-mensen.tsx:302]: Ah I get the naming now. You want to pass in the current value to show in the display. Then in the context of the difference indicator it doesn't care if this value is  related to moving average. It is related to the difference value no matter what that is.

So I would suggest to call it `currentValue` or `currentAbsoluteValue`, because `current` is a more important distinction than `moving average`

VWSCoronaDashboard8 (changes requested):

packages/app/src/components/difference-indicator.tsx
VWSCoronaDashboard8: I get it now, the name is just not ok see later comments

packages/app/src/components/difference-indicator.tsx
VWSCoronaDashboard8: This could be ok if you pass in the value as "current" to make it generic, but then I think you also need to pass in the text to keep it generic, since you can't tie this component to specific moving average text. 

If possible I would create a separate component to display the current value + text for the moving average, and render that inline next to this component, because this seems to violate the "separation of concerns" principle.

packages/app/src/pages/veiligheidsregio/[code]/positief-geteste-mensen.tsx
Comment (VWSCoronaDashboard14) [reply packages/app/src/pages/veiligheidsregio/[code]/positief-geteste-mensen.tsx:302]: Thanks for the feedback! I wasn't totally done yet but seems like I have to shuffle a couple of things. Yeah the difference text needs a couple of extra metrics now see screenshot below.

Screenshot 2021-05-06 at 14 41 55

Comment (VWSCoronaDashboard8) [reply packages/app/src/pages/veiligheidsregio/[code]/positief-geteste-mensen.tsx:302]: 👍 Looks good. If you're not done yet maybe make it a draft PR ;)

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard4 (approved):

PR #2788, opened 2021-05-05
Fix sewer chart left axes

VWSCoronaDashboard14:
The old implementation was dividing the highest number to use as the second tick but that didn't work when the first tick didn't start at 0.  So added a small calculation that finds the tick needed to be in the middle.

### DIFF
--- a/packages/app/src/components/sewer-chart/sewer-chart.tsx
+++ b/packages/app/src/components/sewer-chart/sewer-chart.tsx
@@ -171,9 +171,9 @@ export function SewerChart(props: SewerChartProps) {
 
   /**
    * Manually set Y axis values, as per design.
-   * The middle calculation finds first the difference between the higehst
-   * and the lowest, then divide it by 2 and add the lowest value
-   * to find value exact value in between.
+   * The middle calculation finds first the difference between the highest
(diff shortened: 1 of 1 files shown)


packages/app/src/components/sewer-chart/sewer-chart.tsx
```diff
@@ -328,10 +332,12 @@ export function SewerChart(props: SewerChartProps) {
             <AxisLeft
               scale={scales.yScale}
               tickValues={tickValuesY}
+              numTicks={tickValuesY.length}
+              tickFormat={(x) => formatNumber(x as number)}
               hideTicks
               hideAxisLine
               stroke={colors.data.axis}
-              tickFormat={(x) => formatNumber(x as number)}
+              // tickFormat={(x) => formatNumber(x as number)}
```
VWSCoronaDashboard4, line 340: Was this intended? Or should it be reverted / deleted?

VWSCoronaDashboard4 (approved):

packages/app/src/components/sewer-chart/sewer-chart.tsx
VWSCoronaDashboard14: Oops should be deleted, thanks for noticing.

PR #2789, opened 2021-05-05
FAQ and Cijferverantwoording schema refactor and migration

VWSCoronaDashboard1:
## Summary

The items for the FAQ and Cijferverantwoording schema's have been refactored into separate documents. This way the communication department can publish these items separately.
Two migrations have been added to implement these changes.

## Motivation

Feature request.

## Detailed design

Separate documents have been defined for faq questions and 'figure explanation' items.
The migrations retrieve the current question and item objects and save them as documents, the resulting ids are then placed as references in the specific pages.

Also, as part of this PR all of the schema documents and objects have been turned into named exports, placed in barrel files and these barrel imports are being used in the schema.ts to tie everything together. Unused document types and objects were removed.

Lastly, the sanity dependencies have been upgraded to the latest versions.

### Governance

- [ ] Documentation is added


packages/cms/src/desk-structure.ts
```diff
@@ -65,21 +66,36 @@ export default () =>
...
+        .child(
+          S.list()
+            .title('Groepen en Vragen')
+            .items([
               addListItem(
                 MdQuestionAnswer,
-                'Veelgestelde vragen',
-                'veelgesteldeVragen'
+                'CijferVerantwoording pagina',
+                'cijferVerantwoording'
```
VWSCoronaDashboard8, line 92: Just wondering; Are these three type names all with different formats? 🤔

packages/cms/src/migrations/sprint22/faq.ts
```diff
@@ -0,0 +1,80 @@
...
+      .map((patch: any) => `${patch.id} => ${JSON.stringify(patch.patch)}`)
+      .join('\n')
+  );
+  const transaction = createTransaction(patches);
+
+  await commitTransaction(transaction);
+
+  return migrateNextBatch();
+};
+
```
VWSCoronaDashboard8, line 76: nit; The functions in this file are all anonymous function expressions assigned to named const. In the rest of the codebase we use named function declarations, which has my preference for readability and other reasons. 

Arrow / lambda functions are useful for inline expressions and closures, but I see no reason to use them.

The very small functions like `const commitTransaction = (tx: any) => tx.commit();` are better written inline I think, because ` tx.commit()` is more clear then `commitTransaction(tx)` since you're not completely sure what the function does when you read that.

packages/cms/src/migrations/sprint22/faq.ts
```diff
@@ -0,0 +1,80 @@
...
+migrateNextBatch().catch((err: any) => {
```
VWSCoronaDashboard8, line 77: I think you can expect error to be of type Error? Unless you're dealing with shitty libraries that throw strings or other things that are considered bad practice.

packages/cms/src/migrations/sprint22/figure-explanation.ts
```diff
@@ -0,0 +1,84 @@
+import { getClient } from '../../client';
+
+const client = getClient();
+
+const fetchVerantwoording = () =>
+  client.fetch(`*[_type == 'cijferVerantwoording']`);
```
VWSCoronaDashboard8, line 6: Unless used more than once, I don't see how this functions helps. I think it only obfuscates things. 

It seems to be only used once...

packages/cms/src/migrations/sprint22/figure-explanation.ts
```diff
@@ -0,0 +1,84 @@
...
+        ifRevisionID: doc._rev,
+      },
+    }))
+    .filter((x) => x !== undefined);
+};
+
+const createTransaction = (patches: any[]) =>
+  patches.reduce(
+    (tx, patch) => tx.patch(patch.id, patch.patch),
+    client.transaction()
```
VWSCoronaDashboard8, line 30: These functions seem to be repeated. I think they should be generic and moved to a shared location to make it DRY

packages/cms/src/schemas/schema.ts
```diff
@@ -55,47 +17,12 @@ export default createSchema({
...
-    titleDescriptionBlock,
-    decoratedLink,
+    ...Object.values(allObjects),
 
-    /* LOCALE HELPERS */
-    localeString,
-    localeBlock,
-    localeRichContentBlock,
-    localeText,
+    ...Object.values(localeHelpers),
```
VWSCoronaDashboard8, line 26: Nice cleanup! 👍

packages/cms/src/schemas/schema.ts
```diff
@@ -6,47 +6,9 @@ import createSchema from 'part:@sanity/base/schema-creator';
...
-import lockdown from './restrictions/lockdown';
-import restriction from './restrictions/restriction';
-import restrictionCategory from './restrictions/restriction-category';
-// routekaart en maatregelen
-import restrictionGroup from './restrictions/restriction-group';
-import restrictionGroupLockdown from './restrictions/restriction-group-lockdown';
-import roadmap from './restrictions/roadmap';
+import * as localeHelpers from './locale';
+import * as allObjects from './objects';
+import * as restrictionDocuments from './restrictions';
```
VWSCoronaDashboard8, line 11: nit; I think the names of the folders are sufficient, as the concat below doesn't care. It prevents you from coming up with new names for things. So you could write 

```suggestion
import * as locale from './locale';
import * as objects from './objects';
import * as restrictions from './restrictions';
```
Then also rename allDocuments to documents I guess. Currently it's a little odd that there's both allDocuments and restrictionDocuments...

VWSCoronaDashboard8 (approved):

packages/cms/src/desk-structure.ts
VWSCoronaDashboard1: One is a name, the other is a title.

packages/cms/src/migrations/sprint22/faq.ts
VWSCoronaDashboard1: I used the previous migration as a starting off point, that migration was based on an example from the Sanity docs.
But sure, I can rename them...

packages/cms/src/migrations/sprint22/figure-explanation.ts
VWSCoronaDashboard1: To be honest, these functions only get run twice. Once to perform the migration on develop and once on production. I'm not sure its worth the effort to start refactoring things this deeply.
Unless we'll need these functions exactly like this in future migrations, which is hard to anticipate.

VWSCoronaDashboard8 (approved):

packages/cms/src/schemas/documents/pages/cijfer-verantwoording-page.ts
```diff
@@ -1,4 +1,4 @@
-export default {
+export const cijferVerantwoording = {
```
VWSCoronaDashboard4, line 1: why are all the others in English and this one in Dutch?

packages/cms/src/schemas/documents/pages/over-risico-niveaus-page.ts
```diff
@@ -1,4 +1,4 @@
-export default {
+export const overRisicoNiveaus = {
```
VWSCoronaDashboard4, line 1: risiconiveaus is one word, so lowercase `n` for `niveaus`

VWSCoronaDashboard4 (changes requested):

packages/cms/src/schemas/documents/pages/cijfer-verantwoording-page.ts
VWSCoronaDashboard1: It's named after the schema, I thought it would become confusing if I started mixing the two. Renaming the schema involves a migration, so I think that needs to be moved to a separate ticket.

VWSCoronaDashboard10 (approved):

PR #2794, opened 2021-05-06
Add keyboard interaction to time series charts

VWSCoronaDashboard10:


packages/app/src/components/time-series-chart/logic/hover-state.ts
```diff
@@ -112,13 +111,37 @@ export function useHoverState<T extends TimestampedValue>({
     [xScale, valuesDateUnix]
   );
 
+  const setPointThrottled = useMemo(
```
VWSCoronaDashboard8, line 114: It seems odd to return a callback function from useMemo. Wouldn't useCallback be a better choice? The data processing functions within this throttle will be called anyway for each setPoint call, no?

packages/app/src/components/time-series-chart/logic/use-keyboad-index.ts
```diff
@@ -0,0 +1,72 @@
+import { Dispatch, SetStateAction, useMemo, useState } from 'react';
+import { useHotkey } from '~/utils/hotkey/use-hotkey';
+
+export function useKeyboardIndex(
```
VWSCoronaDashboard8, line 4: This name is a little vague to me. Sounds like the index somehow has to do with multiple keyboards.

Maybe `useKeyboardNavigation` or `useIterationByKeyboard` would work? And instead of `index` possibly use `position`?

packages/app/src/components/time-series-chart/logic/use-keyboad-index.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  return useMemo(
+    () => ({
+      enable: () => setIsEnabled(true),
+      disable: () => setIsEnabled(false),
+    }),
+    []
+  );
+
+  // return [isEnabled ? index : undefined, modifiers] as const;
```
VWSCoronaDashboard8, line 67: comment

packages/app/src/components/time-series-chart/logic/use-keyboad-index.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+function rotate(maxIndex: number, index: number) {
```
VWSCoronaDashboard8, line 70: For me a rotate function would take a value and a rotation and the apply that rotation to the value. That's not what is happening here, since index will not change if it falls within the range.

I would call this something like `wrap` or `wrapAround`. Also it's more common to pass the number you're wrapping first and then the size.

VWSCoronaDashboard8: Also it seems like maxIndex is not really the maximum allowed index, because if you pass in that value, the function will return 0. So it seems like maxIndex is more like array "length" in this case where maxIndex would be length - 1

packages/app/src/components/time-series-chart/components/tooltip/tooltip-series-list.tsx
```diff
@@ -86,11 +90,19 @@ export function TooltipSeriesList<T extends TimestampedValue>({
       )}
       <TooltipList hasTwoColumns={hasTwoColumns}>
         {seriesConfig.map((x, index) => {
+          /**
+           * The key is unique for every date to make sure a screenreader
+           * will read `[label]: [value]`. Otherwise it would read the
+           * changed content which would only be `[value]` and thus miss a some
```
VWSCoronaDashboard1, line 96:
```suggestion
           * changed content which would only be `[value]` and thus miss some
```

packages/app/src/components/time-series-chart/logic/use-keyboad-index.ts
VWSCoronaDashboard1: Perhaps 'cycle' is an appropriate term here? or 'roundRobin'?

VWSCoronaDashboard1 (approved):

packages/app/src/components/time-series-chart/logic/hover-state.ts
VWSCoronaDashboard10: No, there's an actual difference here. The function passed into `useMemo` is triggered only when it detects changes in the dep array. When this function is called it'll return a throttled function which will be assigned to the `setPointThrottled` variable.

We could also create a throttled function within a useCallback hook but that would call `throttle()` on every render and simply ignore its result when the dep array didn't change. It works, but results in unnecessary function invocations.

VWSCoronaDashboard10: We could call this micro-optimization though 😛

VWSCoronaDashboard1 (approved):

packages/app/src/components/time-series-chart/logic/hover-state.ts
VWSCoronaDashboard8: Haha yeah I'm pretty sure the overhead of function invocation in Javascript can be fully neglected.

VWSCoronaDashboard8 (approved):

PR #2795, opened 2021-05-06
Design update vaccine clock

VWSCoronaDashboard4:
## Summary

Design update vaccine clock

Screenshot vaccine clock

### DIFF
--- a/packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
+++ b/packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
@@ -13,33 +13,60 @@ const RADIUS = 55;
 const CONTAINER_WIDTH = RADIUS * 2 + 12;
 const CONTAINER_HEIGHT = RADIUS * 2 + 12;
 
+const MINUTE = 60;
+const FRAMES_PER_MINUTE = 25;
+
 interface VaccineTickerProps {
   data: NlVaccineAdministeredRateMovingAverageValue;
 }
 
 export function VaccineTicker({ data }: VaccineTickerProps) {
-  const { siteText, formatPercentage, formatNumber } = useIntl();
+  const { siteText, formatNumber } = useIntl();
+
+  const dosesPerMinute = data.doses_per_second * 60;
 
   const isMountedRef = useIsMountedRef();
-  const tickDuration = data.seconds_per_dose * 1000;
-  const tickCount = Math.floor(data.doses_per_second * 60);
+  const [timerOffset, setTimerOffset] = useState(0);
+  const [animationProgress, setAnimationProgress] = useState(0);
   const [counter, setCounter] = useState(0);
 
   useEffect(() => {
     let timeoutId: number;
-    function tick() {
+    function animate() {
       if (!isMountedRef.current) return;
 
-      setCounter((x) => x + 1);
+      const now = new Date();
 
-      timeoutId = window.setTimeout(
-        () => requestAnimationFrame(tick),
-        tickDuration
-      );
+      /*
+        Use the offset at page boot to offset the clock
+        Opening page => clock at 0
+      */
+      if (timerOffset === 0) {
+        setTimerOffset(now.getSeconds());
+      }
+
+      const seconds =
+        (now.getSeconds() +
+          MINUTE -
+          timerOffset +
+          now.getMilliseconds() / 1000) %
+        MINUTE;
+
+      /* Reset at the first frame of the new clock cycle */
+      if (Math.ceil(seconds - (FRAMES_PER_MINUTE * 1.5) / 1000) === 0) {
+        setCounter((x) => x + 1);
+      }
+
+      setAnimationProgress(seconds / MINUTE);
+
+      timeoutId = (setTimeout(
+        animate,
+        1000 / FRAMES_PER_MINUTE
+      ) as unknown) as number;
     }
-    tick();
+    animate();
     return () => window.clearTimeout(timeoutId);
-  }, [isMountedRef, tickDuration]);
+  }, [isMountedRef, timerOffset]);
 
   return (
     <Box
@@ -56,33 +83,28 @@ export function VaccineTicker({ data }: VaccineTickerProps) {
           height={CONTAINER_HEIGHT}
           position="relative"
         >
-          <TickCircle
-            numTicks={tickCount}
-            currentTick={counter % tickCount}
-            tickDuration={tickDuration}
-            tickWidth={2}
-            tickLength={10}
-          />
+          <Clock progress={animationProgress} />
 
-          <Box
-            position="absolute"
-            top="calc(50% - 35px)"
-            left="calc(50% - 35px)"
-          >
-            <Shapes counter={counter} tickDuration={tickDuration} />
-          </Box>
+          <Counter
+            progress={animationProgress}
+            dosesPerMinute={dosesPerMinute}
+          />
         </Box>
       </div>
 
       <Box my={2} pl={{ xs: 3, md: 4 }}>
         <Text fontSize="1.625rem" m={0}>
-          {replaceComponentsInText(siteText.vaccinaties.clock.title, {
-            seconds: (
-              <InlineText color={colors.data.primary} fontWeight="bold">
-                {formatPercentage(data.seconds_per_dose)}
-              </InlineText>
-            ),
-          })}
+          <Shapes counter={counter} />
+          {replaceComponentsInText(
+            siteText.vaccinaties.clock.title_doses_per_minute,
+            {
+              dosesPerMinute: (
+                <InlineText color={colors.data.primary} fontWeight="bold">
+                  {formatNumber(dosesPerMinute)}
+                </InlineText>
+              ),
+            }
+          )}
         </Text>
         <Text m={0}>
           {replaceComponentsInText(siteText.vaccinaties.clock.description, {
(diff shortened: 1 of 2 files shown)


packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
```diff
@@ -13,33 +13,60 @@ const RADIUS = 55;
...
+      if (Math.ceil(seconds - (FRAMES_PER_MINUTE * 1.5) / 1000) === 0) {
+        setCounter((x) => x + 1);
+      }
+
+      setAnimationProgress(seconds / MINUTE);
+
+      timeoutId = (setTimeout(
+        animate,
+        1000 / FRAMES_PER_MINUTE
+      ) as unknown) as number;
```
VWSCoronaDashboard1, line 65: I'm still not sure why we're not just using a requestAnimationFrame call here, instead of the setTimeout.

VWSCoronaDashboard1 (approved):

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
VWSCoronaDashboard10: what's the advantage of requestAnimationFrame with this animation? It's a very lightweight animation..

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
```diff
@@ -13,33 +13,60 @@ const RADIUS = 55;
...
+  const { siteText, formatNumber } = useIntl();
+
+  const dosesPerMinute = data.doses_per_second * 60;
 
   const isMountedRef = useIsMountedRef();
-  const tickDuration = data.seconds_per_dose * 1000;
-  const tickCount = Math.floor(data.doses_per_second * 60);
+  const [timerOffset, setTimerOffset] = useState(0);
+  const [animationProgress, setAnimationProgress] = useState(0);
   const [counter, setCounter] = useState(0);
```
VWSCoronaDashboard10, line 31: Might as well rename counter to shapeIndex since that is its only purpose now

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
```diff
@@ -13,33 +13,60 @@ const RADIUS = 55;
...
-      setCounter((x) => x + 1);
+      const now = new Date();
+
+      /*
+        Use the offset at page boot to offset the clock
+        Opening page => clock at 0
+      */
+      if (timerOffset === 0) {
+        setTimerOffset(now.getSeconds());
+      }
```
VWSCoronaDashboard10, line 46: nit; looks like the offset can be instantiated once in a ref on render, then you won't need to check it every tick
```tsx
const timerOffsetRef = useRef(new Date().getSeconds());
```

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
```diff
@@ -13,33 +13,60 @@ const RADIUS = 55;
...
+          timerOffset +
+          now.getMilliseconds() / 1000) %
+        MINUTE;
 
-      timeoutId = window.setTimeout(
-        () => requestAnimationFrame(tick),
-        tickDuration
-      );
+      /* Reset at the first frame of the new clock cycle */
+      if (Math.ceil(seconds - (FRAMES_PER_MINUTE * 1.5) / 1000) === 0) {
```
VWSCoronaDashboard10, line 56: what does that 1.5 mean?

VWSCoronaDashboard10 (approved):
couple of comments

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
VWSCoronaDashboard8: Yeah that's also what I was thinking. With request animation frame you are rendering whenever the browser says it wants to render a frame. Now you're choosing a fixed framerate, and it seems like you are not calculating the position of the counter based on absolute time. So while it might not make much of a different here, you are not accurately representing time. It all depends on the device rendering this an if it is able to exactly render 25 frames per second.  Any system delay of the setTimeout callback will distort the time.

An accurate counter would I think take Date.now() on page load, and then use the request animation frame callback which then checks the current time and compares it with the start time to know the position of the counter. Then you are accurate even if the frames appear at irregular intervals. And a person with a fast computer would get 100fps or something.

It would not necessarily require more code, so it would have my preference, but I can also live with this.

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
```diff
@@ -13,33 +13,60 @@ const RADIUS = 55;
...
+    function animate() {
       if (!isMountedRef.current) return;
 
-      setCounter((x) => x + 1);
+      const now = new Date();
+
+      /*
+        Use the offset at page boot to offset the clock
+        Opening page => clock at 0
+      */
```
VWSCoronaDashboard8, line 43: The comment is a bit confusing and not using /** */ format. I don't quite understand what you're trying to say here. 

If timerOffset is 0 then set it to now.getSeconds() ? I don't get the logic behind it. Maybe you can choose a better word for offset. Position maybe?

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
```diff
@@ -13,33 +13,60 @@ const RADIUS = 55;
...
+      if (timerOffset === 0) {
+        setTimerOffset(now.getSeconds());
+      }
+
+      const seconds =
+        (now.getSeconds() +
+          MINUTE -
+          timerOffset +
+          now.getMilliseconds() / 1000) %
```
VWSCoronaDashboard8, line 52: why are you calling getSeconds one time and getMilliseconds / 1000 another time? Wouldn't that be the same as calling getSeconds once and using the value twice?

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
```diff
@@ -13,33 +13,60 @@ const RADIUS = 55;
 const CONTAINER_WIDTH = RADIUS * 2 + 12;
 const CONTAINER_HEIGHT = RADIUS * 2 + 12;
 
+const MINUTE = 60;
```
VWSCoronaDashboard8, line 16:
```suggestion
const MINUTE_IN_SECONDS = 60;
```

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
```diff
@@ -13,33 +13,60 @@ const RADIUS = 55;
...
+          MINUTE -
+          timerOffset +
+          now.getMilliseconds() / 1000) %
+        MINUTE;
 
-      timeoutId = window.setTimeout(
-        () => requestAnimationFrame(tick),
-        tickDuration
-      );
+      /* Reset at the first frame of the new clock cycle */
```
VWSCoronaDashboard8, line 55: Comment format

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
```diff
@@ -98,122 +120,111 @@ export function VaccineTicker({ data }: VaccineTickerProps) {
...
-  currentTick: number;
-  tickDuration: number;
-  tickWidth: number;
-  tickLength: number;
-}) {
-  const stepRadius = (2 * Math.PI) / numTicks;
-  const stepDegree = 360 / numTicks;
-
+interface ClockProps {
+  progress: number;
```
VWSCoronaDashboard8, line 124:
```suggestion
  progressPercentage: number;
```

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
```diff
@@ -98,122 +120,111 @@ export function VaccineTicker({ data }: VaccineTickerProps) {
...
-}: {
-  counter: number;
-  tickDuration: number;
-}) {
+interface CounterProps {
+  progress: number;
+  dosesPerMinute: number;
+}
+function Counter({ progress, dosesPerMinute }: CounterProps) {
+  const pause = 3; // percentage of 1 minute
```
VWSCoronaDashboard8, line 187: pauseNumFrames, pauseNumSeconds?

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
VWSCoronaDashboard8: Or positionPercentage

packages/cms/src/lokalize/key-mutations.csv
```diff
@@ -1 +1,2 @@
 timestamp,action,key
+2021-05-05T14:36:25.308Z,add,vaccinaties.clock.title_doses_per_minute
```
VWSCoronaDashboard8, line 2: Seems like you typed a space in the path of the CLI. Guess we need to strip that on input, but this will probably create an issue somewhere.

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
VWSCoronaDashboard4: Not exactly, getSeconds give you whole seconds; and getMilliseconds gives you just milliseconds (0-999). To combine them into one a division by 1000 is needed.

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
VWSCoronaDashboard4: It's not really a position?

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
VWSCoronaDashboard8: I meant isn't `now.getMilliseconds() / 1000` equivalent to `now.getSeconds()`?

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
VWSCoronaDashboard4: I've considered using request animation frame, and it should work. There is one caveat, to control the refresh-rate of the counter (for US/design/readability) I've introduce a variable FRAMES_PER_MINUTE to optionally slow down the refresh rate of the whole animation. (And for the clock a 1fps should be sufficient.)

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
VWSCoronaDashboard4: Good catch!

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
VWSCoronaDashboard4: I've added it in the comment

packages/app/src/domain/vaccine/components/vaccine-ticker.tsx
VWSCoronaDashboard4: no, they're not equal

VWSCoronaDashboard14 (approved):

VWSCoronaDashboard8 (approved):

PR #2797, opened 2021-05-06
Fix zero state for charts

VWSCoronaDashboard1:
## Summary

When given a domain of [0,0] the chart was rendered in a strange way.
It turned out that vsix scaleLinear doesn't handle this situation properly. When given [0,0] the resulting scale function will return the start range divided by two. Instead of just the whole number. This is why the chart is rendered like this:
image

Setting the max domain to infinity in the case of a domain of [0,0] turned out to fix it for now:
image

Perhaps design can come up with some sort of 'empty state' for a chart like this. Because it does look a tad 'empty'...

### Governance

- [ ] Documentation is added

### DIFF
--- a/packages/app/src/components/time-series-chart/logic/scales.ts
+++ b/packages/app/src/components/time-series-chart/logic/scales.ts
@@ -65,8 +65,11 @@ export function useScales<T extends TimestampedValue>(args: {
       round: true, // round the output values so we render on round pixels
     });
 
+    // For some reason vsix-scaleLinear doesn't handle een domain of [0,0] correctly.
+    // In that particular case calling yScale(0) will return the (bounds.height / 2), instead of just bounds.height.
+    // A work-around turns out to be setting the max value to Infinity.
     const yScale = scaleLinear({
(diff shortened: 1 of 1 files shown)


VWSCoronaDashboard4 (approved):

packages/app/src/components/time-series-chart/logic/scales.ts
```diff
@@ -65,8 +65,11 @@ export function useScales<T extends TimestampedValue>(args: {
       round: true, // round the output values so we render on round pixels
     });
 
+    // For some reason vsix-scaleLinear doesn't handle een domain of [0,0] correctly.
+    // In that particular case calling yScale(0) will return the (bounds.height / 2), instead of just bounds.height.
+    // A work-around turns out to be setting the max value to Infinity.
```
VWSCoronaDashboard8, line 70: nit; prefer `/** */` for multi-line comments as it makes it much easier to edit. Also there's a typo vsix although you could pronounce it the same I guess :)

VWSCoronaDashboard8 (approved):

PR #2809, opened 2021-05-07
Fullscreen charts

VWSCoronaDashboard10:
- (almost) All charts are now fully responsive, both vertically as horizontally.
-  choropleth tooltip positioning logic has been improved (almost identical to the logic of time series chart)
- Fix typing of `*.svg` imports
- Introducted a `useReponsiveContainer` hook to manage vertical/horizontal resizing
- Created a generic IconButton wrapper component to make icons clickable
- Made the RadioGroup a controlled component
- Created a reusable Modal component


packages/app/src/utils/use-responsive-container.tsx
```diff
@@ -0,0 +1,88 @@
...
+    ),
+    [ref, minHeight]
+  );
+
+  return { width, height, ResponsiveContainer, ref };
+}
+
+type Size = {
+  width: number | undefined;
+  height: number | undefined;
```
VWSCoronaDashboard8, line 49:
```suggestion
  width?: number;
  height?: number;
```

packages/app/src/utils/use-responsive-container.tsx
```diff
@@ -0,0 +1,88 @@
...
+};
+
+function useResizeObserver<T extends HTMLElement | SVGSVGElement>() {
+  const [size, setSize] = useState<Size>({
+    width: undefined,
+    height: undefined,
```
VWSCoronaDashboard8, line 55: Hmm, maybe you can make with and height required but type it as <Size | undefined> ?  To me it makes more sense to only have them individually optional of it is really possible to have one of them defined at runtime.

packages/app/src/utils/use-responsive-container.tsx
```diff
@@ -0,0 +1,88 @@
...
+  });
+
+  const [node, setNode] = useState<T | null>(null);
```
VWSCoronaDashboard8, line 58: Why null? Is it useCallbackRef that requires it?

packages/app/src/components/choropleth-tile.tsx
```diff
@@ -10,19 +10,27 @@ import {
...
     title: string;
     thresholds: ChoroplethThresholdsValue[];
   };
   metadata?: MetadataProps;
   valueAnnotation?: string;
-}
+} & (
+    | {
+        onChartRegionChange: (v: RegionControlOption) => void;
+        chartRegion: 'municipal' | 'region';
```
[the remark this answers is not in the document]
VWSCoronaDashboard10: Agree, but I'll leave it for another PR since it's out of scope.

packages/app/src/components/time-series-chart/time-series-chart.tsx
```diff
@@ -281,125 +286,125 @@ export function TimeSeriesChart<
...
-
-          {benchmark && (
-            <Benchmark
-              value={benchmark.value}
-              label={benchmark.label}
-              top={yScale(benchmark.value)}
-              width={bounds.width}
+      <ResponsiveContainer>
+        <Box position="relative" css={css({ userSelect: 'none' })}>
+          <ChartContainer
```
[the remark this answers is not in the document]
VWSCoronaDashboard10: It's wrapping more than the chartContainer and it shouldn't be part of the responsiveContainer because that should not be bothered by user-select stuff.

packages/app/src/utils/use-responsive-container.tsx
VWSCoronaDashboard10: Yep, just as using a `useRef` for a DOM reference is requiring a `null` argument (`useRef<HTMLElement>(null)`)

PR #2817, opened 2021-05-11
Filter timespan annotation from legend when it isn't visible on the chart

VWSCoronaDashboard13:
## Summary

Removing the timespan annotation from the legend when it doesn't overlap with the chart domain

## Motivation

Client wanted to remove the timespan annotation from the legend on IC hospital beds

## Detailed design

Updated the `useLegendItems` hook to accept a domain which is used to determine whether the timespanAnnotation is visible. Used a `maxNumItems` variable that collects all of the potential items (so not filtering out items based on the domain) to determine if the legend should be rendered

### DIFF
--- a/packages/app/src/components/time-series-chart/logic/legend.tsx
+++ b/packages/app/src/components/time-series-chart/logic/legend.tsx
@@ -25,7 +25,8 @@ export function useLegendItems<T extends TimestampedValue>(
     if (dataOptions?.timespanAnnotations) {
       for (const annotation of dataOptions.timespanAnnotations) {
         const annotationVisible =
-          first(domain) <= annotation.end && annotation.start <= last(domain);
+          (first(domain) as number) <= annotation.end &&
+          annotation.start <= (last(domain) as number);
 
         items.push({
           label: annotation.label,
--- a/packages/app/src/components/time-series-chart/logic/legend.tsx
+++ b/packages/app/src/components/time-series-chart/logic/legend.tsx
@@ -20,9 +20,17 @@ export function useLegendItems<T extends TimestampedValue>(
     }));
 
     /**
+     * Maximum number of legend items
+     *
+     * Used to determine if there are enough items to render a legend
+     */
+    let maxNumItems: number = items.length;
+
+    /**
      * Add annotations to the legend
      */
     if (dataOptions?.timespanAnnotations) {
+      maxNumItems += dataOptions?.timespanAnnotations.length;
       for (const annotation of dataOptions.timespanAnnotations) {
         const annotationVisible =
           (first(domain) as number) <= annotation.end &&
@@ -46,10 +54,7 @@ export function useLegendItems<T extends TimestampedValue>(
      * This prevents us from having to manually set (and possibly forget to set)
      * a boolean on the chart props.
      */
-    const numAnnotations = dataOptions?.timespanAnnotations
-      ? dataOptions?.timespanAnnotations.length
-      : 0;
-    return items.length + numAnnotations > 1 ? items : [];
+    return maxNumItems > 1 ? items : [];
   }, [config, dataOptions, domain]);
 
   return legendItems;
--- a/packages/app/src/components/time-series-chart/logic/legend.tsx
+++ b/packages/app/src/components/time-series-chart/logic/legend.tsx
@@ -32,11 +32,11 @@ export function useLegendItems<T extends TimestampedValue>(
     if (dataOptions?.timespanAnnotations) {
       maxNumItems += dataOptions?.timespanAnnotations.length;
       for (const annotation of dataOptions.timespanAnnotations) {
-        const annotationVisible =
+        const isAnnotationVisible =
           (first(domain) as number) <= annotation.end &&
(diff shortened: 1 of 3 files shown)


packages/app/src/components/time-series-chart/logic/legend.tsx
```diff
@@ -22,10 +24,15 @@ export function useLegendItems<T extends TimestampedValue>(
...
       for (const annotation of dataOptions.timespanAnnotations) {
+        const annotationVisible =
+          (first(domain) as number) <= annotation.end &&
+          annotation.start <= (last(domain) as number);
+
         items.push({
           label: annotation.label,
           shape: 'custom',
           shapeComponent: <TimespanAnnotationIcon />,
+          hidden: !annotationVisible,
```
VWSCoronaDashboard8, line 35: It is unclear to me why you are injecting the item as hidden, to then use that flag to filter it out in the legend component. It seems like you could also simply omit the item here by not calling the push function. Am I overlooking something?

VWSCoronaDashboard1 (approved):

packages/app/src/components/time-series-chart/logic/legend.tsx
VWSCoronaDashboard8: Oops 😅 RTFM!

packages/app/src/components/time-series-chart/logic/legend.tsx
```diff
@@ -17,16 +19,30 @@ export function useLegendItems<T extends TimestampedValue>(
...
      */
     if (dataOptions?.timespanAnnotations) {
+      maxNumItems += dataOptions?.timespanAnnotations.length;
       for (const annotation of dataOptions.timespanAnnotations) {
-        items.push({
-          label: annotation.label,
-          shape: 'custom',
-          shapeComponent: <TimespanAnnotationIcon />,
-        } as LegendItem);
+        const annotationVisible =
```
VWSCoronaDashboard8, line 35: We definitely intent to prefix all booleans in normal variables.

```suggestion
        const isAnnotationVisible =
```

packages/app/src/components/time-series-chart/logic/legend.tsx
```diff
@@ -17,16 +19,30 @@ export function useLegendItems<T extends TimestampedValue>(
       shapeComponent: <SeriesIcon config={x} />,
     }));
 
+    /**
+     * Maximum number of legend items
+     *
+     * Used to determine if there are enough items to render a legend
+     */
+    let maxNumItems: number = items.length;
```
VWSCoronaDashboard10, line 27: nit; type will be inferred from `items.length`
```suggestion
    let maxNumItems = items.length;
```

packages/app/src/components/time-series-chart/logic/legend.tsx
```diff
@@ -17,16 +19,30 @@ export function useLegendItems<T extends TimestampedValue>(
...
+      maxNumItems += dataOptions?.timespanAnnotations.length;
       for (const annotation of dataOptions.timespanAnnotations) {
-        items.push({
-          label: annotation.label,
-          shape: 'custom',
-          shapeComponent: <TimespanAnnotationIcon />,
-        } as LegendItem);
+        const isAnnotationVisible =
+          (first(domain) as number) <= annotation.end &&
+          annotation.start <= (last(domain) as number);
```
VWSCoronaDashboard10, line 37: for readability it might be nice to save the first and last values to variables;
```tsx
if (dataOptions?.timespanAnnotations) {
  const start = first(domain) as number;
  const end = last(domain) as number;

  for (const annotation of dataOptions.timespanAnnotations) {
    const isAnnotationVisible =
      start <= annotation.end && annotation.start <= end;

      // ...
  }
}

File: packages/app/src/components/time-series-chart/logic/legend.tsx [hunk 5]
```diff
@@ -38,8 +54,8 @@ export function useLegendItems<T extends TimestampedValue>(
      * This prevents us from having to manually set (and possibly forget to set)
      * a boolean on the chart props.
      */
-    return items.length > 1 ? items : [];
-  }, [config, dataOptions]);
+    return maxNumItems > 1 ? items : [];
+  }, [config, dataOptions, domain]);
```
VWSCoronaDashboard10, line 58: What is the reason for not using the length of the items anymore?

packages/app/src/components/time-series-chart/logic/legend.tsx
```diff
@@ -17,16 +19,30 @@ export function useLegendItems<T extends TimestampedValue>(
...
+     *
+     * Used to determine if there are enough items to render a legend
+     */
+    let maxNumItems: number = items.length;
+
     /**
      * Add annotations to the legend
      */
     if (dataOptions?.timespanAnnotations) {
+      maxNumItems += dataOptions?.timespanAnnotations.length;
```
VWSCoronaDashboard10, line 33:
```suggestion
      maxNumItems += dataOptions.timespanAnnotations.length;
```

VWSCoronaDashboard10 (approved):

PR #2820, opened 2021-05-11
Update collapsible button design and animation

VWSCoronaDashboard14:
- Update the collapsible button design 
- Updated the expand animation
- Reduced the size of the risk level map dropdown button


VWSCoronaDashboard1 (approved):

packages/app/src/components/collapsible/collapsible-button.tsx
```diff
@@ -7,43 +7,80 @@ import {
...
 }
 
 export const CollapsibleButton = ({
   label,
   children,
+  icon,
 }: CollapsibleButtonProps) => {
-  const { ref: contentRef, height: contentHeight = 0 } = useResizeObserver();
-  const { ref: buttonRef, height: buttonHeight = 0 } = useResizeObserver();
+  const {
```
VWSCoronaDashboard4, line 24: Do you need two resize observers? If you use one, you can read out the other's width/height using the `buttonRef.current.offsetWidth` etc?

packages/app/src/components/collapsible/collapsible-button.tsx
```diff
@@ -7,43 +7,80 @@ import {
...
+    ref: contentRef,
+    height: contentHeight = 0,
+    width: contentWidth = 0,
+  } = useResizeObserver();
+
+  const {
+    ref: buttonRef,
+    width: buttonWidth = 0,
+    height: buttonHeight = 0,
+  } = useResizeObserver();
```
VWSCoronaDashboard8, line 34: This doesn't look like a very nice API to use. I think we should refactor `useResizeObserver` to return a tuple with fallbacks for width/height set to 0. Then you'd be able to name the return values easily.

I'll put it on the backlog...

packages/app/src/components/collapsible/collapsible-button.tsx
```diff
@@ -7,43 +7,80 @@ import {
...
 
+  /**
+   * Calculate the clip path where the content needs to animate to,
+   * this is alligned with the position of the button percentage wise based of the content wrapper
+   */
+  const clipPathCalculation = useMemo(() => {
+    if (!buttonWidth || !contentWidth || !buttonHeight || !contentHeight)
+      return '0 0, 100% 0, 100% 0, 0 0';
+
+    const width = ((buttonWidth / contentWidth) * 100 - 100) / 2;
```
VWSCoronaDashboard4, line 47: What is this magic 100?

In general, I think a few comments here and there helps debugging when a future dev works with this code.

packages/app/src/components/collapsible/collapsible-button.tsx
```diff
@@ -55,62 +92,116 @@ export const CollapsibleButton = ({
...
+    mx: 'auto',
+  })
+);
+
+const Container = styled(Box).attrs({ as: 'section' })<{
+  isOpen: boolean;
+  buttonWidth: number;
+  contentWidth: number;
+  contentHeight: number;
+  buttonHeight: number;
```
VWSCoronaDashboard8, line 108: nit; ordered it would read a little better

packages/app/src/components/collapsible/collapsible-button.tsx
```diff
@@ -7,43 +7,80 @@ import {
...
-        <div
-          ref={buttonRef}
-          css={css({ display: 'flex', justifyContent: 'center' })}
-        >
-          <DisclosureButton>
-            {label}
-            <Chevron open={isOpen} />
-          </DisclosureButton>
-        </div>
+        <ButtonContainer>
```
VWSCoronaDashboard4, line 75: I think Disclose is mentioning / advising against wrapping their button like this. If it works, it works, but I guess it could break when Disclosure updates.

VWSCoronaDashboard8 (approved):

packages/app/src/components/collapsible/collapsible-button.tsx
```diff
@@ -55,62 +92,116 @@ export const CollapsibleButton = ({
...
-    },
-    '[data-reach-disclosure-button][data-state="open"]:after': {
-      content: '""',
-      position: 'absolute',
-      bottom: 0,
-      left: 0,
-      right: 0,
-      bg: 'lightGray',
-      height: '1px',
+      outlineColor: x.isOpen ? 'rgba(0, 0, 0, 0)' : 'blue',
```
VWSCoronaDashboard4, line 169:
```suggestion
      outlineColor: x.isOpen ? 'transparent' : 'blue',
```

Or better, use the rgb values for `blue`, and use that as the transparent color?

packages/app/src/components/collapsible/collapsible-button.tsx
VWSCoronaDashboard8: Hmm it's a node_module. I guess we could wrap or fork it if we want to. Not sure if it's worth it.

packages/app/src/assets/grafiek.svg
```diff
@@ -0,0 +1,7 @@
+<svg focusable="false" width="29" height="22" viewBox="0 0 29 22" fill="currentColor" xmlns="http://www.w3.org/2000/svg">
```
VWSCoronaDashboard8, line 1: I don't see a reason to name our assets in Dutch unless it's something very specific to the domain that we don't have good translation for. So I suggest to rename this to chart.svg

packages/app/src/components/collapsible/collapsible-button.tsx
VWSCoronaDashboard14: The fallback by 0 was only used because somebody else used the same structure on this component so I followed along. Since it only gets used in the memo and if something yields 0 it automatically fails the check because `0` is falsy. So I believe I could remove the default 0.

packages/app/src/components/collapsible/collapsible-button.tsx
VWSCoronaDashboard14: The blue border doesn't animate/transition so in my opinion transparent is a bit nicer and easier to read. 👍

packages/app/src/components/collapsible/collapsible-button.tsx
VWSCoronaDashboard14: I believe I do since the button and content don't know anything of each other and are not related in terms of the structure of the html.

PR #2823, opened 2021-05-12
Apply dynamic height as style instead of classname

VWSCoronaDashboard10:


VWSCoronaDashboard1 (approved):

VWSCoronaDashboard8 (approved):

packages/app/src/utils/use-responsive-container.tsx
```diff
@@ -33,7 +33,7 @@ export function useResponsiveContainer(initialWidth: number, minHeight = 0) {
       height?: string | number;
     }) => (
       <Box ref={ref} height="100%" minHeight={minHeight} position="relative">
-        <Box position="absolute" width="100%" height={height}>
+        <Box position="absolute" width="100%" style={{ height }}>
```
VWSCoronaDashboard8, line 36: You might want to place a comment here, because someone could easily change it back thinking it was a mistake

PR #2826, opened 2021-05-12
Sort time series in backend data before validation

VWSCoronaDashboard1:
## Summary

Some of the new custom validations require the data to be sorted, unfortunately, this wasn't the case yet...

## Motivation

Bug fix.

### Governance

- [ ] Documentation is added

### DIFF
--- a/packages/cli/src/scripts/validate-json-all.ts
+++ b/packages/cli/src/scripts/validate-json-all.ts
@@ -135,6 +135,7 @@ async function validate(schemaName: string, schemaInfo: SchemaInfoItem) {
       }
     } catch (e) {
       console.group();
+      console.error(e);
       console.error(chalk.bgRed.bold(`  ${fileName} cannot be parsed  \n`));
       console.groupEnd();
       return false;
--- a/packages/cli/src/scripts/validate-json-all.ts
+++ b/packages/cli/src/scripts/validate-json-all.ts
@@ -1,7 +1,9 @@
+import { sortTimeSeriesInDataInPlace } from '@corona-dashboard/common';
 import chalk from 'chalk';
 import fs from 'fs';
 import meow from 'meow';
 import path from 'path';
+import { isDefined } from 'ts-is-present';
 import { schemaDirectory } from '../config';
 import {
   createValidateFunction,
@@ -10,6 +12,7 @@ import {
   SchemaInfo,
   SchemaInfoItem,
 } from '../schema';
+import { JSONObject } from '../schema/custom-validations';
 
 const cli = meow(
   `
@@ -114,7 +117,7 @@ async function validate(schemaName: string, schemaInfo: SchemaInfoItem) {
       encoding: 'utf8',
     });
 
-    let data: any = null;
+    let data: JSONObject | undefined;
     try {
       data = JSON.parse(contentAsString);
     } catch (e) {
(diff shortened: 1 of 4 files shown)


packages/cli/src/scripts/validate-json-single.ts
```diff
@@ -123,16 +124,20 @@ createValidateFunction(rootSchema, schemaBasePath).then((validateFunction) => {
     process.exit(1);
   }
 
-  const { isValid, schemaErrors } = executeValidations(
-    validateFunction,
-    jsonData,
-    schemaInfo
-  );
+  if (isDefined(jsonData)) {
+    sortTimeSeriesInDataInPlace(jsonData);
```
VWSCoronaDashboard8, line 128: nit; I think you could place everything in the try/catch, and then you don't need to deal with undefined jsonData, because I don't think a successful parse call will return undefined.

VWSCoronaDashboard8 (approved):

packages/cli/src/scripts/validate-json-all.ts
```diff
@@ -114,26 +116,26 @@ async function validate(schemaName: string, schemaInfo: SchemaInfoItem) {
...
+        console.error(schemaErrors);
+        console.error(chalk.bgRed.bold(`  ${fileName} is invalid  \n`));
+        console.groupEnd();
+        return false;
+      }
+    } catch (e) {
       console.group();
-      console.error(schemaErrors);
-      console.error(chalk.bgRed.bold(`  ${fileName} is invalid  \n`));
+      console.error(chalk.bgRed.bold(`  ${fileName} cannot be parsed  \n`));
```
VWSCoronaDashboard10, line 138: the catch can catch more errors than only the json parsing..
```suggestion
      console.error(e);
```

packages/cli/src/scripts/validate-json-single.ts
```diff
@@ -113,27 +114,28 @@ createValidateFunction(rootSchema, schemaBasePath).then((validateFunction) => {
...
+    );
+
+    if (!isValid) {
+      console.error(schemaErrors);
+      logError(`  ${jsonFileName} is invalid  \n`);
+      process.exit(1);
+    }
   } catch (e) {
     console.group();
     logError(`  ${fileName} cannot be parsed  \n`);
```
VWSCoronaDashboard10, line 135: same here I guess

VWSCoronaDashboard10 (approved):

packages/cli/src/scripts/validate-json-all.ts
VWSCoronaDashboard8: nit; I think it's pretty rare that that stack trace gives you information that the message doesn't. Ans also we have a util for printing errors in red. So all these lines can be simplified to this IMO:

```
logError(`Failed to process file ${fileName}: ${err.message}`)
```

packages/cli/src/scripts/validate-json-single.ts
```diff
@@ -113,27 +114,29 @@ createValidateFunction(rootSchema, schemaBasePath).then((validateFunction) => {
...
+      console.error(schemaErrors);
+      logError(`  ${jsonFileName} is invalid  \n`);
+      process.exit(1);
+    }
   } catch (e) {
     console.group();
+    console.error(e);
     logError(`  ${fileName} cannot be parsed  \n`);
     console.groupEnd();
     process.exit(1);
```
VWSCoronaDashboard8, line 138: Same comment here

VWSCoronaDashboard8 (approved):

PR #2827, opened 2021-05-12
Fix production dataset export

VWSCoronaDashboard10:


VWSCoronaDashboard1 (approved):

packages/cms/src/lokalize/export.ts
```diff
@@ -19,16 +19,20 @@ const cli = meow(
...
 `,
   {
     flags: {
       drafts: {
         type: 'boolean',
-        alias: 'd',
+      },
+      dataset: {
+        type: 'string',
+        isRequired: true,
```
VWSCoronaDashboard8, line 35: I don't see why you need to make it required. I think this might the lokalize:add which calls lokalize:export after.

I think defaulting to development would be fine.

VWSCoronaDashboard8 (approved):

packages/cms/src/lokalize/export.ts
VWSCoronaDashboard10: yep correct, allready rolled it back

PR #2837, opened 2021-05-13
Line chart all zero state

VWSCoronaDashboard1:
## Summary

Add exception for drawing the number of gridlines in the case of al zero values:
image

In this case a gridline on the top and bottom is drawn

## Motivation

Design request.

### Governance

- [ ] Documentation is added


packages/app/src/components/time-series-chart/logic/scales.ts
```diff
@@ -66,9 +66,9 @@ export function useScales<T extends TimestampedValue>(args: {
     });
 
     /**
-    For some reason visx-scaleLinear doesn't handle een domain of [0,0] correctly.
-    In that particular case calling yScale(0) will return the (bounds.height / 2), instead of just bounds.height.
-    A work-around turns out to be setting the max value to Infinity.
+      For some reason visx-scaleLinear doesn't handle een domain of [0,0] correctly.
+      In that particular case calling yScale(0) will return the (bounds.height / 2), instead of just bounds.height.
+      A work-around turns out to be setting the max value to Infinity.
     */
```
VWSCoronaDashboard8, line 72: nit; I would prefer to stick with the same comment format everywhere, with every line starting with "*". You can also use something like VSCode re-wrap extension to wrap the text nicely and never manually have to format it.

VWSCoronaDashboard8 (approved):

packages/app/src/components/time-series-chart/components/axes.tsx
```diff
@@ -124,6 +125,22 @@ export const Axes = memo(function Axes({
   const isLongStartLabel = formatXAxis(startUnix).length > 6;
   const isLongEndLabel = formatXAxis(endUnix).length > 6;
 
+  /**
+   * We make an exception for the situation where all the values in the chart are zero.
+   * In that case the top range has been set to Infinity, but we want to draw exactly
+   * two gridlines in this case (at the top and bottom of the chart). So therefore we
+   * check for this case here and create a scale and gridline count accordingly.
+   */
+  const isAllZeroState = yScale.domain()[1] === Infinity;
```
VWSCoronaDashboard10, line 134: a scale's domain can hold more than one value, therefore `[1]` isn't necessarily the last one.
```tsx
  import last from 'lodash/last'
  // ...
  const isAllZeroState = last(yScale.domain()) === Infinity;
```

VWSCoronaDashboard10 (approved):

packages/app/src/components/time-series-chart/components/axes.tsx
VWSCoronaDashboard4: Or pass a boolean through the props eg `allValuesZero` , instead of deriving it from the domain?

VWSCoronaDashboard4 (approved):

packages/app/src/components/time-series-chart/components/axes.tsx
VWSCoronaDashboard10: Isn't it only derivable from the domain?

VWSCoronaDashboard1: Currently, it is only derivable through the domain. I added Infinity explicitly for a [0,0] domain in order to have some kind of way of recognizing this edge case. (Also, adding the Infinity made sure the rendering of the chart didn't screw up like before)

VWSCoronaDashboard1: Oh, and in this particular (edge) case, the domain definitely consists of only two items.

VWSCoronaDashboard1: > Or pass a boolean through the props eg `allValuesZero` , instead of deriving it from the domain?

This might be more explicit I guess, I'll have a look.

packages/app/src/components/time-series-chart/components/axes.tsx
```diff
@@ -52,6 +52,12 @@ type AxesProps = {
    * it will move the Y-axis to the left.
    */
   xRangePadding?: number;
+
+  /**
+   * Indicates if the chart contains a series that only has values of zero.
+   * (In case this needs special rendering)
+   */
+  allZeroValues?: boolean;
```
VWSCoronaDashboard8: This is not a clear name for a boolean, give it a prefix please

VWSCoronaDashboard1: you were looking at an old commit. It was already renamed to hasAllZeroValues

PR #2839, opened 2021-05-13
Improve lokalize cli flow

VWSCoronaDashboard8:
## Summary

This aims to simplify how we add or remove Lokalize keys to/from Sanity.

* Simplified flow by targeting the keys directly without mentioning subjects
* lokalize:add now accepts the `--key` or `-k` flag. This will be the initial value
* lokalize:delete now accepts the `--key` or `-k` flag. This will be the initial value for single key input. If omitted the user receives a multi-select prompt.
* lokalize:delete now uses autocomplete for the multiselect prompt, which makes it easy to find keys.
* The JSON text export is now integral to the add/delete scripts, skipping the export when no mutations where made.
* Added the ability to exit / abort the CLI scripts at any point.
* Print only essential info to the console
* Adds validation on key string allowing only lower-snake-cased strings with dot notation.


VWSCoronaDashboard1 (approved):

packages/cms/src/lokalize/add-text.ts
```diff
@@ -1,120 +1,150 @@
...
+         * lower-snake-case paths in dot notation.
+         *
+         * @TODO improve the regexp so that the key is not allowed to end in a
+         * dot. Also we should enforce at least 1 dot to have subject+path
+         * separation in Sanity.
+         * https://regex101.com/r/RJaXUv/1
+         */
+        return (
+          existingKeys.find((key) => key === x) === undefined &&
+          /^([a-z_]+\.?)+$/.test(x)
```
VWSCoronaDashboard4, line 136:
```suggestion
          /^([a-z_\.]+[a-z_]$/.test(x)
```

and then you can remove the `todo`

packages/cms/src/lokalize/delete-text.ts
```diff
@@ -10,78 +10,131 @@
...
+ * There is currently no native way to exit prompts on ctrl-c. This is a
+ * workaround that needs to be added to every prompts instance. For more info
+ * see: https://github.com/terkelg/prompts/issues/252#issuecomment-778683666
+ */
+function onState(state: { aborted: boolean }) {
+  if (state.aborted) {
+    if (deletionsCounter > 0) {
+      console.log(`Marked ${deletionsCounter} documents for deletion`);
+      console.log('Updating text export...');
+      exportLokalizeTexts().finally(() => {
```
VWSCoronaDashboard4, line 131: not sure if, when you abort this, you want it to still export stuff?

VWSCoronaDashboard4 (approved):
Good improvements!

packages/cms/src/lokalize/delete-text.ts
VWSCoronaDashboard8: If there were deletions that means those were written to te mutation log. That log is used in the export. I think you always want the json file to reflect the state of Sanity + mutations.

packages/cms/src/lokalize/add-text.ts
VWSCoronaDashboard8: Nice, this works (without the parenthesis)

packages/cms/src/lokalize/delete-text.ts
VWSCoronaDashboard8: Technically this could hardly happen, because the number is only incremented at the very end of the CLI, but I'll leave it in for safety should the flow later be extended, and it can't hurt.

packages/cms/src/lokalize/delete-text.ts
```diff
@@ -10,78 +10,131 @@
...
   console.error('An error occurred:', err.message);
   process.exit(1);
 });
+
+/**
+ * There is currently no native way to exit prompts on ctrl-c. This is a
+ * workaround that needs to be added to every prompts instance. For more info
+ * see: https://github.com/terkelg/prompts/issues/252#issuecomment-778683666
+ */
+function onState(state: { aborted: boolean }) {
```
VWSCoronaDashboard10, line 126: the onCancel method isn't triggered either? 
```tsx
prompts([...], { onCancel: () => ... })

File: packages/cms/src/lokalize/logic/fetch.ts [hunk 4]
```diff
@@ -0,0 +1,14 @@
+import { getClient } from '../../client';
+import { LokalizeText } from '../types';
+
+export async function fetchExistingKeys() {
```
VWSCoronaDashboard10, line 4: Would it help us in the future to rename this function to something which explicitly tells us it includes draft documents?

packages/cms/src/lokalize/logic/export.ts
```diff
@@ -0,0 +1,67 @@
...
+  /**
+   * The client will load drafts by default because it is authenticated with a
+   * token. If the `drafts` flag is not set to true, we will manually
+   * exclude draft-documents on query-level.
+   */
+  const draftsQueryPart = includeDrafts ? '' : '&& !(_id in path("drafts.**"))';
+
+  const documents: LokalizeText[] = await client.fetch(
+    `*[_type == 'lokalizeText' ${draftsQueryPart}] | order(key asc)`
+  );
```
VWSCoronaDashboard10, line 48: can't you call the `fetchExistingKeys` util?

packages/cms/src/lokalize/key-mutations.csv
```diff
@@ -44,3 +44,8 @@ timestamp,action,key
 2021-05-10T08:57:14.416Z,add,veiligheidsregio_positief_geteste_personen_ggd.linechart_positivetests_legend_label_moving_average
 2021-05-10T11:07:55.275Z,add,common.risiconiveau_singular
 2021-05-10T11:08:57.458Z,add,common.risiconiveau_plural
+2021-05-13T12:53:21.451Z,add,aljsdf.alds
+2021-05-13T12:53:57.581Z,add,aljsdf.aldsadf
+2021-05-13T13:28:09.646Z,add,aljsdf.aldsf
+2021-05-13T13:54:57.322Z,delete,aljsdf.alds
+2021-05-13T13:56:12.988Z,delete,aljsdf.aldsadf
```
VWSCoronaDashboard10, line 51: It looks like you missed a delete (add 3, delete 2)

VWSCoronaDashboard10 (approved):

packages/cms/src/lokalize/add-text.ts
```diff
@@ -1,120 +1,146 @@
...
+      initial: initialKey,
+      validate: (x: string) => {
+        /**
+         * Validation requires the key to be new and also to only contain
+         * lower-snake-case paths in dot notation.
+         * https://regex101.com/r/vsvHEi/1
+         */
+        return (
+          existingKeys.find((key) => key === x) === undefined &&
+          /^[a-z_\.]+[a-z_]+$/.test(x)
```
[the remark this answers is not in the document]
VWSCoronaDashboard8: Awesome thanks!

packages/cms/src/lokalize/delete-text.ts
VWSCoronaDashboard8: When does onCancel happen? I tested which with Ctrl-C and that worked with the onState handler. Maybe onCancel is just a more specific handler and onState more of a catch-all for state changes?

VWSCoronaDashboard10: I don't know, it was more or less a question for you 😛

packages/cms/src/lokalize/logic/export.ts
VWSCoronaDashboard8: I don't think so, because documents are actually used further down. The function only returns the keys.

packages/cms/src/lokalize/key-mutations.csv
VWSCoronaDashboard8: I must have somehow deleted it from the dataset already, because it's not there anymore.

packages/cms/src/lokalize/logic/fetch.ts
VWSCoronaDashboard8: I think if you want to fetch all existing keys that kind of implies that you also want to include the drafts. I have a hard time imagining a situation where you want to check against all the keys but only for the published documents.

VWSCoronaDashboard1: When in doubt, add a comment?

PR #2843, opened 2021-05-14
Server to support custom header management

VWSCoronaDashboard4:
## Summary

This PR solves a few things to get nextjs to run on node in production, so  target is the `topic/production-on-node branch`.

Setup of a custom express/nextjs server to do the following:
* Sanity images go via express instead of via nextjs rewrite; this is mainly add/remove specific HTTP headers.
* Prevent a side effect of a nextjs bug (that is not fixed in recent versions) wrongfully expose `/en` and `/nl` as routes with the new i18n setup. These are redirected to the homepage.

Note that the server code (and also next config) is written in JS, not TS.


packages/app/next-server.js
```diff
@@ -0,0 +1,121 @@
...
+});
+
+if (
+  !process.env.NEXT_PUBLIC_SANITY_DATASET ||
+  !process.env.NEXT_PUBLIC_SANITY_PROJECT_ID
+) {
+  throw new Error(
+    'Provide NEXT_PUBLIC_SANITY_DATASET and NEXT_PUBLIC_SANITY_PROJECT_ID'
+  );
+}
```
VWSCoronaDashboard8, line 22: Two lines of assert I think would be cleaner and more informative, since then also you know which of the two is missing.

packages/app/next-server.js
```diff
@@ -0,0 +1,121 @@
...
+  server.get('/nl/*', function (_, res) {
+    res.redirect('/');
+  });
+
+  /**
+   * Set these headers for all non-Sanity-cdn routes
+   */
+  server.use(function (req, res) {
+    const isHtml = req.url.indexOf('.') === -1;
+    setResponseHeaders(res, 15768000, isHtml);
```
VWSCoronaDashboard8, line 83: It would help if this number was stored in a variable with a clear name

packages/app/next-server.js
```diff
@@ -0,0 +1,121 @@
...
+if (
+  !process.env.NEXT_PUBLIC_SANITY_DATASET ||
+  !process.env.NEXT_PUBLIC_SANITY_PROJECT_ID
+) {
+  throw new Error(
+    'Provide NEXT_PUBLIC_SANITY_DATASET and NEXT_PUBLIC_SANITY_PROJECT_ID'
+  );
+}
+
+const dev = process.env.NODE_ENV !== 'production';
```
VWSCoronaDashboard8, line 24: For some reason it feels more logical and common to use it the other way around. And name it as a boolean please. `isProduction` or alternatively as a constant `IS_PRODUCTION`

VWSCoronaDashboard8 (approved):

PR #2853, opened 2021-05-17
Add top header for NL and VR behavior page behind a feature flag

VWSCoronaDashboard14:
Add a temporary component that serves as a "page" behind a feature flag for the new behavior page
Added a data-driven text in the tiles
Moved the calculation functions to a separate file (could also separate them per function)


packages/app/src/domain/behavior/behavior-page-national.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { ContentHeader } from '~/components/content-header';
+import { Tile } from '~/components/tile';
+import { TileList } from '~/components/tile-list';
+import { TwoKpiSection } from '~/components/two-kpi-section';
+import { Heading, InlineText, Text } from '~/components/typography';
+import { useFormatAndSortBehavior } from '~/domain/behavior/behavior-logic';
+import { useIntl } from '~/intl';
+import { replaceComponentsInText } from '~/utils/replace-components-in-text';
+interface BehaviourPageNationalProps {
+  data: any;
```
VWSCoronaDashboard4, line 12: I would give these proper types.

packages/app/src/domain/behavior/behavior-page-safety-region.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { ContentHeader } from '~/components/content-header';
+import { Tile } from '~/components/tile';
+import { TileList } from '~/components/tile-list';
+import { TwoKpiSection } from '~/components/two-kpi-section';
+import { Heading, InlineText, Text } from '~/components/typography';
+import { useIntl } from '~/intl';
+import { replaceComponentsInText } from '~/utils/replace-components-in-text';
+
+interface BehaviorPageSafetyRegionProps {
+  data: any;
```
VWSCoronaDashboard4, line 12: `any` => actual types

VWSCoronaDashboard4 (approved):

packages/app/src/domain/behavior/behavior-logic.tsx
```diff
@@ -0,0 +1,85 @@
...
+  behavior: BehaviorValue
+): {
+  sortedCompliance: BehaviorFormatted[];
+  sortedSupport: BehaviorFormatted[];
+} {
+  const compliance = useFormatBehaviorType(behavior, 'compliance');
+  const support = useFormatBehaviorType(behavior, 'support');
+
+  return sortBehavior(compliance, support);
+}
```
VWSCoronaDashboard8, line 85: This whole file seems to be very specific to the behavior table component. So you might want to rename this to `useBehaviorTableItems` or something and possibly more it to the be nested inside the table.

VWSCoronaDashboard14: The top header also uses the sorted data, so not sure about it to make it part of the table. That was the idea to place it above the table.

PR #2854, opened 2021-05-17
Schemas for situations

VWSCoronaDashboard4:
## Summary

Schemas for situations


packages/app/schema/nl/situations.json
```diff
@@ -0,0 +1,137 @@
...
+        "home": {
+          "type": ["number", "null"]
+        },
+        "visit": {
+          "type": ["number", "null"]
+        },
+        "work": {
+          "type": ["number", "null"]
+        },
+        "school_daycare": {
```
VWSCoronaDashboard8, line 44: daycare doesn't seem to be one word

packages/app/schema/nl/situations.json
```diff
@@ -0,0 +1,137 @@
...
+      "title": "nl_situations_value",
+      "type": "object",
+      "properties": {
+        "date_start_unix": {
+          "type": "integer"
+        },
+        "date_end_unix": {
+          "type": "integer"
+        },
+        "complete_data_count": {
```
VWSCoronaDashboard8, line 29: This doesn't tell me much, because it's not saying what data. Is it the total number of reported infections over all the categories? We use the word "total" in other places so might be more fitting.

packages/app/schema/nl/situations.json
```diff
@@ -0,0 +1,137 @@
...
+        "visit": {
+          "type": ["number", "null"]
+        },
+        "work": {
+          "type": ["number", "null"]
+        },
+        "school_daycare": {
+          "type": ["number", "null"]
+        },
+        "general_practitioner": {
```
VWSCoronaDashboard8, line 47: This is called doctor in other places of the data, so I suggest to keep it consistent.

packages/app/schema/nl/situations.json
```diff
@@ -0,0 +1,137 @@
...
+          "type": ["number", "null"]
+        },
+        "hospital": {
+          "type": ["number", "null"]
+        },
+        "health_care_other": {
```
VWSCoronaDashboard8, line 53: This seems to be the only category about health_care so does it need the suffix other?

packages/app/schema/nl/situations.json
```diff
@@ -0,0 +1,137 @@
...
+        "nursing_home": {
+          "type": ["number", "null"]
+        },
+        "residential_care_disabled": {
+          "type": ["number", "null"]
+        },
+        "residential_care_other": {
+          "type": ["number", "null"]
+        },
+        "daycare_elderly_disabled": {
```
VWSCoronaDashboard8, line 65: day_care

packages/app/schema/nl/situations.json
```diff
@@ -0,0 +1,137 @@
...
+          "type": ["number", "null"]
+        },
+        "daycare_other": {
```
VWSCoronaDashboard8, line 68: idem

packages/app/schema/nl/situations.json
```diff
@@ -0,0 +1,137 @@
...
+          "type": ["number", "null"]
+        },
+        "hospice": {
+          "type": ["number", "null"]
+        },
+        "gathering": {
```
VWSCoronaDashboard8, line 74: Maybe social_gathering would be a little clearer if that is what the data is about, because technically a lot of the others would also be gatherings.

packages/app/schema/nl/situations.json
```diff
@@ -0,0 +1,137 @@
...
+          "type": ["number", "null"]
+        },
+        "travel": {
+          "type": ["number", "null"]
+        },
+        "flight": {
+          "type": ["number", "null"]
+        },
```
VWSCoronaDashboard8, line 82: Interesting that they choose to split travel and flight

packages/app/schema/nl/situations.json
```diff
@@ -0,0 +1,137 @@
...
+      },
+      "required": [
+        "date_start_unix",
+        "date_end_unix",
+        "complete_data_count",
+        "situation_known_percentage",
+        "home",
+        "visit",
+        "work",
+        "school_daycare",
```
VWSCoronaDashboard8, line 113: You have two other categories starting with daycare, so wouldn't it make sense to call this one day_care_school? Or if it is school _and_ day maybe make that more explicit; `school_and_day_care`.

packages/app/schema/nl/situations.json
```diff
@@ -0,0 +1,137 @@
...
+        "daycare_elderly_disabled",
+        "daycare_other",
+        "hospice",
+        "gathering",
+        "travel",
+        "flight",
+        "hospitality",
+        "student_activity",
+        "leisure",
+        "religious_activity",
```
VWSCoronaDashboard8, line 129: These two could be grouped under activity. So activity_student and activity_religious.

packages/app/schema/nl/situations.json
```diff
@@ -0,0 +1,137 @@
...
+        "health_care_other",
+        "nursing_home",
+        "residential_care_disabled",
+        "residential_care_other",
+        "daycare_elderly_disabled",
+        "daycare_other",
+        "hospice",
+        "gathering",
+        "travel",
+        "flight",
```
VWSCoronaDashboard8, line 125: Maybe this can be travel_flight and travel_other

packages/app/schema/vr/situations.json
```diff
@@ -0,0 +1,153 @@
...
+          "type": "integer"
+        },
+        "date_end_unix": {
+          "type": "integer"
+        },
+        "vrcode": {
+          "type": "string",
+          "equalsRootProperty": "code"
+        },
+        "complete_data": {
```
VWSCoronaDashboard8, line 36: This doesn't read like a boolean

packages/app/schema/nl/situations.json
```diff
@@ -0,0 +1,137 @@
...
+        },
+        "other": {
+          "type": ["number", "null"]
+        }
+      },
+      "required": [
+        "date_start_unix",
+        "date_end_unix",
+        "complete_data_count",
+        "situation_known_percentage",
```
VWSCoronaDashboard8, line 109: nit; maybe plural reads a little better `situations_known_percentage`. Especially since you also have investigations_count. But that one doesn't show up in this list. Is that intentional?

packages/app/schema/vr_collection/situations.json
```diff
@@ -0,0 +1,26 @@
...
+      "type": "integer"
+    },
+    "date_end_unix": {
+      "type": "integer"
+    },
+    "vrcode": {
+      "type": "string",
+      "pattern": "^VR[0-9]+$"
+    },
+    "complete_data": {
```
VWSCoronaDashboard8, line 22: Before it was a boolean but here it is an integer. In any case the name isn't helping me much in understanding what the value is about.

VWSCoronaDashboard8 (changes requested):

packages/app/schema/nl/situations.json
VWSCoronaDashboard4: It's not prefixed for eg hospitals as `health_care_hospital`; I'll swap it to make it `other_health_care`

packages/app/schema/nl/situations.json
VWSCoronaDashboard4: Not sure if these should be grouped; they are just parts of the words.

packages/app/schema/nl/situations.json
VWSCoronaDashboard4: There is no grouping going on for similar items in this list, so why just for these two?

packages/app/schema/nl/__index.json
```diff
@@ -188,6 +188,9 @@
     },
     "vaccine_stock": {
       "$ref": "vaccine_stock.json"
+    },
+    "situations": {
```
VWSCoronaDashboard10, line 192: I think this name is rather vague, how about something like:
- `infection_location`
- `infection_site`
- `site_of_infection`
- `site_of_contamination`
- `contamination_situation`
- `infection_situation`

packages/app/schema/nl/situations.json
```diff
@@ -0,0 +1,137 @@
...
+        "date_end_unix": {
+          "type": "integer"
+        },
+        "regions_with_sufficient_responses_total": {
+          "type": "integer"
+        },
+        "situation_known_percentage": {
+          "type": "number"
+        },
+        "home": {
```
VWSCoronaDashboard10, line 35: Would it make sense to prefix all "location"-properties (like home/work/doctor/hospital/etc) to distinguish them from non-location properties?

VWSCoronaDashboard10 (approved):

packages/app/schema/nl/situations.json
VWSCoronaDashboard8: I don't see a practical motivation within the context of this schema to make a distinction between locations and non-locations. I think without the prefix it's quite clear which ones are about a location and which ones are about an activity. So if the chart doesn't make the distinction I'd keep it without prefix.

packages/app/schema/nl/situations.json
```diff
@@ -0,0 +1,137 @@
...
+          "type": ["number", "null"]
+        },
+        "visit": {
```
VWSCoronaDashboard8, line 38: Since there is also "home" I think we should clarify this one. Because a visit could be in your home, but I'm assuming here you visited someone else... 

Maybe `visiting` or `visit_someone`, `visit_made`... It's a tough one :)

packages/app/schema/nl/situations.json
```diff
@@ -0,0 +1,137 @@
...
+          "type": ["number", "null"]
+        },
+        "other": {
+          "type": ["number", "null"]
+        }
+      },
+      "required": [
+        "date_start_unix",
+        "date_end_unix",
+        "regions_with_sufficient_responses_total",
```
VWSCoronaDashboard8, line 108: nit; I think in this case you can leave out total, since there is no sub-total for this category. You can also use _count but it's redundant because the others are also about count. 

Alternatively you can prefix it with num_ or number_of_ so `num_regions_with_sufficient_responses` so that it reads a little better. Than it's also clear that it's not one of the categories.

VWSCoronaDashboard8 (approved):

packages/app/schema/nl/situations.json
VWSCoronaDashboard8: `on_visit`?

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard14 (approved):

PR #2824, opened 2021-05-12
New 'about escalation levels' page

VWSCoronaDashboard1:
## Summary

This PR adds a scoreboard and new layout to the 'about escalation levels' page:
image

## Motivation

Feature request

## Detailed design

Some minor changes were needed in the choropleth component as well because on this page the map doesn't have any click interaction. Therefore, the 'getLink' functionality needed to become optional.

### Governance

- [ ] Documentation is added
- [ ] Test cases are added or updated
- [ ] I've read the contributing document https://github.com/minvws/.github/blob/master/CONTRIBUTING.md
- [ ] I've read and understand the Code of Conduct https://github.com/minvws/.github/blob/master/CODE_OF_CONDUCT.md
- [ ] I understand that any contributions or suggestions I made may make it into the actual code. I've read the License
      https://github.com/minvws/nl-covid19-notification-app-coordination/blob/master/LICENSE.txt
      and the contributor license agreement https://github.com/minvws/nl-covid19-notification-app-coordination/blob/master/CLA.md

PR #2855, opened 2021-05-17
Refactor delivery and administration chart to use time-series-chart

VWSCoronaDashboard1:
## Summary

The new chart looks practically the same except for some minor visual details.
image

The biggest difference is the fact that in the 'estimated' part the grid lines are also dashed, this is because of the hatched overlay. I tried several approaches to get this exactly the same as the previous incarnation but wasn't successful.

## Motivation

Code cleanup.

## Detailed design

Some minor changes to the time-series-chart were needed in order to get this to work:
- The StackedAreaTrend now also has a `strokeWidth` property (it needed to be zero in this case)
- All series now have a unique id, this was needed for one of the earlier incarnations of the estimate annotation. But I decided to keep it in because I think they will come in handy once the charts will get more 'special fx' added.
- TimespanAnnotionConfig now has a `fill` property. For now, it can either be `solid` or `hatched`.
- The old area-chart component has been removed completely.

### Governance

- [ ] Documentation is added


VWSCoronaDashboard10 (changes requested):
Spotted a couple of bugs and I'm also not entirely sure we should have an 'estimate' type annotation. I cannot imagine we would use that annotation on other charts besides the delivery chart.

We might want to look at an annotation api which allows us to draw lines, points and text on the chart. In the future this api could be used to display "events" on certain dates, and the delivery chart could use it to draw the delivery-specific stuff.

Curious what @VWSCoronaDashboard8 thinks about this.

packages/app/src/static-props/get-data.ts
```diff
@@ -205,6 +216,12 @@ export function getGmData(context: GetStaticPropsContext) {
   return { data, municipalityName };
 }
 
+export function selectData<T extends (...args: any) => any>(
+  dataSelector: T
+): () => ReturnType<T> {
+  return () => dataSelector();
+}
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: Hm, you might actually have a point, hehe. I'll try it out.

packages/app/src/components/time-series-chart/logic/common.ts
```diff
@@ -11,13 +13,27 @@ export interface BenchmarkConfig {
...
+      type: 'estimate';
+      start: number;
+      end: number;
+      label: string;
+      shortLabel?: string;
+      cutValuesForMetricProperties?: string[];
+      shapeComponent?: ReactNode;
+      leftLabel: string;
+      rightLabel: string;
+    };
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: I discussed this with @VWSCoronaDashboard8 and we came to the conclusion that for now, it was OK.

packages/app/src/pages/landelijk/intensive-care-opnames.tsx
```diff
@@ -233,6 +235,7 @@ const IntakeIntensiveCare = (props: StaticProps<typeof getStaticProps>) => {
                 dataOptions={{
                   timespanAnnotations: [
                     {
+                      type: 'bar',
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: I'm open to suggestions.

packages/app/src/components/time-series-chart/components/series.tsx
```diff
@@ -66,6 +68,7 @@ function SeriesUnmemoized<T extends TimestampedValue>({
                   curve={config.curve}
                   getX={getX}
                   getY={getY}
+                  id={`${chartId}_${config.metricProperty}`}
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: 'In theory' being the operative word here. I discussed this particular implementation with @VWSCoronaDashboard8 where I had previously implemented a unique id for each series, but we came to the conclusion that the current implementation was good enough since its highly unlikely the same metric property will be displayed in the same chart twice, only in different visualisations.

packages/app/src/static-props/get-data.ts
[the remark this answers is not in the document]
VWSCoronaDashboard1: I guess this just saves the calling context from having to wrap one function.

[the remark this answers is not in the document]
VWSCoronaDashboard1: Ok, so I removed the `selectData()`, it was meant just for convenience, but I suppose we can live without it.

packages/app/src/domain/vaccine/data-selection/selected-delivery-and-administration-data.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export type DeliveryAndAdministrationData = {
+  values: VaccineDeliveryAndAdministrationsValue[];
+  estimatedRange: [number, number];
+};
+
+export function selectedDeliveryAndAdministrationData(): {
+  deliveryAndAdministration: DeliveryAndAdministrationData;
+} {
+  const nlData = getNlData().data;
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: Ah, the `selected` part is actually a typo...

packages/app/src/components/time-series-chart/components/patterns.tsx
```diff
@@ -0,0 +1,41 @@
+import { TimespanAnnotationConfig } from '../logic';
+
+type PatternsProps = {
+  timespanAnnotations: TimespanAnnotationConfig[] | undefined;
+  chartId: string;
+};
+
+/**
+ * For now only the estimate timespan annotation requires a pattern to be registered
+ * in the defs section, so I'm keeping that logic here.
```
VWSCoronaDashboard8, line 10: If this pattern relies on the defs being there. Why are the definitions configurable and not simply embedded / fixed?

packages/app/src/components/time-series-chart/components/patterns.tsx
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Should there be other components/annotations/etc that require these then I suggest
+ * we come up with a more generic approach to this when it becomes a requirement.
+ *
+ */
+export function Patterns({ timespanAnnotations, chartId }: PatternsProps) {
+  const estimates =
+    timespanAnnotations?.filter((x) => x.type === 'estimate') ?? [];
```
VWSCoronaDashboard8, line 18: If all you need is to know whether there are estimates, than extracting a boolean would be a bit cleaner maybe.

```suggestion
  const hasEstimateAnnotation =
    isDefined(timespanAnnotations?.find((x) => x.type === 'estimate'))
    
   return hasEsimate ? <pattern/> : null
   ```
   
   Also it returns one pattern, so I wonder why this component is called `Patterns` in plural
```

packages/app/src/components/time-series-chart/components/series.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard8: I think this is acceptable. I can't think of a practical situation where you'd want to render the same trend twice, unless you are trying to get a visual rendering as a combination of two trends, but I'd consider that a hack and then I think the proper solution would be to introduce a new trend type.

packages/app/src/static-props/get-data.ts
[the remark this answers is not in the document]
VWSCoronaDashboard1: By the way, the same `selectData()` method is also in this PR https://github.com/minvws/nl-covid19-data-dashboard/pull/2824. So that'll need to be tweaked as well.

packages/app/src/components/time-series-chart/components/timespan-annotation.tsx
```diff
@@ -35,19 +41,65 @@ export function TimespanAnnotation({
...
-      height={height}
-      x={x0}
-      width={width}
-      fill={colors.data.underReported}
-      opacity={1}
-      style={{ mixBlendMode: 'multiply' }}
-    />
-  );
+  switch (config.type) {
+    case 'bar':
```
VWSCoronaDashboard8, line 49: I'm not sure about the bar vs estimate naming. One is very generic and the other is very specific. We will probably only use the stripy stuff for estimates, but that doesn't mean we need to have this hardcoded in the chart. From the perspective of chart API I think it makes sense to keep it abstract as it shouldn't have to care about the concept of estimates.

Some suggestions:

`solid` and `hatched`
`solid` and `striped`

If you want to keep estimate I'd still rename bar to `block` as somehow it doesn't look like a bar to me.

packages/app/src/components/time-series-chart/logic/common.ts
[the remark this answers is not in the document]
VWSCoronaDashboard8: I agree that it would be nice if this can be made a little more generic. I already commented on the name estimate, and in this case we could choose to split this up into two different types of annotations. One is the hatched pattern, and the other could be a `Marker` or `LabelledMarker` which draws a line at a specific point and has props for setting labels on either side of the marker. This would be different from a timespan annotation, because the marker is just 1 point in time and doesn't span anything. So I'd suggest to pass that on a different prop `markers` on the chart.

packages/app/src/components/time-series-chart/logic/common.ts
```diff
@@ -11,13 +13,27 @@ export interface BenchmarkConfig {
...
-}
+export type TimespanAnnotationConfig =
+  | {
+      type: 'bar';
+      start: number;
+      end: number;
+      label: string;
+      shortLabel?: string;
+      cutValuesForMetricProperties?: string[];
+      shapeComponent?: ReactNode;
```
VWSCoronaDashboard8, line 24: I forgot why this annotation now wants a shapeComponent... Can't we keep the same component for all "bar" style legend items?

packages/app/src/components/time-series-chart/logic/common.ts
```diff
@@ -11,13 +13,27 @@ export interface BenchmarkConfig {
...
+    }
+  | {
+      type: 'estimate';
+      start: number;
+      end: number;
+      label: string;
+      shortLabel?: string;
+      cutValuesForMetricProperties?: string[];
+      shapeComponent?: ReactNode;
```
VWSCoronaDashboard8, line 33: And in this case, could we maybe standardize the legend items (and avoid making this configurable) Or is that something design wouldn't tolerate?

packages/app/src/components/time-series-chart/time-series-chart.tsx
```diff
@@ -321,6 +325,12 @@ export function TimeSeriesChart<
             onHover={chartEventHandlers.handleHover}
             onFocus={chartEventHandlers.handleFocus}
             onBlur={chartEventHandlers.handleBlur}
+            defs={
+              <Patterns
+                timespanAnnotations={timespanAnnotations}
+                chartId={chartId}
+              />
```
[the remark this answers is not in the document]
VWSCoronaDashboard8: Yes this makes the public API of the chart very odd, so would be great if we can avoid this.

packages/app/src/static-props/get-data.ts
```diff
@@ -146,18 +147,30 @@ export function getVrData(context: GetStaticPropsContext) {
...
+
+export function loadAndSortVrData(vrcode: string) {
+  const data = loadJsonFromDataFile<Regionaal>(`${vrcode}.json`);
+
+  sortTimeSeriesInDataInPlace(data);
+
+  return data;
+}
+
```
VWSCoronaDashboard10, line 173: I don't think these functions should be exported since they're only used in this module

packages/app/src/domain/vaccine/components/vaccine-delivery-and-administrations-tooltip.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const { siteText, formatNumber, formatDateFromSeconds } = useIntl();
+
+  const isEstimate =
+    data.value.date_start_unix >= estimateRange[0] &&
+    data.value.date_end_unix <= estimateRange[1];
+
+  const firstValue = data.value.total_delivered;
+  const firstConfig = data.config
+    .filter(isLineConfig)
+    .find((x) => x.metricProperty === 'total_delivered');
```
VWSCoronaDashboard8, line 43: Why are you filtering on isLineConfig? Could it be that there is another type of config with this same metricProperty?

packages/app/src/pages/landelijk/vaccinaties.tsx
```diff
@@ -51,15 +53,15 @@ export const getStaticProps = createGetStaticProps(
...
-    'vaccine_administered_estimate',
     'vaccine_administered_ggd',
     'vaccine_administered_hospitals_and_care_institutions',
     'vaccine_administered_doctors',
     'vaccine_administered_ggd_ghor'
   ),
+  () => {
+    const nlData = getNlData().data;
+    return selectDeliveryAndAdministrationData(nlData);
+  },
```
VWSCoronaDashboard10, line 64:
```suggestion
  () => selectDeliveryAndAdministrationData(getNlData().data),
```

packages/app/src/domain/vaccine/components/vaccine-delivery-and-administrations-tooltip.tsx
```diff
@@ -0,0 +1,152 @@
...
+  estimateRange,
+}: {
+  data: TooltipData<VaccineDeliveryAndAdministrationsValue>;
+  estimateRange: [number, number];
+}) {
+  const { siteText, formatNumber, formatDateFromSeconds } = useIntl();
+
+  const isEstimate =
+    data.value.date_start_unix >= estimateRange[0] &&
+    data.value.date_end_unix <= estimateRange[1];
```
VWSCoronaDashboard8, line 38: I believe the tooltip mechanism can pass what annotation is active as arguments to the render function, so maybe you can use that.

VWSCoronaDashboard10 (approved):

packages/app/src/domain/vaccine/components/vaccine-delivery-and-administrations-tooltip.tsx
```diff
@@ -0,0 +1,152 @@
...
+          <TooltipValueContainer>
+            <strong>{formatNumber(data.value.total)}</strong>
+          </TooltipValueContainer>
+        </TooltipListItem>
+      </TooltipList>
+    </>
+  );
+}
+
+function formatLabel(labelKey: string | undefined, text: SiteText) {
```
VWSCoronaDashboard8, line 115: Optional arguments typically come last, and then you con't need the | undefined. 

```suggestion
function formatLabel(text: SiteText, labelKey?: string) {
```

packages/app/src/domain/vaccine/components/vaccine-delivery-and-administrations-tooltip.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const labelText = labelKey
+    ? (text.vaccinaties.data.vaccination_chart.product_names as any)[labelKey]
+    : undefined;
+  return labelText ?? labelKey;
+}
```
VWSCoronaDashboard8, line 120: I find this whole function quite odd. Maybe pass in the product names object typed to record<string>. Something like this:

```suggestion
function formatLabel(productNames: Record<string, string>, productKey?: string) {
  return productKey
    ? productNames[productKey]
    : productKey;
}
```
But I wonder why this function needs to exist in the first place and why productKey is used as a fallback for rendering a product name.

packages/app/src/domain/vaccine/vaccine-delivery-and-administrations-area-chart.tsx
```diff
@@ -1,139 +1,95 @@
...
+        dataOptions={{
+          valueAnnotation: siteText.waarde_annotaties.x_miljoen,
+          timespanAnnotations: [
+            {
+              type: 'estimate',
+              start: data.estimatedRange[0],
+              end: data.estimatedRange[1],
+              label:
+                siteText.vaccinaties.data.vaccination_chart.legend.expected,
+              shapeComponent: <HatchedSquare />,
```
VWSCoronaDashboard8, line 69: I mentioned this earlier. Can't this just be the default / fixed way of rendering an estimate in the legend. I'd rather have it standardized if possible.

packages/app/src/domain/vaccine/vaccine-delivery-and-administrations-area-chart.tsx
```diff
@@ -1,139 +1,95 @@
...
+            metricProperty: x as keyof VaccineDeliveryAndAdministrationsValue,
+            type: 'stacked-area',
+            label: replaceVariablesInText(
+              siteText.vaccinaties.data.vaccination_chart.legend_label,
+              {
+                name: (productNames as any)[x],
+              }
+            ),
+            shortLabel: (productNames as any)[x],
+            color: (colors.data.vaccines as any)[x],
```
VWSCoronaDashboard8, line 110: You are using object lookup, so it's not any really

packages/app/src/pages/landelijk/intensive-care-opnames.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard8: Yes maybe we can make type optional with default to bar. It's a bit annoying having to set bar everywhere because one or two charts are using a different type of annotation.

packages/app/src/static-props/get-data.ts
```diff
@@ -146,18 +146,29 @@ export function getVrData(context: GetStaticPropsContext) {
...
     data,
-    safetyRegionName: safetyRegion?.name || '',
+    safetyRegionName,
   };
 }
 
+export function getVrName(code: string) {
+  const safetyRegion = vrData.find((x) => x.code === code);
+  return safetyRegion?.name || '';
+}
```
VWSCoronaDashboard8: Agree, I don't see why we would want to fall back to an empty string. An assertion seems appropriate to me.

VWSCoronaDashboard8 (changes requested):

packages/app/src/components/time-series-chart/components/chart-container.tsx
```diff
@@ -11,11 +11,13 @@
 import css from '@styled-system/css';
 import { Group } from '@visx/group';
 import React from 'react';
+import { isDefined } from 'ts-is-present';
 import { useIsTouchDevice } from '~/utils/use-is-touch-device';
 import { Padding } from '../logic';
 
 interface ChartContainerProps {
   children: React.ReactNode;
+  defs?: React.ReactNode;
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: I removed that functionality but forgot this prop. I'll remove it.

packages/app/src/components/time-series-chart/components/patterns.tsx
VWSCoronaDashboard1: Patterns are always part of the defs section, not sure what you mean

packages/app/src/components/time-series-chart/components/patterns.tsx
```diff
@@ -0,0 +1,41 @@
+import { TimespanAnnotationConfig } from '../logic';
+
+type PatternsProps = {
+  timespanAnnotations: TimespanAnnotationConfig[] | undefined;
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: I had already added a comment about this...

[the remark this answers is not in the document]
VWSCoronaDashboard1: But, since it confuses you, I'll just rename it.

packages/app/src/components/time-series-chart/components/patterns.tsx
VWSCoronaDashboard1: Again, I had added a comment to the component. But I guess comments aren't read hehe. So I've renamed it.

packages/app/src/components/time-series-chart/logic/common.ts
VWSCoronaDashboard1: Please read the PR comments...

packages/app/src/components/time-series-chart/logic/common.ts
VWSCoronaDashboard1: I don't know, if we don't add the hatched pattern to the legend icon it will just become a grey square. So, either we don't show it or we show it with the pattern...

packages/app/src/domain/vaccine/components/vaccine-delivery-and-administrations-tooltip.tsx
VWSCoronaDashboard1: No, in this case its to narrow the type to a type that has a property called 'metricProperty'

packages/app/src/domain/vaccine/components/vaccine-delivery-and-administrations-tooltip.tsx
VWSCoronaDashboard1: I did not know this, I'll have a look. it sounds cleaner.

packages/app/src/components/time-series-chart/logic/common.ts
VWSCoronaDashboard1: Oh wait, I read this wrong, but the solid annotation doesn't need a shapeComponent I suppose

packages/app/src/domain/vaccine/components/vaccine-delivery-and-administrations-tooltip.tsx
VWSCoronaDashboard1: This was copied from the old tooltip, I didn't look into it that deeply.But you're right, the entire formatLabel function isn't needed. I removed it.

packages/app/src/static-props/get-data.ts
VWSCoronaDashboard1: They're not used in this PR, but they are used in the scoreboard PR where a selection of all the VRs is being loaded.

VWSCoronaDashboard1: (I copied the get-data.ts from that PR)

packages/app/src/components/time-series-chart/components/timespan-annotation.tsx
VWSCoronaDashboard1: So, I'm ok with the re-naming. But now that the estimate annotation is called 'hatched'. I don't think the extra line with the two labels can be part of the annotation anymore. (we discussed this before I started implementing this). I guess we'll have to come up with another data option or the third type of annotation.
For now, I'll implement it as a third type. I think that makes the most sense here.

packages/app/src/domain/vaccine/vaccine-delivery-and-administrations-area-chart.tsx
VWSCoronaDashboard1: Ok, I guess that's possible. I'll make the change.

packages/app/src/components/time-series-chart/components/timespan-annotation.tsx
VWSCoronaDashboard1: Hm, I've implemented it as new type of timespan annotation. But, it already doesn't make sense.
I think we need to add a new type of annotation. Let's call it a 'time-annotation' for now.
I've committed the span implementation, so you can see for yourself. But, it doesn't make sense since the divider only has one x coordinate, not a start and end.
So, this already proves imho, that it needs to be a separate thing. Its getting late, let's discuss this tomorrow morning.

VWSCoronaDashboard1: Ok, so the OCD brain wouldn't let me leave this unfinished. So to avoid not being able to sleep I implemented the divider as a new kind of annotation. I called it TimeAnnotation for now. Naming, etc is open for discussion, obviously. Check it out...

VWSCoronaDashboard1: I don't pass the time annotations to the tooltip yet, for instance. We can discuss whether that makes sense or not.

packages/app/src/components/time-series-chart/time-series-chart.tsx
```diff
@@ -393,6 +393,15 @@ export function TimeSeriesChart<
                 config={x}
               />
             ))}
+            {timeAnnotations?.map((x, index) => (
+              <TimeAnnotation
+                key={index}
+                domain={xScale.domain() as [number, number]}
```
VWSCoronaDashboard10, line 399: I don't think we should cast this to a tuple of numbers because the array could hold more than 2 numbers. It's typed like `number[]` on purpose.
```suggestion
                domain=[first({xScale.domain()), last({xScale.domain())]}
```

packages/app/src/components/time-series-chart/components/timespan-annotation.tsx
```diff
@@ -58,60 +51,29 @@ export function TimespanAnnotation({
...
-            />
-          </Group>
-        </>
+        <Bar
+          pointerEvents="none"
+          height={height}
+          x={x0}
+          width={width}
+          fill={`url(#${chartId}_hatched_pattern)`}
+        />
```
VWSCoronaDashboard10, line 62: I think you should render the hatch pattern over here, extra advantage is that you can get rid of the timespanAnnotation checks in the pattern component, also the id can be defined on this level.
```suggestion
        <>
          <HatchedPattern id={`${chartId}_hatched_pattern`} />
          <Bar
            pointerEvents="none"
            height={height}
            x={x0}
            width={width}
            fill={`url(#${chartId}_hatched_pattern)`}
          />
        </>
```

packages/app/src/components/time-series-chart/components/timespan-annotation.tsx
```diff
@@ -127,3 +89,29 @@ export function TimespanAnnotationIcon({
     </svg>
   );
 }
+
+export function HatchedTimespanAnnotationIcon() {
+  return (
+    <svg height="15" width="15">
```
VWSCoronaDashboard10, line 95: Can you make sure this icon has the same width/height/viewbox implementation as the other icon?
```suggestion
export function HatchedTimespanAnnotationIcon({
  width = 15,
  height = 15,
}: TimespanAnnotationIconProps) {
  return (
    <svg width={width} height={height} viewBox={`0 0 ${width} ${height}`}>
```

packages/app/src/components/time-series-chart/components/timespan-annotation.tsx
```diff
@@ -127,3 +89,29 @@ export function TimespanAnnotationIcon({
...
+          height="4"
+          patternTransform="rotate(-45 0 0)"
+          patternUnits="userSpaceOnUse"
+        >
+          <line
+            x1="0"
+            y1="0"
+            x2="0"
+            y2="5"
+            style={{ stroke: 'grey', strokeWidth: 3 }}
```
VWSCoronaDashboard10, line 109: did you mean `gray`?
```suggestion
            style={{ stroke: 'gray', strokeWidth: 3 }}
```

packages/app/src/components/time-series-chart/logic/common.ts
```diff
@@ -15,26 +14,29 @@ export interface BenchmarkConfig {
 
 export type TimespanAnnotationConfig =
   | {
-      type: 'bar';
+      type: 'solid';
```
VWSCoronaDashboard10, line 17: I think we can get rid of the type property and have a single timespan annotation type with a `fill` prop
```suggestion
      fill?: 'solid' | 'hatched'; // optional so the default can be `solid`
```

packages/app/src/components/time-series-chart/time-series-chart.tsx
```diff
@@ -325,13 +327,11 @@ export function TimeSeriesChart<
...
-              <Patterns
-                timespanAnnotations={timespanAnnotations}
-                chartId={chartId}
-              />
-            }
           >
+            <HatchedPattern
+              timespanAnnotations={timespanAnnotations}
+              chartId={chartId}
+            />
```
VWSCoronaDashboard10, line 334: see other comments about moving this component to the TimespanAnnotation

packages/app/src/domain/vaccine/vaccine-delivery-and-administrations-area-chart.tsx
```diff
@@ -9,8 +11,23 @@ import { VaccineDeliveryAndAdministrationsTooltip } from './components/vaccine-d
...
+  'pfizer',
+  'moderna',
+  'astra_zeneca',
+  'cure_vac',
+  'janssen',
+  'sanofi',
+] as (keyof Pick<
+  VaccineDeliveryAndAdministrationsValue,
+  'pfizer' | 'moderna' | 'astra_zeneca' | 'cure_vac' | 'janssen' | 'sanofi'
+>)[];
```
VWSCoronaDashboard10, line 26: Why is this typed so complex compared to `as const`?
```suggestion
const vaccines = [
  'pfizer',
  'moderna',
  'astra_zeneca',
  'cure_vac',
  'janssen',
  'sanofi',
] as const;
```

packages/app/src/domain/vaccine/vaccine-delivery-and-administrations-area-chart.tsx
```diff
@@ -19,7 +36,13 @@ export function VaccineDeliveryAndAdministrationsAreaChart({
 }) {
   const { siteText, formatNumber } = useIntl();
   const firstValue = first(data.values);
-  const vaccineNames = useVaccineNames(firstValue).reverse();
+  const vaccineNames = useMemo(
+    () =>
+      vaccines
+        .filter((x) => firstValue && firstValue[x] !== undefined)
+        .reverse(),
+    [vaccines, firstValue]
```
VWSCoronaDashboard10, line 44: Unnecessary to add vaccines to the dep array since it's defined outside of the function component and won't change
```suggestion
    [firstValue]
```

VWSCoronaDashboard10 (changes requested):

packages/app/src/components/time-series-chart/components/timespan-annotation.tsx
VWSCoronaDashboard1: Ok, it also means we can only add one hatched annotation, otherwise we're duplicating the pattern. Thats why it was in the defs section in the first place.

packages/app/src/components/time-series-chart/time-series-chart.tsx
VWSCoronaDashboard1: Please expand your code suggestion to also cover the `TimespanAnnotation` component, this also takes the domain as a number tuple. That's where I copied this code from. ;)

packages/app/src/components/time-series-chart/logic/common.ts
VWSCoronaDashboard1: And what if there's a third type next week?

packages/app/src/domain/vaccine/vaccine-delivery-and-administrations-area-chart.tsx
VWSCoronaDashboard1: Because it's been used as an indexer, so simply using strings will make the compiler sad

packages/app/src/components/time-series-chart/time-series-chart.tsx
VWSCoronaDashboard10: Ay yes might be better to turn it around;
```tsx
export function TimespanAnnotation({
  start,
  end,
  domain,
  getX,
  height,
}: {
  start: number;
  end: number;
  domain: number[];
  height: number;
  getX: GetX;
}) {
  const min = first(domain);
  const max = last(domain);
  assert(
    typeof min === 'number' && typeof max === 'number',
    `received invalid domain ${domain}`
  );

File: packages/app/src/components/time-series-chart/components/timespan-annotation.tsx (hunk 24 above)
Comment (VWSCoronaDashboard10) [reply packages/app/src/components/time-series-chart/components/timespan-annotation.tsx:62]: I don't think duplicating a pattern is a big issue :-)

File: packages/app/src/components/time-series-chart/time-series-chart.tsx (hunk 23 above)
Comment (VWSCoronaDashboard1) [reply packages/app/src/components/time-series-chart/time-series-chart.tsx:399]: Addendum, while it is, in theory, possible there's a domain consisting of more than two values. In the time-series-chart all the domains are explicitly created with two values. So, I'm not sure if this changes is really that necessary...

File: packages/app/src/components/time-series-chart/components/timespan-annotation.tsx (hunk 24 above)
Comment (VWSCoronaDashboard1) [reply packages/app/src/components/time-series-chart/components/timespan-annotation.tsx:62]: But I also fail to see why the change is necessary then. I don't think I'm adding an incredible amount of complexity here...

File: packages/app/src/components/time-series-chart/components/timespan-annotation.tsx (hunk 25 above)
Comment (VWSCoronaDashboard1) [reply packages/app/src/components/time-series-chart/components/timespan-annotation.tsx]: Done

File: packages/app/src/components/time-series-chart/components/timespan-annotation.tsx (hunk 26 above)
Comment (VWSCoronaDashboard1) [reply packages/app/src/components/time-series-chart/components/timespan-annotation.tsx]: Fixed

File: packages/app/src/domain/vaccine/vaccine-delivery-and-administrations-area-chart.tsx (hunk 30 above)
Comment (VWSCoronaDashboard1) [reply packages/app/src/domain/vaccine/vaccine-delivery-and-administrations-area-chart.tsx]: Fixed

File: packages/app/src/domain/vaccine/vaccine-delivery-and-administrations-area-chart.tsx (hunk 29 above)
Comment (VWSCoronaDashboard10) [reply packages/app/src/domain/vaccine/vaccine-delivery-and-administrations-area-chart.tsx:26]: It looks like the compiler currently complains: https://github.com/minvws/nl-covid19-data-dashboard/pull/2855/checks?check_run_id=2617693033

I also think the compiler is complaining because the `vaccines` array contains more types of vaccines compared to the vaccines coming from lokalize. On my local checkout the following results in no TS errors:
```tsx
const vaccines = [
  'astra_zeneca',
  'janssen',
  'moderna',
  'pfizer',
  // 'cure_vac',
  // 'sanofi',
] as const;
```

VWSCoronaDashboard1: See line 42, 111, 114 and 115 for reference

packages/app/src/components/time-series-chart/time-series-chart.tsx
VWSCoronaDashboard1: I think the domain check is a good one though, that definitely simplifies the code, so I'm using that

packages/app/src/components/time-series-chart/logic/common.ts
VWSCoronaDashboard10: That's not really an argument 😛  we've been working with a single type for several months, also a solid/hatched fill makes sense configuration-wise imho, but do what feels best.

packages/app/src/components/time-series-chart/time-series-chart.tsx
VWSCoronaDashboard1: Oops, no, I thought your code suggestion was valid. But now I realize that the values are already being clipped. As the comment states, this allows for Infinity to be passed in as an end date.

packages/app/src/components/time-series-chart/logic/common.ts
VWSCoronaDashboard1: We've also been working with a domain that consists of two values for several months, yet you insist on checking for larger domains...
But, if it helps you sleep at night, I'll change this to the fill prop solution. I'm getting a bit tired of the back and forth here.

packages/app/src/domain/vaccine/vaccine-delivery-and-administrations-area-chart.tsx
VWSCoronaDashboard1: If remove those vaccine names then we'll have to make a manual change here once those vaccines are on the market. We added the entire list so that BE only has to add the values and this chart will automatically 'grow' accordingly...

packages/app/src/components/time-series-chart/components/timespan-annotation.tsx
VWSCoronaDashboard10: Well I actually think it is:
- there's an unclear link between a HatchedPattern and a specific case of a timespanannotation
- It's weird that a pattern should receive information about timespan annotations
- The timespanAnnotation component could be rendered by different chart implementations (e.g. vertical bar chart, mini bar chart). It would be a broken component because it has unclear dependencies on a component rendered somewhere else.

packages/app/src/components/time-series-chart/logic/common.ts
VWSCoronaDashboard10: > We've also been working with a domain that consists of two values for several months, yet you insist on checking for larger domains..

that's because I read that in d3 documentation last week ;-)

packages/app/src/domain/vaccine/vaccine-delivery-and-administrations-area-chart.tsx
VWSCoronaDashboard10: Sure, makes sense, but another way of fixing is by adding `cure_vac` and `sanofi` in lokalize. Otherwise BE is still unable to add values without our help.

VWSCoronaDashboard1: They are in the Lokalise, no?

packages/app/src/components/time-series-chart/components/timespan-annotation.tsx
VWSCoronaDashboard1: Hm, I think I placed the hatch-pattern component inside the 'components' directory of the time-series-chart. WHich kind of implies its tied to that component.
Currently, the hatched pattern is tied to the timespan annotation, that's why I placed the check inside that component, with the idea that if more patterns would be needed in the future this logic would be lifted outside this component to more upstream logic that would decide which patterns wound need to be rendered or not.

packages/app/src/domain/vaccine/vaccine-delivery-and-administrations-area-chart.tsx
VWSCoronaDashboard1: At least, in my nl_export.json they are...

packages/app/src/components/time-series-chart/components/timespan-annotation.tsx
VWSCoronaDashboard8: I do remember talking about the labels in combination with the annotation but I don't remember what the outcome/motivation was to be honest. Reading your code, and how specific the implementation for estimate was, I feel it would be better to break up the timespan annotation and the marker line with the labels into two separate concepts.

Do I'm guessing what I call a marker you are here referring to as a TimeAnnotation? Seems fine with me too.

packages/app/src/components/time-series-chart/components/timespan-annotation.tsx
VWSCoronaDashboard10: I understand, but it's still unnecessarily complex if there's a simpler implementation possible?

packages/app/src/components/time-series-chart/logic/common.ts
VWSCoronaDashboard8: I thought you could choose the shapeComponent based on the type of annotation. So when it's estimate you use the hatched pattern, without specifying it in the config. But possibly I'm missing some things here.

packages/app/src/domain/vaccine/vaccine-delivery-and-administrations-area-chart.tsx
VWSCoronaDashboard10: that's odd, because they are not part of the dataset:
image

packages/app/src/components/time-series-chart/components/time-annotation.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const [min, max] = domain;
+  const { position } = config;
+
+  const breakpoints = useBreakpoints(true);
+
+  /**
+   * Clip the position to the domain of the x-axis, so that we can
+   * conveniently pass in things like Infinity for end date.
+   */
+  const clippedPosition = Math.min(Math.max(position, min), max);
```
VWSCoronaDashboard8, line 28: Wouldn't this result in the annotation being positioned on one side of the chart, even though according to its position it should lie outside of the chart / is not visible?

I think if the position of the annotation doesn't fall within the domain, we can simply not draw anything.

packages/app/src/components/time-series-chart/components/time-annotation.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  const x = getX({ __date_unix: clippedPosition });
+
+  const fontSize = !breakpoints.lg ? theme.fontSizes[0] : theme.fontSizes[1];
+
+  switch (config.type) {
+    case 'divider':
```
VWSCoronaDashboard8, line 35: 👍  I'm not sure we'll need others in the future, but this is a clear name.

packages/app/src/components/time-series-chart/logic/common.ts
```diff
@@ -1,39 +1,32 @@
...
+  shortLabel?: string;
+  cutValuesForMetricProperties?: string[];
+};
+
+export type TimeAnnotationConfig = {
+  type: 'divider';
+  position: number;
+  leftLabel: string;
+  rightLabel: string;
+};
```
VWSCoronaDashboard8: 🥳

packages/app/src/domain/vaccine/vaccine-delivery-and-administrations-area-chart.tsx
VWSCoronaDashboard1: Oh, I guess I have an old export locally then. These keys have probably been removed during cleanup, I suppose.

packages/app/src/domain/vaccine/components/vaccine-delivery-and-administrations-tooltip.tsx
```diff
@@ -26,16 +25,14 @@ function isStackedConfig(
...
-  estimateRange: [number, number];
 }) {
+  data.timespanAnnotation;
+
   const { siteText, formatNumber, formatDateFromSeconds } = useIntl();
 
-  const isEstimate =
-    data.value.date_start_unix >= estimateRange[0] &&
-    data.value.date_end_unix <= estimateRange[1];
+  const isEstimate = data.timespanAnnotation?.fill === 'hatched';
```
VWSCoronaDashboard8, line 35: 👍

VWSCoronaDashboard8 (approved):
Nice (team) work on this one!

packages/app/src/domain/vaccine/vaccine-delivery-and-administrations-area-chart.tsx
VWSCoronaDashboard10: whoopsy, must have been my bad then

VWSCoronaDashboard10: don't forget to re-add those vaccines again with the cli :-)

packages/app/src/domain/vaccine/use-vaccine-names.ts
```diff
@@ -1,6 +1,7 @@
 import { assert, NlVaccineAdministeredValue } from '@corona-dashboard/common';
```
VWSCoronaDashboard10, line 1: I think this file `use-vaccine-names` can be removed now, looks like it's not used anymore

packages/app/src/components/time-series-chart/components/time-annotation.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  const breakpoints = useBreakpoints(true);
+
+  if (position < min || position > max) {
+    return null;
+  }
+
+  const x = getX({ __date_unix: position });
+
+  const fontSize = !breakpoints.lg ? theme.fontSizes[0] : theme.fontSizes[1];
```
VWSCoronaDashboard10, line 30: shouldn't this ideally be based on the chart width instead of a breakpoint

packages/app/src/components/time-series-chart/components/time-annotation.tsx
```diff
@@ -0,0 +1,66 @@
...
+  switch (config.type) {
+    case 'divider':
+      return (
+        <Group>
+          <Text
+            fontSize={fontSize}
+            x={x - 15}
+            y={15}
+            textAnchor="end"
+            fill="grey"
```
VWSCoronaDashboard10, line 41: Shouldn't this become a color from the theme?

packages/app/src/components/time-series-chart/components/time-annotation.tsx
```diff
@@ -0,0 +1,66 @@
...
+          >
+            {config.leftLabel}
+          </Text>
+          <Text
+            fontSize={fontSize}
+            x={x + 15}
+            y={15}
+            textAnchor="start"
+            fill="grey"
```
VWSCoronaDashboard10, line 50: idem

packages/app/src/components/time-series-chart/components/time-annotation.tsx
```diff
@@ -0,0 +1,66 @@
...
+          >
+            {config.rightLabel}
+          </Text>
+
+          <Line
+            x1={x}
+            x2={x}
+            y1={height}
+            y2={0}
+            stroke="gray"
```
VWSCoronaDashboard10, line 60: idem

packages/app/src/components/time-series-chart/components/timespan-annotation.tsx
```diff
@@ -38,28 +40,51 @@ export function TimespanAnnotation({
...
+
+      <Bar
+        pointerEvents="none"
+        height={height}
+        x={x0}
+        width={width}
+        fill={
+          fill === 'solid'
+            ? colors.data.underReported
+            : `url(#${chartId}_hatched_pattern)`
```
VWSCoronaDashboard10, line 70:
```tsx
const patternId = `${chartId}_hatched_pattern`;

// ...

<pattern id={patternId}>

// ...

fill={`url(#${patternId})`}
```

packages/app/src/components/time-series-chart/components/timespan-annotation.tsx
```diff
@@ -75,3 +100,37 @@ export function TimespanAnnotationIcon({
...
+          height="4"
+          patternTransform="rotate(-45 0 0)"
+          patternUnits="userSpaceOnUse"
+        >
+          <line
+            x1="0"
+            y1="0"
+            x2="0"
+            y2="5"
+            style={{ stroke: 'gray', strokeWidth: 3 }}
```
VWSCoronaDashboard10, line 128: theme color?

packages/app/src/domain/vaccine/components/vaccine-delivery-and-administrations-tooltip.tsx
```diff
@@ -0,0 +1,142 @@
...
+            </span>
+            <TooltipValueContainer>
+              <strong>
+                {formatNumber(data.value[config.metricProperty] || 0)}
+              </strong>
+            </TooltipValueContainer>
+          </TooltipListItem>
+        ))}
+
+        <Spacer mb={1} />
```
VWSCoronaDashboard10, line 93: I guess this component cannot be a child of a `<ol />`

packages/app/src/domain/vaccine/components/vaccine-delivery-and-administrations-tooltip.tsx
```diff
@@ -0,0 +1,142 @@
...
+const TooltipListItem = styled.li`
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+`;
+
+const TooltipValueContainer = styled.span`
+  font-weight: bold;
+  margin-left: 1em;
+`;
```
VWSCoronaDashboard10, line 142: we have fontweights in our theme, using the SC backtick notation ignores all theme values
```suggestion
const TooltipValueContainer = styled.span(css({
  fontWeight: 'bold',
  ml: '1em' // or pick a theme value
  }));
```

packages/app/src/domain/vaccine/data-selection/select-delivery-and-administration-data.ts
```diff
@@ -0,0 +1,53 @@
...
+      ...x,
+    })),
+    ...vaccine_administered_estimate.values.map((x, index) => ({
+      ...{
+        ...vaccine_delivery_estimate.values[index],
+        total_delivered: vaccine_delivery_estimate.values[index].total,
+      },
+      ...x,
+    })),
+  ];
```
VWSCoronaDashboard10, line 42: I think something like this is easier for the eyes:
```tsx
const values: VaccineDeliveryAndAdministrationsValue[] = [];

for (const [index] of vaccine_administered.values.entries()) {
  values.push({
    total_delivered: vaccine_delivery.values[index].total,
    ...vaccine_delivery.values[index],
    ...vaccine_administered.values[index],
  });
}

for (const [index] of vaccine_administered_estimate.values.entries()) {
  values.push({
    total_delivered: vaccine_delivery_estimate.values[index].total,
    ...vaccine_delivery_estimate.values[index],
    ...vaccine_administered_estimate.values[index],
  });
}

**Review by VWSCoronaDashboard10 — APPROVED** [review]
Couple of clean up thingies but I guess we're there 👍

File: packages/app/src/domain/vaccine/use-vaccine-names.ts (hunk 35 above)
Comment (VWSCoronaDashboard1) [reply packages/app/src/domain/vaccine/use-vaccine-names.ts:1]: I think you're correct ;)

File: packages/app/src/components/time-series-chart/components/time-annotation.tsx (hunk 36 above)
Comment (VWSCoronaDashboard1) [reply packages/app/src/components/time-series-chart/components/time-annotation.tsx:30]: You're probably right. This is also an issue in the x ticks PR. I'm going to implement a `useChartBreakPoints` feature there, so for now I'm going to leave this one as is.

File: packages/app/src/components/time-series-chart/components/timespan-annotation.tsx [hunk 45]
```diff
@@ -43,7 +44,7 @@ export function TimespanAnnotation({
     <>
       {fill === 'hatched' && (
         <pattern
-          id={`${chartId}_hatched_pattern`}
+          id={`${patternId}`}
```
VWSCoronaDashboard10, line 47:
```suggestion
          id={patternId}
```

packages/app/src/domain/vaccine/components/vaccine-delivery-and-administrations-tooltip.tsx
```diff
@@ -90,8 +90,7 @@ export function VaccineDeliveryAndAdministrationsTooltip({
           </TooltipListItem>
         ))}
 
-        <Spacer mb={1} />
-        <TooltipListItem>
+        <TooltipListItem mt={1}>
```
VWSCoronaDashboard10, line 93: it's a styled component so you could go for the following and keep the TooltipListItem simple:
```suggestion
        <TooltipListItem css={css({{ mt: 1 })}>
```

PR #2866, opened 2021-05-19
Refactor population count in schemas

VWSCoronaDashboard8:
## Summary

In order to play nicely with Datatino we decided to nest the population count under a static_values property in the root, because the root can only hold strings or objects.


packages/app/schema/gm/__static_values.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "gm_static_values",
+  "type": "object",
+  "properties": {
+    "population_count": {
+      "type": "integer"
+    }
+  },
+  "required": [],
```
VWSCoronaDashboard4, line 10: `population_count` is optional, is that intended?

VWSCoronaDashboard4 (approved):

packages/app/schema/gm/__static_values.json
VWSCoronaDashboard8: I did this so the FE could be implemented using a feature flag. Not sure if we need it. I think we could make all new metrics (or metric properties) optional so we'll be able to merge them in develop without waiting for BE.

VWSCoronaDashboard1 (approved):

PR #2867, opened 2021-05-19
Add redesigned compliance and support chart

VWSCoronaDashboard13:
## Summary

Adding the redesigned compliance and support chart 

## Detailed design

Created a `redesigned-behavior-line-chart-tile` component to house the new behavior line chart tile which looks very similar to the previous chart however it checks if either support or compliance has enough data to plot a line 

I've added the locale to the gegrad_common subject since the text should not differ on nl vs vr

This does not include the update to the select component. I have made a separate ticket for that https://trello.com/c/nqCpFaH7/1917-behavior-redesign-update-select-component-to-show-optional-icon


packages/app/src/domain/behavior/redesigned-behavior-line-chart-tile.tsx
```diff
@@ -0,0 +1,104 @@
...
+          .filter(isPresent).length > 1;
+
+      return complianceHasEnoughData || supportHasEnoughData
+        ? {
+            id,
+            label,
+          }
+        : undefined;
+    })
+    .filter(isPresent);
```
VWSCoronaDashboard4, line 58: What if `currentId` does not end up in this list? Perhaps init `currentId` 's state after filtering this list, and use the first item to mark it as the current?

VWSCoronaDashboard4 (approved):

packages/app/src/domain/behavior/redesigned-behavior-line-chart-tile.tsx
VWSCoronaDashboard13: yeah, this makes sense! I just updated it and caught another bug too 👍

PR #2870, opened 2021-05-20
Split area chart

VWSCoronaDashboard8:
## Summary

image

image


packages/app/src/components/sewer-chart/new-sewer-chart.tsx
```diff
@@ -0,0 +1,226 @@
...
+                    type: 'line',
+                    metricProperty: 'selected_installation_rna_normalized',
+                    label: selectedInstallation,
+                    color: 'black',
+                    style: 'dashed',
+                  },
+                  {
+                    type: 'split-area',
+                    metricProperty: 'average',
+                    label: 'Weekgemiddelde',
```
VWSCoronaDashboard1, line 100: I think this ought to be a lokalize key

packages/app/src/components/sewer-chart/new-sewer-chart.tsx
```diff
@@ -0,0 +1,226 @@
...
+            ) : (
+              <TimeSeriesChart
+                values={dataAverages.values}
+                timeframe={timeframe}
+                markNearestPointOnly
+                seriesConfig={[
+                  {
+                    type: 'split-area',
+                    metricProperty: 'average',
+                    label: 'Weekgemiddelde',
```
VWSCoronaDashboard1, line 114: Same here

packages/app/src/components/time-series-chart/logic/use-format-series-value.ts
```diff
@@ -34,6 +35,14 @@ export function useFormatSeriesValue() {
...
     ) {
       switch (config.type) {
+        case 'line':
+        case 'split-area':
+        case 'area':
+        case 'bar':
+        case 'split-bar':
+        case 'stacked-area':
+        case 'invisible':
+          return getValueString(value[config.metricProperty], isPercentage);
```
VWSCoronaDashboard1, line 45: Since 'range' is the only exception here, this entire block can be removed since the default case already returns the given `getValueString()`

VWSCoronaDashboard1 (approved):

packages/app/src/components/sewer-chart/new-sewer-chart.tsx
```diff
@@ -0,0 +1,226 @@
...
+      {(timeframe) => (
+        <>
+          {dataPerInstallation && (
+            <div css={css({ alignSelf: 'flex-start', mb: 2 })}>
+              <Select
+                options={options}
+                onChange={onChange}
+                onClear={onClear}
+                value={selectedInstallation}
+                placeholder={text.selectPlaceholder || '__no_placeholder'}
```
VWSCoronaDashboard10, line 71: why fallback to `__no_placeholder`?

packages/app/src/components/sewer-chart/new-sewer-chart.tsx
```diff
@@ -0,0 +1,226 @@
+import {
```
VWSCoronaDashboard10, line 1: Would it make sense to move this file to a domain directory?

packages/app/src/components/time-series-chart/logic/legend.tsx
```diff
@@ -18,12 +18,13 @@ export function useLegendItems<T extends TimestampedValue>(
...
     const items = config.filter(isVisible).flatMap<LegendItem>((x) => {
       switch (x.type) {
+        case 'split-area':
         case 'split-bar':
           return x.splitPoints.map((v) => ({
             color: v.color,
             label: v.label,
             shape: 'custom',
-            shapeComponent: <SeriesIcon config={x} value={v.value - 1} />,
+            shapeComponent: <SeriesIcon config={x} value={v.value} />,
```
VWSCoronaDashboard10, line 27: I remember the `value - 1` was necessary in order to display the correct color, maybe I'm mistaken but I guess it's worth it to make sure it isn't broken?

VWSCoronaDashboard10 (approved):

packages/app/src/components/time-series-chart/logic/legend.tsx
VWSCoronaDashboard8: Ah right, I think I've solved this elsewhere by turning > into >=. But yeah should be QAed

packages/app/src/components/sewer-chart/new-sewer-chart.tsx
VWSCoronaDashboard8: Yes I think so, but maybe let's do that after the feature is enabled on production and we can retire the old sewer chart code that is still in components also.

packages/app/src/components/sewer-chart/new-sewer-chart.tsx
VWSCoronaDashboard8: It's a bit of a hack. The placeholder is mandatory on the Select component, but the select is not rendered for NL (only VR and GM) therefor the NewSewerChart has the placeholder as an optional string. Effectively the __no_placeholder string is therefor never used, because in is only missing on NL and there the select is not rendered.

I guess a better solution would be to make placeholder optional on Select again, and have a decent fallback there.

VWSCoronaDashboard8: Never mind, I see that placeholder is now optional on Select already (thought I made it mandatory). So I can just remove this bit of code

PR #2873, opened 2021-05-20
Add new choropleth tile behavior page

VWSCoronaDashboard14:
This features 2 choropleths next to each other sharing the same data in the tooltip
Added a new threshold
Kinda worked my way around the Choropleth component and the way he receives data to load in the correct threshold


packages/app/src/components/choropleth/region-thresholds.ts
```diff
@@ -168,6 +168,68 @@ const behaviorThresholds: ChoroplethThresholdsValue[] = [
   },
 ];
 
+const behaviorComplienceThresholds: ChoroplethThresholdsValue[] = [
```
VWSCoronaDashboard1, line 171:
```suggestion
const behaviorComplianceThresholds: ChoroplethThresholdsValue[] = [
```

packages/app/src/components/choropleth/region-thresholds.ts
```diff
@@ -248,6 +310,8 @@ export const regionThresholds = {
     average: sewerThresholds,
   },
   behavior: behaviorThresholds,
+  behavior_compliance: behaviorComplienceThresholds,
```
VWSCoronaDashboard1, line 313:
```suggestion
  behavior_compliance: behaviorComplianceThresholds,
```

packages/app/src/components/choropleth/choropleth.tsx
```diff
@@ -346,7 +346,7 @@ const createSvgMouseOutHandler = (
 ) => {
   return () => {
     if (timeout.current < 0) {
-      timeout.current = window.setTimeout(() => setTooltip(undefined), 500);
+      timeout.current = window.setTimeout(() => setTooltip(undefined), 10);
```
VWSCoronaDashboard1, line 349: why such a small timeout?

packages/app/src/components/choropleth/tooltips/region/behavior-tooltip.tsx
```diff
@@ -0,0 +1,92 @@
...
+    <TooltipContent
+      title={context.vrname}
+      link={reverseRouter.vr.gedrag(context.vrcode)}
+    >
+      <Box maxWidth="15rem">
+        <Text m={0} mb={2} fontWeight="bold">
+          {siteText.gedrag_onderwerpen[currentMetric]}
+        </Text>
+        <TooltipInfo
+          title="Naleving"
```
VWSCoronaDashboard1, line 52: This title should probably come from the siteText

packages/app/src/domain/behavior/behavior-choropleth.tsx
```diff
@@ -0,0 +1,167 @@
...
+import { useReverseRouter } from '~/utils/use-reverse-router';
+import { BehaviorIdentifier, behaviorIdentifiers } from './behavior-types';
+
+interface BehaviorChoroplethProps {
+  title: string;
+  description: string;
+  data: RegionsBehavior[];
+}
+
+export function BehaviorChoropleth({
```
VWSCoronaDashboard1, line 27: This name is not appropriate for a component that shows two choropleths.  At the very least it ought to be called `BehaviorChoropleths`, but perhaps `BehaviorChoroplethsTile` is better.

VWSCoronaDashboard1 (changes requested):

packages/app/src/components/choropleth/choropleth.tsx
VWSCoronaDashboard14: Unfortunately, I just picked a number, couldn't find any drawbacks when lowering it to 10 (and it looks good having not to wait 0.5 seconds for a tooltip to disappear). But it might be good to test it a bit more and if it does impact other choropleths performances.

VWSCoronaDashboard1: AFAICR, the 500ms wait was chosen so that switching between tooltips won't result in a flicker. But if 10ms works too, then fine.

packages/app/src/components/choropleth/safety-region-choropleth.tsx
```diff
@@ -32,6 +32,7 @@ type SafetyRegionChoroplethProps<T, K extends RegionsMetricName> = {
   highlightCode?: string;
   getLink?: (code: string) => string;
   minHeight?: number;
+  hasNoDataFillColor?: string;
```
VWSCoronaDashboard4, line 35: This name indicates a boolean, but the value is a string?

VWSCoronaDashboard4 (approved):

PR #2895, opened 2021-05-28
Fix collapsible border double border bug

VWSCoronaDashboard14:
There was a slight double right border on the collapsible button, due to the fact that I'm copying the width of the button and apply it to the before element. The before the element is needed for the animation and can't be relative to the button.

But since the button was on a certain breakpoint a width with 2 decimals there was a rounding issue with a 1-pixel difference. By applying a min-width with the same width value as the before element aligns better and solves the rounding issue (since we are using the same value twice).

It is a tiny hack, but people won't notice the 1-pixel alignment difference. 

image


packages/app/src/components/collapsible/collapsible-button.tsx
```diff
@@ -144,6 +144,7 @@ const Container = styled(Box).attrs({ as: 'section' })<{
       fontWeight: 'bold',
       cursor: 'pointer',
       transition: 'color 0.2s ease-out',
+      minWidth: x.isOpen ? undefined : x.buttonWidth,
```
VWSCoronaDashboard4, line 147: Maybe annotate this with a comment?

VWSCoronaDashboard4 (approved):

PR #2896, opened 2021-05-28
Center value when there is only a value in the tooltip

VWSCoronaDashboard14:
The alignment of a single value in the tooltip was still `right`, when 4 or more numbers were in the tooltip it looked kinda odd.
This should `center` all the text in it when there is only a value.
This could be seen at the mini-trend tiles on the homepage.

New:
Screenshot 2021-05-28 at 09 52 47

Old:
Screenshot 2021-05-28 at 09 52 52

### DIFF
--- a/packages/app/src/components/time-series-chart/components/tooltip/tooltip-series-list.tsx
+++ b/packages/app/src/components/time-series-chart/components/tooltip/tooltip-series-list.tsx
@@ -169,9 +169,7 @@ function TooltipListItem({
             <VisuallyHidden>
               <InlineText mr={2}>{label}:</InlineText>
             </VisuallyHidden>
-            <TooltipEntryValue
-              displayTooltipValueOnly={displayTooltipValueOnly}
-            >
+            <TooltipEntryValue isCentered={displayTooltipValueOnly}>
               {children}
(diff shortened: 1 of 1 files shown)


VWSCoronaDashboard1 (approved):

packages/app/src/components/time-series-chart/components/tooltip/tooltip-series-list.tsx
```diff
@@ -169,7 +169,11 @@ function TooltipListItem({
             <VisuallyHidden>
               <InlineText mr={2}>{label}:</InlineText>
             </VisuallyHidden>
-            <TooltipEntryValue>{children}</TooltipEntryValue>
+            <TooltipEntryValue
+              displayTooltipValueOnly={displayTooltipValueOnly}
+            >
```
VWSCoronaDashboard10, line 174: `isCentered` makes more sense as api of the TooltipEntryValue component imho;

```suggestion
            <TooltipEntryValue isCentered={displayTooltipValueOnly}>
```

VWSCoronaDashboard10 (approved):

VWSCoronaDashboard4 (approved):

PR #2905, opened 2021-05-28
Add interaction behavior page

VWSCoronaDashboard14:
Add interaction for the behavior page:
- When selecting a metric in the dropdown, it gets set for all the graphs
- When you click on the table it scrolls down the first graph and selects the metric
- Moved the files to make it a bit more readable than it's a redesign
- Update the colors


packages/app/src/domain/behavior/redesign/behavior-line-chart-tile.tsx
```diff
@@ -13,16 +12,21 @@ import { TimeSeriesChart } from '~/components/time-series-chart';
...
 import { colors } from '~/style/theme';
-import { BehaviorIdentifier, behaviorIdentifiers } from './behavior-types';
+import { behaviorIdentifiers } from '../behavior-types';
+import { BehaviorIcon } from '../components/behavior-icon';
 
 interface BehaviorLineChartTileProps {
   values: NationalBehaviorValue[] | RegionalBehaviorValue[];
   metadata: MetadataProps;
+  currentId: any;
+  setCurrentId: any;
```
VWSCoronaDashboard8, line 22: Do not use any, unless you have no other option because you are working with a foreign interface of an external library. These ids can at least be typed to strings

packages/app/src/domain/behavior/redesign/_behavior-page-national.tsx
```diff
@@ -33,6 +35,9 @@ export function BehaviorPageNational({
   const { nl_gedrag } = siteText;
   const behaviorLastValue = data.behavior.last_value;
 
+  const [currentId, setCurrentId] = useState<BehaviorIdentifier>('wash_hands');
```
VWSCoronaDashboard8, line 38: I think this state should ideally live on the url as a query parameter

packages/app/src/domain/behavior/redesign/behavior-line-chart-tile.tsx
VWSCoronaDashboard14: Aaah it seems I've missed those two, good catch.

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard8 (approved):

PR #2929, opened 2021-06-02
Time series chart gapped annotations

VWSCoronaDashboard1:
## Summary

image

Added a hook named `useGappedLineAnnotations` that creates timespan annotations based on the 'null gaps' in the given values collection.

Also added support for showing lines that are exactly 1 timespan long. Normally these wouldn't render since they consist of just one datapoint. But the GappedLinedTrend will detect these and turn them into two data points that consist of the start and end date of the given timespan.

This brought along another interesting hurdle. Inside the series components we deal with `SeriesSIngleValue` instances that only have a `__date` and `__value` property. At that point the fact if its daily or span data is no longer relevant. Yet, for the situation described earlier I needed to turn one date back into a timespan. That's why the timespan is now injected through a prop down to the `GappedLinedTrend `.
If someone has a prettier solution to this, speak up ;)

Previously, undefined values were ignored by the hover logic, but in the case of gapped lines there needed to be an exception since the tooltip and marker need to indicate why the gap is displayed:

image

This PR also contains another small bugfix: The legend would now always render a hatched pattern in the legend, regardless of the timespan annotation fill value. That was fixed.

### Governance

- [ ] Documentation is added

### DIFF
--- a/packages/app/src/components/time-series-chart/components/gapped-line-trend.tsx
+++ b/packages/app/src/components/time-series-chart/components/gapped-line-trend.tsx
@@ -12,12 +12,12 @@ type GappedLinedTrendProps = {
   getY: (v: SeriesSingleValue) => number;
   curve?: 'linear' | 'step';
   id: string;
+  timespan: number;
 };
 
-const DAY_IN_SECONDS = 24 * 60 * 60;
-
 export function GappedLinedTrend(props: GappedLinedTrendProps) {
-  const { series, color, style, strokeWidth, getX, getY, curve, id } = props;
+  const { series, color, style, strokeWidth, getX, getY, curve, id, timespan } =
+    props;
 
   /**
    * Here we loop through the series and each time a null value is encountered a
@@ -41,15 +41,15 @@ export function GappedLinedTrend(props: GappedLinedTrendProps) {
   );
 
   gappedSeries.forEach((item, index, array) => {
-    if (item.length === 1) {
+    if (item.length === 1 && timespan > 1) {
       array[index] = [
         {
           __value: item[0].__value,
-          __date_unix: item[0].__date_unix - 3 * DAY_IN_SECONDS,
+          __date_unix: item[0].__date_unix - timespan / 2,
         },
         {
           __value: item[0].__value,
-          __date_unix: item[0].__date_unix + 3 * DAY_IN_SECONDS,
+          __date_unix: item[0].__date_unix + timespan / 2,
         },
       ];
     }
--- a/packages/app/src/components/time-series-chart/components/gapped-line-trend.tsx
+++ b/packages/app/src/components/time-series-chart/components/gapped-line-trend.tsx
@@ -40,16 +40,17 @@ export function GappedLinedTrend(props: GappedLinedTrendProps) {
     [[]]
   );
 
+  const halfOfTimespan = timespan / 2;
   gappedSeries.forEach((item, index, array) => {
     if (item.length === 1 && timespan > 1) {
       array[index] = [
         {
           __value: item[0].__value,
-          __date_unix: item[0].__date_unix - timespan / 2,
+          __date_unix: item[0].__date_unix - halfOfTimespan,
         },
         {
           __value: item[0].__value,
-          __date_unix: item[0].__date_unix + timespan / 2,
+          __date_unix: item[0].__date_unix + halfOfTimespan,
         },
       ];
     }
--- a/packages/app/src/components/time-series-chart/components/gapped-line-trend.tsx
+++ b/packages/app/src/components/time-series-chart/components/gapped-line-trend.tsx
@@ -42,7 +42,7 @@ export function GappedLinedTrend(props: GappedLinedTrendProps) {
 
   const halfOfTimespan = timespan / 2;
   gappedSeries.forEach((item, index, array) => {
-    if (item.length === 1 && timespan > 1) {
+    if (item.length === 1 && timespan > 0) {
       array[index] = [
         {
           __value: item[0].__value,
--- a/packages/app/src/components/time-series-chart/components/gapped-line-trend.tsx
+++ b/packages/app/src/components/time-series-chart/components/gapped-line-trend.tsx
@@ -14,6 +14,8 @@ type GappedLinedTrendProps = {
   id: string;
 };
 
+const DAY_IN_SECONDS = 24 * 60 * 60;
+
 export function GappedLinedTrend(props: GappedLinedTrendProps) {
   const { series, color, style, strokeWidth, getX, getY, curve, id } = props;
 
@@ -37,6 +39,21 @@ export function GappedLinedTrend(props: GappedLinedTrendProps) {
     },
     [[]]
   );
+
+  gappedSeries.forEach((item, index, array) => {
+    if (item.length === 1) {
+      array[index] = [
+        {
+          __value: item[0].__value,
+          __date_unix: item[0].__date_unix - 3 * DAY_IN_SECONDS,
+        },
+        {
+          __value: item[0].__value,
+          __date_unix: item[0].__date_unix + 3 * DAY_IN_SECONDS,
+        },
+      ];
+    }
+  });
 
   return (
     <>
--- a/packages/app/src/components/time-series-chart/components/point-markers.tsx
+++ b/packages/app/src/components/time-series-chart/components/point-markers.tsx
@@ -15,6 +15,12 @@ interface PointMarkerProps<T extends TimestampedValue> {
   points: HoveredPoint<T>[];
 }
 
+function hasColor<T extends TimestampedValue>(
+  hoverPoint: HoveredPoint<T>
+): hoverPoint is HoveredPoint<T> & { color: string } {
+  return isDefined(hoverPoint.color);
+}
+
 export function PointMarkers<T extends TimestampedValue>(
   props: PointMarkerProps<T>
 ) {
@@ -31,7 +37,7 @@ export function PointMarkers<T extends TimestampedValue>(
       }}
     >
       {points
-        .filter((x) => isDefined(x.color))
+        .filter(hasColor)
         .map((point, index) => (
           <PointMarker
             color={point.color}
--- a/packages/app/src/components/time-series-chart/components/point-markers.tsx
+++ b/packages/app/src/components/time-series-chart/components/point-markers.tsx
@@ -1,7 +1,6 @@
 import { TimestampedValue } from '@corona-dashboard/common';
 import { transparentize } from 'polished';
 import styled from 'styled-components';
-import { isDefined } from 'ts-is-present';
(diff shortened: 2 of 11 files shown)


packages/app/src/pages/landelijk/vaccinaties.tsx
```diff
@@ -85,6 +86,12 @@ const VaccinationPage = (props: StaticProps<typeof getStaticProps>) => {
 
   const vaccinationPerAgeGroupFeature = useFeature('vaccinationPerAgegroup');
 
+  const gappedLineAnnotations = useGappedLineAnnotations(
```
VWSCoronaDashboard8, line 89: Are you sure you want to merge the changes in this file? :)

packages/app/src/components/time-series-chart/components/gapped-line-trend.tsx
```diff
@@ -38,6 +40,22 @@ export function GappedLinedTrend(props: GappedLinedTrendProps) {
     [[]]
   );
 
+  const halfOfTimespan = timespan / 2;
+  gappedSeries.forEach((item, index, array) => {
+    if (item.length === 1 && timespan > 1) {
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: No, the timespan can't be zero. It's 1 for daily data, and otherwise the span is calculated based on the given data.

packages/app/src/components/time-series-chart/logic/use-gapped-line-annotations.ts
```diff
@@ -0,0 +1,60 @@
...
+import { isDefined, isPresent } from 'ts-is-present';
+import { TimespanAnnotationConfig } from './common';
+
+function isDateValue(value: TimestampedValue): value is DateValue {
+  return isDefined((value as DateValue).date_unix);
+}
+
+function isDateSpanValue(value: TimestampedValue): value is DateSpanValue {
+  return isDefined((value as DateSpanValue).date_start_unix);
+}
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: They look the same as `isDateSeries` or `isDateSpanSeries`, but these functions deal with arrays, not values. I can move them to the same location as these others.

packages/app/src/components/time-series-chart/logic/use-gapped-line-annotations.ts
```diff
@@ -0,0 +1,60 @@
...
+              ? item.date_start_unix
+              : NaN;
+
+            const endDate = isDateValue(item)
+              ? item.date_unix
+              : isDateSpanValue(item)
+              ? item.date_end_unix
+              : NaN;
+
+            let current = last(newItems);
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: Sure

packages/app/src/components/time-series-chart/logic/use-timespan.ts
```diff
@@ -0,0 +1,23 @@
...
+  isDateSeries,
+  isDateSpanSeries,
+  TimestampedValue,
+} from '@corona-dashboard/common';
+import { useMemo } from 'react';
+
+export function useTimespan<T extends TimestampedValue>(values: T[]) {
+  return useMemo(() => {
+    if (!values.length) {
+      return NaN;
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: I don't think it will clean up anything, but for consistency's sake I'm fine with return undefined.

packages/app/src/components/time-series-chart/logic/use-timespan.ts
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    if (isDateSeries(values)) {
+      return 1;
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: Fair enough!

packages/app/src/pages/landelijk/vaccinaties.tsx
VWSCoronaDashboard1: Oops, well spotted, that was my test page. I'll revert that.

packages/app/src/components/time-series-chart/components/gapped-line-trend.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard1: As discussed in the comments below, the default is now zero instead of 1

packages/app/src/components/time-series-chart/logic/use-gapped-line-annotations.ts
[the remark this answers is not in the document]
VWSCoronaDashboard1: Haha, my bad, they already existed there as well. I removed these and imported the correct ones.

packages/app/src/components/time-series-chart/logic/use-gapped-line-annotations.ts
```diff
@@ -0,0 +1,60 @@
...
+            if (!isDefined(current) || !isNaN(current.end)) {
+              current = { start: startDate, end: NaN, label };
+              newItems.push(current);
+            }
+
+            if (
+              index === array.length - 1 ||
+              isPresent(array[index + 1]?.[property])
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: I opted for a code comment instead

packages/app/src/components/time-series-chart/logic/use-timespan.ts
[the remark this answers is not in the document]
VWSCoronaDashboard1: I'm returning zero actually, it makes the code easier and makes sense too. If there are no values, then there is no timespan either.

packages/app/src/components/time-series-chart/components/point-markers.tsx
```diff
@@ -30,6 +37,7 @@ export function PointMarkers<T extends TimestampedValue>(
       }}
     >
       {points
+        .filter(hasColor)
```
VWSCoronaDashboard8, line 40: What was the idea behind this? Seems useful to add a comment here. If you pass point data to pointmarkers, and color is optional, I wouldn't expect the ones without colors to be completely omitted.

packages/app/src/components/time-series-chart/logic/hover-state.ts
```diff
@@ -243,8 +254,8 @@ export function useHoverState<T extends TimestampedValue>({
             return {
               seriesValue,
               x: xScale(xValue),
-              y: yScale(yValue),
-              color: config.color,
+              y: yValue ? yScale(yValue) : -Infinity,
+              color: yValue ? config.color : undefined,
```
VWSCoronaDashboard8, line 258: Could it also work to pass "transparent" as the color. Perhaps then maybe the hasColor check + filter is not needed?

packages/app/src/components/time-series-chart/logic/use-gapped-line-annotations.ts
```diff
@@ -38,23 +34,42 @@ export function useGappedLineAnnotations<T extends TimestampedValue>(
...
-            let current = last(newItems);
-            if (!isDefined(current) || !isNaN(current.end)) {
-              current = { start: startDate, end: NaN, label };
-              newItems.push(current);
+            let currentAnnotation = last(newItems);
+            if (
+              !isDefined(currentAnnotation) ||
+              !isNaN(currentAnnotation.end)
+            ) {
+              currentAnnotation = { start: startDate, end: NaN, label };
```
VWSCoronaDashboard8, line 42: nit; I feel the NaNs in this part could also be replaced with undefined.

VWSCoronaDashboard8 (approved):

packages/app/src/components/time-series-chart/logic/hover-state.ts
VWSCoronaDashboard1: That's worth a try, I'll give it a shot!

packages/app/src/components/time-series-chart/logic/use-gapped-line-annotations.ts
VWSCoronaDashboard1: Nope, because I'm constructing a TimespanAnnotation, which has an 'end' property that is a number. I think changing that to `number|undefined` will yield a bunch of extra undefined checks further downstream. When all I'm doing here is constructing these configs in multiple steps. IMHO, assigning NaN here initially kind of neatly solves that problem here locally.

packages/app/src/components/time-series-chart/components/point-markers.tsx
VWSCoronaDashboard1: Changed in favor of using a transparent color. As discussed below.

packages/app/src/components/time-series-chart/logic/use-timespan.ts
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    if (isDateSeries(values)) {
+      return 0;
+    } else if (isDateSpanSeries(values)) {
+      const value = values[0];
+      return value.date_end_unix - value.date_start_unix;
+    }
+
+    assert(false, 'Encountered an invalid time series');
```
VWSCoronaDashboard10, line 22:
```suggestion
    throw new Error('Encountered an invalid time series');
```

packages/app/src/components/time-series-chart/logic/use-timespan.ts
```diff
@@ -0,0 +1,24 @@
...
+  return useMemo(() => {
+    if (!values.length) {
+      return 0;
+    }
+
+    if (isDateSeries(values)) {
+      return 0;
+    } else if (isDateSpanSeries(values)) {
+      const value = values[0];
+      return value.date_end_unix - value.date_start_unix;
```
VWSCoronaDashboard10, line 19: Does this assume that the timespan is equal for every value? Wouldn't this clash with the mixed values of @VWSCoronaDashboard8 sewer chart?

VWSCoronaDashboard10 (approved):

packages/app/src/components/time-series-chart/logic/use-timespan.ts
VWSCoronaDashboard1: Indeed, until now I thought a series would always have a consistent timespan. I was not aware of variable lengths, if that is so, then I guess this hook will have to calculate the spans for each item and pass that down.

VWSCoronaDashboard1: I'm double checking with @VWSCoronaDashboard8...

VWSCoronaDashboard8: @VWSCoronaDashboard1 @VWSCoronaDashboard10 This won't be an issue. If we have timespans they are always assumed to be consistent throughout the series.

What I'm doing in the sewer chart is converting datespan values to date, by picking the middle of the week. So when the data is merged, there are not spans but only single date values.

PR #2938, opened 2021-06-03
Create situations legenda

VWSCoronaDashboard10:
I had to refactor some parts in order to get the choropleth legenda to be responsive. Also a bug popped up in the `use-choropleth-color-scale` which has been fixed.

image

### DIFF
--- a/packages/app/src/components/choropleth-legenda.tsx
+++ b/packages/app/src/components/choropleth-legenda.tsx
@@ -10,14 +10,12 @@ interface ChoroplethLegendaProps {
   title: string;
   thresholds: ChoroplethThresholdsValue[];
   valueAnnotation?: string;
-  endLabel?: string;
 }
 
 export function ChoroplethLegenda({
   title,
   thresholds,
   valueAnnotation,
-  endLabel,
 }: ChoroplethLegendaProps) {
   const { width = 0, ref } = useResizeObserver<HTMLSpanElement>();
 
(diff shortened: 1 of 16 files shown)


VWSCoronaDashboard1 (approved):

packages/app/src/components/choropleth-legenda.tsx
```diff
@@ -1,36 +1,52 @@
...
+          const isLast = index === thresholds.length - 1;
+          return (
+            <Item key={color + threshold}>
+              <LegendaColor color={color} first={isFirst} last={isLast} />
+              <Label
+                css={isFirst ? css({ transform: 'translateX(0%)' }) : undefined}
+              >
+                {label ?? threshold}
+              </Label>
+              {endLabel && isLast && (
```
VWSCoronaDashboard4, line 42: Why not move this out of the `map` structure?

packages/app/src/components/choropleth-legenda.tsx
```diff
@@ -1,36 +1,52 @@
...
-          </Item>
-        ))}
+        {thresholds.map(({ color, threshold, label }, index) => {
+          const isFirst = index === 0;
+          const isLast = index === thresholds.length - 1;
+          return (
+            <Item key={color + threshold}>
+              <LegendaColor color={color} first={isFirst} last={isLast} />
+              <Label
+                css={isFirst ? css({ transform: 'translateX(0%)' }) : undefined}
```
VWSCoronaDashboard4, line 38: ~~What does this magic `translateX(0%)` do? Might be good to add a comment?~~

I see it's an override of the other transforms. Perhaps pass as an alignment prop to the Label component.

packages/app/src/components/choropleth-legenda.tsx
```diff
@@ -1,36 +1,52 @@
...
+  endLabel?: string;
 }
 
 export function ChoroplethLegenda({
   title,
   thresholds,
   valueAnnotation,
+  endLabel,
 }: ChoroplethLegendaProps) {
+  const { width = 0, ref } = useResizeObserver<HTMLSpanElement>();
```
VWSCoronaDashboard4, line 22: Wouldn't this be a possible source for odd loops? You are observing the size of a node that is restricted in size a parent node that you control the margins of, causing a change in the size of the node ... etc?

packages/app/src/components/choropleth-legenda.tsx
VWSCoronaDashboard10: yep, good idea 👍

packages/app/src/components/choropleth-legenda.tsx
VWSCoronaDashboard10: Nope, the ref is placed on another element so they're not connected.

PR #2939, opened 2021-06-04
Add data coverage tile situations page

VWSCoronaDashboard14:
Adding a new tile, that has 2 states either enough or not enough data
New Lokalize texts
Some mock data
Had to do some fancy replaceComponentsInText in replaceComponentsInText because the whole text had to be bold inside, couldn't find a better way.

**Mandatory screenshots**
_Enough data:_
Screenshot 2021-06-04 at 11 15 31

_Not enough data:_
Screenshot 2021-06-04 at 11 34 22


packages/app/src/domain/situations/situations-data-coverage-tile.tsx
```diff
@@ -0,0 +1,102 @@
...
+          data.has_sufficient_data
+            ? descriptionEnoughCoverage
+            : descriptionNotEnoughCoverage,
+          {
+            date_text: (
+              <InlineText fontWeight="bold">
+                {replaceComponentsInText(dateText, {
+                  date_start_unix: (
+                    <InlineText>
+                      {formatDateFromSeconds(data.date_start_unix)}
```
VWSCoronaDashboard10, line 53: There's a `intl.formatDateSpan()` utility to format date spans in a single month as `van 3 tot 5 april` instead of `van 3 april tot 5 april`, usage:
```
const [start, end] = intl.formatDateSpan(
    { seconds: value.date_start_unix },
    { seconds: value.date_end_unix }
  );
```

Also the replaceable variable `date_start_unix`/`date_end_unix` should be renamed to something without `_unix` since it's not a unix we're replacing but a readable string.

packages/app/src/domain/situations/situations-data-coverage-tile.tsx
```diff
@@ -0,0 +1,102 @@
...
+    mr: 2,
+    mb: 1,
+    borderRadius: '50%',
+    backgroundColor: x.hasSufficientData ? 'data.primary' : 'gray',
+    color: 'white',
+
+    svg: {
+      p: x.hasSufficientData ? undefined : '2px',
+      mt: x.hasSufficientData ? '1px' : undefined,
+    },
```
VWSCoronaDashboard10, line 100: It looks like the two icons have different dimensions. You should request design to fix that instead of playing around with paddings and margins. Icon sizes/dimensions should be predictable.

packages/app/src/domain/situations/situations-data-coverage-tile.tsx
```diff
@@ -0,0 +1,102 @@
...
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+    width: 22,
+    height: 22,
+    mr: 2,
+    mb: 1,
+    borderRadius: '50%',
+    backgroundColor: x.hasSufficientData ? 'data.primary' : 'gray',
+    color: 'white',
```
VWSCoronaDashboard10, line 95: You can inherit the background color from the containing component, this will work when you move the `color` to the svg style block.
```suggestion
    backgroundColor: 'currentcolor',
    svg: {
      color: 'white',
```

packages/app/src/domain/situations/situations-data-coverage-tile.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { Heading, InlineText, Text } from '~/components/typography';
+import { useIntl } from '~/intl';
+import { replaceComponentsInText } from '~/utils/replace-components-in-text';
+import { mockDataCoverage } from '~/domain/situations/logic/mock-data';
+
+interface SituationsDataCoverageTileProps {
+  title: string;
+  descriptionEnoughCoverage: string;
+  descriptionNotEnoughCoverage: string;
+  dateText: string;
```
VWSCoronaDashboard10, line 15: Why are you injecting these texts instead of reading them from inside this component?

packages/app/src/domain/situations/situations-data-coverage-tile.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+const IndicatorCircle = styled.div<{ hasSufficientData: boolean }>((x) =>
+  css({
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+    width: 22,
+    height: 22,
+    mr: 2,
+    mb: 1,
```
VWSCoronaDashboard10, line 92: I think a lot of vertical margins and paddings can be removed, also on the CoverageIndicator component and the mb on the Heading component.

Instead a wrapping `<Box spacing={2}>` component can define all spacings between those components.

packages/app/src/pages/veiligheidsregio/[code]/brononderzoek.tsx
```diff
@@ -102,6 +104,16 @@ export default function BrononderzoekPage(
...
           <ArticleStrip articles={content.articles} />
+          <SituationsDataCoverageTile
+            title={vrText.tile_coverage.title}
+            descriptionEnoughCoverage={
+              vrText.tile_coverage.description_enough_coverage
+            }
+            descriptionNotEnoughCoverage={
+              vrText.tile_coverage.description_not_enough_coverage
+            }
+            dateText={vrText.tile_coverage.date_text}
```
Comment (VWSCoronaDashboard10) [inline packages/app/src/pages/veiligheidsregio/[code]/brononderzoek.tsx:115]: Like I mentioned earlier: these texts can be read from inside the component. It's not shareable component in the component directory so it's not a problem imho.

packages/cms/src/lokalize/key-mutations.csv
```diff
@@ -34,3 +34,9 @@ timestamp,action,key
 2021-06-03T09:34:40.512Z,add,brononderzoek.situaties_kaarten_overzicht.legenda.onvoldoende_data
 2021-06-03T09:34:41.530Z,add,brononderzoek.situaties_kaarten_overzicht.select_help_text
 2021-06-03T09:34:42.543Z,add,brononderzoek.situaties_kaarten_overzicht.titel
+2021-06-04T08:49:55.177Z,add,vr_brononderzoek.tile_coverage.title
+2021-06-04T08:53:07.423Z,add,vr_brononderzoek.tile_coverage.description_enough_coverage
+2021-06-04T08:53:42.448Z,add,vr_brononderzoek.tile_coverage.description_not_enough_coverage
+2021-06-04T08:54:05.513Z,add,vr_brononderzoek.tile_coverage.date_text
+2021-06-04T08:54:38.695Z,add,vr_brononderzoek.tile_coverage.title_enough_coverage
+2021-06-04T08:54:58.663Z,add,vr_brononderzoek.tile_coverage.title_not_enough_coverage
```
VWSCoronaDashboard10, line 42: I think it's not necessary to have these texts under `vr_brononderzoek`. Imho they can be placed under something like `brononderzoek.veiligheidsregio_dekking`.
Also within "brononderzoek" I tried to stick to dutch keys, such as `.titel` and `.beschrijving`. Not sure what we really want though? @VWSCoronaDashboard1 @VWSCoronaDashboard4 @VWSCoronaDashboard8

VWSCoronaDashboard10 (changes requested):

packages/cms/src/lokalize/key-mutations.csv
VWSCoronaDashboard4: I'm fine with the first suggestion, though having a separate main key is how other NL/VR keys are organised.

The second suggestion, with the other keys in the repo being 50%-ish EN/NL, I'd prefer EN keys for new keys. Though, consistency within one subject would be better, so for this feature stick to NL?

VWSCoronaDashboard4 (commented):
Looks good apart from the suggestions by others!

packages/app/src/domain/situations/situations-data-coverage-tile.tsx
```diff
@@ -1,69 +1,63 @@
...
 interface SituationsDataCoverageTileProps {
-  title: string;
-  descriptionEnoughCoverage: string;
-  descriptionNotEnoughCoverage: string;
-  dateText: string;
+  data: {
+    has_sufficient_data: boolean;
+    date_start_unix: number;
+    date_end_unix: number;
+  };
```
VWSCoronaDashboard10, line 15: This can be typed as SituationsValue

VWSCoronaDashboard10 (approved):

PR #2943, opened 2021-06-04
Add situations over time chart

VWSCoronaDashboard1:
## Summary

This PR adds the situations over time chart on the safety region page:

image

## Motivation

Feature request

### Governance

- [ ] Documentation is added


packages/app/src/components/time-series-chart/logic/use-gapped-line-annotations.ts
```diff
@@ -4,7 +4,8 @@ import {
   isDateValue,
   TimestampedValue,
 } from '@corona-dashboard/common';
-import { last } from 'lodash';
+import isBoolean from 'lodash/isBoolean';
+import last from 'lodash/last';
```
VWSCoronaDashboard8, line 8: Didn't we have a plugin rewriting these imports? I'm assuming you can keep it consistent with other imports, otherwise, we'll have to change the others if they're problematic.

packages/app/src/components/time-series-chart/logic/use-gapped-line-annotations.ts
```diff
@@ -73,3 +76,10 @@ function formatLabel(
   const end = formatDateFromSeconds(annotation.end, 'axis');
   return `${start} - ${end}: ${annotation.label}`;
 }
+
+function createIsValidCheck(value: unknown) {
+  if (isBoolean(value)) {
+    return (val: unknown) => val === true;
+  }
+  return (val: unknown) => isPresent(val);
+}
```
VWSCoronaDashboard8, line 85: I find this a little odd, because it seems to be an optimization yet the create function is called for every value in the array with the result only then to be used twice. The work done in the validation is I think pretty negligible so this feels premature.

I'd suggest using this:

```
function isValid(value: unknown) {
  return isBoolean(value) ? value === true : isPresent(value)
}
```

packages/app/src/domain/situations/situations-over-time-chart.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+const seriesColors: Record<SituationKey, string> = {
+  home_and_visits: colors.data.multiseries.cyan,
+  work: colors.data.multiseries.turquoise,
+  school_and_day_care: colors.data.multiseries.turquoise_dark,
+  health_care: colors.data.multiseries.yellow,
+  gathering: colors.data.multiseries.yellow_dark,
+  travel: colors.data.multiseries.orange,
+  hospitality: colors.data.multiseries.orange_dark,
+  other: colors.data.multiseries.magenta,
```
VWSCoronaDashboard8, line 87: nit; maybe destructure multiseries to avoid so much repetition.

VWSCoronaDashboard8 (approved):

packages/app/src/components/time-series-chart/logic/use-gapped-line-annotations.ts
VWSCoronaDashboard1: Yea, that was just a dumb mistake, the creation should have been outside of the loop, lol. Fixed.

packages/app/src/domain/situations/situations-over-time-chart.tsx
VWSCoronaDashboard1: Fixed

packages/app/src/components/time-series-chart/logic/use-gapped-line-annotations.ts
```diff
@@ -17,52 +17,50 @@ export function useGappedLineAnnotations<T extends TimestampedValue>(
...
-              index === array.length - 1 ||
-              isPresent(array[index + 1]?.[property])
-            ) {
-              currentAnnotation.end = endDate;
-              currentAnnotation.label = formatLabel(
-                currentAnnotation,
-                formatDateFromSeconds
-              );
-            }
+          // If this is either the last item in the list or the next
```
VWSCoronaDashboard4, line 45: Maybe convert this to a multi-line comment?

packages/app/src/components/time-series-chart/logic/use-gapped-line-annotations.ts
```diff
@@ -17,52 +17,50 @@ export function useGappedLineAnnotations<T extends TimestampedValue>(
...
-              ? item.date_start_unix
-              : NaN;
+  return useMemo(() => {
+    const sample = first(values)?.[property];
+    const isValid = createIsValidCheck(sample);
+    return values.reduce<TimespanAnnotationConfig[]>(
+      (newItems, item, index, array) => {
+        const value = item[property] as unknown;
+        if (!isValid(value)) {
+          const startDate = isDateValue(item)
```
VWSCoronaDashboard4, line 27: I'd add a bit of commentary here to explain what is going on, for a future developer.

VWSCoronaDashboard4 (approved):

packages/app/src/components/time-series-chart/logic/use-gapped-line-annotations.ts
VWSCoronaDashboard1: I've added a big commentary block on top of the hook that, hopefully, clearly states what it's doing.

packages/app/src/components/time-series-chart/logic/use-gapped-line-annotations.ts
```diff
@@ -73,3 +91,10 @@ function formatLabel(
   const end = formatDateFromSeconds(annotation.end, 'axis');
   return `${start} - ${end}: ${annotation.label}`;
 }
+
+function createIsValidCheck(value: unknown) {
+  if (isBoolean(value)) {
+    return (val: unknown) => val === true;
+  }
+  return (val: unknown) => isPresent(val);
+}
```
VWSCoronaDashboard10, line 100: I don't think the check on the first value (`first(values)?.[property]`) makes a lot of sense? The `createIsValidCheck` could be simplified to:

```suggestion
function hasValue(value: unknown) {
  return isBoolean(value) ? val === true : isPresent(val)
}
```

packages/app/src/components/time-series-chart/logic/use-gapped-line-annotations.ts
```diff
@@ -4,65 +4,83 @@ import {
...
-              index === array.length - 1 ||
-              isPresent(array[index + 1]?.[property])
-            ) {
-              currentAnnotation.end = endDate;
-              currentAnnotation.label = formatLabel(
-                currentAnnotation,
-                formatDateFromSeconds
-              );
-            }
+          /** If this is either the last item in the list or the next
```
VWSCoronaDashboard10, line 64:
```suggestion
          /** 
           * If this is either the last item in the list or the next
```

packages/app/src/domain/situations/situations-over-time-chart.tsx
```diff
@@ -0,0 +1,99 @@
...
+    GappedLineSeriesDefinition<VrSituationsValue>
+  >((situation) => ({
+    type: 'gapped-line',
+    metricProperty: situation.id,
+    color: seriesColors[situation.id],
+    label: situation.title,
+    shape: 'line',
+  }));
+
+  const compareList = list.concat();
```
VWSCoronaDashboard10, line 52: Why this?

VWSCoronaDashboard10 (approved):

packages/app/src/components/time-series-chart/logic/use-gapped-line-annotations.ts
VWSCoronaDashboard1: Why doesn't it make sense? I only have to check the value once.

packages/app/src/domain/situations/situations-over-time-chart.tsx
VWSCoronaDashboard1: Yup, that became superfluous, hehe. There used to be another list, I'll remove this.

packages/app/src/components/time-series-chart/logic/use-gapped-line-annotations.ts
VWSCoronaDashboard10: It complicates code while it probably won't improve performance

VWSCoronaDashboard10: > it's not going to happen that the first item is a boolean and the second a number

and that ☝️  is an assumption, in theory we could create a schema which is of type `boolean | number`.

PR #2947, opened 2021-06-07
Behavior page improvements

VWSCoronaDashboard14:
Back from QA there were several improvements
- The curfew returns null from the backend, there is a filter in place that adds the key name that contains null to the `keysWithoutData` for the choropleth
- Add spacing between choropleths on smaller screens
- Rules the table are now in the correct font

### DIFF
--- a/packages/app/src/domain/behavior/redesign/behavior-choropleths-tile.tsx
+++ b/packages/app/src/domain/behavior/redesign/behavior-choropleths-tile.tsx
@@ -48,9 +48,9 @@ export function BehaviorChoroplethsTile({
    * Lastly we remove all the duplicates in the array and add it to all the keys without data
    */
   const idsThatContainNull = Object.keys(firstRegionData)
-    .filter((key) => firstRegionData[key as keyof RegionsBehavior] === null) // check if a value has null
-    .map((item) => item.slice(0, item.indexOf('_'))) // Slice everything after the underscore to find the behaviorIdentifier
-    .filter((item, pos) => item.indexOf(item) == pos); // Remove duplicate names
+    .filter((key) => firstRegionData[key as keyof RegionsBehavior] === null)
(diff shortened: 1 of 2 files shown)


packages/app/src/domain/behavior/redesign/behavior-choropleths-tile.tsx
```diff
@@ -34,12 +34,26 @@ export function BehaviorChoroplethsTile({
   setCurrentId,
 }: BehaviorChoroplethsTileProps) {
   const { siteText } = useIntl();
+  const firstRegionData = data[0];
 
-  // Find all the keys that only doesn't exist on VR level but does on NL
+  // Find all the keys that doesn't exist on VR level but does on NL
```
VWSCoronaDashboard1, line 39:
```suggestion
  // Find all the keys that don't exist on VR level but do on NL
```

packages/app/src/domain/behavior/redesign/behavior-choropleths-tile.tsx
```diff
@@ -34,12 +34,26 @@ export function BehaviorChoroplethsTile({
...
   const keysWithoutData = behaviorIdentifiers.filter(
-    (item) => !Object.keys(data[0]).find((a) => a.includes(item))
+    (item) => !Object.keys(firstRegionData).find((a) => a.includes(item))
   );
 
+  /**
+   * Since e.g. the curfew has no data anymore and returns null that also needs to be filtered out
+   * First we check if there are some keys that contains a value of null
```
VWSCoronaDashboard1, line 46:
```suggestion
   * First we check if there are some keys that contain a value of null
```

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard4 (approved):

PR #2948, opened 2021-06-07
Add error boundary

VWSCoronaDashboard1:
## Summary

This PR adds an error boundary. I used react-error-boundary for the basis of the component, added a default implementation for the whole dashboard to this and wrapped as many appropriate components in the wrapper.

The error boundary shows a generic message, an error report, and a button that allows the user to copy the error report to their clipboard.

The component looks like this when it encounters an error:
image

## Motivation

Feature request.

### Governance

- [ ] Documentation is added


packages/app/src/components/error-boundary/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+  );
+}
+
+function ErrorFallback({ error }: { error: Error }) {
+  const { siteText } = useIntl();
+  const [clipboardState, setClipboardState] =
+    useState<'init' | 'copied' | 'error'>('init');
+  const errorReport = formatErrorMessage(error);
+
+  const copyErrorReport = () => {
```
VWSCoronaDashboard8, line 25:
```suggestion
  function copyErrorReport(){
```

packages/app/src/components/error-boundary/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+  return `url: ${window.location.href}
+platform: ${navigator.platform}
+userAgent: ${navigator.userAgent}
+browser dimensions: ${window.innerWidth}x${window.innerHeight}
+screen resolution: ${window.screen.width * window.devicePixelRatio}x${
+    window.screen.height * window.devicePixelRatio
+  }
+error: ${error.message}
+stacktrace:
+${error.stack}`;
```
VWSCoronaDashboard8, line 80: This doesn't look nice. Maybe use https://www.npmjs.com/package/dedent I suspect we can use it in more places than this

packages/app/src/components/error-boundary/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+}
+
+function copyErrorMessage(errorReport: string) {
+  try {
+    return navigator.clipboard.writeText(errorReport);
+  } catch (e) {
+    return Promise.reject(e);
+  }
```
VWSCoronaDashboard8, line 88: This looks odd. You're placing a try/catch around a promised function so I don't think it'll throw. And if writeText returns a promise what are you trying to achieve that is different from 

```
function copyErrorMessage(errorReport: string) {
   return navigator.clipboard.writeText(errorReport)
}
```

VWSCoronaDashboard8 (approved):

packages/app/src/components/error-boundary/index.tsx
VWSCoronaDashboard1: Yea, it's kinda my paranoia about the error boundary component throwing errors (which would be kinda dumb), in theory, the clipboard object could throw an error before it created its promise or whatever. lol.

VWSCoronaDashboard10 (changes requested):

packages/app/src/components/error-boundary/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+      () => {
+        setClipboardState('error');
+      }
+    );
+  }
+
+  const mail = siteText.common.foutmelding_email_adres;
+
+  const subject = encodeURIComponent('Foutmelding op corona dashboard');
+  const body = 'test'; //;encodeURIComponent(errorReport);
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: Oops.

packages/app/src/components/error-boundary/index.tsx
VWSCoronaDashboard1: That's printing it out as a JSON object, I don't think that's much of an improvement to be honest...

VWSCoronaDashboard1: Oh, and the stacktrace gets serialized with literal '\n' characters in it that way hehe

VWSCoronaDashboard8 (approved):

packages/app/src/components/age-demographic/age-demographic.tsx
```diff
@@ -1,5 +1,6 @@
 import { Box } from '~/components/base';
 import { Tooltip, useTooltip } from '~/components/tooltip';
+import { ErrorBoundary } from '../error-boundary';
```
VWSCoronaDashboard4, line 3:
```suggestion
import { ErrorBoundary } from '~/components/error-boundary';
```

packages/app/src/components/error-boundary/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+    borderColor: 'red',
+    padding: 2,
+    borderRadius: '8px',
+  })
+);
+
+const Button = styled.button(
+  css({
+    bg: 'transparent',
+    border: '1px solid silver',
```
VWSCoronaDashboard4, line 96: Not a color from the theme? Same for the ErrorReport below

packages/app/src/components/error-boundary/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+          {siteText.common.foutmelding_kon_niet_gekopieerd_worden}
+        </InlineText>
+      )}
+      <ErrorReport>{errorReport}</ErrorReport>
+    </ErrorBox>
+  );
+}
+
+function formatErrorMessage(error: Error) {
+  return `url: ${window.location.href}
```
VWSCoronaDashboard4, line 69: This could look a bit cleaner perhaps using an object/array that you `join` or `reduce` to a string.

packages/app/src/components/choropleth-tile.tsx
```diff
@@ -99,7 +100,9 @@ export function ChoroplethTile({
           flexDirection="column"
           height="100%"
         >
-          <Box height="100%">{children}</Box>
+          <Box height="100%">
+            <ErrorBoundary>{children}</ErrorBoundary>
```
VWSCoronaDashboard4, line 104: The choropleth tile is not the only component that includes choropleths. Eg topical, situations, behavior use a different mechanism. Let's add this wrapper there as well

VWSCoronaDashboard4 (changes requested):

packages/app/src/components/error-boundary/index.tsx
```diff
@@ -66,16 +66,19 @@ function ErrorFallback({ error }: { error: Error }) {
...
+  report.push(
+    `screen resolution: ${window.screen.width * window.devicePixelRatio}x${
+      window.screen.height * window.devicePixelRatio
+    }`
+  );
+  report.push(`error: ${error.message}`);
+  report.push(`stacktrace:`);
+  report.push(error.stack ?? 'No stack trace available');
+  return report.join('\n');
 }
```
VWSCoronaDashboard10: I think an object reads better..
```tsx
  const report = {
    'url': window.location.href,
    'platform': navigator.platform,
    'userAgent': navigator.userAgent,
    'browser dimensions': `${window.innerWidth}x${window.innerHeight}`,
    'screen resolution': `${window.screen.width * window.devicePixelRatio}x${
      window.screen.height * window.devicePixelRatio
    }`,
    'error': error.message,
    'stacktrace': error.stack ?? 'No stack trace available',
  }

  return Object.entries(report).map(([key, value]) => `${key}: ${value}`).join('\n')
}
```

or maybe, if you don't like that approach and want to stick to an array:
```tsx
  return [
    `url: ${window.location.href}`,
    `platform: ${navigator.platform}`,
    `userAgent: ${navigator.userAgent}`,
    `browser dimensions: ${window.innerWidth}x${window.innerHeight}`,
    `screen resolution: ${window.screen.width * window.devicePixelRatio}x${
      window.screen.height * window.devicePixelRatio
    }`,
    `error: ${error.message}`,
    `stacktrace:`,
    error.stack ?? 'No stack trace available',
  ].join('\n');
}
```

VWSCoronaDashboard10 (approved):

packages/app/src/components/error-boundary/index.tsx
VWSCoronaDashboard8: What was the problem with using something like dedent? I don't see how any of this is cleaner 🤷‍♂️ 

Also I don't see why userAgent is camel-cased and some others have spaces. If there are no spaces I guess declaring it as an object which is then formatted to a string could also be ok. But I still think a template string is nicer.

VWSCoronaDashboard1: As much as I absolutely LOVE discussing the perfect way of formatting a string, I'm gonna leave it as it is right now with the last tweak made to the naming and casing. If this prevents someone else from sleeping soundly at night, feel free to change it in a subsequent PR ;)

PR #2955, opened 2021-06-08
Add internal and external link for sanity rich content

VWSCoronaDashboard14:
The wish was that there could be internal and external links from Sanity. This checks based on the route if it should be an internal or external link.


packages/app/src/components/cms/rich-content.tsx
```diff
@@ -73,3 +75,19 @@ function InlineAttachmentMark(props: {
     </a>
   );
 }
+
+const isExternalURL = (url: string) => /^https?:\/\//.test(url);
```
VWSCoronaDashboard1, line 79: I think we agreed that `isAbsoluteUrl` is probably a better term.

VWSCoronaDashboard1 (approved):

packages/app/src/utils/is-absolute-url.ts
```diff
@@ -0,0 +1 @@
+export const isAbsoluteUrl = (url: string) => /^https?:\/\//.test(url);
```
VWSCoronaDashboard8, line 1: nit; I would put this under utils/url.ts  to be able to group function based on subject, but I guess we are not doing that yet for most utils.

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard8 (approved):

packages/app/src/utils/is-absolute-url.ts
VWSCoronaDashboard14: It might be better to pick it as a separate story and do it properly for everything.

PR #2960, opened 2021-06-08
Situations data coverage choropleth tweaks

VWSCoronaDashboard1:
## Summary

- Fix centering of the icon within the bubble for the legend items
- Prevent hidden button from receiving first focus when opening the choropleth in fullscreen mode

## Motivation

Design request.

## Detailed design

In the case of this choropleth the first button in the modal turned out to be the hidden 'tab navigation' button. Default behavior of the `useFocusTrap` is apparently to give focus on the first focusable item it can find. In previous cases this was a visible button, only here it was setting focus, and thereby revealing, the 'tab navigation' button.
So, I've added an ID to the close button and pass this ID on to the focusSelector option of the `useFocusTrap` hook, that way the close button always receives first focus.

### Governance

- [ ] Documentation is added

### DIFF
--- a/packages/app/src/components/fullscreen-chart-tile.tsx
+++ b/packages/app/src/components/fullscreen-chart-tile.tsx
@@ -7,7 +7,6 @@ import { useIntl } from '~/intl';
 import { replaceVariablesInText } from '~/utils/replace-variables-in-text';
 import { useBreakpoints } from '~/utils/use-breakpoints';
 import { usePrevious } from '~/utils/use-previous';
-import { useUniqueId } from '~/utils/use-unique-id';
 import { Spacer } from './base';
 import { IconButton } from './icon-button';
 import { Metadata, MetadataProps } from './metadata';
@@ -40,8 +39,6 @@ export function FullscreenChartTile({
     { subject: intl.siteText.common.grafiek_singular }
   );
 
-  const closeButtonId = useUniqueId();
-
   const tile = (
     <Tile height="100%">
       {children}
@@ -68,7 +65,6 @@ export function FullscreenChartTile({
             title={label}
             onClick={() => setIsFullscreen((x) => !x)}
             size={36}
-            id={isFullscreen ? closeButtonId : undefined}
           >
             {isFullscreen ? <CloseIcon /> : <ExpandIcon />}
           </IconButton>
@@ -90,7 +86,6 @@ export function FullscreenChartTile({
         <Modal
           id="chart-tile-container"
           onClose={() => setIsFullscreen(false)}
-          focusSelector={`#${closeButtonId}`}
           isFullheight
         >
           {tile}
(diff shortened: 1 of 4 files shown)


VWSCoronaDashboard8 (approved):

packages/app/src/components/fullscreen-chart-tile.tsx
```diff
@@ -65,6 +68,7 @@ export function FullscreenChartTile({
             title={label}
             onClick={() => setIsFullscreen((x) => !x)}
             size={36}
+            id={isFullscreen ? closeButtonId : undefined}
```
VWSCoronaDashboard10, line 71: I think these elements (the `tile` variable) get rendered twice in the DOM meaning there are duplicate id's.

VWSCoronaDashboard10 (changes requested):
I think the focus thing can be implemented a little bit cleaner:

```diff
diff --git a/packages/app/src/components/modal.tsx b/packages/app/src/components/modal.tsx
index 7b531621..213b2a28 100644
--- a/packages/app/src/components/modal.tsx
+++ b/packages/app/src/components/modal.tsx
@@ -15,7 +15,7 @@ interface ModalProps {
 export function Modal({ id, children, onClose, isFullheight }: ModalProps) {
   const clickRef = useRef<HTMLDivElement>(null);
   useHotkey('esc', onClose);
-  const focusRef = useFocusTrap();
+  const focusRef = useFocusTrap(true, { focusSelector: `#${id}` });
 
   return (
     <ModalPortal id={id}>
@@ -25,6 +25,7 @@ export function Modal({ id, children, onClose, isFullheight }: ModalProps) {
           css={css({ p: 5, height: isFullheight ? '100%' : undefined })}
           onClick={(evt) => evt.target === clickRef.current && onClose()}
           ref={clickRef}
+          id={id}
         >
           {children}
         </div>
```

This tells the focus trap to set focus on the containing div (inside the modal), which in fact cannot receive focus so there's no visible focus. But a single tab-press will move the focus to the first element.

If instead it is a requirement to have the initial focus on the close button (can't imagine that, but who knows), then you could also move the close button to be the first child of the modal content instead of being the last.

packages/app/src/components/fullscreen-chart-tile.tsx
VWSCoronaDashboard1: Yup, and setting the focus on the containing div of the modal worked too. Cheers!

packages/app/src/components/modal.tsx
```diff
@@ -4,26 +4,21 @@ import { ReactNode, useEffect, useRef, useState } from 'react';
...
-export function Modal({
-  id,
-  children,
-  onClose,
-  isFullheight,
-  focusSelector,
-}: ModalProps) {
+export function Modal({ id, children, onClose, isFullheight }: ModalProps) {
   const clickRef = useRef<HTMLDivElement>(null);
+  const divId = useUniqueId();
```
VWSCoronaDashboard10, line 18: nit;
```suggestion
  const focusId = useUniqueId();
```

VWSCoronaDashboard10 (approved):

PR #2986, opened 2021-06-10
Clean up obsolete behavior page and components

VWSCoronaDashboard10:
Removed the old behavior page related stuff and did some refactoring of the data-shaping hooks which didn't really match anymore with the new setup.


packages/app/src/domain/behavior/behavior-line-chart-tile.tsx
```diff
@@ -2,121 +2,100 @@ import {
...
+  values: NationalBehaviorValue[] | RegionalBehaviorValue[]
+) {
+  /**
+   * We'll only render behaviors with 2 or more values, otherwise it cannot
+   * result in a "line" in our line-chart.
+   */
+  const complianceHasEnoughData =
+    (values as NationalBehaviorValue[])
+      .map((x) => x[`${id}_compliance` as const])
+      .filter(isPresent).length > 1;
```
VWSCoronaDashboard1, line 93: Is this still necessary? With the gapped lines for the situations chart the LineTrend now supports rendering series of just one item. Rendering it as a shirt line.

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard4 (approved):

packages/app/src/domain/behavior/behavior-line-chart-tile.tsx
```diff
@@ -39,11 +39,7 @@ export function BehaviorLineChartTile({
     <ChartTile title={chartText.title} metadata={metadata}>
       <Text css={css({ maxWidth: '30em' })}>{chartText.description}</Text>
       <Box>
-        <SelectBehavior
-          value={currentId}
-          onChange={setCurrentId}
-          filter={(id) => behaviorHasEnoughLineData(id, values)}
-        />
+        <SelectBehavior value={currentId} onChange={setCurrentId} />
```
VWSCoronaDashboard8, line 42: By removing the filter, shouldn't the trend type for this chart be set to gapped-line?

VWSCoronaDashboard10: Depends on what we want: either join null values or skip null values. Since it has always been a "join" I'd rather not change the line type.
FYI: the single point lines-feature has been implemented in the normal line-trend (the gapped-line trend depends on the line-trend)

VWSCoronaDashboard14 (approved):

PR #2990, opened 2021-06-11
Fix initial SSR size of choropleth

VWSCoronaDashboard1:
## Summary

The choropleths on the situations overview were initially rendered on a very small scale. This was because the minHeight was set, but not the initialWidth accordingly. So, for example, a height of 280 was set, yet the initial width had a default of 850, resulting in a tiny map.

I've fixed this by making the initialWidth a multiplication of the minHeight and 0.9. (I think the ratio for the maps is between 0.8/0.9 to 1, so I chose to round up)

## Unresolved questions

### Governance

- [ ] Documentation is added

### DIFF
--- a/packages/app/src/components/choropleth/choropleth.tsx
+++ b/packages/app/src/components/choropleth/choropleth.tsx
@@ -139,8 +139,8 @@ const ChoroplethMap: <T1, T3>(
     hoverRef,
     description,
     renderHighlight,
-    initialWidth = 850,
     minHeight = 500,
+    initialWidth = 0.9 * minHeight,
     showTooltipOnFocus,
   } = props;
 


packages/app/src/components/choropleth/choropleth.tsx
```diff
@@ -139,8 +139,8 @@ const ChoroplethMap: <T1, T3>(
     hoverRef,
     description,
     renderHighlight,
-    initialWidth = 850,
     minHeight = 500,
+    initialWidth = 0.9 * minHeight,
```
VWSCoronaDashboard8, line 143: I don't quite understand this fix. The initial width default goes from 850 to 450 and that somehow makes the component render bigger in SSR?

VWSCoronaDashboard1: No, in the case of the situations overview choropleths a minHeight of 280 is set, resulting in a container of 280x850. I should have stated more clearly in the description. (fixed that)

VWSCoronaDashboard1: So, in the case of the bigger choropleths, the problem wasn't as explicit as in the case of the small ones.

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard14 (approved):

PR #2995, opened 2021-06-14
Update time series chart dashed line in the legend

VWSCoronaDashboard14:
On iOS, the dashed line in the legend didn't show up. This was due to the fact that the width (15px) was too short resulting that it looked like a solid line. By using the linear-gradient technique it shows a perfect dashed line again. The reason why I choose to set it on a pseudo-element was that I could easily enable/disable it by only changing the `content`.

Live:             |  After the fix:
:-------------------------:|:-------------------------:
IMG-4436  |  IMG-4435


VWSCoronaDashboard10 (commented):
This style of line is already available as a series icon, it would require a change in the chart component:

```tsx
// file domain/tested/infected-per-age-group/infected-per-age-group.tsx

/* Static legend contains always enabled items and the under reported item */
  const staticLegendItems = seriesConfig
    .filter((item) => alwayEnabled.includes(item.metricProperty))
    .map<LegendItem>((item) => ({
      label: item.label,
      shape: 'custom' as const,
      shapeComponent: <SeriesIcon config={item} />,
    }))
    .concat([
      {
        shape: 'square' as const,
        color: colors.data.underReported,
        label: text.line_chart_legend_inaccurate_label,
      },
    ]);

**Review by VWSCoronaDashboard1 — COMMENTED** [review]
Wouldn't it also be solved if you just make the dashed array slightly smaller?

**Comment by VWSCoronaDashboard14** [comment]
> Wouldn't it also be solved if you just make the dashed array slightly smaller?

Not sure, have to check later today.

**Comment by VWSCoronaDashboard10** [comment]
But still we already have a SeriesIcon to use for these kind of things; with the following legend-configuration the SeriesIcon is used:

```tsx
  shape: 'custom' as const,
  shapeComponent: <SeriesIcon config={item} />,
```

VWSCoronaDashboard1:
> But still we already have a SeriesIcon to use for these kind of things; with the following legend-configuration the SeriesIcon is used:
> 
> ```tsx
>   shape: 'custom' as const,
>   shapeComponent: <SeriesIcon config={item} />,
> ```

Yes, and AFAIK, the icon is being used. It's just that there's a small visual glitch in it on iPhone when displaying a dashed line.

VWSCoronaDashboard10:
The fact that he's changing css inside the legend component already shows that that SeriesIcon is not used?

PR #3000, opened 2021-06-15
Add variants page

VWSCoronaDashboard14:
- Add the variants page behind a feature flag
- Add the top header on the variants page
- Add mock data based on the schema
- Added a page in Sanity with articles + handy links
- Some new Lokalize keys

**Note**  the mock data can be tweaked later on, I just needed it to not error straight way.

**Note**: Only the key `date_of_insertion` has to be `date_of_insertion_unix` in the schema. Once it is there I will update it here. -> I've added that in this branch

Screenshot 2021-06-15 at 11 20 48


packages/cms/src/schemas/objects/decorated-link.ts
```diff
@@ -1,4 +1,4 @@
-export const decoratedLink = {
+export const link = {
```
VWSCoronaDashboard1, line 1: I don't understand this rename. decoratedLink is renamed to link, while the filename is called decorated-link and the schema name is also decoratedLink...

packages/cms/src/schemas/objects/link.ts
```diff
@@ -0,0 +1,28 @@
+export const decoratedLink = {
```
VWSCoronaDashboard1, line 1: And here the variable is called decoratedLink, but the file is called link and the schema name is link.

VWSCoronaDashboard1 (approved):

packages/cms/src/schemas/objects/decorated-link.ts
VWSCoronaDashboard14: Oops, I've messed that up, that should be at least be switched.

packages/cms/src/lokalize/key-mutations.csv
```diff
@@ -9,3 +9,20 @@ timestamp,action,key
...
+2021-06-14T10:39:47.012Z,add,covid_varianten.rivm.href
+2021-06-14T10:40:04.912Z,add,covid_varianten.rivm.text
+2021-06-14T10:46:31.291Z,delete,covid_varianten.rivm.text
+2021-06-14T10:46:31.291Z,delete,covid_varianten.rivm.download
+2021-06-14T10:46:31.293Z,delete,covid_varianten.rivm.href
+2021-06-14T10:46:54.548Z,add,covid_varianten.bronnen.rivm.text
+2021-06-14T10:47:07.016Z,add,covid_varianten.bronnen.rivm.href
+2021-06-14T10:47:21.585Z,add,covid_varianten.bronnen.rivm.download
+2021-06-15T08:31:00.698Z,add,covid_varianten.informatie_blok.artikelen_titel
+2021-06-15T08:31:20.696Z,add,covid_varianten.informatie_block.nuttige_links_titel
```
VWSCoronaDashboard8, line 28: typo

packages/app/src/assets/chevron-2.svg
```diff
@@ -0,0 +1,3 @@
+<svg focusable="false" width="36" height="36" viewBox="0 0 36 36" fill="none" xmlns="http://www.w3.org/2000/svg">
```
VWSCoronaDashboard10, line 1: I don't really like the name `chevron-2`

packages/app/src/config/features.ts
```diff
@@ -41,4 +41,9 @@ export const features: Feature[] = [
     metricScopes: ['vr', 'vr_collection'],
     metricName: 'situations',
   },
+  {
+    name: 'variantsPagePage',
```
VWSCoronaDashboard10, line 45:
```suggestion
    name: 'variantsPage',
```

packages/app/src/domain/variants/logic/mock-data.ts
```diff
@@ -0,0 +1,59 @@
+const NOW_IN_SECONDS = Math.round(Date.now() / 1000);
+const ONE_WEEK_IN_SECONDS = 60 * 60 * 24 * 7;
+
+function getVariantsDataValue(date_end_unix: number) {
+  const value = {
```
VWSCoronaDashboard10, line 5:
```suggestion
  const value: YouCanAddATypeOverHere = {
```

packages/app/src/pages/landelijk/covid-varianten.tsx
```diff
@@ -0,0 +1,155 @@
...
+    () => {
+      const data = selectNlPageMetricData('variants')();
+      data.selectedNlData.variants =
+        data.selectedNlData.variants || mockVariantsData();
+
+      return data;
+    },
+    () => {
+      return mockVariantsData();
+    },
```
VWSCoronaDashboard10, line 45: I don't think this belongs here? It will inject values/last_value on the props?

packages/app/src/pages/landelijk/covid-varianten.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+  const text = siteText.covid_varianten;
+
+  const metadata = {
+    ...siteText.nationaal_metadata,
+    title: text.metadata.title,
+    description: text.metadata.description,
+  };
+
+  assert(data.variants, 'no situations data found');
```
VWSCoronaDashboard10, line 76:
```suggestion
  assert(data.variants, 'no variants data found');
```

packages/app/src/pages/landelijk/covid-varianten.tsx
```diff
@@ -0,0 +1,155 @@
...
+                        cover={article.cover}
+                        slug={article.slug.current}
+                      />
+                    ))}
+                  </Box>
+                </Tile>
+              </Box>
+            )}
+
+            {content.page.pageLinks && (
```
VWSCoronaDashboard10, line 122: It appears you're checking the existence of an array, maybe also check on length?

packages/app/src/utils/is-absolute-url.ts
```diff
@@ -0,0 +1 @@
+export const isAbsoluteUrl = (url: string) => /^https?:\/\//.test(url);
```
VWSCoronaDashboard10, line 1: Javascript has this beatiful function keyword:
```suggestion
export function isAbsoluteUrl(url: string) {
  return /^https?:\/\//.test(url);
}
```

VWSCoronaDashboard10 (changes requested):

packages/app/src/utils/is-absolute-url.ts
```diff
@@ -1 +1,3 @@
-export const isAbsoluteUrl = (url: string) => /^https?:\/\//.test(url);
+export function isAbsoluteUrl(url: string) {
+  return /^https?:\/\//.test(url);
```
VWSCoronaDashboard10, line 2: I just realize a url starting with `//` is also an absolute url
```suggestion
  return /^(https?:)?\/\//.test(url);
```

PR #3004, opened 2021-06-16
Add articles & causes of death info tile to country-wide deceased monitor section

VWSCoronaDashboard16:
Update the deceased monitor section on the country-wide deceased page:
- Add an article strip
- Add a tile for information about causes of death

image


packages/app/src/domain/deceased/deceased-monitor-section.tsx
```diff
@@ -3,23 +3,32 @@ import {
...
   showDataMessage?: boolean;
+  showCauseMessage?: boolean;
+  articles?: ArticleSummary[];
 }) {
   const { siteText } = useIntl();
   const text = siteText.section_sterftemonitor;
 
   return (
     <>
+      {articles && <ArticleStrip articles={articles} />}
```
VWSCoronaDashboard10, line 30: Not sure whether it will happen, but might be wise to also check for the articles length

packages/app/src/queries/create-page-articles-query.ts
```diff
@@ -13,9 +13,10 @@ type ArticlePageSchema =
 
 export function createPageArticlesQuery(
   schemaName: ArticlePageSchema,
-  locale: string
+  locale: string,
+  name = 'articles'
```
VWSCoronaDashboard10, line 17: nit; The name `name` is rather vague, how about something like `fieldName`?

VWSCoronaDashboard10 (approved):

packages/app/src/domain/deceased/deceased-monitor-section.tsx
VWSCoronaDashboard10: Ah I see the length is also checked inside the ArticleStrip so I guess it's good for now.

VWSCoronaDashboard14 (approved):

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard1 (approved):

PR #3006, opened 2021-06-16
Fix collapsibles not opening when linked to from the same page

VWSCoronaDashboard16:
Collapsible sections currently only open if their hash is already present on page load. If there is a link to a collapsible section on the same page, clicking that link moves to the section, but does not open the collapsible. This PR makes sure collapsibles also open in those cases.

You can find such a link in the development data on the `/verantwoording` page, within 'Vaccinaties', click on 'gedragsonderzoek van het RIVM'.

### DIFF
--- a/packages/app/src/components/collapsible/collapsible-section.tsx
+++ b/packages/app/src/components/collapsible/collapsible-section.tsx
@@ -92,6 +92,10 @@ const Panel = styled((props) => <DisclosurePanel {...props} />)(
   })
 );
 
+function locationHashEquals(id?: string) {
+  return window.location.hash === `#${id}`;
+}
+
 interface CollapsibleSectionProps extends BoxProps {
   summary: string;
   children: ReactNode;
@@ -116,12 +120,8 @@ export const CollapsibleSection = ({
    * If so, the collapsible needs to be opened.
    */
   useEffect(() => {
-    function isOpenedByHash() {
-      return window.location.hash.substr(1) === id;
-    }
-
     function handleHashChange() {
-      if (isOpenedByHash()) {
+      if (locationHashEquals(id)) {
         setIsOpen(true);
       }
     }
@@ -129,10 +129,10 @@ export const CollapsibleSection = ({
     window.addEventListener('hashchange', handleHashChange);
 
     /**
-     * Since we are open by default, we need to close
-     * all sections which are not opened by a hash now
+     * Since we are open by default, we need to close all sections which are not
+     * opened by a hash now
      */
-    setIsOpen(isOpenedByHash());
+    setIsOpen(locationHashEquals(id));
 
     return () => window.removeEventListener('hashchange', handleHashChange);
   }, [id]);
--- a/packages/app/src/components/collapsible/collapsible-section.tsx
+++ b/packages/app/src/components/collapsible/collapsible-section.tsx
@@ -92,7 +92,7 @@ const Panel = styled((props) => <DisclosurePanel {...props} />)(
   })
 );
 
-function locationHashEquals(id?: string) {
+function locationHashEquals(id: string) {
   return window.location.hash === `#${id}`;
 }
 
@@ -121,7 +121,7 @@ export const CollapsibleSection = ({
    */
   useEffect(() => {
     function handleHashChange() {
-      if (locationHashEquals(id)) {
+      if (id && locationHashEquals(id)) {
         setIsOpen(true);
       }
     }
@@ -132,7 +132,7 @@ export const CollapsibleSection = ({
      * Since we are open by default, we need to close all sections which are not
      * opened by a hash now
      */
-    setIsOpen(locationHashEquals(id));
+    setIsOpen(!!id && locationHashEquals(id));
 
     return () => window.removeEventListener('hashchange', handleHashChange);
   }, [id]);
--- a/packages/app/src/components/collapsible/collapsible-section.tsx
+++ b/packages/app/src/components/collapsible/collapsible-section.tsx
@@ -105,7 +105,7 @@ export const CollapsibleSection = ({
   id,
   hideBorder,
 }: CollapsibleSectionProps) => {
-  const [open, setOpen] = useState(true);
+  const [open, setOpen] = useState(false);
   const { wrapperRef } = useSetLinkTabbability(open);
 
   const { ref, height: contentHeight } = useResizeObserver();
@@ -115,8 +115,18 @@ export const CollapsibleSection = ({
    * If so, the collapsible needs to be opened.
    */
   useEffect(() => {
-    const isOpenedByQueryParam = window.location.hash.substr(1) === id;
-    setOpen(isOpenedByQueryParam);
+    const onHashChange = function () {
+      const isOpenedByQueryParam = window.location.hash.substr(1) === id;
+
+      if (isOpenedByQueryParam) {
+        setOpen(true);
+      }
+    };
+
+    window.addEventListener('hashchange', onHashChange);
+    onHashChange();
+
+    return () => window.removeEventListener('hashchange', onHashChange);
   }, [id]);
 
   return (
--- a/packages/app/src/components/collapsible/collapsible-section.tsx
+++ b/packages/app/src/components/collapsible/collapsible-section.tsx
@@ -105,8 +105,8 @@ export const CollapsibleSection = ({
   id,
   hideBorder,
 }: CollapsibleSectionProps) => {
-  const [open, setOpen] = useState(false);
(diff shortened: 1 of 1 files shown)


packages/app/src/components/collapsible/collapsible-section.tsx
```diff
@@ -105,7 +105,7 @@ export const CollapsibleSection = ({
   id,
   hideBorder,
 }: CollapsibleSectionProps) => {
-  const [open, setOpen] = useState(true);
+  const [open, setOpen] = useState(false);
```
VWSCoronaDashboard16, line 108: @VWSCoronaDashboard10 I see you once explicitly changed this to `true`. Do I break any invisible functionality by changing it back? (it seems to work fine)

VWSCoronaDashboard8: While you're changing this code anyway maybe use a better name for the boolean. We normally prefix them so it's clear that it's a boolean. "open" can be anything, and I would expect it to be a function most of all. 

```suggestion
  const [isOpen, setIsOpen] = useState(false);
```

packages/app/src/components/collapsible/collapsible-section.tsx
```diff
@@ -115,8 +115,18 @@ export const CollapsibleSection = ({
    * If so, the collapsible needs to be opened.
    */
   useEffect(() => {
-    const isOpenedByQueryParam = window.location.hash.substr(1) === id;
-    setOpen(isOpenedByQueryParam);
+    const onHashChange = function () {
```
VWSCoronaDashboard8, line 118: We use this pattern of `onEvenName`/`handleEventName`, where the `on` is used for prop names, and `handle` is used for the function actually handling the event. It makes things more clear especially when you need to locally handle the event and then call another handler from the props.

Also we prefer named functions over unnamed lambdas and usually only use lambdas inline

 ```suggestion
    function handleHashChange()  {
```

File: packages/app/src/components/collapsible/collapsible-section.tsx (hunk 1 above)
Comment (VWSCoronaDashboard14) [reply packages/app/src/components/collapsible/collapsible-section.tsx:108]: If I recall correctly that is due to the fact when Javascript is disabled, they are by default open.

File: packages/app/src/components/collapsible/collapsible-section.tsx [hunk 3]
```diff
@@ -115,8 +115,18 @@ export const CollapsibleSection = ({
...
+      const isOpenedByQueryParam = window.location.hash.substr(1) === id;
```
VWSCoronaDashboard8, line 119: nit; I think this is a little more readable. Param is kind of implied by the use of query already.

```suggestion
      const isOpenedByQuery = window.location.hash.includes(id);
```

packages/app/src/components/collapsible/collapsible-section.tsx
```diff
@@ -115,8 +115,18 @@ export const CollapsibleSection = ({
...
+
+      if (isOpenedByQueryParam) {
+        setOpen(true);
+      }
+    };
+
+    window.addEventListener('hashchange', onHashChange);
+    onHashChange();
```
VWSCoronaDashboard8, line 127: I suspect you can do this

```
function isOpenedByQuery(id: string) { 
   return window.location.hash.includes(id);
}
```

```
const [isOpen, setIsOpen] = useState(isOpenedByQuery(id))
```

Then you can omit the `const isOpenedByQueryParam` and you no longer need to call the handler onHashChange in the effect, because that feels an odd way to set what appears to be the initial state.

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard8 (changes requested):

packages/app/src/components/collapsible/collapsible-section.tsx
VWSCoronaDashboard10: yep that's the use case for initially open collapsibles

packages/app/src/components/collapsible/collapsible-section.tsx
VWSCoronaDashboard10: Nope, that won't work because server-side there's no window.location

packages/app/src/components/collapsible/collapsible-section.tsx
VWSCoronaDashboard8: Would be great to have that in a comment then :)

packages/app/src/components/collapsible/collapsible-section.tsx
VWSCoronaDashboard8: Why are we reading from window.location anyway. Can't we get the param from the nextjs route api? If the server is rendering a page for a url with a query parameter, then that should define the state of the collapsible, right?

VWSCoronaDashboard10: It's reading the hash which is not part of the router, hashes aren't available server-side either.

VWSCoronaDashboard8: Ah right, I didn't know that, obviously 👍 

So in short, we render the collapsibles open by default for non-JS, and we need to call the hash handler in use-effect to close them again client-side if no hash is set on the url.

packages/app/src/components/collapsible/collapsible-section.tsx
VWSCoronaDashboard16: I think the current method is most resilient. Using `includes()` or even `endsWith()` will be risky when IDs have overlapping parts. In the case of `includes()`, if you go to `http://localhost:3000/verantwoording#sterftemonitor-cbs` the 'Sterfte' section will also open, for example.

VWSCoronaDashboard8: 👍

VWSCoronaDashboard10: Agree. For readability this is also an option:
```suggestion
      const isOpenedByQueryParam = window.location.hash === `#${id}`;
```

packages/app/src/components/collapsible/collapsible-section.tsx
```diff
@@ -115,8 +116,25 @@ export const CollapsibleSection = ({
...
+      if (isOpenedByHash()) {
+        setIsOpen(true);
+      }
+    }
+
+    window.addEventListener('hashchange', handleHashChange);
+
+    /**
+     * Since we are open by default, we need to close
+     * all sections which are not opened by a hash now
```
VWSCoronaDashboard8, line 133: tip; there's this plugin called re-wrap in vscode that can reformat comments like these, so you don't have to mess with line endings yourself.

VWSCoronaDashboard8 (approved):

packages/app/src/components/collapsible/collapsible-section.tsx
```diff
@@ -115,8 +116,25 @@ export const CollapsibleSection = ({
    * If so, the collapsible needs to be opened.
    */
   useEffect(() => {
-    const isOpenedByQueryParam = window.location.hash.substr(1) === id;
-    setOpen(isOpenedByQueryParam);
+    function isOpenedByHash() {
+      return window.location.hash.substr(1) === id;
+    }
```
VWSCoronaDashboard8, line 121: With the current use-case there is not really a need to place it in a function anymore. A variable like you had before I guess makes a little more sense.

VWSCoronaDashboard10: It does make a little bit of sense because otherwise the `window.location.hash.substr(1) === id` should be duplicated (one in the effect and another in the event handler).

An improvement could be to create a function outside of the component's scope;

```tsx
function equalsLocationHash(id: string) {
  return window.location.hash.substr(1) === id
}

File: packages/app/src/components/collapsible/collapsible-section.tsx (hunk 3 above)
Comment (VWSCoronaDashboard16) [reply packages/app/src/components/collapsible/collapsible-section.tsx]: Even that would not be functionally equivalent. This would open all collapsibles without an `id` if the hash would be literally '#undefined', while the former also protects against that particular case.

Then again, I don't think a hash will ever be 'undefined', so I think we can go for readability here.

File: packages/app/src/components/collapsible/collapsible-section.tsx [hunk 7]
```diff
@@ -92,6 +92,10 @@ const Panel = styled((props) => <DisclosurePanel {...props} />)(
   })
 );
 
+function locationHashEquals(id?: string) {
```
VWSCoronaDashboard10, line 95: id should not be undefined

VWSCoronaDashboard10 (approved):

PR #3010, opened 2021-06-16
Add variants over time chart

VWSCoronaDashboard14:
Add the new variants over time chart some key takeaways:
- Only when variants have from the backend a `_is_variant_of_concern` suffix it should show in the chart
- When the highest value is 10% or below the domain should stay at 0-10, same goes for 80+ it goes to 0-100% But this is now the case for every chart that makes use of percentage (this is discussed with design and hopefully shouldn't bring any issues to the table)

**Note** is currently on hold and waiting for the `variants table branch` where the named variant keys live. So I still have to change that once that get's merged in.

Screenshot with not so nice mock data
Screenshot 2021-06-16 at 15 56 27


packages/app/src/domain/variants/variants-over-time/variants-over-time.tsx
```diff
@@ -0,0 +1,112 @@
+import { NlVariantsValue } from '@corona-dashboard/common';
```
VWSCoronaDashboard1, line 1: I think you can rename this file to index.tsx and remove the barrel file since that one only contains one export, which is this component.

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard1 (approved):

PR #3012, opened 2021-06-16
Add transform to day timestamps for vaccine coverage chart

VWSCoronaDashboard8:
## Summary

We would prefer to show day timestamps after all. This is a quick fix.

### DIFF
--- a/packages/app/src/pages/landelijk/vaccinaties.tsx
+++ b/packages/app/src/pages/landelijk/vaccinaties.tsx
@@ -1,4 +1,7 @@
-import { NlVaccineCoveragePerAgeGroupValue } from '@corona-dashboard/common';
+import {
+  NlVaccineCoveragePerAgeGroupValue,
+  NlVaccineCoverageValue,
+} from '@corona-dashboard/common';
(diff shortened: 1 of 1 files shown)


packages/app/src/pages/landelijk/vaccinaties.tsx
```diff
@@ -390,3 +393,12 @@ function mockCoverageData(): { values: NlVaccineCoveragePerAgeGroupValue[] } {
...
   }
 }
+
+function transformToDayTimestamps(values: NlVaccineCoverageValue[]) {
+  return values.map((x) => ({
+    ...x,
+    date_unix: x.date_end_unix,
+    date_start_unix: undefined,
+    date_end_unix: undefined,
+  }));
```
VWSCoronaDashboard10, line 403: Can you omit the date_start/end keys completely?
```tsx
  return values.map(({ date_start_unix, date_end_unix, ...value }) => ({
    ...value,
    date_unix: x.date_end_unix,
  }));
```

Otherwise, not sure it would happen, a `'date_start_unix' in value` would be true though not expected.

VWSCoronaDashboard10 (approved):

PR #3014, opened 2021-06-17
Implement time series non-interactive option

VWSCoronaDashboard8:
## Summary

For the sewer split area chart, when selecting a location, we would like to display two trends of which only one is interactive, meaning the other one will not have hover state nor show up in the tooltip. It basically serves as a backdrop to give some context to the values of the other trend.

In order to achieve this I had to make sure the bisect function (which finds the nearest time series index on hover) only looks at samples that have a value filled for the interactive trends, because in case of the sewer data the interactive trend is more sparse than the non-interactive trend. 

Without this additional filtering step, the tooltip/hover state would disappear a lot when hovering over areas where the interactive trend has no values.


packages/app/src/components/time-series-chart/components/tooltip/tooltip-series-list.tsx
```diff
@@ -149,15 +149,23 @@ interface TooltipListItemProps {
...
   children,
   icon,
   label,
   displayTooltipValueOnly,
+  isVisuallyHidden,
 }: TooltipListItemProps) {
-  return (
+  return isVisuallyHidden ? (
+    <VisuallyHidden>
+      <li>
```
VWSCoronaDashboard10, line 164: You're wrapping a div around the list item. By adjusting the the VisuallyHidden we can pass an `as` prop:
```tsx
<VisuallyHidden as='li'>{label}:{children}</VisuallyHidden>

/**
 * visually-hidden.tsx
 */
interface VisuallyHiddenProps {
  children: ReactNode;
  as?: ComponentProps<typeof StyledVisuallyHidden>['as'];
}

export function VisuallyHidden({ children, as }: VisuallyHiddenProps) {
  return <StyledVisuallyHidden as={as}>{children}</StyledVisuallyHidden>;
}

const StyledVisuallyHidden = styled.div(
  css({
    position: 'absolute',
    overflow: 'hidden',
    width: '1px',
    height: '1px',
    padding: 0,
    border: 0,
    margin: -1,
    clip: 'rect(0, 0, 0, 0)',
  })
);
```

VWSCoronaDashboard8: 👍 That's quite a lot of changes to avoid a wrapper div, but I think you're right. It is both more elegant and semantically correct.

packages/app/src/components/time-series-chart/logic/hover-state.ts
```diff
@@ -100,16 +138,42 @@ export function useHoverState<T extends TimestampedValue>({
...
+      );
+
+      const timestamp = interactiveValuesDateUnix[index];
+
+      const indexInAllValues = allValuesDateUnix.findIndex(
+        (x) => x === timestamp
+      );
+
+      assert(
+        isDefined(indexInAllValues),
```
VWSCoronaDashboard1, line 170: `findIndex` returns -1 when the given item isn't found. I don't think `isDefined` checks for this, no?

packages/app/src/components/time-series-chart/logic/hover-state.ts
```diff
@@ -414,3 +481,11 @@ function findActiveTimespanAnnotationIndex(
...
   }
 }
+
+function hasSomeFilledProperties(value: Record<string, unknown>) {
+  const entries = Object.entries(value).filter(([_, value]) =>
+    isPresent(value)
+  );
+
+  return entries.length > 0;
+}
```
VWSCoronaDashboard1, line 491:
```suggestion
function hasSomeFilledProperties(value: Record<string, unknown>) {
  return Object.entries(value).some(([_, value]) =>
    isPresent(value)
  );
}
```

VWSCoronaDashboard1 (approved):

packages/app/src/components/time-series-chart/logic/hover-state.ts
VWSCoronaDashboard1: Wait, this will probably also work, even shorter hehe:

function hasSomeFilledProperties(value: Record&lt;string, unknown&gt;) {  return Object.values(value).some(isPresent);}

packages/app/src/components/time-series-chart/logic/hover-state.ts
VWSCoronaDashboard8: Good catch!

packages/app/src/components/time-series-chart/logic/hover-state.ts
VWSCoronaDashboard8: It used to be a longer chain that used the both elements of the entry but yes now its obsolete.

PR #3015, opened 2021-06-17
Add dropdown with sorting options to scoreboard

VWSCoronaDashboard13:
## Summary

Added a dropdown the scoreboard sections to allow user to choose sort order

## Detailed design

Created a const object with the sort options and their sort functions and used the keys to populate the dropdown options. Assigned the selected value to a sortOption state value and used to select the appropriate sort function that is applied to the entries in all sections

### DIFF
--- a/packages/app/src/domain/escalation-level/scoreboard/data-selection/select-scoreboard-data.ts
+++ b/packages/app/src/domain/escalation-level/scoreboard/data-selection/select-scoreboard-data.ts
@@ -3,7 +3,7 @@ import { isDefined } from 'ts-is-present';
 import { vrData } from '~/data/vr';
 import { EscalationLevel } from '~/domain/restrictions/type';
 import { loadAndSortVrData } from '~/static-props/get-data';
-import { ScoreboardRow } from '..';
+import { ScoreboardRow } from '../logic';
 
 const escalationLevels: EscalationLevel[] = [1, 2, 3, 4, null];
 
--- a/packages/app/src/domain/escalation-level/scoreboard/data-selection/select-scoreboard-data.ts
+++ b/packages/app/src/domain/escalation-level/scoreboard/data-selection/select-scoreboard-data.ts
@@ -3,12 +3,12 @@ import { isDefined } from 'ts-is-present';
 import { vrData } from '~/data/vr';
 import { EscalationLevel } from '~/domain/restrictions/type';
 import { loadAndSortVrData } from '~/static-props/get-data';
-import { ScoreboardRow } from '../logic';
+import { ScoreboardRowData } from '../logic';
 
 const escalationLevels: EscalationLevel[] = [1, 2, 3, 4, null];
 
 export function selectScoreboardData() {
-  const scoreboardRows = vrData.reduce<ScoreboardRow[]>(
+  const scoreboardRows = vrData.reduce<ScoreboardRowData[]>(
     (sbData, vr) => {
       const vrData = loadAndSortVrData(vr.code);
       const index =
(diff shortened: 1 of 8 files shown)


packages/app/src/domain/escalation-level/scoreboard/logic/sort-options.tsx
```diff
@@ -0,0 +1,78 @@
...
+    return 0;
+  },
+  location_z_to_a: (a: VrScoreboardData, b: VrScoreboardData) => {
+    if (a.safetyRegionName > b.safetyRegionName) return -1;
+    if (b.safetyRegionName > a.safetyRegionName) return 1;
+    return 0;
+  },
+  positively_tested_high_to_low: (a: VrScoreboardData, b: VrScoreboardData) => {
+    if (
+      (a.data.positive_tested_per_100k as number) >
```
VWSCoronaDashboard1, line 16: you can't assume this is always a valid number, better to write it like this:
```suggestion
      (a.data.positive_tested_per_100k ?? -Infinity) >
```

VWSCoronaDashboard1: Same goes for the rest of the casts of course

VWSCoronaDashboard1 (changes requested):

packages/app/src/domain/escalation-level/scoreboard/logic/sort-options.tsx
```diff
@@ -0,0 +1,78 @@
...
+      (a.data.hospital_admissions_per_million as number)
+    )
+      return 1;
+    return 0;
+  },
+} as const;
+
+export const scoreboardSortIdentifiers = Object.keys(
+  scoreboardSortOptions
+) as SortIdentifier[];
```
VWSCoronaDashboard8, line 76: nit; This const is a little unnecessary I think. You only use it in one place, so you might as well use Object.keys() there so it's a little more transparent and you have less abstraction.

packages/app/src/domain/escalation-level/scoreboard/index.tsx
```diff
@@ -30,6 +27,20 @@ export function Scoreboard({
   maxHospitalAdmissionsPerMillion,
   maxPositiveTestedPer100k,
 }: ScoreboardProps) {
+  const [sortOption, setSortOption] = useState<SortIdentifier>(
+    Object.keys(scoreboardSortOptions)[0] as SortIdentifier
```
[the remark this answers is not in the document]
VWSCoronaDashboard8: Reading the code below it seems like the definition is fine so I think you could write this instead
```suggestion
    'location_a_to_z'
```

VWSCoronaDashboard8 (approved):

packages/app/src/domain/escalation-level/scoreboard/data-selection/select-scoreboard-data.ts
```diff
@@ -3,7 +3,7 @@ import { isDefined } from 'ts-is-present';
 import { vrData } from '~/data/vr';
 import { EscalationLevel } from '~/domain/restrictions/type';
 import { loadAndSortVrData } from '~/static-props/get-data';
-import { ScoreboardRow } from '..';
+import { ScoreboardRow } from '../logic';
```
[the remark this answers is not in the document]
VWSCoronaDashboard13: Good point! changing it to ScoreboardRowData 👍

VWSCoronaDashboard1 (approved):

PR #3019, opened 2021-06-18
Make sidebar date wrap on a new line 

VWSCoronaDashboard14:
Now when the value in the sidebar gets too big and the date doesn't fit next to it anymore, it jumps to a new line.

Screenshot:
Screenshot 2021-06-18 at 09 29 27


VWSCoronaDashboard16 (approved):
Would've expected a `white-space: nowrap` on the date string itself but this gives a cleaner look 👌

PR #3029, opened 2021-06-18
Converting g number chart to TimeSeriesChart

VWSCoronaDashboard13:
## Summary

Updating the g-number chart to use the time series chart

## Motivation

Cleanup

## Detailed design

- Added the ability to show negative values on the timeSeriesChart
- Added a 0 axis highlighter when there are negative and positive values on the chart
- Added a negativeColor to the bar chart config
- Updated bar trend position logic
- **!! Updated g-number schema. Previously there had been some confusion when discussing the g number schema and it ended up using time span data which I updated the vertical bar chart to account for. Now that it is using the TimeSeriesChart, the data needs to be represented as a daily value !!** 

## Drawbacks

Hoping I didn't break anything 🤞


packages/app/src/components/time-series-chart/time-series-chart.tsx
```diff
@@ -315,6 +328,7 @@ export function TimeSeriesChart<
     }
   }, [onSeriesClick, seriesConfig, tooltipData]);
 
+  const highlightZero = yScale.domain()[0] < 0 && yScale.domain()[1] > 0;
```
VWSCoronaDashboard8, line 331: nit; Maybe a little cleaner to use lodash first/last for this. In theory domain is not guaranteed to be 2 values I think.

VWSCoronaDashboard8:
Closed in favor of #3195

PR #3033, opened 2021-06-21
Variants table QA tweaks

VWSCoronaDashboard1:
## Summary

- Changed the column order on desktop view
- Added a min-width of 2px to the percentage chart so that very small percentages are still visible
- Added a max-width of 20em to the percentage chart
- On mobile the entire row is now clickable. I've had to jump through a few hoops there in order to keep the Disclosure button in (we want to use that for its accessibility features), so if you're wondering why I'm fiddling with a forwardRef there. Well, that's why.
- Aligned the chevron to the right on mobile view.
- Gave all cells and headers a top vertical alignment so things look nice when content is wrapping.

## Motivation

QA comments.

### Governance

- [ ] Documentation is added

### DIFF
--- a/packages/app/src/domain/variants/variants-table-tile/components/desktop-variants-table.tsx
+++ b/packages/app/src/domain/variants/variants-table-tile/components/desktop-variants-table.tsx
@@ -13,10 +13,10 @@ import { VariantRow } from '../logic/use-variants-table-data';
 
 const columnKeys = [
   'variant_titel',
-  'eerst_gevonden',
-  'aantal_monsters',
   'percentage',
   'vorige_meeting',
+  'aantal_monsters',
+  'eerst_gevonden',
 ] as const;
 
 type DesktopVariantsTableProps = {
@@ -43,7 +43,13 @@ export function DesktopVariantsTable(props: DesktopVariantsTableProps) {
           <tr key={row.variant}>
             <VariantNameCell variant={row.variant} text={text} />
             <Cell>
-              <InlineText>{row.countryOfOrigin}</InlineText>
+              <PercentageBarWithNumber
+                percentage={row.percentage}
+                color={row.color}
+              />
+            </Cell>
+            <Cell>
+              {row.difference && <VariantDifference value={row.difference} />}
             </Cell>
             <Cell>
               <NumberOfSamples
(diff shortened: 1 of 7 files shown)


packages/app/src/domain/variants/variants-table-tile/components/mobile-variants-table.tsx
```diff
@@ -110,19 +118,20 @@ function MobileVariantRow(props: MobileVariantRowProps) {
   );
 }
 
-const Chevron = styled((props) => <DisclosureButton {...props} />)(
+const Chevron = styled(
+  forwardRef((props, ref) => <DisclosureButton {...(props as any)} ref={ref} />)
+)(
```
VWSCoronaDashboard10, line 123: Wouldn't `styled(DisclosureButton)` be sufficient? Cause the DisclosureButton obviously is able to receive a ref and styled-component is obviously able to pass on this ref. The only reason you had to use the forwardRef is because of the anonymous function between SC and the Disclosure button.

packages/app/src/domain/variants/variants-table-tile/components/desktop-variants-table.tsx
```diff
@@ -43,7 +44,15 @@ export function DesktopVariantsTable(props: DesktopVariantsTableProps) {
...
-              <InlineText>{row.countryOfOrigin}</InlineText>
+              <Box maxWidth="20em">
+                <PercentageBarWithNumber
+                  percentage={row.percentage}
+                  color={row.color}
+                />
+              </Box>
+            </Cell>
+            <Cell>
+              {row.difference && <VariantDifference value={row.difference} />}
```
VWSCoronaDashboard10, line 55: The VariantDifference looks like it's able to handle an undefined `value` (though not properly typed). When it receives undefined it will display a dash. Don't we want that behavior?

VWSCoronaDashboard10 (approved):

packages/app/src/domain/variants/variants-table-tile/components/mobile-variants-table.tsx
VWSCoronaDashboard1: I tried sticking a ref directly on the button, but that didn;t work, that's why I had tor resort to this trickery.

VWSCoronaDashboard1: WHat you state as obvious, I thought was obvious too, but allas...

VWSCoronaDashboard8 (approved):

PR #3037, opened 2021-06-22
Add new header information block

VWSCoronaDashboard14:


VWSCoronaDashboard8 (approved):
Looks great 👍  There appears to be a lot of new code (which I guess has been copied to be nested components of the new block), and I assume all of the old components like the original metadata can be deleted together with the content header when we fully transitioned to this new information block?

I guess this might still be pending the header-level discussion, but I'll approve it anyway.

VWSCoronaDashboard14:
> Looks great 👍 There appears to be a lot of new code (which I guess has been copied to be nested components of the new block), and I assume all of the old components like the original metadata can be deleted together with the content header when we fully transitioned to this new information block?
> 
> I guess this might still be pending the header-level discussion, but I'll approve it anyway.

Yeah, thanks! It is not entirely done yet, there are still some action points before this can be merged, but the design needs to check this out first. 

Some things like changing the heading hierarchy on certain pages (we landed on a solution but haven't added it yet), adding new Lokalise keys, removing the old header component (including the metadata), and changing the last 10 pages with the new header component. So almost done but it is a good work in progress draft.

VWSCoronaDashboard8:
@VWSCoronaDashboard14 If it's not ready for review yet please keep it a draft PR, because I expect this to be ready if not and it prevents us from reviewing things twice.

PR #3040, opened 2021-06-23
Make non icon value in tooltip align to the left

VWSCoronaDashboard14:
Currently, when there is no icon selected the text will be aligned all the same in the tooltip. By this change, it will be aligned with the icons on the far left. The request was by changing it in the variants over time graph but I'm in favor of applying it everywhere where this pattern has been used.

**See example screenshots where the change is also applicable.**
Screenshot 2021-06-23 at 08 22 51
Screenshot 2021-06-23 at 08 21 07
Screenshot 2021-06-23 at 08 20 51

### DIFF
--- a/packages/app/src/components/time-series-chart/components/tooltip/tooltip-series-list.tsx
+++ b/packages/app/src/components/time-series-chart/components/tooltip/tooltip-series-list.tsx
@@ -172,7 +172,7 @@ function TooltipListItem({
   ) : (
     <Box
       as="li"
-      spacing={2}
+      spacing={icon ? 2 : 0}
       spacingHorizontal
       display="flex"
       alignItems="stretch"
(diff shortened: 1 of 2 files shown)


VWSCoronaDashboard4 (approved):

packages/app/src/components/time-series-chart/components/tooltip/tooltip-series-list.tsx
```diff
@@ -195,7 +195,7 @@ function TooltipListItem({
               {icon}
             </Box>
           ) : (
-            <Box width="1em" mt={1} />
+            <Box width={0} mt={1} />
```
VWSCoronaDashboard8, line 198: Wouldn't it render the same if you completely removed the Box here, or the `width={0}` part?

VWSCoronaDashboard8 (approved):

PR #3051, opened 2021-06-24
Accessibility annotations

VWSCoronaDashboard4:
## Summary

Adds references to all charts and choropleths to describe content and functionality for assistive softweare using `aria-label` and visually hidden description referenced using `aria-describedby`.


packages/app/src/components/age-demographic/age-demographic-chart.tsx
```diff
@@ -74,6 +81,9 @@ function AgeDemographicChartWithGenerics<T extends AgeDemographicDefaultValue>({
 
   const { formatPercentage } = useIntl();
 
+  const { label, description, describedById } =
+    useAccessibilityOptions(accessibility);
```
VWSCoronaDashboard8, line 85: The names confuse me. `accessibility` is typed as `AccessibilityOptions` but then used as input for `useAccessibilityOptions`. The result from `useAccessibilityOptions` do not appear to be options to me.

I would suggest to name the property `accessibilityOptions` and the hook something like `useAccessibility`.

packages/app/src/components/time-series-chart/time-series-chart.tsx
```diff
@@ -75,14 +79,17 @@ export type TimeSeriesChartProps<
   T extends TimestampedValue,
   C extends SeriesConfig<T>
 > = {
+  /**
+   * Accessibility options are not optional
+   */
```
VWSCoronaDashboard8, line 84: This comment seems redundant since you can read from the syntax

packages/app/src/components/age-demographic/age-demographic-chart.tsx
VWSCoronaDashboard8: I think `accessibilityOptions` is a bit long for a prop name. My vote would be `a11yOptions` because I also hate typing the word but maybe that's just me.

VWSCoronaDashboard4: I don't think either is a better prop name. `Options` is a sort of synonym to `prop`; and we don't suffix other props or options with `options`.

Shortening it to `a11y` just to type less can be prevented by using an editor that autocompletes the props.

I can adjust the hook to `useAccessibility`; I had considered using `useAccessibilityTexts`; maybe something in that realm makes more sense?

packages/app/src/components/time-series-chart/time-series-chart.tsx
VWSCoronaDashboard4: I think I should extend / clarify the comment a bit. I can totally see a future developer making it optional in TS just because there is no immediate need to add it, since the effects not noticeable visually.

packages/app/src/components/age-demographic/age-demographic-chart.tsx
VWSCoronaDashboard10: How about an api where pas an element and an object with aria props:
```tsx
const a11y = useA11y(a11yOptions);

return (
  <Tile>
    {a11y.descriptionElement}
    <svg {...a11y.describedProps} />
  </Tile>
)

Comment (0x80) [reply packages/app/src/components/age-demographic/age-demographic-chart.tsx:85]: I don't agree that props == options. Props I think are like function arguments. Options are a special type of function argument that is optional and has properties that are all optional.

The use hook names normally have their name refer to what they return, and not what you pass in. So that's why I think useAccessiblityOptions is not a good choice because you pass in the options and return non-optional elements.

Whether we use a shorthand a11y or not, I don't mind much.

Comment (VWSCoronaDashboard8) [reply packages/app/src/components/age-demographic/age-demographic-chart.tsx:85]: I don't agree that props == options. Props I think are like function arguments. Options are a special type of function argument that is optional and has properties that are all optional.

The use hook names normally have their name refer to what they return, and not what you pass in. So that's why I think useAccessiblityOptions is not a good choice because you pass in the options and return non-optional elements.

Whether we use a shorthand a11y or not, I don't mind much.

File: packages/app/src/components/time-series-chart/time-series-chart.tsx (hunk 2 above)
Comment (VWSCoronaDashboard8) [reply packages/app/src/components/time-series-chart/time-series-chart.tsx:84]: Yes that would help. Maybe something like "Accessibility arguments should not be optional". And maybe consider naming them something different from options if they're not optional.

File: packages/app/src/utils/use-accessibility-options.ts [hunk 3]
```diff
@@ -0,0 +1,75 @@
+import { assert } from '@corona-dashboard/common';
+import { isDefined } from 'ts-is-present';
+import { useIntl } from '~/intl';
+
+export interface AccessibilityOptions {
```
VWSCoronaDashboard8, line 5:
```suggestion
export interface AccessibilityConfig {
```

or

```suggestion
export interface AccessibilityDefinition {
```

...or something else without options :)

packages/app/src/utils/use-accessibility-options.ts
```diff
@@ -0,0 +1,75 @@
...
+  key: string;
```
VWSCoronaDashboard10, line 6: you can make this more strict and get rid of type casting on line 26

```suggestion
  key: keyof typeof siteText.accessibility.charts
```

packages/app/src/utils/use-accessibility-options.ts
```diff
@@ -0,0 +1,75 @@
...
+   *  `An accessibility label needs to be provided for ${options.key}`
+   * );
+   */
+
+  /**
+   * There needs to be a description, either constructed by a Lokalize text
+   * or by providing interactive features
+   */
+  assert(
+    chartDescription.length > 10 || options.features?.length,
```
VWSCoronaDashboard10, line 43: what is a length of 10 based on? We could also write a custom validation rule in sanity to enforce a minimum length?

VWSCoronaDashboard10 (approved):

packages/app/src/utils/use-accessibility-options.ts
VWSCoronaDashboard4: this unfortunately does not work, since the `siteText` is not available there

VWSCoronaDashboard10: ah I see, try importing the type `SiteText` instead ;-)
```tsx
key: SiteText['accessibility']['charts']

VWSCoronaDashboard4: shoot, Mondays! 😅

VWSCoronaDashboard8 (approved):

PR #3056, opened 2021-06-24
Visual update full screen functionality

VWSCoronaDashboard14:
The button is always visible and gray, and the background color is a bit darker. 

Screenshot 2021-06-24 at 16 15 33


packages/app/src/components/fullscreen-chart-tile.tsx
```diff
@@ -56,8 +55,9 @@ export function FullscreenChartTile({
             position: 'absolute',
             top: '10px',
             right: '10px',
-            color: isFullscreen || isButtonVisible ? 'silver' : 'white',
-            '&:focus-within, &:hover': { color: 'gray' },
+            color: 'gray',
+
+            '&:hover': { color: 'silver' },
```
VWSCoronaDashboard4, line 60: Not on focus?

VWSCoronaDashboard4 (approved):

VWSCoronaDashboard1 (approved):

packages/app/src/components/fullscreen-chart-tile.tsx
VWSCoronaDashboard14: The focus has some interesting behavior on this button,  when the modal opens the focus stays on the page and not travels to the modal. I could add it but I left it intentionally out.

VWSCoronaDashboard4: That would be a good follow up task/PR; to look at the focus containment; I'll add a task for it.

VWSCoronaDashboard10 (approved):

VWSCoronaDashboard8 (approved):

PR #3069, opened 2021-06-28
Add getLocalisedCountryNames util function

VWSCoronaDashboard1:
## Summary

The international pages will need a list of localised country names in various places. This PR adds a util function which can be used in the `getStaticProps` logic to generate a countrycode->localised-countryname dictionary based on the specified data.

### Governance

- [ ] Documentation is added

### DIFF
--- a/packages/app/src/static-props/utils/__tests__/get-localised-country-names.spec.ts
+++ b/packages/app/src/static-props/utils/__tests__/get-localised-country-names.spec.ts
@@ -1,0 +1,101 @@
+import { getLocalisedCountryNames } from '../get-localised-country-names';
+
+describe('StaticProps::Util: getLocalisedCountryNames', () => {
+  let currentLocale = 'nl';
+
+  beforeEach(() => {
+    currentLocale = process.env.NEXT_PUBLIC_LOCALE ?? 'nl';
+  });
+
+  afterEach(() => {
+    process.env.NEXT_PUBLIC_LOCALE = currentLocale;
+  });
+
+  it('Should return the Dutch country names', () => {
+    const values = [
+      {
+        iso_code: 'nld',
+        value: 1,
+      },
+      {
+        iso_code: 'nor',
+        value: 2,
+      },
+      {
+        iso_code: 'afg',
+        value: 3,
+      },
+    ];
+    const result = getLocalisedCountryNames(values, 'iso_code');
+
+    expect(result).toEqual({
+      nld: 'Nederland',
+      nor: 'Noorwegen',
+      afg: 'Afghanistan',
+    });
+  });
+
+  it('Should return the English country names', () => {
+    process.env.NEXT_PUBLIC_LOCALE = 'en';
+    const values = [
+      {
+        iso_code: 'nld',
+        value: 1,
+      },
+      {
+        iso_code: 'nor',
+        value: 2,
+      },
+      {
+        iso_code: 'afg',
+        value: 3,
+      },
+    ];
+    const result = getLocalisedCountryNames(values, 'iso_code');
+
+    expect(result).toEqual({
+      nld: 'Netherlands',
+      nor: 'Norway',
+      afg: 'Afghanistan',
+    });
+  });
+
+  it('Should use the specified function to retrieve the iso code for the given value when dealing with nested objects', () => {
+    const values = [
+      {
+        isoObj: { iso_code: 'nld' },
+        value: 1,
+      },
+      {
+        isoObj: { iso_code: 'nor' },
+        value: 2,
+      },
+      {
+        isoObj: { iso_code: 'afg' },
+        value: 3,
+      },
+    ];
+    const result = getLocalisedCountryNames(
+      values,
+      (value) => value.isoObj.iso_code
+    );
+
+    expect(result).toEqual({
+      nld: 'Nederland',
+      nor: 'Noorwegen',
+      afg: 'Afghanistan',
+    });
+  });
+
+  it('Should Error when an iso code could not be matched', () => {
+    const values = [
+      {
+        iso_code: 'nlr',
+        value: 1,
+      },
+    ];
+    expect(() => getLocalisedCountryNames(values, 'iso_code')).toThrowError(
+      /Unable to find country data for ISO code nlr/
+    );
+  });
+});
--- a/packages/app/src/static-props/utils/get-localised-country-names.ts
+++ b/packages/app/src/static-props/utils/get-localised-country-names.ts
@@ -42,13 +42,12 @@ export function getLocalisedCountryNames<T extends Record<string, any>>(
     (process.env.NEXT_PUBLIC_LOCALE as LocaleCode) ?? 'nl';
 
   return Object.fromEntries(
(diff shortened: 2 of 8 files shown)


packages/app/src/static-props/utils/__tests__/get-localised-country-names.spec.ts
```diff
@@ -0,0 +1,101 @@
+import { getLocalisedCountryNames } from '../get-localised-country-names';
```
VWSCoronaDashboard8, line 1: nit; I think lokalised is spelled lokalized we want to stick to American English.

packages/app/src/static-props/utils/get-localised-country-names.ts
```diff
@@ -0,0 +1,58 @@
+import { assert, KeysOfType } from '@corona-dashboard/common';
+import { isFunction } from 'lodash';
+import { loadJsonFromDataFile } from './load-json-from-data-file';
+
+type LocaleCode = 'nl' | 'en';
+
+type CountryData = {
+  en: string;
+  nl: string;
+  iso_a3: string;
```
VWSCoronaDashboard8, line 10: I think this can just be called `code`. Since the format can be documented somewhere and I suspect if we use a lowercase version it is probably not officially iso_a3 anymore.

packages/app/src/static-props/utils/get-localised-country-names.ts
```diff
@@ -0,0 +1,58 @@
...
+};
+
+const countryData = loadJsonFromDataFile<CountryData[]>(
+  'country-data.json',
+  'static-json'
+);
+
+function findCountryName(iso: string, locale: LocaleCode) {
```
VWSCoronaDashboard8, line 18:
```suggestion
function getCountryName(code: string, locale: LocaleCode) {
```

packages/app/src/static-props/utils/get-localised-country-names.ts
```diff
@@ -0,0 +1,58 @@
...
+      return [isoCode, findCountryName(isoCode, locale)];
+    })
+  ) as Record<string, string>;
+}
+
+function isFunctionParam<T extends Record<string, any>>(
+  value: KeysOfType<T, string, true> | JoinFunction<T>
+): value is JoinFunction<T> {
+  return isFunction(value);
+}
```
VWSCoronaDashboard8, line 58: Do you already have a use-case for this joinStrategy complexity? I find this all very difficult to follow. 

On one hand you are using complex types like KeysOfType and on the other hand there is the use of `any`. 

The concept behind this PR sounds simple yet the code seems complex, so this makes me wonder if there would be a simple approach.

packages/common/src/types/common.ts
```diff
@@ -0,0 +1,13 @@
+export type KeysOfType<T, U, B = false> = {
```
VWSCoronaDashboard8, line 1: If we really need this, please add a comment about what it does exactly, maybe with an example even.

packages/app/src/static-props/utils/get-localised-country-names.ts
VWSCoronaDashboard1: This is the data shape of the static json file. I can convert it of course, but I'm not sure if it really adds or subtracts anything. If you insist, I'll make the change though.

packages/app/src/static-props/utils/get-localised-country-names.ts
VWSCoronaDashboard1: Well, yea, if I recall correctly, the time series data will contain nested objects. So, that's why I added the extra function because simply using a property name won't really work there.
The KeysOfType simply makes sure that the property name you're passing in is of type string. If not, you'll have to cast to string. The `Record<string, any>` in that regard makes sense, because the value you're extracting from can have properties with all sorts of types, but all we're interested in here are the string properties.

VWSCoronaDashboard10: but can't we simply get a list/map of countries and filter them on the calling context? That defeats the need of all the complex typing?

packages/common/src/types/common.ts
VWSCoronaDashboard1: I'm using this in European choropleth as well, in that case I can make sure that `getFIllColor` and other internal methods are strongly typed because I'll be sure I'm receiving a string. We've had this discussion before and I really don't understand what the problem is... This simply ensures that the values we're working with are strings.
I'll add some comments and a link to where I got the type from though.

packages/app/src/static-props/utils/get-localised-country-names.ts
VWSCoronaDashboard1: I don't understand your proposition.

VWSCoronaDashboard1: As far as I see, this is doing that exactly. This method is meant to generate that very list.

VWSCoronaDashboard1: I think the unit tests make it quite evident actually what the input and output are.

VWSCoronaDashboard10: I might be mistaken, but in my head the following is what the getCountries function should do. What's all the rest for?
```tsx
// data:
const data = [
  {
    "en": "Afghanistan",
    "nl": "Afghanistan",
    "iso_a3": "AFG"
  },
  {
    "en": "Albania",
    "nl": "Albanië",
    "iso_a3": "ALB"
  }
];


// util:
function getCountries() {
  return Object.fromEntries(
    data.map(x => [x.iso_a3.toLowerCase(), x[locale]])
  ) as Record<string, string>
}


// usage:
getCountries()

Comment (VWSCoronaDashboard1) [reply packages/app/src/static-props/utils/get-localised-country-names.ts]: no, the country list is a list of all the countries in the world. As we don't know yet which countries exactly are there, it generates a lookup of all the countries that are actually in the provided data.

Comment (VWSCoronaDashboard1) [reply packages/app/src/static-props/utils/get-localised-country-names.ts]: I mean, if you want we can just include the entire list. I was trying to get the data as small as possible

Comment (VWSCoronaDashboard10) [reply packages/app/src/static-props/utils/get-localised-country-names.ts]: With the above code we could do this on the callling context:
```tsx
// given that a json value has the following shape:
const lastValue = {
  date_unix: 123123123,
  afg: {
    infected: 300,
    moving_average: 300,
  },
  alb: {
    infected: 300,
    moving_average: 300,
  },
};

// then we can get a list of all country codes with data:
const countryCodesWithData = Object.keys(lastValue).filter(
  (x) => x !== "date_unix"
);

// and with that list we can get the list of countries with data:
const countriesWithData = getCountries().filter((x) =>
  countryCodesWithData.includes(x)
);
```

VWSCoronaDashboard1: "given that a json value has the following shape", exactly, initially this was the case, but then this morning it was decided we'd have nested structures as well, That's exactly why I added the extra optional function...
The original function was simply a property lookup, I just added one extra option.

PR #3070, opened 2021-06-28
Set up international page

VWSCoronaDashboard10:
Set up the international page.


VWSCoronaDashboard1 (approved):

packages/app/src/queries/create-page-articles-query.ts
```diff
@@ -4,6 +4,7 @@ type ArticlePageSchema =
   | 'hospitalPage'
   | 'intensiveCarePage'
   | 'positiveTestsPage'
+  | 'in_positiveTestsPage'
```
VWSCoronaDashboard8, line 7: 😒   I remember now that I wanted to propose getting rid of all those pages in the CMS. Since they are only used to define articles so we could instead take a similar approach to how timeline annotations are added in the CMS. Add an articles element which is scoped to a certain page via props/identifier. I'll add a note to start some discussion later.

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard14 (approved):

VWSCoronaDashboard4 (approved):

PR #3072, opened 2021-06-28
Add new information header element international page

VWSCoronaDashboard14:
This new information is a copy of https://github.com/minvws/nl-covid19-data-dashboard/pull/3037 but wasn't reviewed yet. So it might be good to review the whole code here.

It renders the information based on what is available for it. So if the articles and/or the useful links for example are missing it would display differently. I've added some screenshots to give a better overview of how dynamic it is.

**How it should be on the new international page:**
Screenshot 2021-06-28 at 16 25 55

**How it would look without articles**
Screenshot 2021-06-28 at 16 26 25

**How it would look without links**
Screenshot 2021-06-28 at 16 27 02

**Without reference link**
Screenshot 2021-06-28 at 16 27 14


packages/app/src/components/information-block/articles.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+export function Articles({ articles }: ArticlesProps) {
+  const { siteText } = useIntl();
+
+  return (
+    <>
+      <InlineText
+        mb={2}
+        fontSize={2}
+        as="span"
```
VWSCoronaDashboard1, line 24: InlineText already is a span

packages/app/src/components/information-block/information-block.tsx
```diff
@@ -0,0 +1,131 @@
...
+    title: string;
+    href: string;
+  }[];
+  headingLevel?: HeadingLevel;
+  metadata?: MetadataProps;
+  referenceLink?: string;
+  id?: string;
+}
+
+export function InformationBlock({
```
VWSCoronaDashboard1, line 27: `InformationBlock` might be a tad too general. Perhaps `PageInformationBlock` or `SectionInformationBlock`?

packages/app/src/components/information-block/information-block.tsx
```diff
@@ -0,0 +1,131 @@
+import css from '@styled-system/css';
```
VWSCoronaDashboard1, line 1: This file should be renamed to `index.tsx` since it already lives in a directory called information-block. The existing index.tsx now only contains one export.

packages/app/src/components/information-block/metadata.tsx
```diff
@@ -0,0 +1,237 @@
...
+  icon: React.ReactNode;
+  referenceLink: string;
+}
+
+function MetadataReference({ icon, referenceLink }: metadataReferenceProps) {
+  const { siteText } = useIntl();
+
+  return (
+    <Box display="flex" alignItems="flex-start" color="annotation">
+      <Box as="span" minWidth="1.8rem" mt="3px">
```
VWSCoronaDashboard1, line 220: There is a styled component called `Span`, is that not usable instead of a Box as  span?

packages/app/src/components/information-block/useful-links.tsx
```diff
@@ -0,0 +1,157 @@
...
+import ChevronLargeIcon from '~/assets/chevron-large.svg';
+import ExternalLinkIcon from '~/assets/external-link-2.svg';
+import { Box } from '~/components/base';
+import { ExternalLink } from '~/components/external-link';
+import { InlineText } from '~/components/typography';
+import { useIntl } from '~/intl';
+import { asResponsiveArray } from '~/style/utils';
+import { isAbsoluteUrl } from '~/utils/is-absolute-url';
+import { Link } from '~/utils/link';
+interface usefulLinksProps {
```
VWSCoronaDashboard1, line 12: nit: put a space between the imports and the interface

packages/app/src/components/information-block/useful-links.tsx
```diff
@@ -0,0 +1,157 @@
...
+      borderTop="1px solid"
+      borderTopColor="border"
+      width="100%"
+      pt={3}
+      mt={3}
+    >
+      <InlineText
+        mb={2}
+        fontSize={2}
+        as="span"
```
VWSCoronaDashboard1, line 33: Same here, the InlineText already is a span

packages/app/src/components/information-block/useful-links.tsx
```diff
@@ -0,0 +1,157 @@
...
+              </>
+            )}
+          </ListItem>
+        ))}
+      </OrderedList>
+    </Box>
+  );
+}
+
+function TitleWidthIcon({
```
VWSCoronaDashboard1, line 68: Typo? I'm guessing this ought to be called `TitleWithIcon`?

VWSCoronaDashboard1 (approved):

packages/app/src/components/information-block/articles.tsx
VWSCoronaDashboard14: Aah, I've copied the old metadata from the content header. Will also remove them there.

packages/app/src/components/page-information-block/metadata.tsx
```diff
@@ -0,0 +1,237 @@
...
+  icon: React.ReactNode;
+  referenceLink: string;
+}
+
+function MetadataReference({ icon, referenceLink }: metadataReferenceProps) {
+  const { siteText } = useIntl();
+
+  return (
+    <Box display="flex" alignItems="flex-start" color="annotation">
+      <Box minWidth="1.8rem" mt="3px">
```
VWSCoronaDashboard10, line 220: same here

packages/app/src/components/page-information-block/page-information-block.tsx
```diff
@@ -0,0 +1,131 @@
...
+  usefulLinks,
+  metadata,
+  referenceLink,
+  id,
+}: InformationBlockProps) {
+  return (
+    <header
+      id={id}
+      css={css({
+        mt: id ? asResponsiveArray({ _: 4, md: 5 }) : 0,
```
VWSCoronaDashboard10, line 42: please no outside margins. Also setting a margin based on the existence of an `id` prop is rather surprising from the calling context. This style must be lifted.

packages/app/src/components/page-information-block/page-information-block.tsx
```diff
@@ -0,0 +1,131 @@
...
+            />
+          ) : (
+            <Box display="flex" flexWrap="nowrap" alignItems="center">
+              <Heading mb={3} lineHeight={1.3} level={headingLevel}>
+                {title}
+              </Heading>
+            </Box>
+          )}
+        </>
+      )}
```
VWSCoronaDashboard10, line 63: fragment wrapper is unnecessary (`<> </>`)

packages/app/src/components/page-information-block/page-information-block.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+      <Tile>
+        <Box
+          display={{ _: 'block', md: 'grid' }}
```
VWSCoronaDashboard10, line 67: If you stick to grid- of flex-display on mobile. you can use the row-gap for spacing between elements, instead of the outside margin you applied on the Metadata component. Or, if you'd like to keep it like this, apply the responsive bottom margin on a metadata wrapper element inside this component.

packages/app/src/components/page-information-block/useful-links.tsx
```diff
@@ -0,0 +1,148 @@
...
+export function UsefulLinks({ links }: usefulLinksProps) {
+  const { siteText } = useIntl();
+
+  return (
+    <Box
+      borderTop="1px solid"
+      borderTopColor="border"
+      width="100%"
+      pt={3}
+      mt={3}
```
VWSCoronaDashboard10, line 29: no outside margins :-)

packages/app/src/components/page-information-block/useful-links.tsx
```diff
@@ -0,0 +1,148 @@
...
+    >
+      <InlineText
+        mb={2}
```
VWSCoronaDashboard10, line 32: this mb can be replaced with a `spacing={2}` on the container Box

packages/app/src/components/page-information-block/useful-links.tsx
```diff
@@ -0,0 +1,148 @@
...
+              <>
+                <Link href={link.href} passHref>
+                  <a>
+                    <TitleWithIcon
+                      title={link.title}
+                      icon={<ChevronLargeIcon />}
+                    />
+                  </a>
+                </Link>
+              </>
```
VWSCoronaDashboard10, line 56: unnecessary fragment

packages/app/src/components/page-information-block/useful-links.tsx
```diff
@@ -0,0 +1,148 @@
...
+      textDecoration: 'none',
+
+      '&:hover': {
+        textDecoration: 'underline',
+      },
+    },
+  })
+);
+
+const IconContainer = styled.div(
```
VWSCoronaDashboard10, line 137: Better not to render divs inside spans
```suggestion
const IconContainer = styled.span(
```
this might need an additional display style though (e.h. display:inline-block) you'll figure out whether its necessary or not.

packages/app/src/components/page-information-block/useful-links.tsx
```diff
@@ -0,0 +1,148 @@
...
+              <IconContainer>{icon}</IconContainer>
+            </InlineText>
+          ) : (
+            `${word} `
+          )}
+        </InlineText>
+      ))}
+    </>
+  );
+}
```
VWSCoronaDashboard10, line 97: Here's a simpler version:

```tsx
function TitleWithIcon({ title, icon }: { title: string; icon?: JSX.Element }) {
  const [lastWord, ...splittedWords] = title.split(' ').reverse();
  const titleWithoutLastWord = splittedWords.reverse().join(' ');

  return (
    <InlineText fontFamily="body" fontSize="1rem">
      {titleWithoutLastWord}{' '}
      <InlineText display="inline-flex" position="relative">
        {lastWord}
        <IconContainer>{icon}</IconContainer>
      </InlineText>
    </InlineText>
  );
}
```

Only extra change needed outside this file is a bugfix on the (inline-)text component, otherwise the `position` props on Text and InlineText won't work (though they were added to the types 🤷‍♂️ )

```diff
diff --git a/packages/app/src/components/typography/text.tsx b/packages/app/src/components/typography/text.tsx
index 87e716067..e1c05b834 100644
--- a/packages/app/src/components/typography/text.tsx
+++ b/packages/app/src/components/typography/text.tsx
@@ -11,6 +11,7 @@ import {
   ColorProps,
   PositionProps,
   TypographyProps,
+  position,
 } from 'styled-system';
 import { styledShouldForwardProp } from '~/utils/styled-should-forward-prop';
 import {
@@ -32,14 +33,14 @@ export type TextProps = SpaceProps &
 export const Text = styled.p.withConfig({
   shouldForwardProp: styledShouldForwardProp,
 })<TextProps>(
-  compose(margin, padding, typography, color, layout),
+  compose(margin, padding, typography, color, layout, position),
   textTransform
 );
 
 export const InlineText = styled.span.withConfig({
   shouldForwardProp: styledShouldForwardProp,
 })<TextProps>(
-  compose(margin, padding, typography, color, layout),
+  compose(margin, padding, typography, color, layout, position),
   textTransform
 );
```

VWSCoronaDashboard10 (changes requested):

VWSCoronaDashboard8 (approved):

packages/app/src/components/information-block/information-block.tsx
VWSCoronaDashboard8: I remember we started adopting a pattern where we deliberately do not put the main component inside of an index file, because it makes it confusing to have multiple index files open.

So you can still create an index file, but I would advise to only use it for exporting the component from information-block.tsx

VWSCoronaDashboard1: He already did that. I suggested renaming this to index.tsx because otherwise it would only contain one export. I was unaware of the aforementioned convention.

VWSCoronaDashboard14: No worries, I'll change it back

packages/app/src/components/page-information-block/useful-links.tsx
VWSCoronaDashboard14: Clever

packages/app/src/components/page-information-block/page-information-block.tsx
VWSCoronaDashboard14: Ah, yeah this was from my old branch when content blocks are a bit lower on the page. Then it needed a bit more breathing room from the component that is above it. And since they all have an ID to scroll to them I figured it was an easy way of figuring out it wasn't the first element on the page. But I see that I might be a little strange.

VWSCoronaDashboard10: I understand, but you should configure that margin outside of this component. Component should only define spacing inside the container.

PR #3075, opened 2021-06-28
Timeline annotations CMS structure

VWSCoronaDashboard8:
## Summary

This enables the CMS to store configurations for data elements on the dashboard. This will be initially used for time series chart timeline events. Since the approach is aimed at a broader picture I've written some notes about it [here](https://github.com/minvws/nl-covid19-data-dashboard/blob/topic/timeline-annotations-cms/packages/cms/src/elements/NOTES.md)

~~This PR still contains some sketch files / code which is only used for injecting mock data into the CMS for testing and QA. I don't think it hurts, and we can take it out before this gets merged.~~ Took out the sketches because they caused compile errors in use outside of ts-node.

**Not in this PR**: Connecting the actual charts with the CMS data. I think that can be task on its own.


packages/cms/src/elements/elements-list-item.ts
```diff
@@ -0,0 +1,57 @@
...
+
+      return documentStore
+        .listenQuery(
+          `*[_type in $types]{ scope, metricName, metricProperty }`,
+          { types }
+        )
+        .pipe(
+          map((doc: { scope: string }[], index) => {
+            const scopes = uniq(doc.map((x) => x.scope));
+            console.log('scopes', scopes);
```
VWSCoronaDashboard10, line 22: console.log

packages/cms/src/elements/elements-list-item.ts
```diff
@@ -0,0 +1,57 @@
...
+    .title('Elements')
+    .icon(FaChartLine)
+    .child(() => {
+      const types = ['timeSeries', 'kpi', 'choropleth'];
+
+      return documentStore
+        .listenQuery(
+          `*[_type in $types]{ scope, metricName, metricProperty }`,
+          { types }
+        )
```
VWSCoronaDashboard10, line 18: do we need to handle drafts here?

packages/cms/src/elements/schemas/choropleth.ts
```diff
@@ -0,0 +1,9 @@
+import { commonFields, commonPreview } from './shared';
+
+export const choropleth = {
+  name: 'choropleth',
```
VWSCoronaDashboard10, line 4: shouldn't these be typed something like `choropleth_element`?

packages/cms/src/elements/schemas/shared/index.ts
```diff
@@ -0,0 +1,72 @@
+import { snakeCase } from 'change-case';
+import { isDefined } from 'ts-is-present';
+
+export const scopes = [
+  { title: 'Actueel', value: 'ac' },
+  { title: 'Landelijk', value: 'nl' },
+  { title: 'Veiligheidsregio', value: 'vr' },
+  { title: 'Gemeente', value: 'gm' },
+  { title: 'Europe', value: 'eu' },
```
VWSCoronaDashboard10, line 9: do you mean international?

```suggestion
  { title: 'Internationaal', value: 'in' },
```

If not, then go for dutch:
```suggestion
  { title: 'Europa', value: 'eu' },
```

packages/cms/src/elements/schemas/shared/index.ts
```diff
@@ -0,0 +1,72 @@
...
+        .join(' - '),
+      subtitle: [x.scope, x.metricName, snakeCase(x.type), x.metricProperty]
+        .filter(isDefined)
+        .join('.'),
+    };
+  },
+};
+
+/**
+ * By maping the metric name to a title we can make the UI a little more
```
VWSCoronaDashboard10, line 61:
```suggestion
 * By mapping the metric name to a title we can make the UI a little more
```

VWSCoronaDashboard10 (approved):

packages/cms/src/elements/NOTES.md
```diff
@@ -0,0 +1,139 @@
...
+like to have configurable by cms (yet) we should be able to programmatically
+look up via the id elements.
+
+## Lokalize Texts Structure
+
+The nice thing about these ids is that we could start using the same path logic
+for lokalize texts. Lokalize text documents are stored with opaque ids so they
+do not clash with the element ids.
+
+If we manage to clean up lokalize structure is becomes possible to join the
```
VWSCoronaDashboard1, line 122:
```suggestion
If we manage to clean up the lokalize structure it becomes possible to join the
```

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard4 (approved):

packages/cms/src/elements/elements-list-item.ts
VWSCoronaDashboard8: Not sure, I think it can't hurt. In theory, if we don't include drafts, I think you could have a draft-only of a certain element and it would therefore not be accessible via this list.

packages/cms/src/elements/schemas/choropleth.ts
VWSCoronaDashboard8: I don't know. Do you expect we need to store other documents with similar type names in the CMS?

packages/cms/src/elements/schemas/shared/index.ts
```diff
@@ -0,0 +1,72 @@
...
+  },
+  {
+    title: 'Metric Name',
+    name: 'metricName',
+    type: 'string',
+    readOnly: true,
+    hidden: true,
+  },
+  {
+    title: 'Metric Property', // should be optional
```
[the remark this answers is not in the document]
VWSCoronaDashboard8: It was just a reminder that these should be optional. I'll add validation = required to the others and remove the comment

packages/common/src/utils/timeframe/index.ts
```diff
@@ -4,7 +4,7 @@ import {
   isDateSeries,
   isDateSpanSeries,
   TimestampedValue,
-} from '@corona-dashboard/common';
+} from '../../';
```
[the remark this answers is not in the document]
VWSCoronaDashboard8: I didn't make it. It opened a world of hurt. Somehow the jest runner wasn't able to resole the path to ~/data-sorting.

PR #3078, opened 2021-06-29
Add schema for international tested overall

VWSCoronaDashboard8:
## Summary

These are the first schema definitions for international data. Currently only for `tested_overall.infected_per_100k` aka confirmed cases.

- Time series data is split into a single file per country, similar to what we do already with VR and GM
- Country files have a `IN_` prefix because everything ends up in one folder, and otherwise we have no way to pick out just the country specific files.
- Choropleth data is defined in IN_COLLECTION.json similar to what we do already with VR and GM

## TODO after this PR:

- Define a fixed list of countries and their codes, to enforce them in filenames via enum.

### DIFF
--- a/packages/app/schema/in_collection/tested_overall.json
+++ b/packages/app/schema/in_collection/tested_overall.json
@@ -5,7 +5,7 @@
   "additionalProperties": false,
   "required": [
     "country_code",
-    "infected_average",
+    "infected",
     "infected_per_100k_average",
     "date_start_unix",
     "date_end_unix",
@@ -16,7 +16,7 @@
       "type": "string",
       "pattern": "^[A-Z]+$"
     },
-    "infected_average": {
+    "infected": {
       "type": "integer"
     },
     "infected_per_100k_average": {
--- a/packages/app/schema/in_collection/tested_overall.json
+++ b/packages/app/schema/in_collection/tested_overall.json
@@ -5,7 +5,7 @@
   "additionalProperties": false,
   "required": [
     "country_code",
-    "infected",
+    "infected_average",
     "infected_per_100k_average",
     "date_start_unix",
     "date_end_unix",
@@ -16,7 +16,7 @@
       "type": "string",
       "pattern": "^[A-Z]+$"
     },
-    "infected": {
+    "infected_average": {
       "type": "integer"
     },
     "infected_per_100k_average": {
--- a/packages/app/schema/in_collection/tested_overall.json
+++ b/packages/app/schema/in_collection/tested_overall.json
@@ -1,27 +1,27 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
   "type": "object",
-  "title": "regions_tested_overall",
+  "title": "in_tested_overall",
   "additionalProperties": false,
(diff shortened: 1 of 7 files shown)


packages/app/schema/in_collection/tested_overall.json
```diff
@@ -0,0 +1,31 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "title": "regions_tested_overall",
```
VWSCoronaDashboard1, line 4:
```suggestion
  "title": "countries_tested_overall",
```

packages/cli/src/schema/schema-info.ts
```diff
@@ -26,6 +26,18 @@ export function getSchemaInfo(
   const fileList = fs.readdirSync(jsonDirectory);
 
   return {
+    in: {
+      files: getFileNames(fileList, /^IN_[A-Z]+.json$/),
+      basePath: jsonDirectory,
+      customValidations: [
+        createChoroplethValidation(
+          path.join(defaultJsonDirectory, 'IN_COLLECTION.json'),
+          'vrcode'
```
VWSCoronaDashboard1, line 35:
```suggestion
          'country_code'
```

VWSCoronaDashboard1 (approved):

packages/app/schema/in_collection/tested_overall.json
VWSCoronaDashboard8: Good catch. It should be `in_tested_overall` then.

VWSCoronaDashboard1: ah yes, sorry ;)

VWSCoronaDashboard14 (approved):

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard1 (changes requested):
You forgot to run `yarn generate-typescript` ;)

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard1 (approved):

packages/app/schema/in_collection/tested_overall.json
```diff
@@ -0,0 +1,35 @@
...
+    "infected_average",
+    "infected_per_100k_average",
+    "date_start_unix",
+    "date_end_unix",
+    "date_of_insertion_unix"
+  ],
+  "properties": {
+    "country_code": {
+      "type": "string",
+      "pattern": "^[A-Z]+$"
```
VWSCoronaDashboard1, line 17: I think its been established that we'll receive ISO A3 codes, so perhaps we can put a check on the country_code for a min and max char count of 3?

VWSCoronaDashboard1 (approved):

packages/app/schema/in_collection/tested_overall.json
VWSCoronaDashboard8: I think I would prefer to use an enum here with the exact selection / whitelist of country codes.

VWSCoronaDashboard1 (approved):

PR #3081, opened 2021-06-30
Reduce choropleth path length

VWSCoronaDashboard10:
Cut off an unnecessary level of detail of the paths strings. A path value of `M177.08821511867188` becomes `M177`.
This change reduces the size of an entire (single language) export with ~800MB. Some pages are 60% down in size. The homepage is 50% down in size. 💥 ✨

### DIFF
--- a/packages/app/src/components/choropleth/choropleth.tsx
+++ b/packages/app/src/components/choropleth/choropleth.tsx
@@ -319,7 +319,20 @@ function MercatorGroup<G extends Geometry, P>(props: MercatorGroupProps<G, P>) {
       {({ features }) => (
         <g>
           {features.map(
-            ({ feature, path, index }) => path && render(feature, path, index)
+            ({ feature, path, index }) =>
+              path &&
+              render(
+                feature,
+                /**
+                 * Cut off an unnecessary level of detail of the paths strings.
+                 * A path value of `M177.08821511867188` becomes `M177`.
+                 */
+                path.replace(
+                  /\d+\.\d+/g,
+                  (x) => Math.round(parseFloat(x)) + ''
+                ),
+                index
+              )
           )}
         </g>
       )}


packages/app/src/components/choropleth/choropleth.tsx
```diff
@@ -319,7 +319,20 @@ function MercatorGroup<G extends Geometry, P>(props: MercatorGroupProps<G, P>) {
...
+              render(
+                feature,
+                /**
+                 * Cut off an unnecessary level of detail of the paths strings.
+                 * A path value of `M177.08821511867188` becomes `M177`.
+                 */
+                path.replace(
+                  /\d+\.\d+/g,
+                  (x) => Math.round(parseFloat(x)) + ''
+                ),
```
VWSCoronaDashboard1, line 333: This works of course. But might it perhaps be a better option to do this transform in the `/components/choropleth/topology` file?
Right after the transform from topojson to geojson, I mean.
Roughly speaking like this:
```
export const municipalGeo = transformGeoJSONPaths(topojson.feature(
  topology,
  topology.objects.municipalities
)) as MunicipalGeoJSON;
```

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard14 (approved):
👏

packages/app/src/components/choropleth/choropleth.tsx
VWSCoronaDashboard8: If we never use this level of detail, I also think it makes more sense to cut it from the source.

VWSCoronaDashboard8 (approved):

packages/app/src/components/choropleth/choropleth.tsx
VWSCoronaDashboard10: We might be able to improve the size of the topojson yes, but I don't think that will have the same impact as the change of this PR. I think that with an even more optimized topojson file the outputted paths could still be rather detailed, it's just the output of some d3-geo function?

I'll just push this change to develop, if anyone else knows how to improve it even more: feel free of course.

VWSCoronaDashboard1: I wasn't suggesting we optimize the topojson. I suggested we let topo-client here convert the topojson to geojson and then immediately simplify the geojson. All you need to do is loop through the feature.coordinates find the coords and do something like `+value.toFixed(2)` to them.
But I'm fine with emerging this for now.

PR #3092, opened 2021-07-01
Enable source maps in production

VWSCoronaDashboard8:


VWSCoronaDashboard1 (approved):

packages/app/next.config.js
```diff
@@ -12,6 +12,12 @@ const nextConfig = {
    */
   reactStrictMode: true,
 
+  /**
+   * Enable source maps in production, because we want people to report readable
+   * stack traces from the error boundaries feature.
+   */
+  productionBrowserSourceMaps: true,
```
VWSCoronaDashboard10, line 19: does this out of the box write pretty stack traces to the error dumps?

VWSCoronaDashboard1: AFAIK, if sourcemaps are configured they automatically end up in your stacktrace.

"Source maps are natively supported by both Google Chrome and Firefox, as well as IE11. This means that any minified JavaScript files with a source map comment at the bottom will automatically link to it’s sources, so long as the mapping and source files are in the correct locations."

VWSCoronaDashboard8: Yes that's what I'm hoping for at least :)

VWSCoronaDashboard10: wonderful ✨

PR #3104, opened 2021-07-05
Build tweaks/optimizations

VWSCoronaDashboard1:
## Summary

`@coronadashboard/common` was pulling in all of `date-fns`. Solved that by using named imports. Since that lib is compiled using Typescript we can't use any fancy babel re-writing plugins. Added the `date-fns` babel plugin to the app though, so any usage there will be re-written correctly.

Added `DuplicatePackageCheckerPlugin` to `next.config.js` and resolved a bunch of duplicate packages using Webpack aliases.

Upgraded all of the `visx` packages (they were also responsible for a few duplicates because they were nailed down to a specific version).

Upgraded d3-array and added it to the transpile list.

## Motivation

Trying to fight bundles sizes?.

### Governance

- [ ] Documentation is added


VWSCoronaDashboard14 (approved):

packages/app/next.config.js
```diff
@@ -48,10 +53,36 @@ const nextConfig = {
...
+      ['d3-color', '../../node_modules/d3-interpolate/node_modules/d3-color'],
+      ['d3-geo', '../../node_modules/d3-geo'],
+      ['d3-interpolate', '../../node_modules/d3-interpolate'],
+      ['balanced-match', '../../node_modules/balanced-match'],
+      ['date-fns', 'node_modules/date-fns'],
+    ];
+
+    duplicatePackageResolves.forEach(([packageName, resolvedPath]) => {
+      config.resolve.alias[packageName] = path.resolve(__dirname, resolvedPath);
+    });
```
VWSCoronaDashboard10, line 76: this looks a bit like a hack, can't we just fix our dependency versioning? Or use yarn resolutions which exists for these kind of situations?

VWSCoronaDashboard1: This is not a hack, this is one of the recommended ways to solve duplicates:
https://www.npmjs.com/package/duplicate-package-checker-webpack-plugin#resolving-duplicate-packages-in-your-bundle

VWSCoronaDashboard1: As for 'fixing' our dependency versioning, we don't always have 100% control over this. In most cases, these are sub dependencies of dependencies.

VWSCoronaDashboard10: Ah and I see yarn resolutions is also one of the recommended ways. Don't you agree that solution is a bit more elegant?

VWSCoronaDashboard1: I don't have a clear preference, I think both methods have their pros and cons. But if it helps you sleep at night, I don't have a problem with using yarn resolutions. ;)

VWSCoronaDashboard1: I'll make the changes

VWSCoronaDashboard1: Alas, yarn resolutions don't seem to fix the situation, so I'll stick with the 'ugly' solution for now...

VWSCoronaDashboard8 (approved):

PR #3109, opened 2021-07-05
Fetch data element configurations from CMS

VWSCoronaDashboard8:
## Summary

This connects the new TimeSeries timeline events with the CMS "elements" definitions. 

Each page defines in what elements it requires based on scope and metricNames / subjects. Then a getter function grabs the configuration for each specific element and possibly does some data conversion.

In the case of the chart, there are only the timeline events to grab, so the getter function is specific for this. It also converts the timestamps from the CMS readable format to our internal unix timestamps. This API is not in any way final. As we would add more types of configurations and elements I think we probably need to look at how we want to handle this data in general, but this is a first step.

The first time series chart on `landelijk/positief-geteste-mensen` is showing events that are fetched from the CMS. 

If events can't be found based on the chart's request, the getter function will return undefined. I think we do not need to fail if events are missing, but this is up for discussion AFAIC.

I have renamed the metricScope in the feature config and used a different name because metricScope in data is slightly different from metric scope as it is used for the CMS elements. `vr_collection` in the CMS is still scope `vr` but would be element type `choropleth` for example.


packages/app/src/queries/create-page-elements-query.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export function createElementsQuery(
+  scope: MetricScope,
+  metricNames: string[],
+  locale: string
+) {
+  const query = `// groq
+    {
+      'timeSeries': *[
+        ${NO_DRAFTS}
```
VWSCoronaDashboard4, line 18: This no drafts part should not be necessary; in the configuration of the sanity client those are toggled off.

VWSCoronaDashboard4 (approved):

VWSCoronaDashboard1 (approved):

packages/app/src/pages/landelijk/positief-geteste-mensen.tsx
```diff
@@ -114,7 +129,7 @@ const PositivelyTestedPeople = (props: StaticProps<typeof getStaticProps>) => {
             }}
             reference={text.reference}
           />
-          {content.main?.articles && (
+          {content.main.articles && (
```
VWSCoronaDashboard10, line 132: did you validate the main object always exists, also with an empty articles result?

packages/app/src/queries/create-page-elements-query.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  return timelineEvents
+    ? timelineEvents.map(
+        (x) =>
+          ({
+            title: x.title,
+            description: x.description,
+            start: new Date(x.date).getTime() / 1000,
+            end: x.dateEnd ? new Date(x.dateEnd).getTime() / 1000 : undefined,
+          } as TimelineEventConfig)
```
VWSCoronaDashboard10, line 96: nit;
```suggestion
    ? timelineEvents.map<TimelineEventConfig>(
        (x) =>
          ({
            title: x.title,
            description: x.description,
            start: new Date(x.date).getTime() / 1000,
            end: x.dateEnd ? new Date(x.dateEnd).getTime() / 1000 : undefined,
          })
```

VWSCoronaDashboard10 (approved):

packages/app/src/pages/landelijk/positief-geteste-mensen.tsx
VWSCoronaDashboard8: Not really verified, but if I understand groq query structures correctly this property will always be there yes. I put a separate task in tech-dept to look at this https://trello.com/c/37aLC28I/2129-fix-typing-for-creategetcontent-on-articles. 

I'll re-add the ? for now, because I need to apply this type of code on more charts and I don't have time to verify this.

packages/app/src/queries/create-page-elements-query.ts
VWSCoronaDashboard8: Can't seem to find this anywhere in lib/sanity.ts. Could you verify / elaborate on this?

VWSCoronaDashboard10: Unauthenticated requests won't include drafts, and the lib/sanity.ts is an unauthenticated client.

PR #3110, opened 2021-07-06
Add country select to Internationaal page

VWSCoronaDashboard4:
## Summary

Add country select to Internationaal page


packages/app/src/domain/international/select-countries/context.tsx
```diff
@@ -0,0 +1,200 @@
...
+     */
+    isEnabled: showResults,
+  });
+
+  useEffect(() => {
+    /**
+     * On input-change we'll reset the focus index to 0. It's possible that
+     * there is a stronger hit among the VR hits (2nd column). If so, we won't
+     * reset the index to 0, instead it will be set to the index of that hit.
+     */
```
VWSCoronaDashboard1, line 136: This comment mentions a VR?

packages/app/src/domain/international/select-countries/country-code.ts
```diff
@@ -0,0 +1,68 @@
...
+  'nor',
+  'pol',
+  'prt',
+  'rou',
+  'svk',
+  'svn',
+  'swe',
+];
+
+export type CountryCode =
```
VWSCoronaDashboard1, line 36:
```suggestion
export type CountryCode = typeof countryCodes[number];
```
(And type countryCodes to `countryCodes = [....] as const`)

VWSCoronaDashboard1 (approved):

packages/app/src/pages/internationaal/positief-geteste-mensen.tsx
```diff
@@ -137,3 +212,77 @@ export default function PositiefGetesteMensenPage(
...
+      );
+
+      if (!isDefined(objectMatch)) {
+        objectMatch = {
+          date_start_unix: value.date_start_unix,
+          date_end_unix: value.date_end_unix,
+        } as CompiledCountriesValue;
+
+        flattenedData.push(objectMatch);
+      }
```
VWSCoronaDashboard8, line 242: If we don't allow BE to leave out values (so report null when data is not available) this could in theory be simplified a lot. Then in a sorted dataset all samples would line up. In the validation state we could verify that all arrays contain the same number of items.

packages/app/src/pages/internationaal/positief-geteste-mensen.tsx
```diff
@@ -137,3 +212,77 @@ export default function PositiefGetesteMensenPage(
...
+
+function selectedCountriesToSeriesConfig(
+  selectedCountries: CountryCode[],
+  countryNames: Record<CountryCode, string>,
+  selectionColors: string[]
+): LineSeriesDefinition<CompiledCountriesValue>[] {
+  return [
+    {
+      type: 'line' as const,
+      metricProperty: 'nld' as CountryCode,
```
VWSCoronaDashboard8, line 263: nit; I don't think these two "as" typecasts are needed since you are already casting the whole object

packages/app/src/pages/internationaal/positief-geteste-mensen.tsx
```diff
@@ -137,3 +212,77 @@ export default function PositiefGetesteMensenPage(
...
+  countryCodes: CountryCode[],
+  countryNames: Record<CountryCode, string>,
+  data: CompiledCountriesValue[]
+): Country[] {
+  const lastValues = last(data);
+  return countryCodes.map((countryCode) => ({
+    code: countryCode,
+    name: countryNames[countryCode],
+    lastValue: lastValues?.[countryCode] ?? 0,
+  }));
```
VWSCoronaDashboard8, line 287: I think an assert is more appropriate here than a fallback to 0

```suggestion
  const lastValues = last(data);
  assert(lastValues, 'Data appears to be empty')
  
  return countryCodes.map((countryCode) => ({
    code: countryCode,
    name: countryNames[countryCode],
    lastValue: lastValues[countryCode],
  }));
```

packages/app/src/static-props/get-data.ts
```diff
@@ -295,3 +296,21 @@ export function createGetChoroplethData<T1, T2, T3>(settings?: {
     };
   };
 }
+
+export function getInData(countryCodes: string[]) {
+  if (!countryCodes) {
+    throw Error('No valid countryCodes found in context');
+  }
```
VWSCoronaDashboard8, line 308: nit; an assert reads a little cleaner I think and does effectively the same.

But countryCodes is not an optional parameter so I wonder why you even have this check. Should it check for an empty array instead maybe? Or if the check is needed at runtime, we can tell TS that this is an optional argument.

packages/app/src/static-props/get-data.ts
```diff
@@ -297,13 +302,13 @@ export function createGetChoroplethData<T1, T2, T3>(settings?: {
   };
 }
 
-export function getInData(countryCodes: string[]) {
+export function getInData(countryCodes: CountryCode[]) {
   if (!countryCodes) {
     throw Error('No valid countryCodes found in context');
   }
```
VWSCoronaDashboard1: Since we're adding a static list of country codes and the argument is not typed as optional, I think you can skip this check.

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard1 (approved):

PR #3113, opened 2021-07-06
Limit image width in query parameter to sanity

VWSCoronaDashboard8:
## Summary

This checks the `w` query parameter server side to see if the requested width for image size exceeds our maximum. If so it rewrites the url to use the maximum.

We could also choose to return a status 400 instead. I don't have a strong preference.

There is also some semi-random cleanup related stuff in here.


packages/app/next-server.js
```diff
@@ -42,14 +52,24 @@ const SANITY_PATH = `${process.env.NEXT_PUBLIC_SANITY_PROJECT_ID}/${process.env.
...
+          /^\/cms-(images|files)/,
+          `/$1/${SANITY_PATH}`
+        );
+
+        if (req.query.w > MAX_IMAGE_WIDTH) {
+          const newQuery = { ...req.query, w: MAX_IMAGE_WIDTH };
+          return `${newPath.split('?')[0]}?${querystring.stringify(newQuery)}`;
+        } else {
+          return newPath;
+        }
```
klaasman, line 72: Isn't it still possible for script kiddies to request a very large image width `?h={99999}`?
Maybe we should strip all query parameters and only allow one of the widths which are part of the `imageResizeTargets`.

VWSCoronaDashboard8: Good point, I didn't think about the h parameter because we're not using it. I'll remove that one all together.

I don't think we need to strip other parameters. The others can't hurt AFAICT and also I don't see much of an advantage in limiting the request to defined sizes.

VWSCoronaDashboard10: Well, in theory it's still possible to fire 2048 requests for a single image:
- `?w=1`
- `?w=2`
- `?w=3`
- `?w=4`
- ...
- `?w=2048`

VWSCoronaDashboard8: Yes, so 2k extra image requests, which are then cached by the CDN. I don't see this as a big danger. But of course we could write a function that snaps the w to the nearest match.

We are currently not sticking to our own defined list of image sizes. There are some part in the code that request 1200 and smaller images of seemingly arbitrary size.

VWSCoronaDashboard1 (approved):

packages/app/next-server.js
VWSCoronaDashboard4: The biggest culprit are large sizes that take some time before they time out. If we can 404 those early, it saves open connections.

packages/app/next-server.js
```diff
@@ -42,14 +52,39 @@ const SANITY_PATH = `${process.env.NEXT_PUBLIC_SANITY_PROJECT_ID}/${process.env.
...
+          const limitedQuery = omit(
+            {
+              ...req.query,
+              w: shouldLimitWidth ? MAX_IMAGE_WIDTH : req.query.w,
+            },
+            ['h']
+          );
+
+          return `${newPath.split('?')[0]}?${querystring.stringify(
+            limitedQuery
```
VWSCoronaDashboard4, line 83: I don't think we should rewrite the URL. We should give a 4xx error instead (which is in line with what is happening, we don't have images that large, so eg Not Found).

Doing a soft redirect (without using a 3xx response) gives odd behaviour on the CDN side.

VWSCoronaDashboard8: I don't see how this would create odd behavior on the CDN, since we are rewriting the url already as part of the proxy. 

The correct response otherwise would be 400 I think (bad request).

packages/app/next-server.js
```diff
@@ -159,3 +139,12 @@ const SANITY_PATH = `${process.env.NEXT_PUBLIC_SANITY_PROJECT_ID}/${process.env.
     res.removeHeader('X-Powered-By');
   }
 })();
+
+function filterImageRequests(pathname, req) {
+  /**
+   * Only allow images using our maximum used dimensions.
+   * Disallow `h` parameter.
+   */
+  const allowRequest = req.query.w <= MAX_IMAGE_SIZE && !req.query.h;
```
VWSCoronaDashboard8, line 148: Now you're also blocking queries that do not have a `w`. I'm not sure that's intended?

packages/app/next-server.js
```diff
@@ -14,9 +14,9 @@ const { imageResizeTargets, assert } = require('@corona-dashboard/common');
 const { last, omit } = require('lodash');
 const querystring = require('querystring');
 
-const MAX_IMAGE_WIDTH = last(imageResizeTargets);
+const MAX_IMAGE_SIZE = last(imageResizeTargets);
```
VWSCoronaDashboard8, line 17: nit; This parameter is used to define the maximum width. In theory the resulting image can be a portrait and therefor still be higher than this. So it seems to me that width is a clearer name.

packages/app/next-server.js
```diff
@@ -124,3 +139,12 @@ const SANITY_PATH = `${process.env.NEXT_PUBLIC_SANITY_PROJECT_ID}/${process.env.
...
   }
 })();
+
+function filterImageRequests(pathname, req) {
+  /**
+   * Only allow images using our maximum used dimensions.
+   * Disallow `h` parameter.
+   */
+  const allowRequest = req.query.w <= MAX_IMAGE_SIZE && !req.query.h;
+  return allowRequest;
```
VWSCoronaDashboard8, line 149: What is the response going to be when you return a boolean? 

Maybe add a `res.status(400).end()` to be explicit?

VWSCoronaDashboard4: It gives a not found; not sure if there is access to the `res` here...

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard10 (approved):

PR #3115, opened 2021-07-06
Make Errorboundary collect component info 

VWSCoronaDashboard1:
## Summary

This adds a hook called `useComponentPropsReport` which is used by the `ErrorBoundary`.

The hook searches the children of the `ErrorBoundary` for components that are registered with it.
Once it finds one of them it extracts the props that are configured for the specified component.
For example, this is registered for the `TimeSeriesChart`:
`TimeSeriesChart: ['values', 'dataOptions', 'seriesConfig', 'timeframe']`

So, in the case of the `TimeSeriesChart` all of the data and series config is also added to the error report.
This makes the report a lot longer of course, but I figured, the more information, the better.

The values of these props are then serialized along with the normal report that the `ErrorBoundary`
generates from the Error, stack trace, and environment.

Additionally, its is also possible to set some extra error report prop on the ErrorBoundary itself. This prop is
called `extraPropsReport` and is a plain object that a dev can just stick any value on that they deem
interesting enough to be included in the error report.

I had to jump through some hoops to pass the additional props report to the `ErrorFallback`, since
this is passed on to the `react-error-boundary` as class name only, I created a context to pass it on
to the `ErrorFallback` instance.

## Motivation

A more detailed error report is nice.

### Governance

- [ ] Documentation is added


packages/app/src/components/error-boundary/index.tsx
```diff
@@ -77,7 +110,16 @@ function formatErrorReport(error: Error) {
...
+  ];
+
+  if (isDefined(props)) {
+    report.push('component props:');
+    for (const prop in props) {
+      report.push(`${prop}: ${JSON.stringify(props[prop], null, '\t')}`);
+    }
+  }
+
+  return report.join('\n');
```
VWSCoronaDashboard4, line 122: This join could be replaced by a JSON.stringify, then you can remove the current stringify from the loop.

VWSCoronaDashboard1: Actually, that way the entire report would be serialized as JSON, which looks a tad crappy...

packages/app/src/components/error-boundary/use-component-props-report.ts
```diff
@@ -0,0 +1,161 @@
+import { isArray, isObject } from 'lodash';
+import {
+  MutableRefObject,
+  ReactElement,
+  ReactNode,
+  useCallback,
+  useRef,
+} from 'react';
+import { isDefined } from 'ts-is-present';
+import { EuropeChoroplethProps } from '../choropleth/europe-choropleth';
```
VWSCoronaDashboard8, line 10: nit; Not part of this PR, but if we have two letter prefixes for most choropleths, then europe chould use `Eu` :)

packages/app/src/components/error-boundary/use-component-props-report.ts
```diff
@@ -0,0 +1,161 @@
...
+);
+chartPropsMap.set(
+  'EuropeChoropleth',
+  keyCheck<EuropeChoroplethProps<any>>(['data', 'metricProperty'])
+);
+chartPropsMap.set(
+  'StackedChart',
+  keyCheck<StackedChartProps<any>>(['values', 'config', 'timeframe'])
+);
+chartPropsMap.set(
```
VWSCoronaDashboard8, line 59: This error boundary component props thing seems generic enough to not tie it to "charts" specifically. So maybe just name this `componentPropsMap`? or `selectedComponentProps` the fact that it's a map is not so interesting maybe.

packages/app/src/components/error-boundary/use-component-props-report.ts
```diff
@@ -0,0 +1,161 @@
...
+  keyCheck<GmChoroplethProps<any, any>>([
+    'data',
+    'metricName',
+    'metricProperty',
+  ])
+);
+chartPropsMap.set(
+  'EuropeChoropleth',
+  keyCheck<EuropeChoroplethProps<any>>(['data', 'metricProperty'])
+);
```
VWSCoronaDashboard8, line 54: You could make all of this more DRY if you defined a function with an interface like:

```
const someResult = selectComponentProps([
 'MunicipalityChoropleth', [
    'data',
    'metricName',
    'metricProperty',
  ]],
  ['EuropeChoropleth', ['data', 'metricProperty']], ...etc)
```

You then only need to internally in the function map `'EuropeChoropleth'` to type `EuropeChoroplethProps`

VWSCoronaDashboard8 (commented):
I didn't quite finish reviewing yet, but I'll submit what I have so far...

packages/app/src/components/error-boundary/use-component-props-report.ts
VWSCoronaDashboard8: For typing it might be easier to pass in the actual type, so EuropeChoroplethProps instead of 'EuropeChoropleth', and then look up the string from the type.

VWSCoronaDashboard8: Or wait, even better. Pass in the actual component, extract the props type from that (there's a React type to do that I believe I've seen @VWSCoronaDashboard10 use it before), and then I think you wouldn't need the string at all maybe?

packages/app/src/domain/hospital/admissions-per-age-group/admissions-per-age-group.tsx
```diff
@@ -99,36 +100,38 @@ export function AdmissionsPerAgeGroup({
...
-              label: text.line_chart_legend_inaccurate_label,
-              shortLabel: text.tooltip_labels.inaccurate,
-            },
-          ],
-        }}
-      />
-      <Legend items={staticLegendItems} />
-    </>
+    <ErrorBoundary>
+      <>
```
VWSCoronaDashboard8, line 104: nit; Maybe you can make ErrorBoundary handle children like a normal React component, where multiple children are allowed. This wrapper fragment should technically not be needed.

VWSCoronaDashboard8 (approved):

PR #3122, opened 2021-07-07
Fix sanity locale-string validation

VWSCoronaDashboard10:


packages/cms/src/schemas/locale/locale-string.ts
```diff
@@ -10,3 +10,11 @@ export const localeString = {
     type: 'string',
   })),
 };
+
+export function localeStringValidation(getRules: (fieldRule: any) => any) {
+  return (Rule: any) =>
+    Rule.fields({
+      nl: (FieldRule: any) => getRules(FieldRule).type('string'),
+      en: (FieldRule: any) => getRules(FieldRule).type('string'),
+    });
```
VWSCoronaDashboard8, line 19: Looks like you wrote argument/instance names with capital letters as if they were types.

I would suggest to write Rule and FieldRule with small letters here.

VWSCoronaDashboard8: Nice solution though 👍

VWSCoronaDashboard8 (approved):

packages/cms/src/schemas/locale/locale-string.ts
VWSCoronaDashboard10: It's based on Sanity documentation. I guess we could see it as an initialisation of a class?

image

VWSCoronaDashboard10: Ah fuck that, I'll go for lowercase.

VWSCoronaDashboard8: Yes, it appears that Rule is a class and Rule.required() etc are calls to static class methods. Then it makes sense.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static

VWSCoronaDashboard4 (approved):

PR #3126, opened 2021-07-07
Connect other charts to timeline config

VWSCoronaDashboard8:
## Summary

This connects all charts to the CMS for which communication requested to add timeline events.

I had to implement a workaround. The real solution I think is to make the articles query not include "articles" as part of it's response, but leave that name for where the query is configured (now named fix_this).

Then in places where now "ggd" and "main" are used, it can be "ggd_articles" and "main_articles"

### DIFF
--- a/packages/app/src/pages/veiligheidsregio/[code]/positief-geteste-mensen.tsx
+++ b/packages/app/src/pages/veiligheidsregio/[code]/positief-geteste-mensen.tsx
@@ -387,7 +387,13 @@ const PositivelyTestedPeople = (props: StaticProps<typeof getStaticProps>) => {
                         .infected_percentage,
                   },
                 ]}
-                dataOptions={{ isPercentage: true }}
+                dataOptions={{
+                  isPercentage: true,
+                  timelineEvents: getTimelineEvents(
+                    content.elements.timeSeries,
+                    'tested_ggd'
+                  ),
+                }}
               />
             )}
           </ChartTile>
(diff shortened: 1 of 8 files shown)


VWSCoronaDashboard4 (approved):

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard1 (approved):

packages/app/src/pages/veiligheidsregio/[code]/positief-geteste-mensen.tsx
```diff
@@ -62,6 +70,11 @@ export const getStaticProps = createGetStaticProps(
         locale,
         'ggdArticles'
       )},
+      "elements": ${createElementsQuery(
+        'vr',
+        ['tested_overall', 'tested_ggd'],
```
Comment (VWSCoronaDashboard10) [inline packages/app/src/pages/veiligheidsregio/[code]/positief-geteste-mensen.tsx:75]: shouldn't this be only 'tested_overall'?

VWSCoronaDashboard10 (approved):

packages/app/src/pages/veiligheidsregio/[code]/positief-geteste-mensen.tsx
Comment (VWSCoronaDashboard8) [reply packages/app/src/pages/veiligheidsregio/[code]/positief-geteste-mensen.tsx:75]: Not for VR but it does for GM. Nothing bad happens when you try to fetch elements that don't exist though, but I'll change the GM page

PR #3132, opened 2021-07-07
Mutate Lokalize via JSON edits

VWSCoronaDashboard10:
## Summary

This allows us to do any type of text mutation by editing our local `locale/nl_export.json` file. The file now contains document ids as part of the keys. This document is compared to a reference file, created during export, located in `.lokalize-reference/nl_export.json`.

Document ids are stripped from the JSON when it is loaded into the app, and a type interface is created from the JSON at `site-text.d.ts`. This is what the application sees.

Any part of the JSON can be renamed and moved, and should be detected as such. For these types of mutations are first simulated locally in your branch, and get finalized when the branch is merged. Finalizing a move mutation means mutating the key + subject properties of the text document. 

Because we are mutating the original documents, the edit history and drafts of those documents will remain. 

This new approach also replaces the previous lokalize:add/delete commands.


VWSCoronaDashboard8 (approved):

packages/cms/src/lokalize/apply-json-edits.ts
```diff
@@ -0,0 +1,147 @@
...
+    ...mutations.move.map(
+      (mutation) =>
+        ({
+          title: `move:\t\t${mutation.key} → ${mutation.moveTo}`,
+          value: { type: 'move', mutation },
+        } as const)
+    ),
+  ];
+
+  if (!choices.length) return console.log('\nNo changes to sync\n');
```
VWSCoronaDashboard1, line 51: Don't you need something like `process.exit(0)` here?

packages/cms/src/lokalize/apply-json-edits.ts
```diff
@@ -0,0 +1,147 @@
...
+
+  const response = (await prompts({
+    type: 'multiselect',
+    name: 'keys',
+    message: 'Select the mutations to sync to Sanity:',
+    choices,
+    onState,
+  })) as { keys: typeof choices[number]['value'][] };
+
+  if (!response.keys.length) return console.log('\nNo mutations selected\n');
```
VWSCoronaDashboard1, line 61: Same here. no `process.exit(0)`?

packages/cms/src/lokalize/apply-json-edits.ts
```diff
@@ -0,0 +1,147 @@
...
+        documentId: document._id,
+      });
+    }
+
+    if (choice.type === 'move') {
+      const { key, documentId, moveTo } = choice.mutation;
+
+      await appendTextMutation({ action: 'move', key, documentId, moveTo });
+    }
+  }
```
VWSCoronaDashboard1, line 96: Is it necessary to run these calls sequentially? Or can they be gathered in a `Promise.all` to speed things up a nodge?

packages/cms/src/lokalize/logic/export.ts
```diff
@@ -63,4 +77,64 @@ export async function exportLokalizeTexts(
...
+  const json = prettier.format(JSON.stringify(data), { parser: 'json' });
+  return new Promise<void>((resolve, reject) =>
+    fs.writeFile(path, json, { encoding: 'utf8' }, (err) =>
+      err ? reject(err) : resolve()
+    )
+  );
+}
+
+export async function generateTypes() {
+  const data = flatten(
```
VWSCoronaDashboard1, line 104: nit: I think flatten is generic: `flatten<any, Record<string, string>>`

packages/cms/src/lokalize/logic/export.ts
```diff
@@ -63,4 +77,64 @@ export async function exportLokalizeTexts(
...
+
+  const body = prettier.format(
+    `
+      /**
+       * This file was auto-generated from the lokalize export script.
+       */
+      export interface SiteText ${textsTypeString}
+    `,
+    { parser: 'typescript' }
+  );
```
VWSCoronaDashboard1, line 129: NICE!

packages/cms/src/lokalize/logic/files.ts
```diff
@@ -0,0 +1,43 @@
...
+    flatten<unknown, Record<string, string>>(texts),
+    (_value, key) => (key.includes('.') ? key : `__root.${key}`)
+  );
+
+  return flatTexts as Record<string, string>;
+}
+
+export async function readReferenceTexts() {
+  /**
+   * @TODO Make is possible to choose the English file instead.
```
VWSCoronaDashboard1, line 24:
```suggestion
   * @TODO Make it possible to choose the English file instead.
```

packages/cms/src/lokalize/sync-after-feature.ts
```diff
@@ -3,70 +3,102 @@
...
-  const collapsedMutations = collapseTextMutations(mutations);
+  const addDeleteMutations = getCollapsedAddDeleteMutations(mutations);
+  const moveMutations = getCollapsedMoveMutations(mutations);
 
-  const additions = collapsedMutations.filter((x) => x.action === 'add');
+  /**
+   * We synchronize move mutations before deletions. That way we prevent
+   * accidental deletions in possible edge cases where the same key was deleted
+   * and moved. The mutations collapsing logic happens for add/delete and move
+   * separately and is therefor not 100% correct in figuring out what should
```
VWSCoronaDashboard1, line 39:
```suggestion
   * separately and is therefore not 100% correct in figuring out what should
```

VWSCoronaDashboard1 (approved):

packages/cms/src/lokalize/apply-json-edits.ts
VWSCoronaDashboard8: Why? process exit code 0 is the default

VWSCoronaDashboard1: Is it? I was under the impression you need to explicitly call `process.exit`, I'm not aware of there being a default, but I could be mistaken of course.

packages/cms/src/lokalize/apply-json-edits.ts
VWSCoronaDashboard8: I doubt it is worth the added complexity in code. The only fairly slow part of this loop is the sanity client create call. Most speed gain we would get from using a transaction, but that complicates things because you don't want to append the text mutations when the transaction fails for some reason.

This loop will only be annoyingly slow if processing a lot of new keys, but in practice I don't see this happening. Large changes in data will be using `move` mutations, and those are not slowing this loop down considerably I think. The `appendTextMutation` call should be very fast.

packages/cms/src/lokalize/apply-json-edits.ts
VWSCoronaDashboard8: If an executable of any kind ends its process without setting an exit code, the code will default to 0 or "exit without error". So of you exit a node.js process with just a return or by letting it come to its natural end, it will effectively be the same as calling process.exit(0) explicitly

PR #3144, opened 2021-07-08
Use dynamic variants schema's for NL and IN

VWSCoronaDashboard8:
## Summary

This implements the variants schema's using the strategy we discussed, taking a similar approach to sewer_per_installation.

I have diverged a little by placing the `date_of_insertion` at the root-level of the schema, since this values would be the same for every variant and every time series value. If this poses a problem we can move these properties to live in the "value" definition.

I have not defined schemas for `in_collection/variants`, because I do not yet know how we're going to handle that in this case. This is something we need to discuss first, since we have not yet used this type of nested data structure in combination with choropleths.

NOTE: This branch will not pass compilation check until we've processed the data to become compatible with the original variants structure, or implement whatever we come up with to connect the data to our charts.


VWSCoronaDashboard1 (approved):

packages/common/src/data-sorting.ts
```diff
@@ -63,6 +65,31 @@ export function sortTimeSeriesInDataInPlace<T>(
...
+       * exists here might crash the app
+       */
+      console.error('variants.values does not exist');
+      return;
+    }
+
+    nestedSeries.values = nestedSeries.values.map((x) => {
+      x.values = sortTimeSeriesValues(x.values) as
+        | NlVariantsVariantValue[]
+        | InVariantsVariantValue[];
```
VWSCoronaDashboard10, line 88: Don't we need to set dates to midday here too? See lines 54-64.

VWSCoronaDashboard10 (approved):

VWSCoronaDashboard4 (approved):

packages/common/src/data-sorting.ts
VWSCoronaDashboard8: Good catch! I think you're right. I've added it.

VWSCoronaDashboard8:
These changes have been merged as part of work on #3167

PR #3167, opened 2021-07-13
Topic/international variants table

VWSCoronaDashboard1:
## Summary

Serverside data shaping for the variants

### Governance

- [ ] Documentation is added


packages/app/src/domain/variants/variants-table-tile/index.tsx
```diff
@@ -47,19 +45,26 @@ export function VariantsTableTile({
       <Box maxWidth="maxWidthText">
         <Markdown
           content={replaceVariablesInText(text.varianten_tabel.omschrijving, {
-            sample_size: data.sample_size,
+            sample_size: '',
```
VWSCoronaDashboard8, line 48: Why not remove the property is sample_size is not used in the text? Or maybe add a comment to explain why you're using an empty string.

packages/app/src/static-props/variants/get-variant-chart-data.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export function getVariantChartData(
+  nlVariants: NlVariants | undefined,
+  variantTranslations: SiteText['covid_varianten']['varianten']
+) {
+  if (!isDefined(nlVariants) || !isDefined(nlVariants.values)) {
+    return EMPTY_VALUES;
+  }
+
+  const vocVariants = nlVariants.values.filter(
```
VWSCoronaDashboard8, line 31: nit; you can avoid the confusing acronym by calling it `variantsOfConcern`. It's not many more characters and it reads better IMO.

packages/app/src/static-props/variants/get-variant-chart-data.ts
```diff
@@ -0,0 +1,86 @@
...
+    .filter((x) => x.endsWith('_percentage'))
+    .map((x) => x.slice(0, x.indexOf('_')));
+
+  const seriesConfig = variantNames.map<
+    LineSeriesDefinition<VariantChartValue>
+  >((x) => {
+    const color = (colors.data.variants as Record<string, string>)[x];
+    const label = (variantTranslations as Record<string, string>)[x];
+    assert(color, `No color specified for variant called "${x}"`);
+    assert(label, `No label specified for variant called "${x}"`);
```
VWSCoronaDashboard8, line 66: nit; The assertions state that these can be undefined so the used types are not really correct. For these of cases I sometimes define a `Dictionary` type:

```
type Dictionary<T> = Record<string, T | undefined>
```

```suggestion
    const color = (colors.data.variants as Dictionary<string>)[x];
    const label = (variantTranslations as Dictionary<string>)[x];
    assert(color, `No color specified for variant called "${x}"`);
    assert(label, `No label specified for variant called "${x}"`);
```

VWSCoronaDashboard8 (approved):

packages/app/src/domain/variants/variants-table-tile/index.tsx
VWSCoronaDashboard1: Oh, I forgot to add that one, this is a mistake, it should have a valid value. I'll fix it.

VWSCoronaDashboard1: Fixed

VWSCoronaDashboard14 (approved):

PR #3170, opened 2021-07-13
Add variants over time stacked area graph and synced country selection on international page

VWSCoronaDashboard14:
This PR adds the stacked area chart to the international variants page, along with the synced country selection state.
I've tweaked the table and chart tiles a little in order to make the data/null/undefined states, hopefully, a little clearer.

Basically, it means the following:
- data: there is data available, so we show the table and/or chart
- undefined: no country has been selected yet, so we show a message explaining this
- null: A country has been selected, but no data is available for this specific country,  so we show a message explaining this


packages/app/src/components/time-series-chart/logic/hover-state.ts
```diff
@@ -367,17 +367,21 @@ export function useHoverState<T extends TimestampedValue>({
...
 
         switch (config.type) {
           case 'stacked-area':
+          case 'gapped-stacked-area':
             return [
               {
                 seriesValue,
                 x: xScale(xValue),
-                y: yScale(yValueB),
+                y: yValueB ? yScale(yValueB) : 0,
```
VWSCoronaDashboard8, line 384: If I understand correctly we are filtering out values from display in the tooltip in other cases, but here we fall back to 0. Why is there a fallback to zero as opposed to not showing anything?

I don't quite understand why we have to make an exception here.

packages/app/src/domain/international/variants-stacked-area-tile/variants-stacked-area-tile.tsx
```diff
@@ -0,0 +1,208 @@
...
+            timeframe={timeframe}
+            seriesConfig={filteredConfig}
+            disableLegend
+            dataOptions={{
+              isPercentage: true,
+              forcedMaximumValue: 100,
+            }}
+            formatTooltip={(context) => {
+              /**
+               * With the current context the other_percentage is the first in to be displayed.
```
VWSCoronaDashboard8, line 122:
```suggestion
               * With the current config the other_percentage is the first in to be displayed.
```

I think it would be useful to explain why you can't solve this by defining the configuration in the preferred order. I'm guessing other == 100% and can therefor would obscure the rest of the series if it were rendered on top. But if everything is rendered stacked on top of each other, like I would expect this chart to do, then I none of the series should ever overlap each other...

packages/app/src/domain/variants/variants-table-tile/index.tsx
```diff
@@ -51,35 +51,82 @@ export function VariantsTableTile({
...
     date_of_insertion_unix: number;
   } | null;
-  children?: ReactNode | null;
+  children?: ReactNode;
+}) {
+  if (!isPresent(data) || !isPresent(dates)) {
+    return (
+      <FullscreenChartTile>
+        <Heading level={3}>{text.titel}</Heading>
+        <Box maxWidth="maxWidthText">
```
VWSCoronaDashboard8, line 60: This looks like a mistake did you mean to write this?
```suggestion
        <Box maxWidth={maxWidthText}>
```

packages/app/src/domain/variants/variants-table-tile/index.tsx
```diff
@@ -51,35 +51,82 @@ export function VariantsTableTile({
...
-    ? {
-        date: [dates.date_start_unix, dates.date_end_unix],
-        source,
-        obtained: dates.date_of_insertion_unix,
-      }
-    : undefined;
+  const metadata: MetadataProps = {
+    date: [dates.date_start_unix, dates.date_end_unix],
+    source,
+    obtained: dates.date_of_insertion_unix,
```
VWSCoronaDashboard8, line 114: This is not a clear name IMO. If it is a date I would at least make that clear. I'm used to using the at suffix for dates and in this project we use unix, so suggestions: `obtained_at`, `obtained_at_unix` ,`date_obtained_unix`...

packages/app/src/pages/internationaal/varianten.tsx
```diff
@@ -121,21 +164,57 @@ export default function VariantenPage(
...
+        : '';
+
+    const noDataMessageChart =
+      variantChart === undefined
+        ? text.selecteer_een_land_omschrijving
+        : variantChart === null
+        ? text.geen_data_omschrijving
+        : '';
+
+    return [noDataMessageTable, noDataMessageChart] as const;
```
VWSCoronaDashboard8, line 218: This feels a little silly. These are all such light-weight operations that the useMemo is probably creating more overhead than it's saving. AFAIC this part can be used directly in the parent context without any hook. The added abstraction / complexity is not worth it IMO.

packages/app/src/domain/variants/variants-table-tile/index.tsx
VWSCoronaDashboard1: I think this property has been part of the metadata component for a pretty long time, but I can change it to the `_at` suffix

packages/app/src/domain/variants/variants-table-tile/index.tsx
VWSCoronaDashboard1: Nope, this is theme property, not an actual value. (This is being used in a lot of other places as well, by the way, I didn't make this up here)

packages/app/src/components/time-series-chart/logic/hover-state.ts
VWSCoronaDashboard1: It's the same pattern as the `gapped-line`, normally we don't have any `null` values, but in this case we do. So, this situation only occurs in the case of 'gapped-stacked-area'.

packages/app/src/domain/international/variants-stacked-area-tile/variants-stacked-area-tile.tsx
VWSCoronaDashboard1: I'll have to check with @VWSCoronaDashboard14 , he wrote this part. But I think you're right in your assumption.

packages/app/src/domain/variants/variants-table-tile/index.tsx
VWSCoronaDashboard8: Sorry I've become a little sloppy in my reviews then :)

packages/app/src/components/metadata.tsx
```diff
@@ -14,15 +14,15 @@ export interface MetadataProps extends MarginBottomProps {
...
+  obtained_at?: number;
   isTileFooter?: boolean;
   datumsText?: string;
 }
 
 export function Metadata({
   date,
   source,
-  obtained,
+  obtained_at,
```
VWSCoronaDashboard8, line 25: My bad. We're in JS / non-API territory here, so maybe find/replace to `obtainedAt`

packages/app/src/domain/international/variants-stacked-area-tile/variants-stacked-area-tile.tsx
VWSCoronaDashboard1: Ok, checked and changed the comment to better clarify the reasoning behind this.

packages/app/src/domain/variants/variants-table-tile/index.tsx
VWSCoronaDashboard1: Ok, fixed

packages/app/src/pages/internationaal/varianten.tsx
VWSCoronaDashboard1: Fixed as well

packages/app/src/components/metadata.tsx
VWSCoronaDashboard1: Ouch, sorry, I'm not awake yet either haha. I'll make the change.

VWSCoronaDashboard1: Fixed

packages/app/src/domain/variants/variants-table-tile/index.tsx
VWSCoronaDashboard1: Its a very fair comment though, the rename does make things more readable again

VWSCoronaDashboard8 (approved):

packages/app/src/domain/international/variants-stacked-area-tile/variants-stacked-area-tile.tsx
```diff
@@ -0,0 +1,209 @@
...
+            }}
+            formatTooltip={(context) => {
+              /**
+               * In the chart the 'other_percentage' stack is rendered on top,
+               * but in the tooltip it needs to be displayed as the last item.
+               * (These are both design decisions)
+               */
+              const reorderContext = {
+                ...context,
+                config: [...context.config, context.config[0]].slice(1),
```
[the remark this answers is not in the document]
VWSCoronaDashboard1: Well, if I interpret the code, this copies the original array into a new one, then adds the first item from the original array and adds it to the end of the new array. Then the slice removes the first item (the one that was added last) from the start of the new array.
But in theory, you're right about the fragility of this code and I'll make the change you suggest.

PR #3177, opened 2021-07-14
Fix close bug mobile and add flags international combo box

VWSCoronaDashboard14:
Added flags
Removed a check where the screen needed to be bigger when closed.
Fixed a bug where when there were no search results and you press enter, the whole page crashes.

### DIFF
--- a/packages/app/src/domain/international/select-countries/select-country-results.tsx
+++ b/packages/app/src/domain/international/select-countries/select-country-results.tsx
@@ -45,6 +45,7 @@ export function SelectCountriesResults() {
                     src={`/icons/flags/${x.id.toLowerCase()}.svg`}
                     width="17"
                     height="13"
+                    alt=""
                     css={css({
                       mr: 2,
                     })}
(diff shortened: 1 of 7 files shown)


packages/app/src/domain/international/select-countries/select-country-results.tsx
```diff
@@ -40,6 +40,15 @@ export function SelectCountriesResults() {
                   <span css={css({ flex: '0 0 24px' })}>
                     {x.data.isSelected ? <CheckedIcon /> : <UncheckedIcon />}
                   </span>
+                  <img
+                    aria-hidden
+                    src={`/icons/flags/${x.id.toLowerCase()}.svg`}
```
VWSCoronaDashboard4, line 45: add an empty alt attribute: `alt=""`

VWSCoronaDashboard4 (approved):

packages/app/src/domain/international/select-countries/select-country-results.tsx
VWSCoronaDashboard14: Done!

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard10 (approved):

PR #3179, opened 2021-07-14
Remove user added query parameters from HTML

VWSCoronaDashboard4:
## Summary

Remove user added query parameters from HTML.

`asPath` gives all query parameters, also ones not defined/used by us in the route. Without this fix, going to an article URL with appended `?foo=<script>...` it would (HTML-endcoded) end up in the HTML. This fix just completely removes it.


VWSCoronaDashboard8 (approved):

VWSCoronaDashboard14 (approved):

VWSCoronaDashboard10 (approved):
I don't really see which issue this change tries to fix .. isn't there still some sort of identical issue but then with the path name?

PR #3189, opened 2021-07-16
Display values with a consistent amount of decimals

VWSCoronaDashboard10:


VWSCoronaDashboard1 (approved):

packages/app/src/components/time-series-chart/logic/use-metric-property-formatters.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export function useMetricPropertyFormatters<T extends TimestampedValue>(
+  seriesConfig: SeriesConfig<T>,
+  values: T[]
+) {
+  const intl = useIntl();
+
+  return useMemo(() => {
+    function createFormatter(values: number[]) {
+      const numOfDecimals = getMaxNumOfDecimals(values);
```
VWSCoronaDashboard8, line 21: nit; If you want to keep it short I'd say `numDecimals` is clear enough but `numberOfDecimals` has my preference I guess

packages/app/src/utils/get-consistent-number-formatter.tsx
```diff
@@ -0,0 +1,28 @@
+export function getConsistentNumberFormatter(
```
VWSCoronaDashboard8, line 1: This name is a bit funky. Maybe we can replace it when we get rid of other formatting functions? If we're going to use this consistently, I think it should eventually just be called `getNumberFormatter`

packages/app/src/components/time-series-chart/logic/use-metric-property-formatters.ts
```diff
@@ -0,0 +1,51 @@
...
+import { useIntl } from '~/intl';
+import { getMaxNumOfDecimals } from '~/utils/get-consistent-number-formatter';
+import { SeriesConfig } from './series';
+
+export type MetricPropertyFormatters<T extends TimestampedValue> = Record<
+  keyof T,
+  (value: number) => string
+>;
+
+export function useMetricPropertyFormatters<T extends TimestampedValue>(
```
VWSCoronaDashboard8, line 13: What you are formatting is the metric value, not the property name. 

Maybe we can just call it `useMetricFormatters` or otherwise `useMetricValueFormatters`

packages/app/src/components/time-series-chart/logic/use-metric-property-formatters.ts
VWSCoronaDashboard8: Do we really want the formatting to depend on the decimals used in the data? I vaguely remember it was decided that we use one faction digit for every decimal value across the dashboard...

packages/app/src/components/time-series-chart/logic/use-metric-property-formatters.ts
VWSCoronaDashboard8: Never mind, this was a bit of a brain fart. Metric property can refer to both the name and the value. I think the original name is fine.

VWSCoronaDashboard8 (approved):

packages/app/src/components/time-series-chart/logic/use-metric-property-formatters.ts
VWSCoronaDashboard10: We (FE) should not round 2 decimals to 1 decimal, BE is responsible for this.. Also the R-number chart has two decimals.

packages/app/src/utils/get-consistent-number-formatter.tsx
VWSCoronaDashboard10: This function isn't even used anymore, will remote it :-)

PR #3199, opened 2021-07-16
Last changes for international variants page

VWSCoronaDashboard1:
## Summary

- Schema fixes according to latest data insights

### Governance

- [ ] Documentation is added

### DIFF
--- a/packages/app/src/static-props/variants/get-variant-table-data.ts
+++ b/packages/app/src/static-props/variants/get-variant-table-data.ts
@@ -2,12 +2,13 @@ import {
   assert,
   Dictionary,
   DifferenceDecimal,
+  InNamedDifference,
   InVariants,
   NlNamedDifference,
   NlVariants,
 } from '@corona-dashboard/common';
 import { first } from 'lodash';
-import { isDefined } from 'ts-is-present';
+import { isDefined, isPresent } from 'ts-is-present';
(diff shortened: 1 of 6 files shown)


packages/app/src/static-props/variants/get-variant-table-data.ts
```diff
@@ -36,11 +37,13 @@ export function getVariantTableData(
   }
 
   function findDifference(name: string) {
-    const difference = namedDifference.variants__percentage.find(
-      (x) => x.name === name
-    );
-    assert(difference, `No variants__percentage found for variant ${name}`);
-    return difference;
+    if (isPresent(namedDifference.variants__percentage)) {
```
VWSCoronaDashboard4, line 40: nothing returned / asserted when `namedDifference.variants__percentage` is not present?

VWSCoronaDashboard4 (approved):

VWSCoronaDashboard8 (approved):

packages/app/src/static-props/variants/get-variant-table-data.ts
VWSCoronaDashboard1: That means it's undefined, which is a valid value in this case. I could explicitly return `undefined`, but it'll be technically the same.

PR #3201, opened 2021-07-19
Vaccine coverage last bits

VWSCoronaDashboard8:
## Summary

- Fix the schema / validation error
- Add an inline tooltip for 60-64
- Fix date in footer

### DIFF
--- a/packages/app/src/pages/landelijk/vaccinaties.tsx
+++ b/packages/app/src/pages/landelijk/vaccinaties.tsx
@@ -200,8 +200,8 @@ const VaccinationPage = (props: StaticProps<typeof getStaticProps>) => {
               }
               metadata={{
                 datumsText: text.datums,
-                date: vaccine_coverage_per_age_group.last_value
-                  ?.date_of_report_unix,
+                date: vaccine_coverage_per_age_group.values[0]
+                  .date_of_report_unix,
                 source: siteText.vaccinaties.vaccination_coverage.bronnen.rivm,
               }}
             >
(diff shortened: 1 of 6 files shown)


packages/app/src/pages/landelijk/vaccinaties.tsx
```diff
@@ -200,8 +200,8 @@ const VaccinationPage = (props: StaticProps<typeof getStaticProps>) => {
               }
               metadata={{
                 datumsText: text.datums,
-                date: vaccine_coverage_per_age_group.last_value
-                  ?.date_of_report_unix,
+                date: vaccine_coverage_per_age_group.values[0]
+                  .date_of_report_unix,
```
VWSCoronaDashboard10, line 204: you're going from the last value to the first value?

VWSCoronaDashboard1 (approved):

packages/app/src/pages/landelijk/vaccinaties.tsx
VWSCoronaDashboard8: There is only one value. This is not a time series

VWSCoronaDashboard10: Ah right, shouldn't it be called `value` instead?

VWSCoronaDashboard1: I'm guessing the array has a length of one. But yea, probably safer to use last().

VWSCoronaDashboard8: Values are used differently here from time series. The array contains a value for each age group so nothing to do with time. I'm taking the timestamp from the first age group. All have the same timestamps, so it's completely redundant in how it was set up.

VWSCoronaDashboard10: Aaah of course. 🤦

PR #3206, opened 2021-07-19
Add for all pages the new page information block

VWSCoronaDashboard14:
Add the new `PageInformationBlock` and changed it a bit there to make it more flexible.
Remove article-strip
Remove old content header
Remove decorated link
Changed a bit in the Sanity vaccine page schema


VWSCoronaDashboard1 (approved):

packages/app/src/components/page-information-block/components/articles.tsx
```diff
@@ -47,6 +48,8 @@ interface ArticleItemProps {
 }
 
 function ArticleItem({ slug, cover, title }: ArticleItemProps) {
+  const splittedWords = title.split(' ');
```
VWSCoronaDashboard8, line 51: nit; Just `words` is clear enough I think. Technically you split a sentence into words. You didn't split the words.

packages/app/src/components/page-information-block/components/articles.tsx
```diff
@@ -58,7 +61,31 @@ function ArticleItem({ slug, cover, title }: ArticleItemProps) {
...
+                      alignItems: 'center',
+                    })}
+                  >
+                    {word}
+                    <ChevronIcon />
+                  </InlineText>
+                ) : (
+                  `${word} `
+                )}
+              </InlineText>
```
VWSCoronaDashboard8, line 86: This seems to result in some funky markup with a span nested in a span. It looks like you're trying to render the last item differently, but you can do that without the nesting I think. 

If you move the "isLastWord" check to the root of the map, then you can decide to return a word or word+chevron each with their own specific InlineText wrapper, avoiding the span in span. Would that be sufficient?

packages/app/src/components/page-information-block/components/header.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { Heading } from '~/components/typography';
+import { VisuallyHidden } from '~/components/visually-hidden';
+import { asResponsiveArray } from '~/style/utils';
+import { useBreakpoints } from '~/utils/use-breakpoints';
+
+type HeaderProps = {
+  title: string;
+  icon: JSX.Element;
+  category?: string;
+  screenReaderCategory?: string;
```
VWSCoronaDashboard8, line 13: Maybe add a comment to explain why the normal category is not suitable for screen readers. From just reading this API that is not clear

packages/app/src/components/page-information-block/components/header.tsx
```diff
@@ -0,0 +1,87 @@
...
+  });
+
+  return (
+    <Box
+      position={{ md: 'absolute' }}
+      left="0"
+      top="-11px"
+      css={css({
+        width: responsiveIconSize,
+        height: responsiveIconSize,
```
VWSCoronaDashboard8, line 76: nit; I think these can also be just props

packages/app/src/components/page-information-block/components/header.tsx
```diff
@@ -0,0 +1,87 @@
...
+function Icon({ children, size }: { children: React.ReactNode; size: string }) {
+  const responsiveIconSize = asResponsiveArray({
+    _: `calc(${size} - 0.5rem)`,
+    md: size,
+  });
+
+  return (
+    <Box
+      position={{ md: 'absolute' }}
+      left="0"
```
VWSCoronaDashboard8, line 72:
```suggestion
      left={0}
```

packages/app/src/components/page-information-block/components/metadata.tsx
```diff
@@ -215,23 +220,56 @@ interface metadataReferenceProps {
 function MetadataReference({ icon, referenceLink }: metadataReferenceProps) {
   const { siteText } = useIntl();
 
+  const splittedWords =
```
VWSCoronaDashboard8, line 223: Same nitpick here about words

packages/app/src/components/page-information-block/components/metadata.tsx
```diff
@@ -215,23 +220,56 @@ interface metadataReferenceProps {
...
+                    position: 'relative',
+                    display: 'flex',
+                    alignItems: 'center',
+                  })}
+                >
+                  {word}
+                  <ChevronIcon />
+                </InlineText>
+              ) : (
+                `${word} `
```
VWSCoronaDashboard8, line 251: ... and nesting

packages/app/src/components/page-information-block/page-information-block.tsx
```diff
@@ -22,78 +25,119 @@ interface InformationBlockProps {
   metadata?: MetadataProps;
   referenceLink?: string;
   id?: string;
+  category?: string;
+  screenReaderCategory?: string;
+  hasExtraMarginTop?: boolean;
```
VWSCoronaDashboard8, line 30: Given that a component should not use outside margins, this prop seems to be an anti-pattern. Can you maybe control the margin / spacing in the context where the component is used?

PR #3207, opened 2021-07-19
Apply some visual tweaks to vaccine coverage chart

VWSCoronaDashboard8:


packages/app/src/domain/vaccine/components/coverage-row.tsx
```diff
@@ -34,7 +34,12 @@ function MobileCoverageRow(props: RowProps) {
   const children = React.Children.toArray(props.children);
 
   return (
-    <Row width="100%" display="flex" flexDirection="column">
+    <Row
+      width="100%"
+      display="flex"
+      flexDirection="column"
+      onClick={() => setIsOpen(!isOpen)}
+    >
```
VWSCoronaDashboard4, line 42: This won't work with keyboard

VWSCoronaDashboard8: There's also the normal button still, so this is an extra. Would that be problematic?

VWSCoronaDashboard1 (approved):

packages/app/src/domain/vaccine/components/coverage-row.tsx
VWSCoronaDashboard10: That would work I guess but I see there's another issue with that button: the click handler is on the wrong component. Can you fix that?

```diff
- <Button>
+ <Button onClick={() => setIsOpen(!isOpen)}>
    <Chevron
      isOpen={isOpen}
-     onClick={() => setIsOpen(!isOpen)}
      color={colors.blue}
    />
</Button>
```

VWSCoronaDashboard4 (approved):

VWSCoronaDashboard10 (approved):

VWSCoronaDashboard10 (approved):

PR #3213, opened 2021-07-19
Variant data shaping fixes for the timeseries chart

VWSCoronaDashboard1:
## Summary

Once we switched from mock data to real data it became apparent a number of wrong assumptions were made.

The value lists for historical data are of different lengths, so aggregating each timespan needed a new approach,
along with the actual determination of the full date range of all the values.

This took some back and forth with the back-end folks, but this is what the end result actually should be.

## Motivation

Bug fixes.

### Governance

- [ ] Documentation is added

### DIFF
--- a/packages/app/src/static-props/variants/get-international-variant-chart-data.ts
+++ b/packages/app/src/static-props/variants/get-international-variant-chart-data.ts
@@ -72,7 +72,11 @@ function createCompleteDateRange(lists: InVariantsVariant[]) {
     }))
     .filter(
       (x, i, arr) =>
-        arr.findIndex((y) => y.date_end_unix === x.date_end_unix) === i
+        arr.findIndex(
+          (y) =>
+            y.date_end_unix === x.date_end_unix &&
+            y.date_start_unix === x.date_start_unix
+        ) === i
     )
     .sort((a, b) => a.date_end_unix - b.date_end_unix);
 }
--- a/packages/app/src/static-props/variants/get-international-variant-chart-data.ts
+++ b/packages/app/src/static-props/variants/get-international-variant-chart-data.ts
@@ -1,4 +1,5 @@
 import { In, InVariants } from '@corona-dashboard/common';
+import { first, last } from 'lodash';
 import { isDefined } from 'ts-is-present';
 import { VariantChartValue } from './get-variant-chart-data';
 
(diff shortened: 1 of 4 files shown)


packages/app/src/static-props/variants/get-international-variant-chart-data.ts
```diff
@@ -25,57 +25,58 @@ export function getVariantChartData(variants: InVariants | undefined) {
...
+        const otherItem = variantOfConcern.values.find(
+          (x) => x.date_end_unix === partialItem.date_end_unix
+        );
+        if (otherItem) {
+          total += otherItem.percentage;
+          item[`${variantOfConcern.name}_percentage`] = otherItem.percentage;
+        }
+        return { item, total };
+      },
+      { item: partialItem, total: 0 }
```
VWSCoronaDashboard8, line 52: I think this part deserves some high-level description of the what and why of these operations. Just to outline the motivation.

packages/app/src/static-props/variants/get-international-variant-chart-data.ts
```diff
@@ -25,57 +25,58 @@ export function getVariantChartData(variants: InVariants | undefined) {
     return EMPTY_VALUES;
   }
 
+  // The different historical value lists are of different lengths,
+  // so here we create a start to end range based on all of them:
```
VWSCoronaDashboard8, line 29: nit; I prefer to use /** and */ for all (block) comments. They stand out more, are easier to (re)format and give editor hints where code is used.

packages/app/src/static-props/variants/get-international-variant-chart-data.ts
```diff
@@ -25,57 +25,58 @@ export function getVariantChartData(variants: InVariants | undefined) {
...
+function createCompleteDateRange(lists: InVariantsVariant[]) {
+  return lists
+    .flatMap((x) => x.values)
+    .map<VariantChartValue>((x) => ({
+      date_start_unix: x.date_start_unix,
+      date_end_unix: x.date_end_unix,
+      sample_size: x.sample_size,
+    }))
+    .filter(
+      (x, i, arr) =>
```
VWSCoronaDashboard8, line 74: nit; Unnecessary use of short-hands

```suggestion
      ({date_start_unix, date_end_unix}, index, array) =>
```

Then you can also use `x` instead of `y`

packages/app/src/static-props/variants/get-international-variant-chart-data.ts
```diff
@@ -25,57 +25,58 @@ export function getVariantChartData(variants: InVariants | undefined) {
...
     variantChart: values,
-    dates: {
-      date_of_insertion_unix: firstVariant.last_value.date_of_insertion_unix,
-      date_start_unix: firstVariant.last_value.date_start_unix,
-      date_end_unix: firstVariant.last_value.date_end_unix,
-    },
   } as const;
 }
+
+function createCompleteDateRange(lists: InVariantsVariant[]) {
```
VWSCoronaDashboard8, line 65: nit; Somehow I think it's more a "get" than a "create", since it's a conversion of data

packages/app/src/static-props/variants/get-international-variant-chart-data.ts
```diff
@@ -25,57 +25,58 @@ export function getVariantChartData(variants: InVariants | undefined) {
...
+  return lists
+    .flatMap((x) => x.values)
+    .map<VariantChartValue>((x) => ({
+      date_start_unix: x.date_start_unix,
+      date_end_unix: x.date_end_unix,
+      sample_size: x.sample_size,
```
VWSCoronaDashboard8, line 71: I'm not quite following how sample size fits in here. It seems to me that the flatmap throws everything on one big pile, and the findIndex will return the first with matching timestamp. So don't sample sizes get mingled then or are size values the same for all items in the supplied `InVariantsVariant[]`?

Also it makes the function name confusing since it's about a date range. What does sample size have to do with it? A comment would help I think.

VWSCoronaDashboard1: I went looking for places where sample_size is actually used. Turns out, nowhere. So the fix was easy, remove the property altogether.

VWSCoronaDashboard8 (approved):

PR #3216, opened 2021-07-20
Restrict access to image api for non-whitelisted query params

VWSCoronaDashboard4:
## Summary

Restrict access to image api for non-whitelisted query params

### DIFF
--- a/packages/app/next-server.js
+++ b/packages/app/next-server.js
@@ -7,17 +7,15 @@ const {
 } = require('http-proxy-middleware');
 const dotenv = require('dotenv');
 const path = require('path');
+const { imageResizeTargets } = require('@corona-dashboard/common');
 
 const SIX_MONTHS_IN_SECONDS = 15768000;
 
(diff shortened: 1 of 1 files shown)


VWSCoronaDashboard1 (approved):

packages/app/next-server.js
```diff
@@ -7,17 +7,15 @@ const {
...
-
-const MAX_IMAGE_WIDTH = last(imageResizeTargets);
-assert(
-  MAX_IMAGE_WIDTH > 0,
-  'Failed to get maximum image width from imageResizeTargets'
-);
+const ALLOWED_SENTRY_IMAGE_PARAMS = {
+  w: imageResizeTargets.map((x) => x.toString()),
+  q: ['65'],
+  auto: ['format'],
```
VWSCoronaDashboard8, line 17: If we only allow one value it doesn't seem to make much sense to let the client pass these values. But I'm still curious to hear why it would be such a bad idea to have the proxy re-formats the url query parameters.

VWSCoronaDashboard8 (approved):

PR #3230, opened 2021-07-21
Update Tooltip style

VWSCoronaDashboard13:
## Summary

Updating the tooltip style

## Motivation

Design tweak

## Detailed design

Making the Tippy tooltip into a styled component to apply the proper boxShadow and used `useBreakpoints` to apply max width since that style gets overridden by Tippy by default.

## Unresolved questions

Would we still prefer to use rem instead of px for this?

### Governance

- [ ] Documentation is added

### DIFF
--- a/packages/app/src/lib/tooltip.tsx
+++ b/packages/app/src/lib/tooltip.tsx
@@ -13,8 +13,7 @@ const StyledTippy = styled(Tippy)(
   css({
     backgroundColor: 'white',
     color: 'black',
-    boxShadow:
-      '0 0 20px 4px rgb(154 161 177 / 15%), 0 4px 80px -8px rgb(36 40 47 / 25%), 0 4px 4px -2px rgb(91 94 105 / 15%)',
+    boxShadow: 'tile',
     '.tippy-arrow': {
       color: 'white',
     },
@@ -37,7 +36,7 @@ export function WithTooltip(props: Omit<TippyProps, 'theme'>) {
     <StyledTippy
       appendTo={getBody}
       onMount={handleMount}
-      maxWidth={breakpoints.sm ? '285px' : '350px'}
+      maxWidth={breakpoints.sm ? '260px' : '285px'}
       {...props}
     />
   );
--- a/packages/app/src/lib/tooltip.tsx
+++ b/packages/app/src/lib/tooltip.tsx
@@ -18,14 +18,14 @@ const StyledTippy = styled(Tippy)(
       color: 'white',
     },
   })
-);
+) as typeof Tippy;
 
 /**
  * Usage:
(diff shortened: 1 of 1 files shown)


packages/app/src/lib/tooltip.tsx
```diff
@@ -1,23 +1,44 @@
...
+);
+
 /**
  * Usage:
  *     <WithTooltip content={<p>message</p>}><button>foo</button></WithTooltip>
  */
 
-export function WithTooltip(props: TippyProps) {
+export function WithTooltip(props: Omit<TippyProps, 'theme'>) {
+  const breakpoints = useBreakpoints(true);
```
VWSCoronaDashboard10, line 29: the initial value `true` won't be necessary because tooltips are never rendered server-side

packages/app/src/lib/tooltip.tsx
```diff
@@ -1,23 +1,44 @@
...
+
   if (!isDefined(props.content)) {
     return <>{props.children}</>;
   }
 
   return (
-    <Tippy theme="light" appendTo={getBody} onMount={handleMount} {...props} />
+    <StyledTippy
+      appendTo={getBody}
+      onMount={handleMount}
```
VWSCoronaDashboard10, line 38: while you're at it, could you fix another issue with the tooltip which is probably fixed with this change:
```suggestion
      {...handleMount ? { onMount: handleMount } : {}}
```

I've assigned you on trello to the related ticket ("Tippy cannot read property apply of undefined")

VWSCoronaDashboard10 (approved):

VWSCoronaDashboard8 (approved):

packages/app/src/lib/tooltip.tsx
VWSCoronaDashboard13: Hmm I'm a bit confused by this. What is this solution doing? I can't apply the spread operator on a function and I'm getting a typescript error that makes it seem like the original solution is correct 
image

VWSCoronaDashboard1 (approved):

packages/app/src/lib/tooltip.tsx
VWSCoronaDashboard10: Ah yes, a bit of explanation would have been useful.
Apparently the tippy library cannot handle an `onMount` prop with an `undefined` value. It's probably checking on existence of the `onMount` key in the props and it will use that value to overwrite some default value.. 

The `handleMount` will be `undefined` in production builds so those builds would throw an error when interacting with a tooltip. 

The proposed solution is to only apply the `onMount` prop when the `handleMount` handler is defined.

I believe the TS error is caused by the wrapping Tippy with styled-components. A fix could be to cast StyledTippy to the type of Tippy:
```tsx
const StyledTippy = styled(Tippy)`` as typeof Tippy;
```

PR #3239, opened 2021-07-22
Add black hover outline international choropleth

VWSCoronaDashboard14:


packages/app/src/components/choropleth/path.tsx
```diff
@@ -133,9 +133,10 @@ const StyledHoverPath = styled.path<{
     transitionDuration: '120ms, 90ms',
     transitionTimingFunction: 'ease-out',
     cursor: x.isClickable ? 'pointer' : 'default',
-    stroke: x.stroke ? '#000' : 'transparent',
+    stroke: x.stroke ? (x.isSelected ? '#000' : 'transparent') : 'transparent',
     strokeWidth: x.isSelected ? 3 : 0,
```
VWSCoronaDashboard1, line 137: doesn't a black outline look very crude with a stroke width of 3? I'm guessing design will probably have an opinion here...

VWSCoronaDashboard1 (approved):

packages/app/src/components/choropleth/path.tsx
VWSCoronaDashboard14: It's definitely there yes, you can't miss it.

Screenshot 2021-07-22 at 15 26 49

VWSCoronaDashboard8 (approved):

packages/app/src/components/choropleth/path.tsx
VWSCoronaDashboard1: Yea, on France its acceptable I guess, but the smaller features will turn into a blob I think ;)

PR #3241, opened 2021-07-22
Clean up isWeeklyData from sidebar-metric

VWSCoronaDashboard14:
It is not needed anymore to set is explicit. Right now we are just looking for a `date_unix` and when not found we can safely assume it is a week range.


VWSCoronaDashboard1 (approved):

packages/app/src/components/sidebar-metric/sidebar-metric.tsx
```diff
@@ -105,20 +105,21 @@ export function SidebarMetric<T extends { difference: unknown }>({
   const config = getMetricConfig(
     scope,
     metricName as unknown as string,
-    metricProperty
+    metricProperty ?? ''
```
VWSCoronaDashboard8, line 108: This looks odd to me. If metricProperty needs to be optional then I think we should make it optional in the function arguments. An empty string should not be treated as a valid metric property.

It looks like you have been able to remove the use of NO_METRIC_PROPERTY so I think it can just be optional in the arguments and handle that inside the function.

VWSCoronaDashboard8 (changes requested):

packages/app/src/metric-config/index.ts
```diff
@@ -21,21 +20,24 @@ export type DataScope = 'nl' | 'vr' | 'gm';
...
   /**
    * Fall back to an empty object so we don't have to specify empty objects in
    * the config file for properties that do not need a config. Since all
    * root-level properties in the MetricConfig are optional, an empty object is
    * still a valid configuration.
    */
-  const config = get(metricConfig, [scope, metricName, metricProperty], {});
+  const config = get(
+    metricConfig,
+    [scope, metricName, metricProperty ?? ''],
```
VWSCoronaDashboard8, line 38: nit; I think this is a little cleaner, because then you don't use '' as a property.

```suggestion
    metricProperty ? [scope, metricName, metricProperty] : [scope, metricName],
```

VWSCoronaDashboard8 (approved):

PR #3243, opened 2021-07-22
Create useCollapsible hook

VWSCoronaDashboard10:
Generic hook for collapsing content. Core features:

  - SSR will always display content
  - a11y aria props are taken care of
  - will animate out of the box

Usage:

The hook exposes the following:
  - `content(children: ReactNode)`
    - wrap the collapsible content with this function
  - `button(children?: ReactNode)`
    - wrap an element (probably a button) with this function which will set
      aria props and a click handler
    - the child is optional, when it's empty it will render a button with a
      chevron
  - `chevron`
    - a chevron-svg which will animate based on internal open/close state
  - `isOpen`
    - current open/close state

### DIFF
--- a/packages/app/src/components/collapsible/collapsible-content.tsx
+++ b/packages/app/src/components/collapsible/collapsible-content.tsx
@@ -1,16 +1,33 @@
-import {
-  Disclosure,
-  DisclosureButton,
-  DisclosurePanel,
-} from '@reach/disclosure';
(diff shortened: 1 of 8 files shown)


VWSCoronaDashboard1 (approved):

VWSCoronaDashboard8 (approved):
💯

PR #3248, opened 2021-07-23
Release/2.25.3

VWSCoronaDashboard8:


VWSCoronaDashboard1 (approved):

VWSCoronaDashboard4 (approved):

packages/app/src/static-props/variants/get-variant-chart-data.ts
```diff
@@ -37,6 +38,7 @@ export function getVariantChartData(variants: NlVariants | undefined) {
       [`${firstVariant.name}_occurrence`]: value.percentage,
```
VWSCoronaDashboard10, line 38: unrelated to the PR but I just noticed that we're writing value.percentange to the _occurencee property, looks like a bug @VWSCoronaDashboard1 
```suggestion
      [`${firstVariant.name}_occurrence`]: value.percentage,
```

PR #3249, opened 2021-07-23
Update header component wrapping text

VWSCoronaDashboard14:
Fix wrapping I didn't see when working with dummy content.

Old:
Screenshot 2021-07-23 at 10 44 48

New:
Screenshot 2021-07-23 at 10 45 01


VWSCoronaDashboard1 (approved):

packages/app/src/components/page-information-block/components/articles.tsx
```diff
@@ -77,7 +77,7 @@ function ArticleItem({ slug, cover, title }: ArticleItemProps) {
                     <ChevronIcon />
                   </InlineText>
                 ) : (
-                  `${word} `
+                  <InlineText>{`${word} `}</InlineText>
```
VWSCoronaDashboard8, line 80: nit; Not the focus of this PR but I think it would look better if InlineText had a margin right instead of an extra space to push the chevron. To me the alignment of the chevron in both screenshots doesn't look ideal, but if design thinks this is ok then maybe just leave it.

VWSCoronaDashboard8 (approved):

PR #3252, opened 2021-07-23
Add script to sync lokalize texts from production to development

VWSCoronaDashboard8:
## Summary

This pulls in all the published texts from the production set that still exist in the development dataset.


VWSCoronaDashboard1 (approved):
I'm not sure if it's needed, but maybe add a prompt to this one that says: 'Hey, this is going to sync all of the production lokalize data to the development dataset, are you sure you wanna do that?'
Or is that unnecessary, you reckon?

VWSCoronaDashboard8:
> I'm not sure if it's needed, but maybe add a prompt to this one that says: 'Hey, this is going to sync all of the production lokalize data to the development dataset, are you sure you wanna do that?'
> Or is that unnecessary, you reckon?

The one scenario I can think of where you don't want to overwrite development, is when you've changed the number of variables used in a text. But since these text changes are not synced to production (for already existing texts), you should manually update the texts in the production set anyway when you make these kinds of changes. 

However you can only update the production set with new variables once the code is about to be released, so if someone runs this script right before a release it could be annoying. A prompt is easy to add, so I'll do that 👍

VWSCoronaDashboard4 (approved):

PR #3266, opened 2021-07-26
Add groups for data explanation page

VWSCoronaDashboard13:
## Summary

Updates schemas for CMS to require a group selection for every data explanation and updated UI to reflect this categorization.

## Motivation

Improved readability

## Detailed design

Modeled schema additions and styling after FAQ page. Any items that are not assigned to a group are grouped together at the bottom of the page without a title. All groups and items within them are sort alphabetically


packages/app/src/pages/verantwoording.tsx
```diff
@@ -80,22 +87,38 @@ const Verantwoording = (props: StaticProps<typeof getStaticProps>) => {
...
-                    <Box mt={3}>
-                      <RichContent blocks={item.content} />
-                    </Box>
-                  )}
-                </CollapsibleSection>
-              ) : null;
-            })}
-          </article>
-        )}
+        {Object.entries(groups)
```
VWSCoronaDashboard4, line 90: I'd sort/map/check this array outside the JSX, it keeps the JSX cleaner.

packages/app/src/pages/verantwoording.tsx
```diff
@@ -80,22 +87,38 @@ const Verantwoording = (props: StaticProps<typeof getStaticProps>) => {
...
+          .sort((a, b) => {
+            if (a[0] === 'null') return 1;
```
VWSCoronaDashboard4, line 92: Strings values 'null' ? Can't we tweak the query to return something better perhaps?

VWSCoronaDashboard1 (commented):
I think this change requires a migration as well. Since the group field is required, every explanation will become invalid. This can be fixed manually in the UI, but a migration will take care of it automatically.

if you have a look at `cms/src/migrations/sprint18`, this migration can probably be copied verbatim (since it uses the same pattern and I think you've copied the schema's from the FAQ bits), but with the correct schema names replaced.

Once the migration has run, you also won't be needing the null checks because you'll be sure that every item will be assigned to a group.

packages/app/src/pages/verantwoording.tsx
VWSCoronaDashboard13: removed most of it! Do you still think it's necessary to sort outside of the jsx?

packages/app/src/pages/verantwoording.tsx
VWSCoronaDashboard13: added a migration script so there should be no nulls 👍

packages/app/src/pages/verantwoording.tsx
VWSCoronaDashboard1: We prefer to do any kind of data manipulation outside of the JSX

packages/cms/src/migrations/sprint28/index.ts
```diff
@@ -0,0 +1,98 @@
...
+ * The first of the 'cijferVerantwoordingGroups' documents is then retrieved.
+ *
+ * Then it changes the '_type' field for all 'cijferVerantwoording.collapsibleList' documents from 'reference' to 'figureExplanationItem' and
+ * assigns the cijferVerantwoordingGroups document to the 'group' field.
+ *
+ * This migration can be run by executing the following command from the packages/cms directory:
+ * sanity exec src/migrations/sprint28 --with-user-token
+ *
+ * to run the migration against the production dataset, prefix the command like this:
+ * SANITY_STUDIO_API_DATASET=production sanity exec src/migrations/sprint28 --with-user-token
```
VWSCoronaDashboard1, line 14: This is no longer true, to run this migration against the production dataset you simply need to change `getClient('development')` to `getClient('production')`,

VWSCoronaDashboard1 (approved):

packages/cms/src/migrations/sprint28/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  client.fetch(`*[_type == 'cijferVerantwoordingGroups'][0]`);
+
+const buildPatches = (docs: any[], group: any) =>
+  docs
+    .map((doc) => ({
+      id: doc._id,
+      patch: {
+        set: {
+          ...doc,
+          _type: 'figureExplanationItem',
```
VWSCoronaDashboard8, line 33: I think this patch / PR could a be good opportunity to move away from the `figureExplanationItem`. I think the name is poorly chosen, as "cijferverantwoording" is about the data and is not tied to any sort of visual representation.

Also for the groups you are using the type `cijferVerantwoordingGroups` which seems inconsistent with the English type name `figureExplanationItem`. So if it's not too much of a hassle, I'd suggest to copy that data over to a new type named something like `cijferVerantwoordingItem`

Also, shouldn't `cijferVerantwoordingGroups` be named `cijferVerantwoordingGroup` instead? Isn't the type defining a single group?

VWSCoronaDashboard8 (changes requested):

packages/cms/src/migrations/sprint28/index.ts
VWSCoronaDashboard8: If you do decide to stick with `figureExplanationItem` that's fine to, but then I'd expect the group to be named `figureExplanationGroup` or `figureExplanationItemGroup` instead.

VWSCoronaDashboard8: I see that the word "figure" is used a lot in the English texts that these items contain. And that figure in English can also just mean statistics without having a visual representation attached to it. So I guess the word figure is fine here after all.

My suggestions for consistent type naming still stand then.

VWSCoronaDashboard13: I can update `figureExplanationItem` to `cijferVerantwoordingItem`. I will say it looks like the other documents like FAQ also use a mix of english and dutch naming so in a way this change will result in inconsistency with those. 

As for the `cijferVerantwoordingGroup` document I think it should be plural since it houses all of the groups? 
```
export const cijferVerantwoordingGroups = {
  name: 'cijferVerantwoordingGroups',
  type: 'document',
  title: 'Cijfer verantwoording groepen',
  fields: [
    {
      name: 'group',
      type: 'localeString',
      title: 'Groepsnaam',
    },
  ],
  preview: {
    select: {
      title: 'group',
    },
    prepare: prepareLocalized,
  },
};
```

VWSCoronaDashboard8: You're right `cijferVerantwoordingGroups` is correct then 👍

VWSCoronaDashboard8 (approved):

PR #3269, opened 2021-07-27
Fix barchart rendering for big datasets

VWSCoronaDashboard1:
## Summary

- Render bar chart as a stepped area chart when bar width is 1 pixel
- Fix centering of the marker dot on the tooltip marker line

image

### Governance

- [ ] Documentation is added


VWSCoronaDashboard13 (approved):
🎉

packages/app/src/components/time-series-chart/components/point-markers.tsx
```diff
@@ -22,10 +22,12 @@ export function PointMarkers<T extends TimestampedValue>(
 
   if (!points.length) return null;
 
+  const x = points[0].x - Math.floor(size / 2);
```
VWSCoronaDashboard8, line 25: nit; Maybe something like `offsetX` or `centerX` is a bit more descriptive. I associate the name x with the location at which the component is drawn. Also below there is a similar calculation for y for which you don't use a variable. Maybe you can place that here as `offsetY` for consistency

VWSCoronaDashboard8 (approved):

packages/app/src/components/time-series-chart/components/point-markers.tsx
VWSCoronaDashboard1: the y calculation is done with a map loop, so that's why it's not on the same level as the x position

PR #3272, opened 2021-07-28
Fix locale sanity issues

VWSCoronaDashboard4:
## Summary

Fix locale sanity issues

* Always specify a domains list, by adding an explicit fallback when no `DOMAIN_EN` is set.
* `getStaticPaths` is behaving odd. We don't use localized routes, but the returned array needs both locales.
* Some pages using our `<Link>` component were broken, especially ones using Rich Content from Sanity. Fixed by always explicitly providing a default locale to `Link`.

### DIFF
--- a/packages/app/src/pages/artikelen/[slug].tsx
+++ b/packages/app/src/pages/artikelen/[slug].tsx
@@ -19,9 +19,23 @@ const articlesQuery = `*[_type == 'article'] {"slug":slug.current}`;
 export async function getStaticPaths() {
   const articlesData = await (await getClient()).fetch(articlesQuery);
 
-  const paths = articlesData.map((article: { slug: string }) => ({
-    params: { slug: article.slug },
-  }));
+  /**
+   * getStaticPaths needs explicit locale routes to function properly...
+   */
+  const paths = articlesData.reduce(
+    (paths: { params: any; locale: string }[], article: { slug: string }) => {
+      return paths.concat([
+        {
+          params: { slug: article.slug },
+          locale: 'en',
+        },
+        {
+          params: { slug: article.slug },
+          locale: 'nl',
+        },
+      ]);
+    }
+  );
 
   // { fallback: false } means other routes should 404.
   return { paths, fallback: false };
--- a/packages/app/src/pages/artikelen/[slug].tsx
+++ b/packages/app/src/pages/artikelen/[slug].tsx
@@ -22,20 +22,16 @@ export async function getStaticPaths() {
   /**
    * getStaticPaths needs explicit locale routes to function properly...
    */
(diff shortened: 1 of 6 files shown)


packages/app/src/utils/link.tsx
```diff
@@ -1,11 +1,13 @@
...
 import NextLink, { LinkProps } from 'next/link';
+import { useIntl } from '~/intl';
 
 /**
  * Link component which disables the default Next/Link scroll behavior
  */
 
 export function Link(props: LinkProps & { children?: React.ReactNode }) {
-  const { locale = false } = props;
+  const defaultLocale = useIntl().locale ?? 'nl';
```
VWSCoronaDashboard10, line 10: would't it be better to set the default in the useIntl hook?

packages/app/src/pages/artikelen/[slug].tsx
```diff
@@ -19,9 +19,23 @@ const articlesQuery = `*[_type == 'article'] {"slug":slug.current}`;
...
+          params: { slug: article.slug },
+          locale: 'en',
+        },
+        {
+          params: { slug: article.slug },
+          locale: 'nl',
+        },
+      ]);
+    }
+  );
```
Comment (VWSCoronaDashboard10) [inline packages/app/src/pages/artikelen/[slug].tsx:38]: A flatMap seems to be more appropriate:
```tsx
// also don't forget to type the articlesData:
type Article { slug: string }
const articlesData: Article[] = await (await getClient()).fetch(articlesQuery);

const paths = articlesData.flatMap((article) => [
  {
    params: { slug: article.slug },
    locale: 'en',
  },
  {
    params: { slug: article.slug },
    locale: 'nl',
  },
]);
```

VWSCoronaDashboard10 (approved):

packages/app/src/utils/link.tsx
VWSCoronaDashboard4: That'd be a good one for a separate PR :-)

VWSCoronaDashboard8: It seems to me like the locale already has a fallback to `nl` in the _app.tsx. Not sure why it isn't working then:

```
  const locale = (process.env.NEXT_PUBLIC_LOCALE || 'nl') as LanguageKey;

  const [text, toggleHotReloadButton, dataset] = useLokalizeText(locale);

  assert(text, `Encountered unknown language: ${locale}`);

  const intlContext = useIntlHelperContext(locale, text, dataset);
```

VWSCoronaDashboard8: I think creating a separate PR for that is a bit unnecessary overhead. Might as well figure it out straight away, since this missing fallback seems to be part of the bug.

PR #3274, opened 2021-07-28
Add new article filters

VWSCoronaDashboard14:
Add new article filters to the CMS and pages It has a hard-coded list of predefined filter options.
Every tag gets at least used once so there is no need for a zero state when there are none selected. When the user enters a wrong query in the URL bar it has a fallback to `alles`.
At the bottom of article detail pages, the user should see the tags related to the article and could click on them.

On the overview page of the articles there is a bar width for all the possible options and going on mobile it shows a select. The filter is using the param routes e.g. `/artikelen?categorie=vaccinaties`

Article overview:
Screenshot 2021-07-28 at 11 21 31

Mobile article overview:
Screenshot 2021-07-28 at 11 46 05

Bottom of article detail:
Screenshot 2021-07-28 at 11 15 03


packages/app/src/domain/topical/common/categories.ts
```diff
@@ -0,0 +1,12 @@
...
+  'infecties',
+  'sterfte',
+  'ziekenhuizen',
+  'kwetsbare_groepen',
+  'vroege_indicatoren',
+  'gedrag',
+] as const;
+export type CategoriesTypes = typeof categories[number];
+
+export const categoryAll = 'alles';
```
VWSCoronaDashboard1, line 12: I wonder if the `categoryAll` is really needed. What if `currentCategory === undefined` simply implies `categoryAll`? The the category query param can be optional, so if the param doesn't exist, all categories are assumed.

packages/cms/src/schemas/fields/article-fields.ts
```diff
@@ -40,6 +40,24 @@ export const ARTICLE_FIELDS = [
     fieldset: 'metadata',
     validation: (rule: Rule) => rule.required(),
   },
+  {
+    title: 'Categorieën instellen',
+    name: 'categories',
+    type: 'array',
```
VWSCoronaDashboard1, line 46: Sanity has support for a specialized tag UI view:
https://www.sanity.io/docs/array-type#layout-2e4724b419b9

VWSCoronaDashboard1: The only caveat is that the array needs to contain simple strings, if I understand correctly.

VWSCoronaDashboard1 (changes requested):

packages/cms/src/schemas/fields/article-fields.ts
VWSCoronaDashboard14: Seems to work.

Screenshot 2021-07-28 at 13 54 45

packages/app/src/components/article-detail.tsx
```diff
@@ -1,14 +1,17 @@
+import { css } from '@styled-system/css';
+import styled from 'styled-components';
 import { ArrowIconLeft } from '~/components/arrow-icon';
 import { Box } from '~/components/base';
 import { ContentBlock } from '~/components/cms/content-block';
 import { Heading, InlineText } from '~/components/typography';
+import { CategoriesTypes } from '~/domain/topical/common/categories';
```
VWSCoronaDashboard8, line 7: If a type has multiple options, it usually still has a singular name. Also `CategoriesTypes` having plurals twice is not correct.

So I think this should be `CategoryType` and maybe to avoid confusion with a real type (since it's just an enum) call it `Category` or `CategoryName`

VWSCoronaDashboard8: If the categories are only used for articles maybe name it `ArticleCategory`

packages/app/src/domain/topical/article-list.tsx
```diff
@@ -23,20 +32,30 @@ export function ArticleList({ articleSummaries }: ArticleListProps) {
...
-        </ArticleBox>
-      ))}
+      {articleSummaries
+        .filter((item) => {
+          if (!isDefined(currentCategory)) return true;
+
+          return (
+            currentCategory === categoryAll ||
+            (isDefined(currentCategory) &&
+              item.categories?.includes(currentCategory))
```
VWSCoronaDashboard8, line 42:
```suggestion
        .filter((item) => isDefined(currentCategory) ?
           currentCategory === categoryAll 
            && item.categories?.includes(currentCategory) : true
```

packages/app/src/domain/topical/common/categories.ts
VWSCoronaDashboard8: I agree, try to use undefined instead. But if you're still keeping categoryAll logic, I think the string should just be part of the array (probably as a special string like __all), so that your type is correct. Because if all is a type of category it should be included in the type.

packages/app/src/pages/artikelen/index.tsx
```diff
@@ -34,24 +45,186 @@ export const getStaticProps = createGetStaticProps(
   })
 );
 
+const allCategories = [categoryAll, ...categories] as unknown as (
+  | CategoriesTypes
+  | typeof categoryAll
+)[];
```
VWSCoronaDashboard8, line 51: .... because then you avoid a mess like this

packages/app/src/pages/artikelen/index.tsx
```diff
@@ -34,24 +45,186 @@ export const getStaticProps = createGetStaticProps(
...
+
+      return {
+        label,
+        value: id,
+      };
+    });
+  }, [siteText]);
+
+  const handleCategoryFilter = useCallback(
+    function setNewParam(item: CategoriesTypes | typeof categoryAll) {
```
VWSCoronaDashboard8, line 72: ... and this

packages/app/src/pages/artikelen/index.tsx
```diff
@@ -34,24 +45,186 @@ export const getStaticProps = createGetStaticProps(
...
+        },
+        undefined,
+        { shallow: true }
+      );
+    },
+    [replace]
+  );
+
+  const currentCategory = [...categories, categoryAll].includes(
+    query.categorie as CategoriesTypes | typeof categoryAll
```
VWSCoronaDashboard8, line 86: etc

VWSCoronaDashboard8 (changes requested):

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard8 (approved):

VWSCoronaDashboard1 (approved):

PR #3275, opened 2021-07-28
PoC: Add an icon package to de-dupe icons across the CMS and front-end

VWSCoronaDashboard:
## Summary

I've forked react-feather and made our own icon system which gets generated automatically from a folder of SVG's.

## Motivation

We are duplicating the icons in the CMS and the front-end. This way we can de-dupe it and have a single source of truth.
There's also not a convenient way to change height, width, color or other attributes. This icon package solves that too.

## Detailed design

`yarn compile` reads all of the svg files in `/src/svg/` and outputs them as React Components
You can read `bin/build.js` for more info on how it works.

## Unresolved questions

- Should Icons be same-size or have the same height but different widths?
- Which attributes do we want to set by default?
- Do we want to provide the SVG wrapper in the build script, or in the icon itself?


packages/icons/bin/build.js
```diff
@@ -0,0 +1,134 @@
...
+  const defaultAttrs = {
+    xmlns: 'http://www.w3.org/2000/svg',
+    width: 'size',
+    height: 'size',
+    viewBox: '0 0 36 36',
+    fill: 'none',
+    stroke: 'color',
+    // strokeWidth: 2,
+    // strokeLinecap: 'round',
+    // strokeLinejoin: 'round',
```
VWSCoronaDashboard, line 75: What do we want/need here by default?

packages/icons/bin/build.js
```diff
@@ -0,0 +1,134 @@
...
+};
+
+icons.forEach((i) => {
+  const location = path.join(rootDir, 'src/icons', `${i}.js`);
+  const ComponentName = i === 'github' ? 'GitHub' : upperCamelCase(i);
+  const defaultAttrs = {
+    xmlns: 'http://www.w3.org/2000/svg',
+    width: 'size',
+    height: 'size',
+    viewBox: '0 0 36 36',
```
VWSCoronaDashboard, line 70: Should icons be equal in height/width or should the width be dynamic? That would be tricky if we generate the default attributes around every SVG files in the build script.

packages/icons/src/icons/alcohol-verkoop.js
```diff
@@ -0,0 +1,30 @@
...
+);
+
+AlcoholVerkoop.propTypes = {
+  color: PropTypes.string,
+  size: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+};
+
+AlcoholVerkoop.displayName = 'AlcoholVerkoop';
+
+export default AlcoholVerkoop;
```
VWSCoronaDashboard, line 30: This is all automatically generated. We could also gitignore it, but I wouldn't see why you don't want to commit this.

packages/icons/src/svg/afstand-sporten.svg
```diff
@@ -0,0 +1 @@
... 8.74559 20.5333 9.03366C20.4624 8.98015 20.4144 8.90238 20.3976 8.81528C20.392 8.78541 20.3902 8.75555 20.392 8.72693C20.3933 8.70266 20.3976 8.6784 20.4039 8.65476C20.4126 8.6224 20.4263 8.5913 20.4443 8.56205C20.5407 8.33372 20.6739 8.1228 20.8394 7.93739C21.0372 7.81918 21.2538 7.7383 21.4784 7.69724C21.6327 7.66924 21.7901 7.65991 21.9475 7.66986C22.1248 7.65991 22.3009 7.70906 22.4465 7.81109C22.592 7.91251 22.6991 8.06058 22.7507 8.23044C22.8378 8.42642 22.9492 8.61058 23.0817 8.77857C23.1707 8.80408 23.4002 8.86754 23.8719 8.86754L23.9098 8.8551C23.893 8.93536 23.8501 9.00753 23.7879 9.06041C23.7256 9.1133 23.6479 9.14441 23.5657 9.14814C23.3642 9.18734 23.1564 9.18734 22.9541 9.14814C22.7762 9.07161 22.5976 8.75306 22.5087 8.58756L22.4577 8.8047L22.4166 8.97207C22.373 9.15187 22.3282 9.33666 22.2915 9.50464C22.5603 9.69938 22.7389 9.82942 22.8546 9.92212C22.9971 10.036 23.0437 10.0926 23.0437 10.1417C23.0637 10.2475 23.0637 10.3552 23.0437 10.4603V11.4676Z"/><path d="M23.9353 25.3433C23.9353 25.9629 23.4338 26.4644 22.8142 26.4644C22.1951 26.4644 21.693 25.9629 21.693 25.3433C21.693 24.7242 22.1951 24.2221 22.8142 24.2221C23.4338 24.2221 23.9353 24.7242 23.9353 25.3433Z"/>
```
VWSCoronaDashboard, line 1: In this file you can see I removed the outer SVG element. That's because the build scripts add a generic SVG wrapper around every icon. We can also choose to keep the outer SVG elements in the files themselves. What do you think?

VWSCoronaDashboard: We're going to remove all of the outer SVG's automatically.

packages/icons/bin/build.js
VWSCoronaDashboard: The script now reads the original width/height properties by parsing the SVG first and uses those properties. However, size expects a number that is equal in width and height so not sure if this is really going to help us.

VWSCoronaDashboard1 (approved):

packages/cms/src/components/icons/icon.tsx
```diff
@@ -69,17 +68,21 @@ export function Icon(props: IconProps) {
...
-                  key={id}
-                  direction="column"
-                  align="center"
-                  onClick={(event) => onChange(createPatchFrom(id))}
-                >
-                  <img src={src} width="36" height="36" />
-                  <Radio checked={value === id} readOnly />
-                </Flex>
-              ))}
+              {allIcons.map(([id, src]) => {
```
VWSCoronaDashboard17, line 71: Should be able to use:

```suggestion
              {allIcons.map(([id, GridIcon]) => {
```

packages/icons/.eslintrc
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "airbnb",
```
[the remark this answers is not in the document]
VWSCoronaDashboard: Sounds like a plan.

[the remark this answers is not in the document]
VWSCoronaDashboard: I've removed the Airbnb config and disabled eslint for now. I don't think it's worth the effort at the moment.

packages/icons/package.json
```diff
@@ -0,0 +1,48 @@
...
+  "version": "0.0.0",
+  "description": "React ready Corona Dashboard Icons",
+  "main": "build/index.js",
+  "module": "dist/index.js",
+  "sideEffects": false,
+  "typings": "dist/index.d.ts",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "compile": "rm -rf src/icons && node bin/build.js",
+    "build:es": "rm -rf dist && babel src --out-dir dist --copy-files",
```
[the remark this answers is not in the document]
VWSCoronaDashboard: Good catch. Done.

PR #3276, opened 2021-07-28
Tweaks on country select dropdown

VWSCoronaDashboard13:
## Summary

- Added one fixed decimal place to all of the values for the countries. This required updating the formatting function in commons
- Updated the number of selected countries to be reflective of all countries rather than just the ones that are shown by passing the value from the context

### DIFF
--- a/packages/common/src/intl/create-formatting.ts
+++ b/packages/common/src/intl/create-formatting.ts
@@ -65,12 +65,15 @@ export function createFormatting(
 ) {
   function formatNumber(
     value: number | string | undefined | null,
-    options?: {
-      maximumFractionDigits?: number;
-      minimumFractionDigits?: number;
-    }
+    numFractionDigits?: number
   ): string {
     if (typeof value === 'undefined' || value === null) return '-';
+    const options = numFractionDigits
+      ? {
+          maximumFractionDigits: numFractionDigits,
+          minimumFractionDigits: numFractionDigits,
+        }
+      : undefined;
 
     return Intl.NumberFormat(languageTag, options).format(Number(value));
   }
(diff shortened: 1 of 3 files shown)


packages/common/src/intl/create-formatting.ts
```diff
@@ -63,13 +63,16 @@ export function createFormatting(
...
 ) {
-  // Number formatting
-  const NumberFormat = new Intl.NumberFormat(languageTag);
-
-  function formatNumber(value: number | string | undefined | null): string {
+  function formatNumber(
+    value: number | string | undefined | null,
+    options?: {
+      maximumFractionDigits?: number;
+      minimumFractionDigits?: number;
```
VWSCoronaDashboard8, line 70: I've mentioned this recently in a PR from @VWSCoronaDashboard1. I think we should simplify the signature for these formatting functions because I think we will always want to use min and max together and with the same value. Also in that case I don't see the point in having an options object. If could just be 

```suggestion
numFractionDigits?: number;
```

And for formatPercentage we can even give it a default of 1.  But wait, now I realize that what you're trying to do is already what `formatPercentage` is doing....? 

I think we should get rid of `formatPercentage` all together (because it is also used for formatting decimal values that are not percentages), and unify all things by just having a `formatNumber` function with a fraction digits min/max set to 0.

But that would greatly inflate the scope of this PR, so I'll create a task for it in the tech dept column.

So for this PR just use the existing formatPercentage function then I guess.

VWSCoronaDashboard8 (approved):

packages/common/src/intl/create-formatting.ts
VWSCoronaDashboard13: I updated the second parameter to the optional numFractionDigits parameter and used it in this instance

VWSCoronaDashboard1 (approved):

PR #3278, opened 2021-07-28
Sanity: Fix 'double language fields' for collapsibles in portable text fields

VWSCoronaDashboard1:
## Summary

For context, this problem:
image

The reason why the duplicate language fields were shown was because of a rather nasty schema mistake in the case of 'collapsible' text blocks in portable text. The regular 'collapsible' schema object was used to define such block, but this was added to a specific language with a parent portable text object.

Like this:
```
PortableText -> English -> Collapsible -> English
                                       -> Dutch
             -> Dutch  -> Collapsible -> English
                                      -> Dutch
```

So, I've created a separate schema object called 'inlineCollapsible' that simply has a content and title property without any localization.

After that a migration had to be written that retrieves all of the schema's that have portable text in them, find the portable text instances that have a collapsible in them and replace them with the new inlineCollapsible.

Some other changes:
- As part of this PR all of the Sanity version shave been upgraded to the latest version as well.
- The localeBlock and localeRichContentBlock now share the same fields. Technically becoming the same thing, in the future  another migration can be written to normalize those. But for now, this works.

### Governance

- [ ] Documentation is added


packages/cms/src/migrations/sprint28/collapsibles.ts
```diff
@@ -0,0 +1,114 @@
...
+
+function whereClause(values: string[]) {
+  return values
+    .filter((x, i) => i > 0)
+    .map(
+      (x) =>
+        `${x}.en[]._type match 'collapsible' || ${x}.nl[]._type match 'collapsible'`
+    )
+    .join(' || ');
+}
```
VWSCoronaDashboard8, line 31: This is really hard to follow because "values" as a name and a type doesn't tell you anything. I think it would help a lot if you typed the argument like `createQuery(schemas: {name: SchemaName, fields: string[]}[])`

VWSCoronaDashboard8 (approved):

packages/cms/src/migrations/sprint28/collapsibles.ts
VWSCoronaDashboard1: Well, since this is a migration and will be run just once. I didn't pay that much attention to details like that, just wanted to get the damned working. But I will do some renaming.

VWSCoronaDashboard1: Done! Hopefully more readable now.

VWSCoronaDashboard13 (approved):

PR #3279, opened 2021-07-29
Update article detail page full width image

VWSCoronaDashboard14:
The full-width gray background was gone also the option in the menu was gone (and is default true now). This is brought back.


VWSCoronaDashboard1 (approved):

packages/app/src/components/article-detail.tsx
```diff
@@ -48,14 +48,20 @@ export function ArticleDetail({ article }: ArticleDetailProps) {
...
-        <ContentBlock>
-          <Box fontSize="1.125rem">
-            <RichContent
-              blocks={article.content}
-              contentWrapper={ContentBlock}
-            />
-          </Box>
-        </ContentBlock>
+        <Box
+          fontSize="1.125rem"
```
VWSCoronaDashboard13, line 52: Should this use the new textVariant prop to set fontSize?

VWSCoronaDashboard14: This branch is quite old, might need to merge in the new develop to have this updated.

PR #3282, opened 2021-07-29
Add command line script to add an element

VWSCoronaDashboard1:
## Summary

Kind of what the title states.
The prompt lets you choose from a list of generated types and metric names (pulled from the schema's).
When finished successfully it checks if an element with the given ID already exists
and if not, it creates it on both development and production.

### Governance

- [ ] Documentation is added

### DIFF
--- a/packages/cms/src/elements/add.ts
+++ b/packages/cms/src/elements/add.ts
@@ -1,0 +1,180 @@
+import prompts from 'prompts';
+import { isDefined } from 'ts-is-present';
+import { getClient } from '../client';
+import {
+  loadSchemaMetricProperties,
+  loadSchemaMetrics,
+  Scope,
+} from './logic/load-schema';
+
+type Element = {
+  scope: Scope;
+  metricName: string;
+  metricProperty: string | undefined;
+  _type: ElementType;
+};
+
+type ElementType = 'choropleth' | 'timeSeries' | 'kpi';
+
+const devClient = getClient('development');
+const prodClient = getClient('production');
+
+(async function run() {
+  const element = await promptForElement();
+
+  if (isDefined(element)) {
+    await saveElement(element);
+  }
+
+  console.log('Have a nice day...');
+})();
+
+async function saveElement(element: Element) {
+  return Promise.all([devClient.create(element), prodClient.create(element)]);
+}
+
+async function promptForElement(): Promise<Element | undefined> {
+  const element: Element = {
+    scope: 'nl',
+    metricName: '',
+    metricProperty: undefined,
+    _type: 'choropleth',
+  };
+
+  const choices = [
+    { title: 'Municipal', value: 'gm' },
+    { title: 'Safety region', value: 'vr' },
+    { title: 'National', value: 'nl' },
+    { title: 'International', value: 'in' },
+  ];
+
+  const scopeResponse = (await prompts({
+    type: 'select',
+    name: 'scope',
+    message: 'Select the scope for this element:',
+    choices,
+    onState,
+  })) as { scope: Scope };
+
+  element.scope = scopeResponse.scope;
+
+  const metricNameResponse = (await prompts({
+    type: 'select',
+    name: 'metricName',
+    message: 'Select the metric name for this element:',
+    choices: loadSchemaMetrics(scopeResponse.scope).map((x) => ({
+      title: x,
+      value: x,
+    })),
+    onState,
+  })) as { metricName: string };
+
+  element.metricName = metricNameResponse.metricName;
+
+  const metricPropertyConfirmationresponse = await prompts([
+    {
+      type: 'confirm',
+      name: 'isConfirmed',
+      message: 'Do you need to select a metric property?',
+      initial: false,
+    },
+  ]);
+
+  if (metricPropertyConfirmationresponse.isConfirmed) {
+    const metricPropertyResponse = (await prompts({
+      type: 'select',
+      name: 'metricProperty',
+      message: 'Select a metric property for this element:',
+      choices: loadSchemaMetricProperties(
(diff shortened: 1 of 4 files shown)


packages/cms/src/elements/add.ts
```diff
@@ -0,0 +1,180 @@
...
+        title: x,
+        value: x,
+      })),
+      onState,
+    })) as { metricProperty: string };
+
+    element.metricProperty = metricPropertyResponse.metricProperty;
+  }
+
+  const elemenTypeChoices = [
```
VWSCoronaDashboard8, line 97: typo

packages/cms/src/elements/logic/load-schema.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export function loadSchemaMetrics(scope: Scope) {
+  const schema = loadJSONFromFile(path.join(schemaPath, scope, '__index.json'));
+
+  return Object.entries<{ type: string } | { $ref: string }>(schema.properties)
+    .filter(pickMetricNames)
+    .map(([key]) => key);
+}
+
+export function loadSchemaMetricProperties(scope: Scope, metricName: string) {
```
VWSCoronaDashboard8, line 16: nit; the fact that it's loading something to get to the metric properties is not so interesting for the consumer I think. This could just be named `getMetricProperties` or something, because then you also know that you're getting.

packages/cms/src/elements/logic/load-schema.ts
```diff
@@ -0,0 +1,44 @@
...
+  { type: string } | { $ref: string }
+]) {
+  if ('type' in value) {
+    return false;
+  } else {
+    return !value.$ref.startsWith('__');
+  }
+}
+
+function loadJSONFromFile(filePath: string) {
```
VWSCoronaDashboard8, line 41:
```suggestion
function loadJsonFromFile(filePath: string) {
```

packages/cms/src/elements/add.ts
```diff
@@ -0,0 +1,180 @@
...
+      name: 'isConfirmed',
+      message: `Is this correct?\n${elementToString(
+        element
+      )}\nWhen choosing 'yes' this element will be saved to both development and production.`,
+      initial: false,
+    },
+  ]);
+
+  if (isCorrectResponse.isConfirmed) {
+    return element;
```
VWSCoronaDashboard8, line 147: An element needs to be stored under a specific id formatted as [scope, metricName, elementType, (optional) metricProperty]. Then these would be run through change-case snakeCase function and joined together using `__`. 

Also see https://coronadashboard.sanity.studio/desk/elements;nl;nl__hospital_nice__time_series

I don't think I have the original script anymore which did this for the initial set, but I'm sure you'll figure it out :)

packages/cms/src/elements/add.ts
```diff
@@ -0,0 +1,180 @@
...
+  const { _type, scope, metricName, metricProperty } = element;
+  const query = !isDefined(metricProperty)
+    ? `*[_type == '${_type}' && scope == '${scope}' && metricName == '${metricName}'][0]`
+    : `*[_type == '${_type}' && scope == '${scope}' && metricName == '${metricName}' && metricProperty == '${metricProperty}'][0]`;
+
+  const devDocument = await devClient.fetch(query);
+  const prodDocument = await prodClient.fetch(query);
+
+  return devDocument === null && prodDocument === null;
+}
```
VWSCoronaDashboard8, line 161: Based on that fixed id, this function can be simplified because you only need to check for documents with a specific id.

packages/cms/src/elements/add.ts
```diff
@@ -0,0 +1,180 @@
...
+
+function elementToString(element: Element) {
+  if (!isDefined(element.metricProperty)) {
+    return `${element.scope}.${element.metricName}.${element._type}`;
+  }
+  return `${element.scope}.${element.metricName}.${element.metricProperty}.${element._type}`;
```
VWSCoronaDashboard8, line 167: This is not correct from the identifier point of view. The optional property comes last. I don't think it matters much because this is just visual feedback but to keep it consistent with what is visible in the CMS I think it's better to match the order of the id.

You might want to prompt the user for the property last as well. I think that makes sense because it's the only optional part, but also not important

VWSCoronaDashboard8 (changes requested):

packages/cms/src/elements/add.ts
VWSCoronaDashboard1: Indeed, way simpler!

packages/cms/src/elements/add.ts
VWSCoronaDashboard1: I agree, I placed the metric property last in the prompt cycle. And fixed the id generation order.

packages/cms/src/elements/add.ts
```diff
@@ -149,22 +153,31 @@ async function promptForElement(): Promise<Element | undefined> {
...
   return devDocument === null && prodDocument === null;
 }
 
-function elementToString(element: Element) {
+function camelToSnakeCase(camelCased: string) {
+  return camelCased.replace(
+    /[A-Z]/g,
+    (letter: string) => `_${letter.toLowerCase()}`
+  );
+}
```
VWSCoronaDashboard8, line 170: No need to roll our own I think. We already use the change-case module which has a method to convert things to snake case

```
import { snakeCase } from 'change-case';
```

VWSCoronaDashboard8 (approved):

packages/cms/src/elements/add.ts
VWSCoronaDashboard1: Ah, sorry, I missed that one, I'll replace it

PR #3284, opened 2021-07-30
Municipal choropleth bounding box feature filter

VWSCoronaDashboard1:
## Summary

Exactly like the title says ;)
The municipality choropleth 'zooms in' when a safety region is selected. But the entire map is still rendered
in the DOM, resulting in a lot of elements that are not shown on screen, yet still in the DOM.
This PR filters the municipality features down to only those that are visible within the map's boundingbox.
Resulting in a slightly smaller DOM, which is, hopefully, a bit more performant on ancient devices.

The `regionBoundingBoxMunicipalities` lookup was pre-generated by looping through all the region features,
calculating a bounding box polygon for each, slightly enlarging this box, and then for each region bbox
it was calculated which municipality geometry was contained in it.
The code for this is included in the `cli` package.

### Governance

- [ ] Documentation is added


packages/cli/src/scripts/calculate-bounding-box-municipalities.ts
```diff
@@ -0,0 +1,104 @@
...
+
+const gmGeo = topojson.feature(
+  topology,
+  topology.objects.municipalities
+) as unknown as MunicipalGeoJSON;
+
+const vrGeo = topojson.feature(
+  topology,
+  topology.objects.vr_collection
+) as unknown as VrGeoJSON;
```
VWSCoronaDashboard8, line 40: I would place this within the run statement so that errors are handled. Especially the file reading.

VWSCoronaDashboard8 (approved):

packages/cli/src/scripts/calculate-bounding-box-municipalities.ts
VWSCoronaDashboard1: Done

packages/app/src/components/choropleth/region-bounding-box-municipalities.ts
```diff
@@ -0,0 +1,1002 @@
+export const regionBoundingBoxMunicipalities: Record<string, string[]> = {
```
VWSCoronaDashboard10, line 1: Can you add a comment at the top of this file that it is generated code and should not be manually mutated?

VWSCoronaDashboard10 (approved):

PR #3289, opened 2021-07-30
Improve use of page article queries

VWSCoronaDashboard8:
## Summary

Initially I planned to refactor the query for articles so that it would return an array of articles that you could attach to any named property, but this is not possible due to the way page types are set up.

This is an attempt to clean things up within the current constraints, by using shared types and improving the naming of things.


packages/app/src/pages/landelijk/rioolwater.tsx
```diff
@@ -72,6 +70,8 @@ const SewerWater = (props: StaticProps<typeof getStaticProps>) => {
     description: text.metadata.description,
   };
 
+  console.log(content.articles);
```
VWSCoronaDashboard1, line 73: remove console

VWSCoronaDashboard14, line 73: console

VWSCoronaDashboard1 (approved):

PR #3301, opened 2021-08-03
Add ay11 country selector dropdown

VWSCoronaDashboard14:
Took some time to figure out, but I managed to create an ay11 dropdown.
And used these 2 links as inspiration (source of truth) whatever you wanna call it:
- https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html
- https://www.w3.org/TR/wai-aria-practices-1.1/

- Fully functional with your keyboard and "hopefully" all the right aria attributes making it a nice experience for people using the screenreader.
- Renamed the other country selector to `multi-country-selector`

Here is a nice photo of it.
Screenshot 2021-08-04 at 17 53 09


packages/app/src/domain/international/select-country/select-country.tsx
```diff
@@ -0,0 +1,211 @@
...
+      onChange(matchingCountries[index]);
+      // When the user presses "enter" we toggle off the focus state for the element.
+      if (inputRef.current) inputRef.current.blur();
+    },
+    maxPossibleItems: matchingCountries.length,
+    handleOnClose,
+  });
+
+  /**
+   * Everytime the user inputs a new value reset the focus index to the beginning of the results.
```
VWSCoronaDashboard1, line 85:
```suggestion
   * Every time the user inputs a new value reset the focus index to the beginning of the results.
```

packages/app/src/domain/international/select-country/select-country.tsx
```diff
@@ -0,0 +1,211 @@
...
+      />
+
+      <OrderedList
+        id={`${prefixId}_list`}
+        tabIndex={-1}
+        role="listbox"
+        aria-labelledby={prefixId}
+        isOpen={isOpen}
+        /**
+         * Set the currently selected country when opened, otherwise fall back to the currently selected country.
```
VWSCoronaDashboard1, line 115: set the currently selected country, otherwise fall back to...the currently selected country?
This comment is confusing ;)

packages/app/src/domain/international/select-country/select-country.tsx
```diff
@@ -0,0 +1,211 @@
...
+    textDecoration: 'none',
+    color: 'black',
+    width: '100%',
+    bg: x.hasFocus ? 'contextualContent' : 'white',
+    transitionProperty: 'background',
+    transitionDuration: x.hasFocus ? '0ms' : '120ms',
+    border: 'none',
+    textAlign: 'left',
+    fontFamily: 'inherit',
+    fontSize: '1em',
```
VWSCoronaDashboard1, line 204: size should come from the theme. Perhaps add a text variant for this specific instance, or choose one that already has these properties.

packages/app/src/domain/international/select-country/select-country.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+const OrderedList = styled.ol<{ isOpen: boolean }>((x) =>
+  css({
+    display: x.isOpen ? 'block' : 'none',
+    position: 'absolute',
+    top: '100%',
+
+    width: '100%',
+
+    border: `1px solid silver`,
```
VWSCoronaDashboard1, line 175: color should come from the theme

packages/app/src/domain/international/select-country/select-country.tsx
```diff
@@ -0,0 +1,211 @@
...
+    borderTopColor: 'transparent',
+    borderBottomLeftRadius: 1,
+    borderBottomRightRadius: 1,
+
+    backgroundColor: 'white',
```
VWSCoronaDashboard1, line 180: color should come from the theme

packages/app/src/domain/international/select-country/select-country.tsx
```diff
@@ -0,0 +1,211 @@
...
+);
+
+const ListItem = styled.li<{ hasFocus?: boolean }>((x) =>
+  css({
+    px: 3,
+    py: 2,
+    display: 'flex',
+    alignItems: 'center',
+    textDecoration: 'none',
+    color: 'black',
```
VWSCoronaDashboard1, line 196: color should come from the theme

packages/app/src/domain/international/select-country/use-hit-selection.ts
```diff
@@ -0,0 +1,105 @@
+import { useRef, useState } from 'react';
```
VWSCoronaDashboard1, line 1: I see there are already two use-hit-selection hooks in the code base. Is this an exact copy? Then perhaps use the existing one. If not, maybe its an idea to investigate whether it is possible to create a generic hook that can be used for all the current instances.

packages/app/src/utils/modulo.ts
```diff
@@ -0,0 +1,3 @@
+export function modulo(current: number, max: number) {
```
VWSCoronaDashboard1, line 1: I don't think `modulo` quite covers the implementation here, it looks like a specialized use of modulo. Perhaps the name can better reflect this?

VWSCoronaDashboard1 (changes requested):

packages/app/src/domain/international/select-country/select-country.tsx
Denniswegereef: Isn't white in the theme?

VWSCoronaDashboard14: Isn't white coming from the theme?

packages/app/src/domain/international/select-country/use-hit-selection.ts
VWSCoronaDashboard14: This isn't an exact copy hit has some specific behavior for this dropdown. I understand that maybe using a more generic approach could be feasible. But on the other end, I'd like to keep this behavior tied to this component.

packages/app/src/utils/positive-modulo.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Always give back a positive number based on your current and maximum.
+ * For example floorMudulo8, 5) => 3
```
VWSCoronaDashboard1, line 3:
```suggestion
 * For example positiveModulo(8, 5) => 3
```

VWSCoronaDashboard14: Ah yeah, that was my old naming

packages/app/src/domain/international/select-country/select-country.tsx
```diff
@@ -107,14 +112,18 @@ export function SelectCountry({
...
 
       <OrderedList
+        ref={containerRef}
         id={`${prefixId}_list`}
         tabIndex={-1}
         role="listbox"
         aria-labelledby={prefixId}
         isOpen={isOpen}
+        onMouseDown={() => setIsMouseDown(true)}
+        onMouseUp={() => setIsMouseDown(false)}
```
VWSCoronaDashboard1, line 125: Have you tested the following scenario with this approach:
Click on the component, drag your mouse pointer outside of the component and THEN release the mouse button.
I suspect that the state won't be reverted in that case. But perhaps this won't really interfere that much, just something to test before release, I suppose.

VWSCoronaDashboard1 (approved):

PR #3306, opened 2021-08-04
Update font size of dropdown message

VWSCoronaDashboard13:


VWSCoronaDashboard14 (approved):

VWSCoronaDashboard1 (approved):

packages/app/src/domain/international/select-countries/select-country-results.tsx
```diff
@@ -113,6 +113,7 @@ const StyledNoHits = styled.div(
     color: 'gray',
     textAlign: 'center',
     p: 3,
+    fontSize: 1,
```
VWSCoronaDashboard1, line 116: Instead of setting a font size here, its probably better to use a variant on the Text element here.
If a suitable variant doesn't exist yet, add one.

VWSCoronaDashboard1 (changes requested):

VWSCoronaDashboard1 (approved):

PR #3309, opened 2021-08-05
Update Icons in sidebar

VWSCoronaDashboard13:
## Summary

Updates the icons used in the sidebar and most headings so they are more consistently sized.

## Motivation

Messy icon layouts 

## Detailed design

Updated the icons to be fixed heights but variable widths (so they aren't square). 

In the sidebar they are wrapped and centered, in the headings only a margin right is applied

Also updated the headings with categories to use a grid to simplify layout styles

Icons are currently colored purple so they are easy to recognize on the site. Will update them to be black when QA has approved the layout styles

## Drawbacks

This doesn't fix the alignment of all icons across the site


VWSCoronaDashboard14:
I think we put `focusable=false` on every SVG.

VWSCoronaDashboard14 (approved):

packages/app/src/domain/vaccine/vaccine-page-introduction.tsx
```diff
@@ -37,11 +36,30 @@ export function VaccinePageIntroduction({
     <Box spacing={4}>
       <Tile>
         <Box spacing={3}>
-          <HeadingWithIcon
-            icon={<VaccinatieIcon />}
-            title={text.title}
-            headingLevel={1}
-          />
+          <Box
```
VWSCoronaDashboard4, line 39: Should this maybe become the new HeadingWithIcon component? The only other use is in `risiconiveau.tsx` , and maybe that can benefit from these alignments as well?

VWSCoronaDashboard4 (approved):

VWSCoronaDashboard1 (approved):

packages/app/src/domain/vaccine/vaccine-page-introduction.tsx
VWSCoronaDashboard13: I decided not to touch the existing HeadingWithIcon component because it wouldn't make sense to apply this across the board. It looks good on risiconieveau and this alignment was pretty specific

PR #3311, opened 2021-08-05
Create a generic flag component and add border to flags

VWSCoronaDashboard14:
Created a generic international image flag component.
Added a subtle border so flags with a lot of white are easier to spot


packages/app/src/domain/international/international-flag.tsx
```diff
@@ -0,0 +1,27 @@
+import css from '@styled-system/css';
+
+interface InternationalFlagProps {
+  country_code: string;
+  width?: number;
+  height?: number;
+}
+
+export function InternationalFlag({
```
VWSCoronaDashboard4, line 9: Why not

```suggestion
export function Flag({
```

packages/app/src/domain/international/international-flag.tsx
```diff
@@ -0,0 +1,27 @@
+import css from '@styled-system/css';
+
+interface InternationalFlagProps {
+  country_code: string;
```
VWSCoronaDashboard4, line 4: 🐫  -case this one ;-)

VWSCoronaDashboard13 (approved):

PR #3323, opened 2021-08-11
Remove white gaps for a gapped-stacked-area chart

VWSCoronaDashboard14:
When a value is null it can't draw an area and leaves a hole in the area graph. Maybe sometimes it has to be the behavior but for the international variants, it certainly didn't. 

Added a prop `renderNullAsZero ` on the `dataConfig` to enable this behavior.

Without `renderNullAsZero `.
Screenshot 2021-08-11 at 10 57 36

With `setNullToZero`
Screenshot 2021-08-11 at 10 57 45

### DIFF
--- a/packages/app/src/components/time-series-chart/logic/common.ts
+++ b/packages/app/src/components/time-series-chart/logic/common.ts
@@ -1,7 +1,8 @@
 import { TimelineEventConfig } from '../components/timeline';
 
 /**
- * The prop setNullToZero only make sense when rendering a gapped-stacked-area, hence why it only works there.
+ * The prop setNullToZero only make sense when rendering a gapped-stacked-area,
+ * hence why it only works in that particular config .
  */
 export interface DataOptions {
   valueAnnotation?: string;
--- a/packages/app/src/components/time-series-chart/logic/common.ts
+++ b/packages/app/src/components/time-series-chart/logic/common.ts
@@ -1,7 +1,7 @@
 import { TimelineEventConfig } from '../components/timeline';
 
 /**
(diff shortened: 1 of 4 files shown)


packages/app/src/components/time-series-chart/logic/common.ts
```diff
@@ -8,6 +12,7 @@ export interface DataOptions {
   timespanAnnotations?: TimespanAnnotationConfig[];
   timeAnnotations?: TimeAnnotationConfig[];
   timelineEvents?: TimelineEventConfig[];
+  setNullToZero?: boolean;
```
VWSCoronaDashboard1, line 15: I think `renderNullAsZero` is a better term, actually. It describes better what its function is.

packages/app/src/components/time-series-chart/logic/series.ts
```diff
@@ -142,6 +142,7 @@ export interface GappedStackedAreaSeriesDefinition<T extends TimestampedValue>
   strokeWidth?: number;
   isNonInteractive?: boolean;
   mixBlendMode?: Property.MixBlendMode;
+  setNullToZero?: boolean;
```
VWSCoronaDashboard1, line 145: I don't think this is used anymore, right? It was moved to the data options.

VWSCoronaDashboard1 (changes requested):

packages/app/src/components/time-series-chart/logic/series.ts
VWSCoronaDashboard14: Oops yeah, forgot to remove it there.

VWSCoronaDashboard1 (approved):

PR #3326, opened 2021-08-11
Filtering Behavior Chart Select Options

VWSCoronaDashboard13:
## Summary

Filtering VR behavior chart select to only show options that have data to display on the graph

## Motivation

Chart looks broken when there is no data

## Detailed design

Created a `useBehaviorChartOptions` hook in the `BehaviorLineChartTile` file which maps through each of the behavior identifiers to check if there are values for it to return the list of options that are conditionally passed to the `BehaviorLineChartTile` and then to the `SelectBehavior` component

The `SelectBehavior` component uses the defined list of behaviorIdentifiers as the default for it's options when the `options` prop is not provided 

## Alternatives

The options could be filtered within the BehaviorLineChartTile, but since this should only be necessary for VR, I thought it made more sense to pass it from the parent which also decides on the initial selected item

### DIFF
--- a/packages/app/src/domain/behavior/behavior-line-chart-tile.tsx
+++ b/packages/app/src/domain/behavior/behavior-line-chart-tile.tsx
@@ -79,19 +79,17 @@ export function BehaviorLineChartTile({
   );
 }
 
-export function useBehaviorChartOptions(
-  values: NlBehaviorValue[] | VrBehaviorValue[]
-) {
+export function useBehaviorChartOptions<T>(values: T[]) {
   const behaviorLookupKeys = useBehaviorLookupKeys();
 
   return useMemo(() => {
     return behaviorLookupKeys
       .map((x) => {
         const complianceValues = values
-          .map((value) => value[x.complianceKey])
+          .map((value: any) => value[x.complianceKey])
           .filter(isPresent);
         const supportValues = values
-          .map((value) => value[x.supportKey])
+          .map((value: any) => value[x.supportKey])
           .filter(isPresent);
 
         if (complianceValues.length && supportValues.length) {
--- a/packages/app/src/domain/behavior/behavior-line-chart-tile.tsx
+++ b/packages/app/src/domain/behavior/behavior-line-chart-tile.tsx
@@ -1,4 +1,6 @@
 import { NlBehaviorValue, VrBehaviorValue } from '@corona-dashboard/common';
+import { useMemo } from 'react';
+import { isDefined, isPresent } from 'ts-is-present';
 import { Box } from '~/components/base';
 import { ChartTile } from '~/components/chart-tile';
 import { MetadataProps } from '~/components/metadata';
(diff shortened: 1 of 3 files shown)


packages/app/src/domain/behavior/components/select-behavior.tsx
```diff
@@ -9,21 +9,28 @@ import { BehaviorIcon } from './behavior-icon';
...
   value: BehaviorIdentifier;
   onChange: (value: BehaviorIdentifier) => void;
+  options?: BehaviorIdentifier[];
 }
 
-export function SelectBehavior({ value, onChange }: SelectBehaviorProps) {
+export function SelectBehavior({
+  value,
+  onChange,
+  options = behaviorIdentifiers as unknown as BehaviorIdentifier[],
```
VWSCoronaDashboard13, line 18: not ideal. Any other suggestions?

packages/app/src/domain/behavior/behavior-line-chart-tile.tsx
```diff
@@ -69,3 +78,26 @@ export function BehaviorLineChartTile({
...
+export function useBehaviorChartOptions(
+  values: NlBehaviorValue[] | VrBehaviorValue[]
+) {
+  const behaviorLookupKeys = useBehaviorLookupKeys();
+
+  return useMemo(() => {
+    return behaviorLookupKeys
+      .map((x) => {
+        const complianceValues = values
+          .map((value) => value[x.complianceKey])
```
VWSCoronaDashboard13, line 91: image

packages/app/src/domain/behavior/behavior-line-chart-tile.tsx
```diff
@@ -69,3 +78,26 @@ export function BehaviorLineChartTile({
...
   );
 }
+
+export function useBehaviorChartOptions(
+  values: NlBehaviorValue[] | VrBehaviorValue[]
+) {
+  const behaviorLookupKeys = useBehaviorLookupKeys();
+
+  return useMemo(() => {
+    return behaviorLookupKeys
```
VWSCoronaDashboard4, line 88: Instead of a useMemo, this can be calculated as part of the static props; so it's not checking this every time on the client side.

packages/app/src/domain/behavior/behavior-line-chart-tile.tsx
VWSCoronaDashboard4: The NL and VR behavior types are not equal. It can solved by using a generic:

```
useBehaviorChartOptions<T>(values: T[])
```

and calling it in the desired spot using `useBehaviorChartOptions<NlBehaviorValue>(...)`

VWSCoronaDashboard4: Do you really need to loop over all data? The schemas and types for VR simply don't contain some dropdown items. I think the dropdown could be filtered by looking at the types?

VWSCoronaDashboard13: Can we assume that there will be some values if they provide the key for a specific behavior?

VWSCoronaDashboard4: there is an empty state already; eg "Avondklok" on the NL page. The VR dropdown tries to display data that will never be there (because it's not in the schema)?

VWSCoronaDashboard13: I wasn't aware. Updated it to just check that the key is available on the type!

packages/app/src/domain/behavior/behavior-line-chart-tile.tsx
VWSCoronaDashboard13: Moved it to getStaticProps 👍

VWSCoronaDashboard4 (approved):

VWSCoronaDashboard1 (approved):

PR #3328, opened 2021-08-12
Refactor the choropleths into one single component 

VWSCoronaDashboard1:
## Summary

This PR removes the `GmChoropleth`, `VrChoropleth` and `InChoropleth` components in favor of a single (semi) generic component.
It follows the patterns of the timeseries chart a little bit, in that it receives an array of data along with a `dataConfig` and `dataOptions` prop that instruct the component how to deal with the incoming data.
The type of map that will be rendered is indicated by the `map` prop, allowing the values `gm`, `vr` and `in` for, respectively, a municipality, safety region or international map.

The topojsons have been consolidated in the sense that they now only have one feature property: `code`.
The dutch topojson used to have the feature names in the properties as well, but since this data is already available in the application and the international map requires localized country names it made more sense to just get rid of the geojson properties and resolve the names in the application.

The tooltip has been refactored into something a bit more generic, only 2 custom choropleth tooltips remain. The rest can be handled by the new generic one.
To implement the generic tooltip the `replaceVariablesInText` has been augmented a little bit, as it now supports 'commands'. Currently only one command is supported: **plural**.
For example, this placeholder: `{{admissions_on_date_of_reporting, plural, patients}}`, will check if the value of `admissions_on_date_of_reporting` is equal to one and in that case will choose the `patients.singular` value to render, otherwise it renders `patients.plural`. (So it expects the `patients` var to be an object of the shape `{plural: string, singular: string}`).
These extra variables can be passed to the tooltip using `dataOptions.tooltipVariables`, this optional var gets merged with the selected `dataItem` and gets passed to the `replaceVariablesInText` method. The text in the generic tooltip gets rendered as markdown now as well.

All existing instances of the old components have been replaced already with this new version, and so the old components have been completely removed from the code base as well.

### Governance

- [ ] Documentation is added


packages/app/src/components/choropleth/components/path.tsx
```diff
@@ -140,8 +140,8 @@ const StyledHoverPath = styled.path<{
     '&:hover, a:focus &': {
       transitionDuration: '0ms',
       fill: x.fill ?? 'none',
-      stroke: x.stroke ?? '#fff',
-      strokeWidth: x.strokeWidth ?? 3,
+      stroke: x.stroke ?? '#ff0000',
```
VWSCoronaDashboard4, line 143: intended color change?

packages/app/src/components/choropleth/logic/topology.ts
```diff
@@ -1,26 +1,33 @@
-import { GmGeoJSON, InGeoJSON, VrGeoJSON } from '@corona-dashboard/common';
-import { FeatureCollection, MultiPolygon } from 'geojson';
+import { FeatureCollection, MultiPolygon, Polygon } from 'geojson';
 import * as topojson from 'topojson-client';
-import inTopology from './geography-europe-simplified.json';
-// Load all the geographical data including the data entries (regions and municipalities)
-import nlTopology from './geography-simplified.topo.json';
+import inTopology from './in.topo.json';
```
VWSCoronaDashboard4, line 3: Are unused topologies compiled out of the bundles for pages that do not need them? I'd expect the inTopology to not be loaded on eg VR pages.

packages/app/src/components/choropleth/logic/use-choropleth-features.ts
```diff
@@ -0,0 +1,130 @@
...
+import { isDefined } from 'ts-is-present';
+import { MapType, vrBoundingBoxGmCodes } from '~/components/choropleth/logic';
+import { getVrForMunicipalityCode } from '~/utils/get-vr-for-municipality-code';
+import { getVrMunicipalsForMunicipalCode } from '~/utils/get-vr-municipals-for-municipal-code';
+import { CodedGeoJSON, gmGeo, inGeo, nlGeo, vrGeo } from './topology';
+import { ChoroplethDataItem } from './types';
+
+export type FeatureType = keyof ChoroplethFeatures;
+
+const boundingBoxCodes = ['ISL', 'NOR', 'ESP', 'GRC', 'CYP'];
```
VWSCoronaDashboard4, line 12: This hardcoded config  could use a comment to explain it a bit

packages/app/src/components/choropleth/logic/use-choropleth-tooltip.ts
```diff
@@ -0,0 +1,238 @@
...
+      }
+
+      /**
+       * Pass the DOM node to fix positioning in Firefox
+       */
+      const coords = localPoint(ref.current, event);
+
+      if (isPresent(coords)) {
+        setTooltip({
+          left: coords.x + 5,
```
VWSCoronaDashboard4, line 170: Magic offsets?

packages/app/src/components/choropleth/logic/use-feature-props.ts
```diff
@@ -0,0 +1,107 @@
...
+
+type FeaturePropFunctions = {
+  fill: GetFeatureProp;
+  stroke: GetFeatureProp;
+  strokeWidth: GetFeatureProp<number>;
+};
+
+export const DEFAULT_STROKE_WIDTH = 0.5;
+
+export function useFeatureProps<T extends ChoroplethDataItem>(
```
VWSCoronaDashboard4, line 22: A bit of readme for this config would be nice

packages/app/src/components/choropleth/index.tsx
```diff
@@ -0,0 +1,372 @@
...
+  boundingBoxPadding?: OptionalBoundingBoxPadding;
+  /**
+   * This defines an optional set of map heights and paddings based
+   * on a set of given width break points
+   */
+  dynamicSizeConfiguration?: DynamicSizeConfigurations;
+};
+
+/**
+ * This is a (semi) generic choropleth component that supports a dutch map of municipalities or safetyregions
```
VWSCoronaDashboard4, line 120:
```suggestion
 * This is a (semi) generic choropleth component that supports a Dutch map of municipalities or safetyregions
```

packages/app/src/components/choropleth/index.tsx
```diff
@@ -0,0 +1,372 @@
...
+ * and a european map.
```
VWSCoronaDashboard4, line 121:
```suggestion
 * and a European map.
```

packages/app/src/components/choropleth/logic/types.ts
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * Select all the item types of all the properties from the GmCollection with an array type that has a gmcode property
+ */
+export type GmDataItem = GmCollection[KeysOfType<
+  GmCollection,
+  { gmcode: string }[]
+>][number];
+
+/**
+ * Here we map a MapType to a corresponding dataitem type
```
VWSCoronaDashboard4, line 65:
```suggestion
 * Here we map a MapType to a corresponding DataItem type
```

packages/app/src/components/choropleth/logic/use-dynamic-size.ts
```diff
@@ -0,0 +1,74 @@
...
+        return 1;
+      })
+      .find((item) => {
+        if (isDynamicConfiguration(item) && isDefined(containerWidth)) {
+          containerWidth >= item.containerWidth;
+        }
+        return true;
+      });
+
+    //This assert cannot ever trigger, since the Tuple will always return the last item, but the compiler doesn't understand this...
```
VWSCoronaDashboard4, line 42:
```suggestion
    /**
     * This assert cannot ever trigger, since the Tuple will always return the last item, but the compiler doesn't understand this...
     */
```

VWSCoronaDashboard4 (approved):
Looking good! I like the improved API of the new `<Chropleth>`

packages/app/src/components/choropleth/components/path.tsx
VWSCoronaDashboard1: Nope, that was used for testing, I'll revert

packages/app/src/components/choropleth/logic/topology.ts
VWSCoronaDashboard1: No idea, but this is how it already was implemented in the previous incarnation. I think there's also an idea to load the topojson using SWR anyway, so I'd like to move that to a different PR.

packages/app/src/components/choropleth/logic/use-choropleth-features.ts
VWSCoronaDashboard1: This was already in the existing international map, but it makes sense, I'll add some comments.

packages/app/src/components/choropleth/logic/use-choropleth-tooltip.ts
VWSCoronaDashboard1: Was part of the existing implementation

packages/app/src/components/choropleth/logic/use-feature-props.ts
VWSCoronaDashboard1: Done

packages/app/src/components/choropleth/components/mercator-hover-group.tsx
```diff
@@ -0,0 +1,74 @@
...
+  return (
+    <Projection projection={projection} data={data} fitExtent={fitExtent}>
+      {({ features }) => (
+        <g>
+          {features
+            .filter(featureHasPath)
+            .map((x) => ({
+              ...x,
+              path: x.path.replace(
+                /\d+\.\d+/g,
```
VWSCoronaDashboard14, line 48: Maybe a small comment why?

VWSCoronaDashboard14 (approved):

packages/app/src/components/choropleth/components/mercator-hover-group.tsx
VWSCoronaDashboard1: Turned this into a separate utils function with appropriate commentary

PR #3335, opened 2021-08-13
Small typography tweaks

VWSCoronaDashboard4:
## Summary

Small typography tweaks

* Mostly adjustments to font size and font weight

### DIFF
--- a/packages/app/src/style/global-style.tsx
+++ b/packages/app/src/style/global-style.tsx
@@ -11,7 +11,7 @@ ${tags} {
     padding: 0;
     border: 0;
     font-size: 100%;
-    font: inherit;
+    font-family: inherit;
     vertical-align: baseline;
   }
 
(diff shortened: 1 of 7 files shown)


packages/app/src/style/global-style.tsx
```diff
@@ -11,7 +11,7 @@ ${tags} {
     padding: 0;
     border: 0;
     font-size: 100%;
-    font: inherit;
+    font-family: inherit;
```
VWSCoronaDashboard4, line 14: I think this was the intend of the original code. Default styling for font-weight etc on elements such as `strong` are back.

VWSCoronaDashboard1 (approved):

packages/app/src/style/global-style.tsx
VWSCoronaDashboard1: Makes sense I suppose

PR #3342, opened 2021-08-16
Add inline tooltip when there are gaps in the line chart behavior page

VWSCoronaDashboard14:
Add a function that trims the left and right `null` values and then checks if there are still `null` values. Based on that render a `InlineTooltip`. 

Screenshot 2021-08-16 at 12 47 40

### DIFF
--- a/packages/app/src/domain/behavior/behavior-line-chart-tile.tsx
+++ b/packages/app/src/domain/behavior/behavior-line-chart-tile.tsx
@@ -2,7 +2,7 @@ import { NlBehaviorValue, VrBehaviorValue } from '@corona-dashboard/common';
 import { dropRightWhile, dropWhile } from 'lodash';
 import { useMemo } from 'react';
 import { isDefined, isPresent } from 'ts-is-present';
-import { Box, Spacer } from '~/components/base';
+import { Box } from '~/components/base';
 import { ChartTile } from '~/components/chart-tile';
 import { InlineTooltip } from '~/components/inline-tooltip';
 import { MetadataProps } from '~/components/metadata';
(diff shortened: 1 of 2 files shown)


packages/app/src/domain/behavior/behavior-line-chart-tile.tsx
```diff
@@ -91,3 +119,22 @@ export function getBehaviorChartOptions<T>(value: T) {
...
+  key: keyof NlBehaviorValue
+) {
+  return useMemo(() => {
+    const trimmedLeftValues = dropWhile(values, (i) => i[key] === null);
+    const trimmedLeftAndRightValues = dropRightWhile(
+      trimmedLeftValues,
+      (i) => i[key] === null
+    );
+
+    return trimmedLeftAndRightValues.some((i) => i[key] === null);
```
VWSCoronaDashboard1, line 138: nit: we have a nice `isPresent` util function for this ;)

packages/app/src/domain/behavior/behavior-line-chart-tile.tsx
```diff
@@ -91,3 +119,22 @@ export function getBehaviorChartOptions<T>(value: T) {
     })
     .filter(isDefined);
 }
+
+/**
+ * Trimm all the null values on the left and the right side of the array.
+ * If there are still null values left we can assume there is a gap in the data.
+ */
+export function useTrimValuesAndFindGap(
```
VWSCoronaDashboard1, line 127: The fact that this hook trims the leading and trailing nulls is an implementation detail that doesn't need to be reflected in the name. I think `useDataHasGaps` suffices.

VWSCoronaDashboard1 (approved):

packages/app/src/domain/behavior/behavior-line-chart-tile.tsx
VWSCoronaDashboard14: Beautiful

PR #3371, opened 2021-08-19
Fix full width image article detail

VWSCoronaDashboard14:
Apparently, by a merge conflict, this got accidentally overwritten, added back.

Screenshot 2021-08-19 at 11 17 42

### DIFF
--- a/packages/app/src/components/article-detail.tsx
+++ b/packages/app/src/components/article-detail.tsx
@@ -50,7 +50,6 @@ export function ArticleDetail({ article }: ArticleDetailProps) {
       </ContentBlock>
       {!!article.content?.length && (
         <Box
-          fontSize="1.125rem"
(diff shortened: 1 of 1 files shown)


packages/app/src/components/article-detail.tsx
```diff
@@ -49,14 +49,20 @@ export function ArticleDetail({ article }: ArticleDetailProps) {
...
-        <ContentBlock>
-          <Box textVariant="body1">
-            <RichContent
-              blocks={article.content}
-              contentWrapper={ContentBlock}
-            />
-          </Box>
-        </ContentBlock>
+        <Box
+          fontSize="1.125rem"
```
VWSCoronaDashboard1, line 53: There isn't a text variant with this size? If not, can you add it?

VWSCoronaDashboard1 (changes requested):

PR #3389, opened 2021-08-24
Update vaccine coverage over time and schema change

VWSCoronaDashboard14:
Added mock data
Moved the component to the 1 folder
Changed the schema so it's in line with what we are getting back from the back-end.
Added birth years with some new Lokalise keys

Screenshot 2021-08-25 at 10 43 20


VWSCoronaDashboard4 (approved):

packages/app/src/domain/vaccine/vaccine-coverage-per-age-group.tsx
```diff
@@ -17,17 +17,17 @@ type Props = {
...
+  '18+',
+  '12-17',
+  '18-30',
+  '31-40',
+  '41-50',
+  '51-60',
+  '61-70',
+  '71-80',
+  '81+',
+].reverse();
```
VWSCoronaDashboard17, line 30: `.reverse()` sorts arrays in place instead of making a copy first. If that is what you want, all good. Otherwise, you should probably first make a copy, eg. `[...SORTING_ORDER].reverse()`

VWSCoronaDashboard14: Sorted it

VWSCoronaDashboard1 (approved):

PR #3401, opened 2021-08-25
Add higher lower and date in reproduction kpi tile

VWSCoronaDashboard14:
Removed the `staticTimespan` since it doesn't get used in the app.
Added an option to show `the old_date_unix` in the text below the KPI bar chart.

### DIFF
--- a/packages/app/src/components/difference-indicator/tile-difference.tsx
+++ b/packages/app/src/components/difference-indicator/tile-difference.tsx
@@ -9,17 +9,18 @@ import { Container, IconContainer } from './containers';
 export function TileDifference({
   value,
   isDecimal,
-  staticTimespan,
   maximumFractionDigits,
   isPercentage,
+  showOldDateUnix,
 }: {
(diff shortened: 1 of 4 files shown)


packages/app/src/components/difference-indicator/tile-difference.tsx
```diff
@@ -31,10 +32,14 @@ export function TileDifference({
...
-  const timespanTextNode = staticTimespan ?? text.vorige_waarde;
+  const timespanTextNode = showOldDateUnix
+    ? formatDateFromSeconds(value.old_date_unix)
+    : text.vorige_waarde;
 
   if (difference > 0) {
-    const splitText = text.toename.split(' ');
+    const splitText = showOldDateUnix
+      ? text.hoger.split(' ')
+      : text.toename.split(' ');
```
VWSCoronaDashboard16, line 42: Why does it depend on showOldDate if we say higher than or more than? Isn't that more a grammar thing based on what kind of value we are talking about?

PR #3414, opened 2021-08-30
PoC: Create generic page type and add conditional fields to Sanity

VWSCoronaDashboard:
## Summary

Trying to reduce the amount of unique schema's we have in Sanity

## Motivation

- Easier to add new pages
- Easier to remove pages
- Easier to group by schema type in Sanity and create an organized CMS
- In the future... add localizations per page?

## Detailed design

Coming soon...


packages/app/src/pages/landelijk/vaccinaties.tsx
```diff
@@ -93,6 +93,8 @@ const VaccinationPage = (props: StaticProps<typeof getStaticProps>) => {
     description: text.metadata.description,
   };
 
+  console.log(content);
```
VWSCoronaDashboard1, line 96: Remove console

packages/cms/src/schemas/documents/pages/article-page-articles.ts
```diff
@@ -0,0 +1,41 @@
...
+
+      const articles = [article0, article1, article2].filter(Boolean);
+      const title = articles.length > 0 ? `${articles.join(', ')}` : '';
+      const subtitle =
+        articles.length > 0 ? `${articles.length} artikelen` : 'Geen artikelen';
+
+      return {
+        title,
+        subtitle,
+      };
```
VWSCoronaDashboard1, line 38: I don't really grasp what's going on here. The `prepare` creates a title and subtitle based off, what I understand, three articles, yet in the `select` 4 articles are used?

packages/cms/src/schemas/documents/pages/article-page-articles.ts
```diff
@@ -0,0 +1,41 @@
+import { HIGHLIGHTED_ARTICLES } from '../../fields/highlighted-articles';
+import { GrArticle } from 'react-icons/gr';
+
+export const pageArticles = {
+  title: 'Page Articles',
+  name: 'articlePageArticle',
```
VWSCoronaDashboard1, line 6: nit: keep the export and document the same. I also think that perhaps `pageArticles` is a better name for the document, instead of `articlePageArticle` which is a bit convoluted. If we prefix all of the documents that have a pageIdentifier with `page` I think we can keep a consistent naming scheme.

packages/cms/src/schemas/documents/pages/article-page-links.ts
```diff
@@ -0,0 +1,47 @@
...
+
+      const links = [link0, link1, link2].filter(Boolean);
+      const title = links.length > 0 ? `${links.join(', ')}` : '';
+      const subtitle =
+        links.length > 0 ? `${links.length} links` : 'Geen links';
+
+      return {
+        title,
+        subtitle,
+      };
```
VWSCoronaDashboard1, line 44: Same comment as on the page articles, I'm not following the intent of this code. (I'm not very smart, so that might be an issue...)

packages/cms/src/schemas/documents/pages/article-page.ts
```diff
@@ -0,0 +1,18 @@
+export const page = {
+  title: 'Page',
+  name: 'articlePage',
```
VWSCoronaDashboard1, line 3: I'd call this `pageIdentifier`, I think it a little clearer for the intent of this document.

VWSCoronaDashboard1 (changes requested):
This is a pretty cool start. As for this comment: `In the future... add localizations per page`
I think if we use the pageidentifier key as the root key for all the lokalize keys that belong to a page, then it should already be possible to create apge that show a subselection of lokalize keys, no?

VWSCoronaDashboard7:
We're ended up going with a slightly different route so this became obsolete. Closing for that reason.

PR #3427, opened 2021-08-31
Add country -> color mapping in MultiSelectCountries

VWSCoronaDashboard17:
## Summary

Adds a country to color mapping in the `MultiSelectCountries` component.

The mapping hook makes sure that colors are persistently mapped to country codes until the country code is no longer used. It will respect ordering of colors and always use the first color available.

I haven't taken color array rollover into account currently as the maximum number of countries selected is the same as the array length of `ORDERED_COLORS`. It would be doable to add although it might be good to discuss how to handle that case (when more countries are selected than colors are available):

- do we rollover to the start (eg. start with the first color again)?
- do we throw an error?
- do we rollover and darken/lighten the color?


VWSCoronaDashboard4 (approved):
Looks neat! The rollover does not seem necessary indeed, as the amount of colors matches the limit. If ever something else is needed, the hook can be extended/abstracted/etc :-)

VWSCoronaDashboard17:
Added tests

VWSCoronaDashboard14 (approved):

PR #3442, opened 2021-09-01
Set maximum number of date ticks to number of full days in time domain of 

VWSCoronaDashboard16:
## Summary

If a time series only spans 2-3 days, we still show 5 date ticks on desktop, which does not make sense. This PR sets the maximum to the number of full days in the time series, so we don't get double date ticks:

Old:
image

New:
image

### DIFF
--- a/packages/app/src/components/time-series-chart/components/axes.tsx
+++ b/packages/app/src/components/time-series-chart/components/axes.tsx
@@ -124,7 +124,12 @@ export const Axes = memo(function Axes({
     [formatPercentage]
   );
 
-  const xTickNumber = breakpoints.sm ? (timeframe === 'all' ? 6 : 5) : 3;
+  const preferredDateTicks = breakpoints.sm ? (timeframe === 'all' ? 6 : 5) : 3;
+  const fullDaysInDomain = Math.floor((endUnix - startUnix) / 86400);
+  const xTickNumber = Math.max(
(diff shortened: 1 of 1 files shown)


packages/app/src/components/time-series-chart/components/axes.tsx
```diff
@@ -124,7 +124,12 @@ export const Axes = memo(function Axes({
     [formatPercentage]
   );
 
-  const xTickNumber = breakpoints.sm ? (timeframe === 'all' ? 6 : 5) : 3;
+  const preferredDateTicks = breakpoints.sm ? (timeframe === 'all' ? 6 : 5) : 3;
```
VWSCoronaDashboard17, line 127: Should this ternary be the other way around? eg. 3 for breakpoints.sm and 6/5 for other breakpoints?

VWSCoronaDashboard16: `breakpoints.sm` being true means screen size is _bigger_ than ~768px. The actual mq used is `screen and (min-width: ${string})`

VWSCoronaDashboard17: Right, that's somewhat confusing but I guess it's in part due to how `styled-system` value arrays work.

VWSCoronaDashboard17 (approved):

PR #3445, opened 2021-09-01
Reduce number of ticks on narrow time series

VWSCoronaDashboard16:
## Summary

The number of date ticks in a time series was determined based on the screen width. However, sometimes time series are displayed narrow on desktop screens as well, and you get the same spacing issues. This PR makes it possible to force a certain number of ticks in those cases, and applies it to the narrow time series we currently have.

### DIFF
--- a/packages/app/src/components/time-series-chart/components/axes.tsx
+++ b/packages/app/src/components/time-series-chart/components/axes.tsx
@@ -16,8 +16,8 @@ import { memo, Ref, useCallback } from 'react';
 import { useIntl } from '~/intl';
 import { colors } from '~/style/theme';
 import { createDate } from '~/utils/create-date';
-import { useBreakpoints } from '~/utils/use-breakpoints';
 import { useIsMounted } from '~/utils/use-is-mounted';
+import { useResizeObserver } from '~/utils/use-resize-observer';
 import { Bounds } from '../logic';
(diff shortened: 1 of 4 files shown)


VWSCoronaDashboard1 (approved):

packages/app/src/components/time-series-chart/components/axes.tsx
```diff
@@ -124,12 +126,16 @@ export const Axes = memo(function Axes({
     [formatPercentage]
   );
 
-  const preferredDateTicks = breakpoints.sm ? (timeframe === 'all' ? 6 : 5) : 3;
-  const fullDaysInDomain = Math.floor((endUnix - startUnix) / 86400);
-  const xTickNumber = Math.max(
-    Math.min(fullDaysInDomain, preferredDateTicks),
-    2
-  );
+  if (!xTickNumber) {
```
VWSCoronaDashboard17, line 129: This can be `0` which is falsey, so `isPresent()` might be better here

VWSCoronaDashboard17 (approved):
Beyond a nit, looks good.

VWSCoronaDashboard14 (approved):

packages/app/src/components/time-series-chart/components/axes.tsx
WDeenik: True

PR #3453, opened 2021-09-01
Fix feature flag schema support

VWSCoronaDashboard1:
## Summary

Add support to the json validator for the feature flags.
In `create-validation-function.ts` the feature flags are now loaded as well and right before a schema
gets parsed the specified metric names as found in `features.ts` are removed from the `required`
lists.
That way the validator will consider these metrics optional, yet during the `generate-typescript` phase
they are required. This way we can disable a feature while at the same keeping the compiler happy
without having to add all sorts of undefined checks everywhere in our code.

## Motivation

Better DX, hopefully

### Governance

- [ ] Documentation is added


packages/cli/src/schema/create-validate-function.ts
```diff
@@ -1,36 +1,82 @@
...
+      disableFeatureFlagMetrics(schema, disabledMetrics);
+    }
+    return schema;
   } catch (e) {
     if (e instanceof Error)
       throw new Error(`Error while parsing file ${schemaPath}:\n${e.message}`);
   }
 }
 
+function disableFeatureFlagMetrics(schema: any, features: Feature[]) {
```
VWSCoronaDashboard17, line 30: Should it maybe check if the feature flag `isEnabled`?

VWSCoronaDashboard1: the features are already filtered at the top of the page where they are imported:
`const disabledMetrics = features.filter((x) => !x.isEnabled);`
So I only disable the metrics for the features that have `isEnabled` set to false.

VWSCoronaDashboard1:
> Would it be a good idea to also merge in the changes in #3440 here as that contains an upgrade to `ajv`?

no, I'm not going to merge in a major version upgrade right before a release ;)

VWSCoronaDashboard17:
Yep, that makes sense now that I know this needs to go into master.

PR #3457, opened 2021-09-02
Only use smooth scroll for sidebar links, use instant scroll for all others

VWSCoronaDashboard16:
## Summary

Only for sidebar links the smooth scroll actually makes sense, since the whole sidebar will stay visible. For all other page transitions the whole page will be replaced, which makes an instant scroll-to-top more logical. The only exception on this (I found while building this) are hash links: they will stay on the same page, or scroll to a specific position on a new page. In both cases a smooth scroll will help the user have some sense of where they are on the page.

## Design + Drawbacks

It took me a while to find the best way of doing it, because it seems like a really clean solution is not really possible. There are 3 main reasons for that:

1. We cannot detect through a route change event which link triggered the route change. Path is not reliable, since there are also links from non-sidebar pages to sidebar pages, and in those cases we don't want a smooth scroll. This makes that we have to fix this through some `onClick` behaviour on the links themselves.
2. Next links do not have something which lets you add your own `onClick` behaviour. I tried adjusting the `onClick` of the link's child (that's how the Next link itself makes it work without adding a wrapper element), but there's _a lot_ of edge cases you need to consider to do that properly, so I quickly abandoned that approach. Resolved to adding a wrapper element for our own `onClick` behaviour anyway. It seems like it does not impact styling of links on the site, but I don't know if there are links somewhere I didn't look at that are susceptible to this.
3. Because hash links still need a smooth scroll, we need to keep the smooth scroll on there through CSS. Since the options for `scrollTo()` behaviour are only 'smooth' or 'auto', we cannot use that to force an instant scroll once the smooth scroll is set through CSS. The only way of doing that is to temporarily overwrite that style. Not that nice, but it works.

If anyone sees another (cleaner) solution of the above points, please say so!

### DIFF
--- a/packages/app/src/components/aside/menu.tsx
+++ b/packages/app/src/components/aside/menu.tsx
@@ -4,14 +4,13 @@ import { NextRouter, useRouter } from 'next/router';
 import { ReactNode } from 'react';
 import styled from 'styled-components';
 import { UrlObject } from 'url';
+import chevronUrl from '~/assets/chevron.svg';
 import { Box } from '~/components/base';
 import { Anchor, Text } from '~/components/typography';
 import { SpaceValue } from '~/style/theme';
 import { asResponsiveArray } from '~/style/utils';
 import { Link } from '~/utils/link';
 import { AsideTitle } from './title';
-
-import chevronUrl from '~/assets/chevron.svg';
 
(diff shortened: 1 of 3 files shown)


VWSCoronaDashboard16:
This approach introduces too many problems.

PR #3469, opened 2021-09-03
Fix spacing on vr current risk level

VWSCoronaDashboard16:
## Summary

Fix spacing on vr current risk level

### DIFF
--- a/packages/app/src/pages/veiligheidsregio/[code]/risiconiveau.tsx
+++ b/packages/app/src/pages/veiligheidsregio/[code]/risiconiveau.tsx
@@ -130,8 +130,13 @@ const RegionalRestrictions = (props: StaticProps<typeof getStaticProps>) => {
           />
 
           <ChartTile title={text.current_escalation_level} disableFullscreen>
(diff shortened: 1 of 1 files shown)


VWSCoronaDashboard1 (approved):

packages/app/src/pages/veiligheidsregio/[code]/risiconiveau.tsx
```diff
@@ -180,6 +177,7 @@ const RegionalRestrictions = (props: StaticProps<typeof getStaticProps>) => {
                       width={{ _: '100%', lg: '50%' }}
                       pl={{ _: 0, lg: 3 }}
                       mb={3}
+                      mt={{ _: 3, lg: 0 }}
```
Comment (VWSCoronaDashboard14) [inline packages/app/src/pages/veiligheidsregio/[code]/risiconiveau.tsx:180]: It's better to remove both margins and add a spacing property on the box element where the flex-direction is set.

VWSCoronaDashboard14 (commented):
It's better to remove both margins and add a spacing property on the box element where the flex-direction is set.

VWSCoronaDashboard14 (approved):

PR #3487, opened 2021-09-07
Add warning tooltips to mini trends

VWSCoronaDashboard16:
## Summary

Adds the possibility of adding warning labels & tooltips to mini trend tiles through Elements in Sanity.

Also refactored the mini trend tile to accept custom charts, so it can be used for the vaccines tile as well. Deduplicates the tile template that was basically a copy/paste for the vaccines tile.

Had to fix the script for adding elements because it used an invalid path on my mac.


packages/app/src/domain/topical/mini-trend-tile.tsx
```diff
@@ -73,24 +85,28 @@ export function MiniTrendTile<T extends TimestampedValue>(
...
+                seriesConfig={[
+                  {
+                    metricProperty,
+                    type: 'area',
+                    label: title,
+                    color: colors.data.primary,
+                  },
+                ]}
+              />
+            )}
```
VWSCoronaDashboard1, line 109: I think this solution is a bit messy. Either we create two components: One that receives a chart instance and one that receives chart data. Or we can ditch the tile that receives chart data and refactor everything into a tile that receives a chart instance and just create a specialized mini-timeseries-chart that contains all the defaults written here.

Having one component support both scenario's is going to result in messy code, imho. Because in the case you assign a `chart` prop, you can still assign `trendData`, from just the API of the component it's never really clear that the chart prop will have precedence, etc.

Let's try and have components keep one focus and one responsibility.

VWSCoronaDashboard1 (changes requested):

packages/app/src/domain/topical/mini-trend-tile.tsx
VWSCoronaDashboard16: Good point. Refactored it back now to include a mini trend chart in the template of the tile itself. That chart can also be used for the vaccines now with just 1 extra property for the timeframe used.

VWSCoronaDashboard1 (approved):

PR #3490, opened 2021-09-07
Create mandatory amount prop for difference tiles  & fix difference labels everywhere

VWSCoronaDashboard16:
## Summary

Made average difference tiles support higher/lower as well as more/less, then went through the site setting all difference labels (higher/lower vs. more/less) correctly.
Also made it mandatory to tell a difference tile which one it should be to prevent issues like this in the future.

FYI currently only the reproduction number & percentage of positive tests are values which are _not_ amounts (as in: the data is not talking about countable things).

### DIFF
--- a/packages/app/src/components/difference-indicator/tile-average-difference.tsx
+++ b/packages/app/src/components/difference-indicator/tile-average-difference.tsx
@@ -1,7 +1,5 @@
 import { DifferenceDecimal, DifferenceInteger } from '@corona-dashboard/common';
-import { Gelijk } from '@corona-dashboard/icons';
-import { Up } from '@corona-dashboard/icons';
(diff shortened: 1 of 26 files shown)


VWSCoronaDashboard1 (approved):
Maybe you should have made isAmount=true the default, it would have saved you adding the prop in a LOT of places ;)

VWSCoronaDashboard16:
My reasoning was that if there's no default, we always have to tell the tile which one it needs to be. It's easy to overlook that the grammar is not completely correct when adding new KPI values / bar scales and the code doesn't have context on what the data is about, so it needs to be told explicitly which one to use.

Took the (small) pain of adding them all now, so we don't have to correct wrong grammar after release anymore in the future.

VWSCoronaDashboard14 (approved):

PR #3491, opened 2021-09-08
Clean up tile difference

VWSCoronaDashboard14:
The tile difference is now a little bit more cleaner it has some new Lokalise keys that support markdown. 
So when they want to have a bold text they can just use `**{{amount}} hoger**`. 
Added a new prop `rendererOptions` to the markdown renderer to support extra renderer options. In this case it was needed because I didn't want to render the text as a paragraph and wanted to replace `bold` with a regular span.

For the end-user almost nothing has been changed. If I iron out the last feedback points I'll start also changing the moving average tile difference to be in line with the same structure as here.

### DIFF
--- a/packages/app/src/components/difference-indicator/tile-difference.tsx
+++ b/packages/app/src/components/difference-indicator/tile-difference.tsx
@@ -32,24 +32,28 @@ export function TileDifference({
   );
 
   let content;
-  let icon;
+  let containerWithIcon;
 
   if (difference > 0) {
     content = hasHigherLowerText ? text.waarde_hoger : text.waarde_meer;
 
-    icon = <ContainerWithIcon icon={<Up />} color={'red'} />;
+    containerWithIcon = <ContainerWithIcon icon={<Up />} color={'red'} />;
   }
 
   if (difference < 0) {
     content = hasHigherLowerText ? text.waarde_lager : text.waarde_minder;
 
-    icon = <ContainerWithIcon icon={<Down />} color={'data.primary'} />;
+    containerWithIcon = (
+      <ContainerWithIcon icon={<Down />} color={'data.primary'} />
+    );
   }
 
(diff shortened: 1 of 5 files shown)


packages/app/src/components/markdown.tsx
```diff
@@ -10,9 +10,10 @@ import { Link } from '~/utils/link';
 import { DisplayOnMatchingQueryCode } from './display-on-matching-query-code';
 import { Message } from './message';
 import { Anchor } from './typography';
-
+import { ElementType } from 'react';
 interface MarkdownProps {
   content: string;
+  renderersOptions?: { [nodeType: string]: ElementType };
```
VWSCoronaDashboard1, line 16:
```suggestion
  rendererOverrides?: { [nodeType: string]: ElementType };
```
Since you're overriding renderers, I don't think the term `options` applies here.

VWSCoronaDashboard1 (approved):

packages/app/src/components/difference-indicator/tile-difference.tsx
```diff
@@ -1,90 +1,98 @@
...
-    ? formatDateFromSeconds(value.old_date_unix)
-    : text.vorige_waarde;
+  let content;
+  let containerWithIcon;
 
   if (difference > 0) {
-    const splitText = isAmount
-      ? text.toename.split(' ')
-      : text.hoger.split(' ');
+    content = isAmount ? text.waarde_hoger : text.waarde_meer;
```
VWSCoronaDashboard16, line 38: This is exactly the wrong way around now.

VWSCoronaDashboard14: I'll fix it in https://github.com/minvws/nl-covid19-data-dashboard/pull/3492

PR #3492, opened 2021-09-08
Update moving average tile difference in line with the other difference

VWSCoronaDashboard14:
Same structure as the `tileDifference` that got merged in this morning to be a bit more consistent. 
Added some new Lokalise keys e.g. `"**{{amount}} lager** dan het gemiddelde van de afgelopen 7 dagen **({{totalAverage}})**` and deleted the old ones to avoid confusion.

### DIFF
--- a/packages/app/src/components/difference-indicator/tile-average-difference.tsx
+++ b/packages/app/src/components/difference-indicator/tile-average-difference.tsx
@@ -1,73 +1,93 @@
 import { DifferenceDecimal, DifferenceInteger } from '@corona-dashboard/common';
 import { Down, Gelijk, Up } from '@corona-dashboard/icons';
 import { InlineText } from '~/components/typography';
+import css from '@styled-system/css';
 import { useIntl } from '~/intl';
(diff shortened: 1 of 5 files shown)


packages/app/src/components/difference-indicator/tile-average-difference.tsx
```diff
@@ -1,73 +1,93 @@
...
-        </InlineText>
-        <InlineText>
-          {text.zeven_daags_gemiddelde}
-          {oldValue}
-        </InlineText>
-      </Container>
+  if (difference > 0) {
+    content = isAmount
+      ? text.zeven_daags_gemiddelde_waarde_hoger
+      : text.zeven_daags_gemiddelde_waarde_meer;
```
VWSCoronaDashboard16, line 36: This should be exactly the other way around

VWSCoronaDashboard16 (changes requested):
Now we're using higher/lower and more/less the wrong way around

VWSCoronaDashboard16 (approved):

PR #3493, opened 2021-09-08
Handle cases where there's no data for vaccine coverage

VWSCoronaDashboard16:
## Summary

When there's no data for vaccination coverage of a certain municipality, we need to show '-' on places where those numbers are normally shown, and the choropleth should color that municipality white.


packages/app/src/domain/vaccine/vaccine-coverage-choropleth-per-gm.tsx
```diff
@@ -200,6 +201,8 @@ export function ChoroplethTooltip<T extends ChoroplethDataItem>(
     tooltipValue = replaceVariablesInText(tooltipValueContent, {
       value: formatPercentage(value.value) + '%',
     });
+  } else if (value === null) {
+    tooltipValue = '-';
```
VWSCoronaDashboard14, line 205: This should be an en dash `–`

VWSCoronaDashboard14 (approved):

packages/app/src/domain/vaccine/vaccine-page-introduction-vr-gm.tsx
```diff
@@ -63,10 +63,10 @@ export function VaccinePageIntroductionVrGm({
                       )
                 }
               />
+            ) : isPresent(data.fully_vaccinated_percentage) ? (
+              <KpiValue percentage={data.fully_vaccinated_percentage} />
             ) : (
-              <KpiValue
-                percentage={data.fully_vaccinated_percentage as number}
-              />
+              <KpiValue text="-" />
```
VWSCoronaDashboard1, line 69: Can this check be done inside KpiValue? I think it's safe to say that a dash needs to be shown whenever the value is null, so instead of doing this check wherever we use KpiValue, perhaps just move the logic inside the component?

VWSCoronaDashboard1 (approved):

packages/app/src/domain/vaccine/vaccine-page-introduction-vr-gm.tsx
VWSCoronaDashboard16: Was actually considering that but decided against it because this is the first and only exception for it and it seemed a bit weird to consider the situation of not getting a value inside KpiValue, since well, it's supposed to be used as a _value_.

But since I was already in doubt and you mentioning it as well, I'll move it there 👍

PR #3494, opened 2021-09-08
Make the logo RO clickable 

VWSCoronaDashboard14:
Had to do something funky with a clip-path to make not the full SVG clickable since it has some empty spacing on the right (which is weird if you can click on an anchor with nothing).
Not sure if it was intentional (since it's already been like this since the beginning of this project) to have that in the SVG. We could either export a new SVG without that or leave this as is.


VWSCoronaDashboard1 (approved):

VWSCoronaDashboard16 (approved):
Seeing that we have to do other weird stuff to get the logo positioned right it might be good to look at the SVGs some time, but this works.

PR #3495, opened 2021-09-08
Implement feature flag data defaults

VWSCoronaDashboard1:
## Summary

This PR adds a method called `initializeFeatureFlaggedData` which intercepts the `getStaticProps` phase when json data is being loaded. It then looks for disabled feature flags and initializes default data on the incoming json data according to the given schema.
This PR does assume that a schema already exists for a feature, since it doesn't really make sense to start merging stuff into develop when the schema isn't even clear yet.
This solution prevents loads of undefined checks in potential data shaping functions that are called during `getStaticProps`.

### Governance

- [ ] Documentation is added


VWSCoronaDashboard14 (approved):

packages/app/src/static-props/feature-flags/initialize-feature-flagged-data.ts
```diff
@@ -0,0 +1,208 @@
...
+      }
+      return [x, initializeProperty(schema.properties[x], schema)];
+    })
+  );
+}
+
+function initializeSimpleProp(type: SimpleMetricType) {
+  switch (type) {
+    case 'boolean':
+      return true;
```
VWSCoronaDashboard16, line 135: I would expect everything to have the default state of an uninitialised variable, and I would expect an uninitialised boolean to be `false`. Why `true`?

VWSCoronaDashboard16 (approved):
Nice! Some suggestions on improving the tests and 1 question, but generally this looks 👌

packages/app/src/static-props/feature-flags/initialize-feature-flagged-data.ts
VWSCoronaDashboard1: This was easier to test. The actual value will never be used anyway.

packages/app/src/static-props/feature-flags/__tests__/initialize-feature-flagged-data.spec.ts
```diff
@@ -77,8 +79,7 @@ describe('initializeFeatureFlaggedData', () => {
 
     initializeFeatureFlaggedData(gmCollection, 'gm_collection');
 
-    expect(gmCollection.testCollection).toBeDefined();
-    expect(gmCollection.testCollection.length).toEqual(2);
+    expect(gmCollection.testCollection).toEqual(initialValues);
```
VWSCoronaDashboard16, line 82: This will always be true since the array instance in the `gmCollection` object is the same as the `initialValues` array, and the values are initialized in-place. We need to put a copy of the initialValues in the `gmCollection`.

VWSCoronaDashboard1: > This will always be true since the array instance in the `gmCollection` object is the same as the `initialValues` array, and the values are initialized in-place. We need to put a copy of the initialValues in the `gmCollection`.

You're right, I fixed and merged it here: https://github.com/minvws/nl-covid19-data-dashboard/pull/3502
I now check a copy of the data.

PR #3505, opened 2021-09-13
Update vaccination per age group table

VWSCoronaDashboard14:
Refactor the vaccine per age group table
- Unfortunately, how it was set up it was impossible to keep the old code so had to rewrite everything.
Created 2 different layouts for mobile and narrow to keep it simple without doing so many checks on breakpoints.
- It has support for `95% or more` what can be used on VR/GM level.
- Changed a bit of the `PercentageBar` what can now accept a background and a diagonal hatched pattern if it is ever needed. This was needed when something is 95% or higher the last 5% is a pattern. See screenshot:
Screenshot 2021-09-15 at 10 37 14

Note: in some weird way it set double quotes in the schema back to single quotes?


packages/app/src/domain/vaccine/vaccine-coverage-per-age-group/components/bar.tsx
```diff
@@ -0,0 +1,36 @@
+import { isPresent } from 'ts-is-present';
+import { PercentageBar } from '~/components/percentage-bar';
+import { parseFullyVaccinatedPercentageLabel } from '~/domain/vaccine/logic/parse-fully-vaccinated-percentage-label';
+
+interface BarProps {
+  value: number | null;
+  color: string;
+  label?: string | null;
+}
```
VWSCoronaDashboard1, line 9:
```suggestion
type BarWithValue = {
  value: number;
};
type BarWithLabel = {
  label: string | null;
};

type BarProps = {color: string;} & (BarWithValue  |  BarWithLabel );

function isBarWithValue(props:BarProps): value is BarWithValue  {
  return 'value' in props;
}
function isBarWithLabel (props:BarProps): value is BarWithLabel {
  return 'label' in props;
}

```

packages/app/src/domain/vaccine/vaccine-coverage-per-age-group/components/bar.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export function Bar({ value, color, label }: BarProps) {
```
VWSCoronaDashboard1, line 11:
```suggestion
export function Bar(props: BarProps) {
```

packages/app/src/domain/vaccine/vaccine-coverage-per-age-group/components/bar.tsx
```diff
@@ -0,0 +1,36 @@
...
+  let parsedVaccinatedLabel;
+  if (isPresent(label)) {
+    parsedVaccinatedLabel = parseFullyVaccinatedPercentageLabel(label);
+  }
```
VWSCoronaDashboard1, line 15:
```suggestion
  const percentageBarValue = isBarWithLabel(props) ? parseFullyVaccinatedPercentageLabel(props.label).value :  isBarWithValue(props) ? props.value : undefined;
```

packages/app/src/domain/vaccine/vaccine-coverage-per-age-group/components/bar.tsx
```diff
@@ -0,0 +1,36 @@
...
+          hasFullWidthStripedBackground
+        />
+      ) : (
+        <PercentageBar
+          percentage={value as number}
+          height={8}
+          color={color}
+          hasFullWidthBackground
+        />
+      )}
```
VWSCoronaDashboard1, line 33:
```suggestion
      {isDefined(percentageBarValue) && (
        <PercentageBar
          percentage={percentageBarValue}
          height={8}
          color={color}
          backgroundStyle={isBarWithLabel(props) ? 'hatched' : 'normal'}
        />
      )}
```

VWSCoronaDashboard1 (changes requested):

packages/app/src/components/percentage-bar.tsx
```diff
@@ -1,35 +1,84 @@
...
+        color={color}
       />
+
+      {hasFullWidthStripedBackground && (
+        <Box
+          css={css({
+            // Created by https://stripesgenerator.com/
+            backgroundImage: `linear-gradient(45deg, ${fullWidthStripedBackgroundColor} 30%, #ffffff 30%, #ffffff 50%, ${fullWidthStripedBackgroundColor} 50%, ${fullWidthStripedBackgroundColor} 80%, #ffffff 80%, #ffffff 100%)`,
+            backgroundSize: '7.07px 7.07px',
+          })}
```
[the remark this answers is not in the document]
VWSCoronaDashboard16: Can't we use transparent instead of white so we don't run into trouble if we ever change our background color to some off-white color?

packages/app/src/domain/vaccine/vaccine-coverage-per-age-group/components/age-group.tsx
```diff
@@ -0,0 +1,31 @@
...
+    total: formatNumber(ageGroupTotal),
+  });
+
+  return (
+    <Box display="flex" flexDirection="column">
+      <InlineText fontWeight="bold">{range}</InlineText>
+      <InlineText variant="label1">
+        {`${birthyear_range}${totalText ? `: ${totalText}` : ''}`}
+      </InlineText>
+    </Box>
```
VWSCoronaDashboard16, line 29: Why use `InlineText` and then make the container flex column? It has the same visual result as using a non-styled container with `Text` elements, right?

packages/app/src/domain/vaccine/vaccine-coverage-per-age-group/components/narrow-percentage.tsx
```diff
@@ -0,0 +1,93 @@
...
+                  }
+                )
+          }
+        />
+      ) : (
+        <PercentageMarkup
+          value={`${formatPercentage(value as number)}%`}
+          color={color}
+          textLabel={textLabel}
+        />
```
VWSCoronaDashboard16, line 57: Instead of doubling this template I would determine the textValue above in a variable and just insert it here as value.

Also, I think we can maybe write a more common function which does this whole value/label formatting. The same is done for the tooltip in the choropleth and we now have multiple lokalize keys for this. We can do that later though.

packages/app/src/domain/vaccine/vaccine-coverage-per-age-group/components/wide-percentage.tsx
```diff
@@ -0,0 +1,86 @@
...
+                )
+          }
+        />
+      ) : (
+        <PercentageMarkup
+          value={`${formatPercentage(value as number)}%`}
+          color={color}
+        />
+      )}
+    </>
```
VWSCoronaDashboard16, line 55: Same here, would put the textValue in a variable so you don't have the double template here.

VWSCoronaDashboard16 (changes requested):

packages/app/src/domain/vaccine/vaccine-coverage-per-age-group/components/narrow-percentage.tsx
VWSCoronaDashboard14: Let's do it later, please.

packages/app/src/domain/vaccine/vaccine-coverage-per-age-group/components/age-group.tsx
VWSCoronaDashboard14: Want to use spans for the text instead of regular paragraph elements. I could either set the spans as block elements or just use a flex-direction to put them horizontally under each other.

packages/app/src/domain/vaccine/vaccine-coverage-per-age-group/components/wide-percentage.tsx
VWSCoronaDashboard14: Keeping this for now since we're gonna get a more generic function for this.

PR #3513, opened 2021-09-15
Display both vaccine metrics in choropleth tooltip

VWSCoronaDashboard16:
## Summary

Refactors the vaccine coverage choropleth tooltip to allow showing secondary metrics next to the main one, switches the main metric to the percentage of people who had at least 1 shot, and adds the vaccine coverage metric as secondary.

image


packages/app/src/domain/vaccine/vaccine-coverage-choropleth-per-gm.tsx
```diff
@@ -134,32 +138,49 @@ export function VaccineCoverageChoroplethPerGm({
 
 type ChoroplethTooltipProps<T extends ChoroplethDataItem> = {
   data: TooltipData<T>;
-  getSecondaryMetric: (d: T) => { sign: string; value: number } | null;
+  getValues?: (
+    d: T,
+    text: Record<string, Record<string, Record<string, string>>>,
```
VWSCoronaDashboard1, line 143: From what I can see you always inject the `siteText.choropleth_tooltip` here, so instead of the nested Records you could type it explicitly to this part of the siteText, by importing the `SiteText` interface and typing the parameter as `SiteText['choropleth_tooltip']`.

packages/app/src/domain/vaccine/logic/get-vaccine-coverage-display-values.ts
```diff
@@ -0,0 +1,74 @@
...
+  formatPercentage: (value: number) => string
+) {
+  return function <T extends ChoroplethDataItem>(
+    data: T,
+    property: keyof T,
+    labelKey: keyof T
+  ) {
+    const parsedValue = isPresent(data[labelKey])
+      ? parseFullyVaccinatedPercentageLabel(data[labelKey] as unknown as string)
+      : (data[property] as unknown as undefined | number | null) || null;
```
VWSCoronaDashboard1, line 55:
```suggestion
      : (data[property] as unknown as undefined | number | null) ?? null;
```
using the `||` operator would assign null in the case where data[property] returns a value of zero (0), because that is considered falseish

packages/app/src/domain/vaccine/logic/get-vaccine-coverage-display-values.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    if (typeof parsedValue === 'number') {
+      return formatPercentage(parsedValue) + '%';
+    } else if (parsedValue !== null && 'sign' in parsedValue) {
```
VWSCoronaDashboard1, line 59:
```suggestion
    } else if (isPresent(parsedValue) && 'sign' in parsedValue) {
```
From what I understand, in theory `parsedValue` can also be undefined, so even with the null check, this would lead to a runtime error.

VWSCoronaDashboard1 (changes requested):

packages/app/src/domain/vaccine/logic/get-vaccine-coverage-display-values.ts
VWSCoronaDashboard16: It should become `null` if the metric is undefined in the data because of the line you mentioned above, but might be good to use isPresent for consistency as well.

VWSCoronaDashboard1: Ah, yes, you're right, it does default to null

PR #3514, opened 2021-09-15
Update toggle KPI block for VR/GM

VWSCoronaDashboard14:
Since the shape of the data is different it was originally intended I'll add a prop where we can add the data a bit more flexibly.


packages/app/src/domain/vaccine/vaccine-coverage-toggle-tile.tsx
```diff
@@ -1,69 +1,57 @@
...
-type AgeGroupTypes = {
-  description_vaccination_grade: string;
-  description_vaccination_one_shot: string;
-  label: string;
+type ageTypes = {
+  fully_vaccinated: number | null;
+  has_one_shot: number | null;
+  birthyear: string;
+  label_fully_vaccinated: string | null;
+  label_has_one_shot: string | null;
```
VWSCoronaDashboard1, line 22:
```suggestion
type AgeTypes = {
  fully_vaccinated: number | null;
  has_one_shot: number | null;
  birthyear: string;
  label_fully_vaccinated: string | null;
  label_has_one_shot: string | null;
```
Also, doesn't this type get generated? Or is this for mock purposes?

VWSCoronaDashboard1 (approved):

packages/app/src/domain/vaccine/vaccine-coverage-toggle-tile.tsx
VWSCoronaDashboard14: We have 2 completely different schemas that need to fit together in 1 component, it was changed last minute where the data should come from. I couldn't find a better solution.

packages/app/src/pages/veiligheidsregio/[code]/vaccinaties.tsx
```diff
@@ -115,15 +115,21 @@ export const VaccinationsVrPage = (
...
+    data.vaccine_coverage_per_age_group.values.filter(
+      (item) => item.age_group_range === '18+'
+    )[0] as VrVaccineCoveragePerAgeGroupValue;
 
-  const gmCodes = gmCodesByVrCode[data.code];
-  const selectedGmCode = gmCodes ? gmCodes[0] : undefined;
+  const filteredAgeGroup12Plus =
+    data.vaccine_coverage_per_age_group.values.filter(
+      (item) => item.age_group_range === '12+'
+    )[0] as VrVaccineCoveragePerAgeGroupValue;
```
Comment (VWSCoronaDashboard16) [inline packages/app/src/pages/veiligheidsregio/[code]/vaccinaties.tsx:134]: If you only take the first value anyway, you should use `values.find()`.

packages/app/src/pages/landelijk/vaccinaties.tsx
```diff
@@ -113,6 +114,17 @@ const VaccinationPage = (props: StaticProps<typeof getStaticProps>) => {
...
+  const sampleData = {
+    age_18_plus_fully_vaccinated: 123,
+    age_18_plus_has_one_shot: 4123411234123,
+    age_18_plus_birthyear: '-2015',
+    age_12_plus_fully_vaccinated: 34,
+    age_12_plus_has_one_shot: 123412341234,
+    age_12_plus_birthyear: '-1990',
+    date_unix: 12341231,
+    date_of_insertion_unix: 12309443,
+  };
```
VWSCoronaDashboard16, line 126: Just a reminder to use the real data before merging

VWSCoronaDashboard16 (approved):

PR #3544, opened 2021-09-21
Add sparkbars component to Actueel MiniTrendTiles

VWSCoronaDashboard17:
## Summary

Adds a `SparkBars` component to the current MiniTrendTiles. This is not the right place and they will need to be moved to the new trend layout when that is finished but this allows us to review the spark-bars.


VWSCoronaDashboard17:
One outstanding question I have is whether it makes sense to add accessibility features to the component to allow folks using screenreaders to 'read' the trend?

VWSCoronaDashboard1 (approved):

PR #3543, opened 2021-09-21
Add new actueel mini trends

VWSCoronaDashboard1:
## Summary

This adds the new designs for the mini trend tiles on the actueel pages:

### Gemeente:
image

### Landelijk:
image

PR #3549, opened 2021-09-21
Implement general vaccine coverage percentage formatter

VWSCoronaDashboard16:
## Summary

The logic of nicely formatting a vaccine coverage percentage value into a user-friendly label was implemented separately on the 4 different locations it was needed. This PR creates one general formatter for this which can be used on all places where we need to do this.


VWSCoronaDashboard1 (commented):
lol, I think we've done a bit of double work here:
https://github.com/minvws/nl-covid19-data-dashboard/pull/3543/files#diff-85597afb1bedc3da6b23779ac12746f5eca0ed2b993fead57ba7a39821888cb6

packages/app/src/domain/vaccine/logic/use-vaccine-coverage-percentage-formatter.ts
```diff
@@ -0,0 +1,77 @@
+import { createFormatting } from '@corona-dashboard/common';
+import { isPresent } from 'ts-is-present';
+import { useIntl } from '~/intl';
+import { SiteText } from '~/locale';
+import { replaceVariablesInText } from '~/utils/replace-variables-in-text';
+import { parseFullyVaccinatedPercentageLabel } from './parse-fully-vaccinated-percentage-label';
+
+export type KeyWithLabel<T> = {
```
VWSCoronaDashboard14, line 8: This is indeed some advanced typing, don't know if it could be simplified but some comments would be nice!

VWSCoronaDashboard14, line 8: This is indeed some advanced typing, don't know if it could be simplified but some comments would be nice!

VWSCoronaDashboard1: I'm not against this solution. I think it might be a tad over-engineered, currently we don't really use this pattern anywhere else, so making it this generic just might be a bit overkill.
But, since its been written now, I'm not going to oppose it.

PR #3557, opened 2021-09-22
Update homepage articles

VWSCoronaDashboard14:
All changes are applicable to the normal topical as well as GM/VR

- Communication can choose now only a weekly message and 1 message
- Communication can choose 3 custom articles at the bottom of the pages
- Add ContentTeaser item

Bigger highlighted items in the middle of the page
Screenshot 2021-09-22 at 15 47 22

At the bottom of the page
Screenshot 2021-09-22 at 15 47 26


packages/app/src/components/teaser-item.tsx
```diff
@@ -0,0 +1,78 @@
...
+          height={variant === 'normal' ? (breakpoints.sm ? 108 : 66) : 66}
+          sizes={[[1200, 438]]}
+        />
+      </Box>
+      <Box maxWidth="25rem" spacing={publicationDate || category ? 2 : 0}>
+        <InlineText
+          variant="overline2"
+          color="bodyLight"
+          css={css({ display: 'block' })}
+        >
```
VWSCoronaDashboard1, line 55: Why not simply use `Text` instead of `InlineText` with the display set to block?

VWSCoronaDashboard1 (approved):

packages/app/src/components/teaser-item.tsx
VWSCoronaDashboard14: Trying the hard way

packages/app/src/components/content-teaser.tsx
```diff
@@ -0,0 +1,73 @@
...
+export interface ContentTeaserProps {
+  title: string;
+  slug: string;
+  cover: ImageBlock;
+  category?: string;
+  publicationDate?: string;
+  variant?: 'small' | 'normal';
+}
+
+export function ContentTeaser({
```
VWSCoronaDashboard16, line 20: Nit: If I see there's an ArticleTeaser and a ContentTeaser I would expect the ContentTeaser to actually contain some small excerpt of the article's content, but it doesn't have that. The ArticleTeaser actually does. Wouldn't it be more logical to switch those names around?

VWSCoronaDashboard16 (approved):

packages/app/src/components/content-teaser.tsx
VWSCoronaDashboard14: Nope, the content teaser is actually content they can choose in the CMS (not related specifically only to articles). But on the homepage, I'll throw the article content in the same component.  The article teaser is for the /artikelen. It wouldn't make sense to switch them around since the content can hold more than only article content. Hope that make sense :)

PR #3570, opened 2021-09-24
Add (hidden) menu headings for screenreaders

VWSCoronaDashboard16:
Adds (hidden) headings to menu's so they are easier to navigatie to and through by screenreader users. Used this w3 example to determine the setup we needed: https://www.w3.org/WAI/tutorials/page-structure/headings/#main-heading-after-navigation

While I was at it I also added `aria-current="page"` to active menu page links so it is clear for screenreader users which page they are at right now.


packages/app/src/domain/layout/nl-layout.tsx
```diff
@@ -128,13 +129,16 @@ export function NlLayout(props: NlLayoutProps) {
             /** re-mount when route changes in order to blur anchors */
             key={router.asPath}
             id="metric-navigation"
-            aria-label={siteText.aria_labels.metriek_navigatie}
+            aria-aria-labelledby="sidebar-title"
```
VWSCoronaDashboard17, line 132:
```suggestion
            aria-labelledby="sidebar-title"
```

VWSCoronaDashboard17 (approved):

VWSCoronaDashboard1 (approved):

PR #3577, opened 2021-09-27
Add workaround for combobox ordering bug

VWSCoronaDashboard16:
## Summary

When typing in different search terms in the combobox, the keyboard navigation order of the options would get scrambled. This seems to be a bug in the reach combobox we are using: https://github.com/reach/reach-ui/issues/840. This PR adds the workaround suggested in that issue, which indeed seems to work.

We should keep an eye on the issue, so we can remove this workaround when the bug is solved.


VWSCoronaDashboard16:
A better workaround was just posted, I'm going to research it a bit first, since this one is not ideal

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard17 (approved):

PR #3583, opened 2021-09-27
Add aria-hidden to value annotation

VWSCoronaDashboard1:
## Summary

The value annotation of the timeseries chart now has an aria-hidden attribute set to true


VWSCoronaDashboard14 (approved):
Nice work!

VWSCoronaDashboard16 (approved):

PR #3586, opened 2021-09-28
Simplify the NL/VR/GM sidebars

VWSCoronaDashboard17:
## Summary

This PR removes all metrics from the sidebars in preparation of refactoring their design.

It includes a fairly significant refactor to static props getters and their typing. There are no more default datasets being selected from the data (which before was the case for the sidebars). Instead, all data keys will need to be passed explicitly.

Support was added for requesting data at nested keys, eg. `selectNlData('difference.sewer__average')`. This uses standard lodash path selectors. The selectors have type completion and the results of the selection is fully typed.

This change saves around 400Mb in the generated `.next/` folder and shaves 2 minutes off the build time.

Additionally, the sidebars are now data-driven and their design was updated to be simpler.


VWSCoronaDashboard1 (approved):
Very nice improvements here. Did you do a full build? I bet this results in quite a bit of size reduction as well.

packages/app/src/pages/landelijk/varianten.tsx
```diff
@@ -30,12 +31,15 @@ export const getStaticProps = withFeatureNotFoundPage(
...
-      delete data.selectedNlData.variants;
+      const data = selectNlData('variants', 'named_difference')();
+
+      const {
+        selectedNlData: { variants, ...rest },
+      } = data;
 
       return {
-        selectedNlData: data.selectedNlData,
+        selectedNlData: rest,
```
VWSCoronaDashboard16, line 41: Since the `named_difference` is also used within this same return and not in the template, is it really necessary returning the rest here?

packages/app/src/static-props/get-data.ts
```diff
@@ -140,23 +126,18 @@ async function replaceReferencesInContent(
...
- */
-export function selectNlPageMetricData<T extends keyof Nl = NlPageMetricNames>(
-  ...additionalMetrics: T[]
-) {
-  return selectNlData(...[...nlPageMetricNames, ...additionalMetrics]);
-}
+type UnionDeepMerge<T extends Record<string, unknown>> = {
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  [K in keyof T]: T[K] extends object ? U.Merge<T[K]> : T[K];
+};
```
VWSCoronaDashboard16, line 132: Would be nice if type 'magic' like this is explained a bit in comments? Also at the places this is used maybe. The ts-toolbelt docs are not that clear on what happens here.

VWSCoronaDashboard16 (approved):
Love the deep selectors for data! Have to take a look into ts-toolbelt, it really seems like a powerful tool.

packages/app/src/static-props/get-data.ts
VWSCoronaDashboard17: No this is my own type magic :) I'm merging unions that sit at keys of objects, one level deep. I'll add some docs.

packages/app/src/pages/landelijk/varianten.tsx
VWSCoronaDashboard17: I didn't really look at what is/isn't used, just copied what was there. I'll update it to disregard the rest.

VWSCoronaDashboard1 (approved):

PR #3587, opened 2021-09-28
Enable SSR fallback again so EN GM/VR pages do not 404 anymore

VWSCoronaDashboard16:
## Summary

We did not render dynamic pages (all VR/GM specific pages) during build. Since we turned the SSR fallback of those dynamic routes off a while ago, the pages return a 404 at the moment.

This PR makes sure the `getStaticPaths()` method of these routes returns both locales, so they are both rendered during build. (https://nextjs.org/docs/advanced-features/i18n-routing#dynamic-routes-and-getstaticprops-pages)

### DIFF
--- a/packages/app/src/static-paths/gm.ts
+++ b/packages/app/src/static-paths/gm.ts
@@ -4,9 +4,12 @@ import { gmData } from '~/data/gm';
  * getStaticPaths creates an array of all the allowed `/gemeente/[code]` routes.
  */
 export function getStaticPaths() {
-  const paths = gmData.map((x) => ({
-    params: { code: x.gemcode },
-  }));
+  const paths = gmData.flatMap((x) =>
+    ['nl', 'en'].map((locale) => ({
+      locale,
+      params: { code: x.gemcode },
+    }))
+  );
 
   return {
     paths,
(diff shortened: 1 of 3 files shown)


VWSCoronaDashboard1 (commented):
I would like to investigate other ways of solving this, this solution results in doubling the build time. I would like to prevent this.

VWSCoronaDashboard16:
Yes it seems the SSR fallback was there for a reason, it makes sense to use SSR for the EN pages since they are not visited that often. We need to look for a way to keep that fallback but give proper 404's when pages do not exist (which is why the fallback was removed)

VWSCoronaDashboard16:
I first thought of adding another StaticProps wrapper like we have for disabled features: `WithNotFoundPage`, that accepts a custom function validating if the page exists or not. However, that requires that wrapper on every dynamic vr/gm page, which is a bit of a hassle and easy to forget for new pages in the future, which introduces this 'invisible' bug again.

In the end chose to actually check for the ENOENT error in the getStaticProps, so we now throw a 404 whenever a file required for rendering the page cannot be found, which actually makes sense.

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard14 (approved):

PR #3588, opened 2021-09-28
Add moving average date start unix

VWSCoronaDashboard14:
For the hospital/icu admissions add a moving average when it starts counting the 7-day average date. So e.g. the average is from 17 September the new property should hold a unix timestamp for the date of 10 September.

I know the key names get a little long, but I will rather be as descriptive as possible.

### DIFF
--- a/packages/app/schema/gm/hospital_nice.json
+++ b/packages/app/schema/gm/hospital_nice.json
@@ -7,6 +7,7 @@
       "required": [
         "admissions_on_date_of_admission",
         "admissions_on_date_of_admission_moving_average",
+        "admissions_on_date_of_admission_moving_average_date_start_unix",
         "admissions_on_date_of_reporting",
         "date_unix",
         "date_of_insertion_unix"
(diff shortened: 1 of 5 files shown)


packages/app/schema/gm/hospital_nice.json
```diff
@@ -7,6 +7,7 @@
       "required": [
         "admissions_on_date_of_admission",
         "admissions_on_date_of_admission_moving_average",
+        "admissions_on_date_of_admission_moving_average_date_start_unix",
```
VWSCoronaDashboard1, line 10: I think you can just suffix this with `date_unix`, because `date_start_unix` is used everywhere else for date spans, so this would be confusing.

VWSCoronaDashboard1 (changes requested):

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard16 (approved):

PR #3596, opened 2021-09-29
Add risk level options in sanity

VWSCoronaDashboard14:
If someone has a better idea of turning this array with a length 1 in actual radio buttons.
Be my guest because I'm either not good at Googling it or it isn't possible out of the box.

Otherwise communication can choose 3 risk levels and a date from where it started.


packages/app/src/queries/topical-page-query.ts
```diff
@@ -10,41 +10,45 @@ export function getTopicalPageQuery(
...
+            ...cover,
+            "asset": cover.asset->
+          }
+        }
+      }[0].articles,
+      'elements': ${createElementsQuery(code, elementNames, locale)},
+      'riskLevel': *[_type == 'escalationLevelPage']{
+		    "level": riskLevel,
+		    "dateFrom": date,
+      }[0]
```
VWSCoronaDashboard16, line 51: I think it would be more logical to put the risk level on the actual page it is used in Sanity.

VWSCoronaDashboard16 (changes requested):

PR #3606, opened 2021-09-30
Data-driven sidebar

VWSCoronaDashboard17:
## Summary

This PR makes the sidebar rendering data driven as opposed to using large compositions of JSX.

## Motivation

Simplify adding new routes to layouts.

## Detailed design

The premise is quite simple: enter a keyed map with strings for categories and items in the sidebar and get back an expanded map containing titles, hrefs and icons. Use the expanded map to render the sidebar components.

For now, I've chosen to only render the easily loopable part of the sidebar using this method. I still need to look into adding support for injecting texts, feature flags and other exceptions. The styling of the sidebar items also still needs to be updated. I'll take on both of those in the next PR.


packages/app/src/domain/layout/logic/sidebar.tsx
```diff
@@ -0,0 +1,136 @@
...
+  reproduction_number: 'reproductiegetal',
+} as const;
+
+type UseSidebarArgs<T extends 'nl' | 'vr' | 'gm'> = {
+  layout: T;
+  map: SidebarMap<T>;
+  code?: T extends 'nl' ? never : string;
+};
+
+export function useSidebar<T extends 'nl' | 'vr' | 'gm'>({
```
VWSCoronaDashboard14, line 78: Since this file exports, a hook might as well call the file use-sidebar.tsx

packages/app/src/domain/layout/logic/sidebar.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+  return useMemo(() => {
+    const getHref = (key: ItemKeys<T>) => {
+      const route =
+        mapKeysToReverseRouter[key as keyof typeof mapKeysToReverseRouter];
+
+      if (layout === 'nl') {
+        return reverseRouter.nl[route]();
+      }
+
```
VWSCoronaDashboard1, line 94: Instead of checking for a valid code using `isPresent(code)`, maybe it makes more sense to simply `assert()` the existence of the code here? Otherwise the method will return undefined which will lead to more nebulous errors downstream.

VWSCoronaDashboard1 (approved):

PR #3615, opened 2021-10-01
Update packages

VWSCoronaDashboard17:
## Summary

Maintenance

### DIFF
--- a/packages/app/src/types/css.d.ts
+++ b/packages/app/src/types/css.d.ts
@@ -9,3 +9,11 @@ declare module 'react' {
     css?: CssFunctionReturnType;
   }
 }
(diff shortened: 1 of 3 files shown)


packages/app/src/types/css.d.ts
```diff
@@ -9,3 +9,11 @@ declare module 'react' {
     css?: CssFunctionReturnType;
   }
 }
+
+declare module '@corona-dashboard/icons' {
```
VWSCoronaDashboard17, line 13: Makes sure we can use the `css` prop on icons.

VWSCoronaDashboard1 (approved):

PR #3623, opened 2021-10-04
Add compression to choropleth images

VWSCoronaDashboard1:
## Summary

As the title states. This compresses the generated choropleth preview images, saving a little bit of bandwidth and disk space


VWSCoronaDashboard17:
Nice one! I think if you use [`sharp`](https://www.npmjs.com/package/sharp) instead of `imagemin`, Next.js will also use that library during production to optimize other (statically imported) images by default. Two birds, one stone.

VWSCoronaDashboard16:
Agree on using sharp might be better. It is probably better performance-wise (in my experience libvips is super fast compared to other image libraries), and imagemin is not maintained anymore.

PR #3628, opened 2021-10-05
Update sidebar with visual tweaks

VWSCoronaDashboard14:
- Change the width of the sidebar on certain breakpoints
- When the screen is narrow and the content isn't visible I'm disabling the isActive check

### DIFF
--- a/packages/app/src/components/aside/menu.tsx
+++ b/packages/app/src/components/aside/menu.tsx
@@ -115,8 +115,8 @@ export function MenuItemLink({ href, icon, title }: MenuItemLinkProps) {
     <li>
       <Link href={href} passHref>
         <StyledAnchor
-          isActive={isActive}
-          aria-current={isActive ? 'page' : undefined}
+          isActive={breakpoints.md && isActive}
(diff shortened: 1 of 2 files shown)


packages/app/src/components/aside/menu.tsx
```diff
@@ -113,10 +115,14 @@ export function MenuItemLink({ href, icon, title }: MenuItemLinkProps) {
     <li>
       <Link href={href} passHref>
         <StyledAnchor
-          isActive={isActive}
-          aria-current={isActive ? 'page' : undefined}
+          isActive={breakpoints.md && isActive}
+          aria-current={breakpoints.md && isActive ? 'page' : undefined}
```
VWSCoronaDashboard17, line 119: `aria-current` should still be there even on smaller breakpoints, the page is still active even if the styling doesn't reflect it.

VWSCoronaDashboard14: Alright! Wasn't totally sure about that one.

VWSCoronaDashboard17 (approved):
🚀

PR #3655, opened 2021-10-13
Update variants schema for historical significance and variants of concert or interest. 

VWSCoronaDashboard14:


VWSCoronaDashboard16:
The reference(s) to the old `is_variant_of_concern` also have to be changed in code still

PR #3666, opened 2021-10-15
Display admissions_on_date_of_admission in Actueel hospital admissions MiniTrendTile

VWSCoronaDashboard17:
- Applying the hotfix just done to master to develop as well.


VWSCoronaDashboard17:
They are auto-generated, so I'm not sure why they are here. This branch was freshly branched from develop so I assume that if you run the generate-typescript script on develop you'll see the same thing.

PR #3755, opened 2021-10-29
Update schema vaccine status

VWSCoronaDashboard14:
We're merging the `not_vaccinated` and `has_one_shot` into one key


VWSCoronaDashboard7:
We're merging #3768 which also contains this code, so closing it for that.

PR #3757, opened 2021-10-29
Cut all values that are not needed on the homepage

VWSCoronaDashboard14:
A new util function that will cut values based on the existing timeframes, this will result that only the last 35 values are being added to the page compared the all the 600+ per metric.

### DIFF
--- a/packages/app/src/pages/actueel/veiligheidsregio/[code].tsx
+++ b/packages/app/src/pages/actueel/veiligheidsregio/[code].tsx
@@ -50,6 +50,7 @@ import {
   selectVrData,
 } from '~/static-props/get-data';
 import { countTrailingNullValues } from '~/utils/count-trailing-null-values';
+import { cutValuesFromTimeframe } from '~/utils/cut-values-from-timeframe';
 import { getBoundaryDateStartUnix } from '~/utils/get-boundary-date-start-unix';
 import { replaceComponentsInText } from '~/utils/replace-components-in-text';
 import { replaceVariablesInText } from '~/utils/replace-variables-in-text';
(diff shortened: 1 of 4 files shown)


VWSCoronaDashboard1 (approved):

packages/app/src/pages/actueel/veiligheidsregio/[code].tsx
```diff
@@ -60,12 +61,22 @@ export { getStaticPaths } from '~/static-paths/vr';
...
-    'difference'
-  ),
+
+  (context) => {
+    const data = selectVrData(
+      'vaccine_coverage_per_age_group',
+      'hospital_nice',
+      'code',
+      'difference'
+    )(context);
```
Comment (VWSCoronaDashboard17) [inline packages/app/src/pages/actueel/veiligheidsregio/[code].tsx:71]: I feel like we do this often (change data from select functions in gSP) enough that maybe the `selectVrData` could take an optional last argument `processors` that can be used to update/change the relevant bits of the data. It could be a function or an array of functions. This construct could help in keeping gSP definitions more clean and readable. WDYT?

VWSCoronaDashboard17 (approved):
👍 Only feedback I have is in a broader sense and doesn't apply to the PR specifically.

packages/app/src/pages/actueel/veiligheidsregio/[code].tsx
Comment (VWSCoronaDashboard14) [reply packages/app/src/pages/actueel/veiligheidsregio/[code].tsx:71]: Yeah, we can put this on the tech debt, could be useful yes but indeed not in the scope of this change!

Comment (VWSCoronaDashboard14) [reply packages/app/src/pages/actueel/veiligheidsregio/[code].tsx:71]: @VWSCoronaDashboard1

PR #3758, opened 2021-10-31
Automatically re-compile app when commons change

VWSCoronaDashboard1:
## Summary

Making changes to the commons package during development previously didn't trigger a re-compile in the app which could lead to confusing situations. I.e. when editing feature flags or theme colors, etc.

This PR adds a build:watch script in the commons package which also gets run by the 'dev' script in the app package.
build:watch simply starts has a chokidar file change watcher on the common/src directory and re-builds the package when changes are detected. Changes to the commons are already automatically detected by the app which will trigger a re-compile.


VWSCoronaDashboard17 (approved):
That's helpful!

VWSCoronaDashboard14 (approved):

PR #3773, opened 2021-11-02
Update README.md

MarByteBeep:
`yarn start` doesn't do anything. It should be `yarn sanity start` or `yarn dev`, which is the same.


VWSCoronaDashboard17 (approved):
Thank you for your contribution!

PR #3776, opened 2021-11-02
Update risk level to be data driven

VWSCoronaDashboard14:
- Warning: this contains dummy data that needs to be removed.
- Add 2 sort of custom KPI tiles to the about-risklevel page. Since the inline KPI's were pretty tailored to the content from the CMS.
- Tweak the banner a bit to work data-driven
- Remove in the CMS the option to choose risk levels 

Screenshot 2021-11-02 at 12 56 45


packages/app/schema/nl/risk_level.json
```diff
@@ -0,0 +1,72 @@
...
+      "type": "array",
+      "minItems": 1,
+      "maxItems": 1,
+      "items": {
+        "$ref": "#/definitions/value"
+      }
+    },
+    "last_value": {
+      "$ref": "#/definitions/value"
+    }
```
VWSCoronaDashboard1, line 16: I guess this will need to have that new `value` shape as well, so without the values array.

packages/app/src/domain/escalation-level/common.ts
```diff
@@ -1,8 +1,5 @@
 import { colors } from '@corona-dashboard/common';
 
-/**
- * @TODO remove when we have this available in the schemas
- */
 export type EscalationLevelType = 1 | 2 | 3;
```
VWSCoronaDashboard1, line 3: Shouldn't this be removed as well? Since it is now in the schema, and therefore in `data.ts`?

VWSCoronaDashboard1 (approved):

packages/app/schema/nl/risk_level.json
VWSCoronaDashboard14: Yes

VWSCoronaDashboard1 (approved):

PR #3777, opened 2021-11-02
Add a border to the spark bars

VWSCoronaDashboard14:
Quite subtle but it's there. Can't simply use a border-bottom since the SVG is a bit wider than the bars.

Screenshot 2021-11-02 at 16 03 02


packages/app/src/components/spark-bars.tsx
```diff
@@ -31,19 +33,22 @@ export function SparkBars<T extends TimestampedValue>(
...
+      })}
     >
       {last7Days.map((d, i) => (
         <rect
           key={'date_unix' in d ? d.date_unix : d.date_start_unix}
           x={i * BAR_WIDTH}
           width={BAR_WIDTH}
           opacity="0.5"
-          y={y(Math.max(0, d[averageProperty]))}
+          y={y(Math.max(0, d[averageProperty])) + BORDER_HEIGHT}
```
VWSCoronaDashboard16, line 51: If we add the border height we shift all bars to be partly behind the border, that's not what we want right?

packages/app/src/components/spark-bars.tsx
```diff
@@ -52,6 +57,13 @@ export function SparkBars<T extends TimestampedValue>(
           }
         />
       ))}
+      <rect
+        x={0}
+        width={BAR_WIDTH * last7Days.length}
+        height={BORDER_HEIGHT}
+        fill={colors.silver}
+        y={BAR_HEIGHT + BORDER_HEIGHT}
```
VWSCoronaDashboard16, line 65: Ah right you add the border height here as well, so the whole SVG gets shifted 2 pixels down. Why?

VWSCoronaDashboard16 (changes requested):
If I read the code correctly you shift all SVG content down by `BORDER_HEIGHT`, which explains the need for having an `overflow: visible`. Why not keep everything on its 'normal' place?

packages/app/src/components/spark-bars.tsx
```diff
@@ -31,19 +33,22 @@ export function SparkBars<T extends TimestampedValue>(
...
-      height={BAR_HEIGHT}
+      height={BAR_HEIGHT + BORDER_HEIGHT}
       role="img"
       aria-hidden="true"
       focusable="false"
-      viewBox={`0 0 ${BAR_WIDTH * 7} ${BAR_HEIGHT}`}
+      viewBox={`0 0 ${BAR_WIDTH * 7} ${BAR_HEIGHT + BORDER_HEIGHT}`}
+      css={css({
+        overflow: 'visible',
+      })}
```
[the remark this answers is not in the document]
VWSCoronaDashboard14: In some weird way yes. When not using this it show a small rendering bug on the second spark bar? But unsure why.

packages/app/src/components/spark-bars.tsx
VWSCoronaDashboard14: Nope, I had to extend the view box by 2 pixels and let the bars render on Y with their height. And let the border render at the height of the bars + the height of the border. By not doing that the border is overlapping the bottom of the bars, but I want to keep the height of the bars.

packages/app/src/components/spark-bars.tsx
VWSCoronaDashboard17: Instead, the `y` scale should probably go from `0` to `height - BORDER_HEIGHT`.

VWSCoronaDashboard16: If I understand correctly the bars should still go from `0` to `BAR_HEIGHT` but we add a border with `BORDER_HEIGHT` below. I would expect that we only need to extend the viewbox and add the border at `y=BAR_HEIGHT`. 

I just don't understand why both the bars as the border are shifted down by `BORDER_HEIGHT`, it only results in all bars starting two pixels lower and the border being outside the SVG:

image

VWSCoronaDashboard16 (approved):

PR #3779, opened 2021-11-03
Add unit tests

VWSCoronaDashboard17:
## Summary

This is meant as a collaborative PR where we can start adding unit tests for utility functions, among others.


Dockerfile
```diff
@@ -51,6 +51,8 @@ RUN yarn workspace @corona-dashboard/common build \
 && yarn workspace @corona-dashboard/cli generate-typescript \
 && yarn workspace @corona-dashboard/icons build
 
+RUN yarn test
```
VWSCoronaDashboard7, line 54: This needlessly creates a new Docker layer, it would therefore be better to bundle it with the previous RUN command. There is also a "test:ci" command, why this one and not the other during the docker build?

VWSCoronaDashboard17: I'll take another look at this

PR #3780, opened 2021-11-04
Properly handle missing content

MarByteBeep:
Account for undefined data. There is a bunch of other pages that aren't handling this well either.


VWSCoronaDashboard16:
Thanks for you contribution!

Note that the data was required on Sanity's side, so in practice not checking for undefined data here was not really a problem. Just last week we refactored the way pages are stored in Sanity, and thus also changed the way these queries work in #3790. We made sure we do check the possibility of missing data in the new queries now, just to be sure, so I am closing this now.

PR #3787, opened 2021-11-05
Add IC admissions vaccine incidence schema

VWSCoronaDashboard17:


VWSCoronaDashboard16:
I believe this one can be closed since we will probably be showing hospital incidence on the IC page as well.

VWSCoronaDashboard1 (approved):

PR #3788, opened 2021-11-05
Update tile selector to show the correct layout based on the viewport

VWSCoronaDashboard14:
This seems to work, comparing when building the develop branch and this branch this one shows the correct layout instantly. The only drawback is that we're indeed shipping a bit more code to the user.


VWSCoronaDashboard1 (approved):

VWSCoronaDashboard16:
I'm missing the point why the IDs are needed, can't we apply the `display: asResponsiveArray(...),` styles directly on the surrounding boxes of the narrow & wide layout?

VWSCoronaDashboard14:
@VWSCoronaDashboard16 didn't even think of that. Yes, that might work even better!

PR #3792, opened 2021-11-08
Make dropdown vaccination choropleth start on new line when there is no space

VWSCoronaDashboard14:


VWSCoronaDashboard16:
This does not seem to work perfectly here 
image

VWSCoronaDashboard14:
@VWSCoronaDashboard16 huh, that's weird. Let me double check

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard16 (approved):
The problem I found only occurs on a very specific resolution that's not used for any particular device, so in practice this state will not occur. For mobile it works as promised 👍

PR #3265, opened 2021-07-26
Improving the readme

VWSCoronaDashboard8:
## Summary

This mainly focusses on the app readme. Updating the documentation around setup instructions, and also simplifying some of the steps so that users are not forced to create a `.env.local` file to get things up and running.

I have moved some texts from `package/app/README` to the root README because that's where they seem to belong since we've migrated to a monorepo.

Issue #2236, opened 2021-03-08
Difficulty getting local environment set up due to Sanity integration


mpboom:
### Describe the bug, issue or concern

While trying to follow the README to set up the application locally, I ran into some issues. The `yarn download` command didn't seem to work properly, for which I opened #2235, but I also had some difficulties with what seem to be the integration with Sanity. If any workaround is available, it would be good to mention it in the README.

### To Reproduce
1. Clone the repository
2. Change directory to `packages/app/`
3. Run `yarn`
4. Run `yarn download` (with fix in #2235 applied)
5. Run `yarn dev`
6. Get greeted with error messages

### Expected behavior
I expect this to work and start the development server without any errors.

### Logs
My `.env.local` is exactly the same as the `.env.local.example`, I only updated the `NEXT_PUBLIC_LOCALE`. I also tried changing `NEXT_PUBLIC_SANITY_PROJECT_ID` to something random, but to no avail. This just raises errors in different places (when it actually tries to fetch data).

Not that I think it'll help that much, but here is the log:
https://gist.github.com/mpboom/6c324a7aaf933715c019094a4f179e94

### Additional context
I believe the issue is here that the local environment doesn't seem to function if you don't enter correct credentials for Sanity. To work on the vaccine page, I applied the following changes so no Sanity-related requests are fired:
https://github.com/mpboom/nl-covid19-data-dashboard/commit/88c2cabf8cb751d43093dc09e3b1f6d03dc0d510
But that, of course, is a very limited and cubersome workaround.

### Governance
- [x] I've read the contributing document https://github.com/minvws/.github/blob/master/CONTRIBUTING.md
- [x] I've read and understand the Code of Conduct https://github.com/minvws/.github/blob/master/CODE_OF_CONDUCT.md
- [x] I understand that any contributions or suggestions I made may make it into the actual code. I've read the License https://github.com/minvws/nl-covid19-notification-app-coordination/blob/master/LICENSE.txt and the contributor license agreement https://github.com/minvws/nl-covid19-notification-app-coordination/blob/master/CLA.md

VWSCoronaDashboard7:
Hi, we're aware of this issue and have discussed a few possible workarounds. Will update this ticket once there's news on that.

VWSCoronaDashboard7:
This shouldn't be an issue anymore.

mpboom:
[...]

PR #3799, opened 2021-11-08
Implement more fixes related to getting started locally

mpboom:
## Summary

In packages/app/.env.local.example, use correct default environment
variables for Sanity-related things to
ensure locally running after copying the .env.local.example properly
works.

The "yarn bootstrap" command actually performs the tasks the README
indicates it should perform. Also clarified some details about getting
started in the README.

Related to #2236.

## Motivation

Since the creation of #2236, care has been taken to make sure you are not required to have valid keys for Sanity in a local environment to be able to run it. However, it still doesn't work according to the `README`:

> By default, the site builds using the development dataset.

Instead, it throws errors if you just copy over the `.env.example` file.

Furthermore, the README indicates that you can use `yarn bootstrap` to:

> Right now, it will download a fresh dataset from production, export the texts
> from lokalize (development dataset) and build the common and icons packages
> for use by the app package.

Yet, the command yields:

```js
"bootstrap": "exit 0",
```

This PR implements some fixes regarding getting started. Hopefully this lowers the bar for outside/new contributors.

## Detailed design

_Not applicable, there aren't many complicated changes here._

## Drawbacks

Drawback? 1) would be that in #3431, the `yarn bootstrap` step has been altered to explicitly exhibit this behavior of doing nothing. However, the issue comes with very little context so I'm very unsure why this changed.

The second drawback 2) I can think of is that this way the Sanity project ID is hardcoded in the example config. I, however, don't think it's a very big deal since #3265 basically hardcoded the project ID throughout the application. Additionally, it'd be trivial to update and really helps new contributors getting up and running.

## Alternatives

An alternative to both drawbacks would be to not update code but rather mention required steps in the documentation (README).

## Unresolved questions
 
_Not applicable._

### Governance

- [ ] Documentation is added


packages/app/package.json
```diff
@@ -138,7 +138,7 @@
   },
   "scripts": {
     "analyze": "cross-env ANALYZE=true yarn build",
-    "bootstrap": "exit 0",
+    "bootstrap": "yarn download && yarn workspace @corona-dashboard/cms lokalize:export",
```
VWSCoronaDashboard17, line 141:
```suggestion
    "bootstrap": "yarn download",
```

The `lokalize:export` script is already ran from the `@corona-dashboard/cms` package when you execute `yarn bootstrap` from the monorepo root.

VWSCoronaDashboard17 (commented):
Thank you for your contribution, this looks like a helpful addition.

Some small points of feedback.

packages/app/package.json
VWSCoronaDashboard16: We removed `yarn download` from the bootstrap command since we often work with development datasets that will be overwritten by the command, while it is a very useful command during development as well. We should update the documentation to make that clear though.

packages/app/README.md
```diff
@@ -8,7 +8,10 @@ React application uses Next.js as its framework.
 First run `yarn install` or simply `yarn` at the root of the repository if you
 haven't already. This installs all the dependencies for all of the packages.
 
-Then from `packages/app` you can run:
+Make sure you have an `.env.local` file in `packages/app` with correct environment variables. To
+get started, you can copy `.env.local.example`.
```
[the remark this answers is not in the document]
mpboom: Good idea! I've added it.

packages/app/package.json
mpboom: OK, so I take this issue can then be entirely resolved by updating the README? I've reverted the change that updated `package.json` and updated the README instead:

1. I removed the `Yarn Bootstrap` section, since it falsely claimed the download and lokalize scripts would be run
2. I updated the listed commands to not include `yarn bootstrap`, but to _do_ include all other relevant commands
3. Some other minor updates

This way, we make no changes to Yarn scripts and local develop environments would not be impacted while still allowing others to properly build and run locally by following the instructions in the README.

Does that work for you?

packages/app/README.md
```diff
@@ -8,16 +8,20 @@ React application uses Next.js as its framework.
...
+get started, you can copy `.env.local.example`:
 
-1. `yarn bootstrap`
-2. `yarn dev`
+```sh
+cp packages/app/.env.local.example packages/app/.env.local
+```
 
-These steps are described in more detail below.
+Then from `packages/app` you can run:
```
VWSCoronaDashboard16, line 18:
```suggestion
Then from the root you can run:
```

packages/app/README.md
```diff
@@ -8,16 +8,20 @@ React application uses Next.js as its framework.
...
 
-### Yarn Bootstrap
+1. `yarn download`
+2. `yarn workspace @corona-dashboard/cms lokalize:export`
+3. `yarn dev`
```
VWSCoronaDashboard16, line 22:
```suggestion
1. `yarn download` - Downloads latest data
2. `yarn bootstrap` - Downloads / builds all other requirements
3. `yarn dev` - Starts the dev server
```

VWSCoronaDashboard16 (commented):
Thanks for updating the docs!

`yarn bootstrap` is still a useful command to run, since it includes _all_ steps needed to get running except for the data download. Furthermore it is probably the command we'll update in the future when more might be needed to get up-and-running, so that way the documentation is more future-proof.

Do you agree with the suggestions I propose here?

mpboom:
@VWSCoronaDashboard16 yes, that definitely works! I've implemented your suggestions. I've made it `repository root` to make it a bit clearer.

VWSCoronaDashboard16 (approved):

PR #3815, opened 2021-11-15
Add schema for booster KPI block

VWSCoronaDashboard14:
This schema contains 2 date ranges for the left block and right block.
If there are any better names for keys, please let me know.


packages/app/schema/nl/booster_shot.json
```diff
@@ -0,0 +1,67 @@
...
+        },
+        "total_date_end_unix": {
+          "type": "integer"
+        },
+        "received_last_seven_days": {
+          "type": "number"
+        },
+        "vaccinated_last_seven_days": {
+          "type": "number"
+        },
```
VWSCoronaDashboard1, line 42: I'm not sure about the naming of these two data points. I don't know too much about the booster feature, and I have a hard time deducting from this name what the value actually represents.

VWSCoronaDashboard1 (changes requested):

VWSCoronaDashboard1 (approved):

PR #3816, opened 2021-11-15
Add booster shot per age group schema

VWSCoronaDashboard14:
Identical setup to [this schema](https://github.com/minvws/nl-covid19-data-dashboard/blob/develop/packages/app/schema/nl/vaccine_coverage_per_age_group.json).
In the `birthyear_range` we are expecting a structure like this so that we can parse it on the front end later.
`-1940` -> `1940 of eerder`
`1941-1950` `1941 - 1950`

We don't do any calculations on when the dates are on the front-end so we're expecting both from the backend.
They are always tied together with this structure, regarding the `birthyear_range` `and age_group_range`.

### DIFF
--- a/packages/app/schema/nl/__index.json
+++ b/packages/app/schema/nl/__index.json
@@ -4,7 +4,7 @@
   "title": "nl",
   "required": [
     "behavior",
-    "nl_booster_shot_per_age_group",
+    "booster_shot_per_age_group",
     "code",
     "corona_melder_app_download",
     "corona_melder_app_warning",
--- a/packages/app/schema/nl/__index.json
+++ b/packages/app/schema/nl/__index.json
@@ -4,6 +4,7 @@
   "title": "nl",
   "required": [
     "behavior",
+    "nl_booster_shot_per_age_group",
     "code",
     "corona_melder_app_download",
     "corona_melder_app_warning",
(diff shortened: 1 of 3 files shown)


packages/app/schema/nl/booster_shot_per_age_group.json
```diff
@@ -0,0 +1,65 @@
...
+            "18-30",
+            "31-40",
+            "41-50",
+            "51-60",
+            "61-70",
+            "71-80",
+            "81+"
+          ]
+        },
+        "received_booster_amount": {
```
VWSCoronaDashboard16, line 43: Tiny nit: I don't think we use the `_amount` suffix anywhere so it breaks a bit with convention.

packages/app/schema/nl/__index.json
```diff
@@ -4,6 +4,7 @@
   "title": "nl",
   "required": [
     "behavior",
+    "nl_booster_shot_per_age_group",
```
VWSCoronaDashboard16, line 7:
```suggestion
    "booster_shot_per_age_group",
```

VWSCoronaDashboard16 (changes requested):

VWSCoronaDashboard16 (approved):

PR #3819, opened 2021-11-15
Add separate schema key for archiving ggd positive tested percentage graph

VWSCoronaDashboard16:
The GGD positive tests percentage graph will be archived, but we still need to get new percentages for the graph with the absolute numbers, since it is displayed in the tooltip there.

This is a proposal for a schema change which adds a key for the values of the separate percentage graph so we can keep on using the normal key for the graph with absolute numbers.

### DIFF
--- a/packages/app/schema/nl/tested_ggd.json
+++ b/packages/app/schema/nl/tested_ggd.json
@@ -52,6 +52,7 @@
         "infected",
         "infected_moving_average",
         "infected_percentage",
+        "infected_percentage_separate",
         "tested_total_moving_average",
         "infected_percentage_moving_average",
         "tested_total",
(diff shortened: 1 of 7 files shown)


VWSCoronaDashboard1 (approved):

packages/app/schema/nl/tested_ggd.json
```diff
@@ -49,6 +52,7 @@
         "infected",
         "infected_moving_average",
         "infected_percentage",
+        "infected_percentage_separate",
```
VWSCoronaDashboard7, line 55: This does not imply _why_ it's separate which could lead to confusion as to why it's not in sync with the other key. Shouldn't we just postfix it 'frozen', 'archived' or something like that?

VWSCoronaDashboard16: Thought about that, but that would lead to a conflict if the percentage we keep on using now will be archived. I tried to describe the graph it belonged to (which is also quite difficult).

VWSCoronaDashboard16:
Included in #3843

PR #3823, opened 2021-11-16
Add booster shot per age group section

VWSCoronaDashboard14:
It has some dummy data that needs to get removed later.
The schemas are also merged in this branch.
Moved the logic for creating the birthyear stuff to the utils folder since it's now getting used in 2 components.
There is no mobile design, everything just shrinks (what actually works well until like 300px screen size).

Screenshot:
Screenshot 2021-11-16 at 17 32 54


packages/app/src/domain/vaccine/vaccine-booster-per-age-group.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+    assert(index >= 0, `No sorting order defined for age group ${ageGroup}`);
+
+    return index;
+  };
+
+  const sortedValues = data.sort(
+    (a, b) =>
+      getSortingOrder(a.age_group_range) - getSortingOrder(b.age_group_range)
+  );
```
VWSCoronaDashboard1, line 41: maybe memo this?

VWSCoronaDashboard1 (approved):

packages/app/src/domain/vaccine/vaccine-booster-per-age-group.tsx
VWSCoronaDashboard14: There is no interactive state, so it wouldn't add much value for now.

PR #3827, opened 2021-11-18
Feature/add confirmed cases to actueel

VWSCoronaDashboard18:
On the homepage with the 'actuele situatie in Nederland' there needs to be a 'positief geteste mensen' tile. with the information about the confirmed cases with a time chart and a positive percentage of confirmed cases tested by the GGD.


VWSCoronaDashboard1 (approved):

Denniswegereef (approved):
Looking good 👍

packages/app/src/components/data-driven-text/components/metric.tsx
```diff
@@ -39,22 +39,19 @@ export function Metric<T extends DataKeys, K = DataFile<T>>({
         | DifferenceDecimal)
     : undefined;
 
-  assert(
```
Denniswegereef, line 42: I think this assert still needs to be here?

packages/app/src/components/data-driven-text/components/metric.tsx
```diff
@@ -39,22 +39,19 @@ export function Metric<T extends DataKeys, K = DataFile<T>>({
...
-    isDefined(differenceValue),
-    `Missing value for difference:${differenceKey}`
-  );
-
```
Denniswegereef, line 46: I think this assert still needs to be here?

VWSCoronaDashboard1, line 46: This assert doesn't need to be here in this form. But I think that it does need AN assert.
But one that checks if `differenceKey` is defined, and in that case the `differenceValue` needs to exist

VWSCoronaDashboard1 (approved):

packages/app/src/components/data-driven-text/components/metric.tsx
```diff
@@ -39,22 +39,26 @@ export function Metric<T extends DataKeys, K = DataFile<T>>({
         | DifferenceDecimal)
     : undefined;
 
-  assert(
-    isDefined(differenceValue),
-    `Missing value for difference:${differenceKey}`
-  );
+  if (isUndefined(differenceKey)) {
```
VWSCoronaDashboard16, line 42: Nit: `!isDefined` saves an extra import ;)

VWSCoronaDashboard16 (approved):
Looks good! Two tiny nitpicks

PR #3828, opened 2021-11-19
Keeping UI elements in the same style

VWSCoronaDashboard19:
* Update styling for UI element: dropdown
* Update styling for UI element: toggle
* Update styling for UI element: legend

### DIFF
--- a/packages/app/src/components/rich-content-select/components/index.tsx
+++ b/packages/app/src/components/rich-content-select/components/index.tsx
@@ -11,32 +11,51 @@ export const SelectBox = styled(Box)(
     display: 'flex',
     justifyContent: 'space-between',
     alignItems: 'center',
-    borderColor: 'border',
+    borderColor: 'silver',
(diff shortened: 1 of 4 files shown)


packages/app/src/components/rich-content-select/components/index.tsx
```diff
@@ -11,32 +11,51 @@ export const SelectBox = styled(Box)(
     display: 'flex',
     justifyContent: 'space-between',
     alignItems: 'center',
-    borderColor: 'border',
+    borderColor: 'silver',
     borderStyle: 'solid',
+    borderRadius: '5px',
```
Denniswegereef, line 16: Would be nice to get the border-radius from the theming!

VWSCoronaDashboard1: There is already a `radii` property on the theme even, so easy fix

VWSCoronaDashboard1 (approved):

PR #3830, opened 2021-11-19
Feature/remove roadmap and lockdown toggle from sanity

VWSCoronaDashboard22:
Removed roadmap and the lockdown toggle from Sanity as they were no longer used. Also removed the show lockdown check from the Maatregelen pages


VWSCoronaDashboard16 (commented):
Looks good! 👌 

I think `restrictionGroup.tsx` can also be deleted since it was only used by the restrictionCategory which you already deleted.

VWSCoronaDashboard22:
Removed restrictionGroup.tsx

VWSCoronaDashboard16 (approved):

PR #3838, opened 2021-11-22
Feature/add aria label to interactive legend buttons

VWSCoronaDashboard22:
Added an aria label to interactive legends which use age. Screen readers were not able to let the user know what unit the number were in.

### DIFF
--- a/packages/app/src/components/interactive-legend.tsx
+++ b/packages/app/src/components/interactive-legend.tsx
@@ -11,7 +11,7 @@ export interface SelectOption<T = string> {
   label: string;
   color: string;
   shape?: 'line' | 'circle' | 'square';
-  ariaLabel?: 'string';
+  legendAriaLabel?: string;
 }
 
 interface InteractiveLegendProps<T = string> {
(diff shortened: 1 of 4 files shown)


packages/app/src/components/interactive-legend.tsx
```diff
@@ -11,6 +11,7 @@ export interface SelectOption<T = string> {
   label: string;
   color: string;
   shape?: 'line' | 'circle' | 'square';
+  ariaLabel?: 'string';
```
VWSCoronaDashboard14, line 14: ` ariaLabel?: 'string';` -> `ariaLabel?: string;`

VWSCoronaDashboard1, line 14:
```suggestion
  legendAriaLabel?: string;
```

VWSCoronaDashboard1 (changes requested):
Are you 100% sure that those are the only charts with interactive legends? I thought there were more, but I could be mistaken...

packages/app/src/components/interactive-legend.tsx
VWSCoronaDashboard22: Fixed SelectOption interace and renamed ariaLabel property

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard14 (approved):

PR #3839, opened 2021-11-22
Add more chart sanity configs

VWSCoronaDashboard1:
## Summary

This PR adds support for the age demographic charts, donut charts and choropleths to Sanity. In order for he editors to be able to add these charts to articles and other portable text content


packages/app/src/components/cms/inline-age-demographic.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  const { siteText } = useIntl();
+
+  const dateUrl = getDataUrl(startDate, endDate, configuration);
+
+  const { data } = useSWRImmutable(dateUrl, (url: string) =>
+    fetch(url).then((_) => _.json())
+  );
+
+  if (!isDefined(data)) {
```
VWSCoronaDashboard19, line 39: Is it convenient to create a seperate file for line 39 to 45 (repeated piece of code), so it can also be used in the `inline-choropleth.tsx` and  `inline-donut-chart.tsx`?

VWSCoronaDashboard1: Sure, I'll make the change

packages/app/src/components/cms/inline-choropleth.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  const dataOptions = {
+    getLink: isDefined(configuration.link)
+      ? get(useReverseRouter, configuration.link, undefined)
+      : undefined,
+    tooltipVariables: parseTooltipVariables(
+      configuration.tooltipVariables,
+      siteText
+    ),
+  };
```
VWSCoronaDashboard16, line 47: Links are not working right now because it's trying to get the links from the reverseRouter hook itself instead of from the router it returns.

```suggestion
  const { siteText } = useIntl();
  
  const reverseRouter = useReverseRouter();

  const dateUrl = getDataUrl(undefined, undefined, configuration, 'choropleth');

  const { data } = useSWRImmutable(dateUrl, (url: string) =>
    fetch(url).then((_) => _.json())
  );

  if (!isDefined(data)) {
    return <InlineLoader />;
  }

  const dataOptions = {
    getLink: isDefined(configuration.link)
      ? get(reverseRouter, configuration.link, undefined)
      : undefined,
    tooltipVariables: parseTooltipVariables(
      configuration.tooltipVariables,
      siteText
    ),
  };
```

packages/cms/src/schemas/documents/choropleth-configuration.ts
```diff
@@ -0,0 +1,176 @@
...
+      fieldset: 'configuration',
+    },
+    {
+      title: 'Hover stroke width',
+      name: 'hoverStrokeWidth',
+      type: 'number',
+      fieldset: 'configuration',
+    },
+    {
+      title: 'Highlight stroke width',
```
VWSCoronaDashboard16, line 107:
```suggestion
      title: 'Highlight stroke',
```

packages/cms/src/schemas/documents/choropleth-configuration.ts
```diff
@@ -0,0 +1,176 @@
...
+      name: 'highlightStrokeWidth',
+      type: 'number',
+      fieldset: 'configuration',
+    },
+    {
+      title: 'Area stroke width',
```
VWSCoronaDashboard16, line 120:
```suggestion
      title: 'Area stroke',
```

packages/cms/src/schemas/documents/choropleth-configuration.ts
```diff
@@ -0,0 +1,176 @@
...
+      name: 'highlightSelection',
+      type: 'boolean',
+      fieldset: 'dataOptions',
+    },
+    {
+      title: 'Selected code',
+      name: 'selectedCode',
+      type: 'string',
+      fieldset: 'dataOptions',
+    },
```
VWSCoronaDashboard16, line 149: None of these options are passed through to the actual choropleth component, so what's the use of having them here?

packages/cms/src/schemas/documents/age-demographic-chart-configuration.ts
```diff
@@ -0,0 +1,91 @@
...
+      type: 'string',
+      title: 'Kleur',
+      inputComponent: ChartColorInput,
+      validation: (rule: Rule) => rule.required(),
+    },
+    {
+      name: 'maxDisplayValue',
+      type: 'number',
+      title: 'Maximum display value',
+    },
```
VWSCoronaDashboard16, line 89: This one is not passed on to the actual component yet

VWSCoronaDashboard16 (changes requested):
Nice one! Found some small bugs though. And I would have expected that the age demographic chart adhered to the content width, or is it by design that it's wider?
image

packages/app/src/components/cms/inline-choropleth.tsx
VWSCoronaDashboard1: haha, good catch, this was a dumb mistake of mine haha

packages/cms/src/schemas/documents/choropleth-configuration.ts
VWSCoronaDashboard1: That's a mistake on my part as well, I was using a simpler implementation as a reference, but I need to add all of these options to the choropleth config. Well spotted, I obviously was sleeping at the end of the is PR hehe

VWSCoronaDashboard16 (approved):

VWSCoronaDashboard1:
> Nice one! Found some small bugs though. And I would have expected that the age demographic chart adhered to the content width, or is it by design that it's wider? 

Yes indeed, this is by design

PR #3869, opened 2021-11-29
Add back button going to topical page to mobile metric menu

VWSCoronaDashboard16:
Adds a back button to the mobile metric menu which goes to the relevant topical page. Also restyle this button (it already existed on metric pages). Also match this styling for the back button on the VR/GM topical pages.

While working on this I noticed the VR/GM topical index pages were broken since we removed `admissions_on_date_of_reporting` from the choropleth data, so I fixed that as well.


VWSCoronaDashboard1 (approved):

packages/common/src/data/reverse-router.ts
```diff
@@ -1,8 +1,13 @@
...
   const reverseRouter = {
     actueel: {
-      vr: (code: string) => `/actueel/veiligheidsregio/${code}`,
-      gm: (code: string) => `/actueel/gemeente/${code}`,
+      vr: (code?: string) =>
+        code
+          ? `/actueel/veiligheidsregio/${code}`
+          : `/actueel/veiligheidsregio`,
+      gm: (code?: string) =>
+        code ? `/actueel/gemeente/${code}` : `/actueel/gemeente`,
```
VWSCoronaDashboard18, line 9: What was the reason for this oneliner? Shouldn't it be the same codestyling as lines 5-7?

VWSCoronaDashboard16: This is all Prettier. It has a set maximum number of characters per line and wraps things to make each line fit that requirement automatically. The VR line is longer so needs an extra line wrap.

VWSCoronaDashboard18 (approved):

PR #3871, opened 2021-11-30
Sidebar and content styling

VWSCoronaDashboard19:
* Reduced width of sidebar so it looks the same as sidebar on other pages
* Conditionally adding margin to generic component

### DIFF
--- a/packages/app/src/domain/topical/mini-tile.tsx
+++ b/packages/app/src/domain/topical/mini-tile.tsx
@@ -12,14 +12,14 @@ export type MiniTileProps = {
   text: ReactNode;
   warning?: string;
   children: ReactNode;
-  hasMargin?: boolean;
+  hasNoMargin?: boolean;
 };
 
 export function MiniTile(props: MiniTileProps) {
-  const { icon, text, title, warning, children, hasMargin } = props;
+  const { icon, text, title, warning, children, hasNoMargin } = props;
 
   return (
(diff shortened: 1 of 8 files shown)


VWSCoronaDashboard18 (approved):

packages/app/src/domain/topical/mini-tile.tsx
```diff
@@ -12,13 +12,14 @@ export type MiniTileProps = {
   text: ReactNode;
   warning?: string;
   children: ReactNode;
+  hasNoMargin?: boolean;
```
VWSCoronaDashboard1, line 15: This naming has a weird negation in it. I would either called it `hasMargin`, and change the implementation, or call it `hideMargin`.
Looking at the code it also seems to only deal with the left margin, so to make it even clearer, the name ought to be either `hasLeftMargin` or `hideLeftMargin`.

packages/app/src/domain/topical/mini-trend-tile.tsx
```diff
@@ -16,6 +16,7 @@ type MiniTrendTileProps<T extends TimestampedValue = TimestampedValue> = {
   seriesConfig: SeriesConfig<T>;
   dataOptions?: DataOptions;
   values: T[];
+  hasNoMargin?: boolean
```
VWSCoronaDashboard1, line 19: Same comment and the previous one concerning the naming.

VWSCoronaDashboard1 (approved):

PR #3877, opened 2021-12-02
Feature/lazy loading component

VWSCoronaDashboard22:
Added a component to detect it a child is in view and render it when it is

### DIFF
--- a/packages/app/src/components/in-view.tsx
+++ b/packages/app/src/components/in-view.tsx
@@ -1,0 +1,20 @@
+import { ReactNode, useRef, useEffect, useState } from 'react';
+import { useIsInView } from '~/utils/use-is-in-view';
+
+interface InViewProps {
+  children: ReactNode;
+  rootMargin?: string;
+  once?: boolean;
+}
+
+export function InView({ rootMargin, children, once = true }: InViewProps) {
+  const ref = useRef(null);
+  const isInView = useIsInView(ref, rootMargin);
+  const [hasChanged, setHasChanged] = useState<boolean>(false);
+
+  useEffect(() => {
+    isInView && setHasChanged(true);
+  }, [isInView]);
+
+  return <div ref={ref}>{(isInView || (once && hasChanged)) && children}</div>;
+}
--- a/packages/app/src/components/in-view.tsx
+++ b/packages/app/src/components/in-view.tsx
@@ -14,13 +14,17 @@ export function InView({
 }: InViewProps) {
   const ref = useRef(null);
   const isInView = useIsInView(ref, rootMargin);
-  const [hasChanged, setHasChanged] = useState<boolean>(false);
+  const hasChanged = useRef(isInView);
 
   useEffect(() => {
-    isInView && setHasChanged(true);
+    if (isInView) {
+      hasChanged.current = true;
+    }
   }, [isInView]);
 
   return (
-    <div ref={ref}>{(isInView || (triggerOnce && hasChanged)) && children}</div>
+    <div ref={ref}>
+      {(isInView || (triggerOnce && hasChanged.current)) && children}
+    </div>
   );
 }
--- a/packages/app/src/components/in-view.tsx
+++ b/packages/app/src/components/in-view.tsx
@@ -1,4 +1,4 @@
-import { ReactNode, useRef, useEffect, useState } from 'react';
+import { ReactNode, useRef, useEffect } from 'react';
 import { useIsInView } from '~/utils/use-is-in-view';
 
 interface InViewProps {
--- a/packages/app/src/components/in-view.tsx
+++ b/packages/app/src/components/in-view.tsx
@@ -4,10 +4,14 @@ import { useIsInView } from '~/utils/use-is-in-view';
 interface InViewProps {
   children: ReactNode;
   rootMargin?: string;
-  once?: boolean;
+  triggerOnce?: boolean;
 }
 
-export function InView({ rootMargin, children, once = true }: InViewProps) {
+export function InView({
+  rootMargin,
+  children,
+  triggerOnce = true,
+}: InViewProps) {
   const ref = useRef(null);
   const isInView = useIsInView(ref, rootMargin);
   const [hasChanged, setHasChanged] = useState<boolean>(false);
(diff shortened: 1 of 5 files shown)


packages/app/src/components/in-view.tsx
```diff
@@ -0,0 +1,20 @@
+import { ReactNode, useRef, useEffect, useState } from 'react';
+import { useIsInView } from '~/utils/use-is-in-view';
+
+interface InViewProps {
+  children: ReactNode;
+  rootMargin?: string;
+  once?: boolean;
```
VWSCoronaDashboard19, line 7: Maybe a better name for this as it is a boolean, instead of `once`. Prefixing it with `is`, `has` or `can`.

VWSCoronaDashboard20: That name is pre-defined by the package we use. I think we are not able to change this.

VWSCoronaDashboard20 (approved):

VWSCoronaDashboard1 (commented):
I have checked out the branch locally and when switching from the vaccination page to positive tests I get this error:
image

packages/app/src/components/in-view.tsx
VWSCoronaDashboard22: Changed prop name

packages/app/src/components/in-view.tsx
```diff
@@ -0,0 +1,26 @@
...
+  children,
+  triggerOnce = true,
+}: InViewProps) {
+  const ref = useRef(null);
+  const isInView = useIsInView(ref, rootMargin);
+  const [hasChanged, setHasChanged] = useState<boolean>(false);
+
+  useEffect(() => {
+    isInView && setHasChanged(true);
+  }, [isInView]);
```
VWSCoronaDashboard1, line 21: Is this really necessary? `isInview ` already returns a stateful value, so if it changes the component will be re-rendered.

packages/app/src/pages/landelijk/vaccinaties.tsx
```diff
@@ -562,72 +569,74 @@ const VaccinationPage = (props: StaticProps<typeof getStaticProps>) => {
...
-                    text.grafiek_draagvlak.leeftijd_jaar,
-                    { ageGroup: '16 - 24' }
-                  ),
-                  color: colors.data.multiseries.cyan,
-                },
-                {
-                  type: 'invisible',
-                  metricProperty: 'percentage_average',
-                  label: siteText.common.totaal,
+            <InView>
```
VWSCoronaDashboard1, line 572: I see you've added to rootMargin as an optional prop on `InView` perhaps use a negative margin roughly the height of a chart so that the chart gets loaded right before it scrolls into view. This might give a smoother UX because it will prevent charts 'plopping into view' a bit.

VWSCoronaDashboard1 (changes requested):
This PR will reduce the DOM size considerably, because the charts are lazy rendered, therefore improving the user experience.  It doesn't lazy *load* the components though, so the component and data will be part of the initial download. Another improvement here could be to dynamically load the chart component and associated data (which was the original plan behind the ticket).
Have you checked the lighthouse scores before and after this improvement?

packages/app/src/components/in-view.tsx
VWSCoronaDashboard22: Changed it to useRef

packages/app/src/pages/landelijk/vaccinaties.tsx
VWSCoronaDashboard22: Added rootMargins to all InView components

packages/app/src/utils/use-is-in-view.ts
```diff
@@ -0,0 +1,25 @@
+import { useState, useEffect, MutableRefObject, useCallback } from 'react';
+
+export function useIsInView(
+  element: MutableRefObject<HTMLElement | null>,
+  rootMargin?: string
+) {
+  const [isInView, setIsInView] = useState(false);
+  const connect = (observer: IntersectionObserver, htmlElement: HTMLElement) =>
+    observer.observe(htmlElement);
```
VWSCoronaDashboard1, line 9:
```suggestion
  const connect = useCallback(
    (observer: IntersectionObserver, htmlElement: HTMLElement) =>
      observer.observe(htmlElement),
    []
  );
```

packages/app/src/utils/use-is-in-view.ts
```diff
@@ -0,0 +1,25 @@
...
+  useEffect(() => {
+    if (!element.current) return;
+
+    const observer = new IntersectionObserver(
+      ([entry]) => setIsInView(entry.isIntersecting),
+      { rootMargin }
+    );
+
+    connect(observer, element.current);
+    return () => element.current && disconnect(observer, element.current);
```
VWSCoronaDashboard1, line 21:
```suggestion
    if (!element.current) return;
    
    const currentElement = element.current;

    const observer = new IntersectionObserver(
      ([entry]) => setIsInView(entry.isIntersecting),
      { rootMargin }
    );

    connect(observer, currentElement );
    return () => currentElement  && disconnect(observer, currentElement );
```

packages/app/src/utils/use-is-in-view.ts
```diff
@@ -0,0 +1,25 @@
...
+  }, []);
```
VWSCoronaDashboard1, line 22:
```suggestion
  }, [connect, disconnect, element, rootMargin]);
```

packages/app/src/utils/use-is-in-view.ts
VWSCoronaDashboard1: By the way, instead of using callbacks you could also move these two functions outside of the component. Since they have no dependencies on any local data.

VWSCoronaDashboard1 (changes requested):

packages/app/src/utils/use-is-in-view.ts
VWSCoronaDashboard1: There was actually a lint warning that described this issue, so I made these change accordingly.

VWSCoronaDashboard1 (approved):

PR #3890, opened 2021-12-03
Feature/loading indicator

VWSCoronaDashboard18:
Added a Loading Indicator and show it while a topical page loads.


VWSCoronaDashboard22:
Please use the proper code formatting

packages/app/src/components/loader/loader.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+        setinViewHeight(heightOfInView);
+      }}
+    >
+      <Box {...wrapperProps} {...inView} position="absolute">
+        <Box spacing={3} m={3} display="flex" flexDirection="column" alignItems="center" justifyContent="center">
+          <Box spacingHorizontal={1} display={'flex'} alignItems="end">
+            {delayTiming.map((delayPercentage) => <MotionBox {...attributes} transition={{...transitionProps, delay: delayPercentage}} />)}
+          </Box>
+          <Text aria-label={loadingText} lineHeight="1.0">{loadingText}</Text>
```
VWSCoronaDashboard22, line 104: Don't use aria-label or aria-labelledby on any other non-interactive content such as p, legend, li, or ul, because it is ignored. https://www.w3.org/TR/using-aria/#label-support

packages/app/src/components/loader/loader.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const inView:InView = {
+    display: 'flex',
+    alignItems: "center",
+    justifyContent: "center",
+    height: inViewHeight,
+  }
+
+  return (
+    <Box {...wrapperProps} {...toTopOverlay} position="absolute"
+      ref={el => {
```
VWSCoronaDashboard22, line 86: Why not do this with css?

packages/app/src/components/loader/loader.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  const attributes:MotionBoxProps = {
+    height: '3rem',
+    width: '0.75rem',
+    bg: colors.data.scale.blue[0],
+    transformOrigin: 'bottom center',
+    animate: { scaleY: [1, .2, .7, .4, 1] },
+  }
+
+  const transitionProps:TransitionProps = {
```
VWSCoronaDashboard22, line 55: Why use TransitionProps here? Types will be inferred

packages/app/src/components/loader/loader.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const [inViewHeight, setinViewHeight] = useState('0');
+
+  const wrapperProps:WrapperProps = {
+    top: '0',
+    right: '0',
+    bottom: '0',
+    left: '0',
+  }
+
+  const toTopOverlay:ToTopOverlay = {
```
VWSCoronaDashboard22, line 72: What's the use of creating two variables to pass into the Box

packages/app/src/components/loader/loading-router.tsx
```diff
@@ -0,0 +1,38 @@
...
+    router.events.on('routeChangeComplete', () => {
+      setIsLoading(false);
+      setHasCompleted(true);
+    });
+  });
+
+  return (
+    <Box position="relative">
+      {children}
+      { isLoading && <Loader/> }
```
VWSCoronaDashboard22, line 35: Maybe render either the loader or the children.

packages/app/src/components/loader/loading-router.tsx
```diff
@@ -0,0 +1,38 @@
+import React, { useEffect, useState }from 'react';
+import { useRouter } from 'next/router'
+import { Loader } from '~/components/loader/loader';
+import { Box } from '~/components/base';
+
+interface LoadingRouterProps {
+  children?: React.ReactNode;
```
VWSCoronaDashboard22, line 7: Why is children optional?

packages/app/src/components/loader/loading-router.tsx
```diff
@@ -0,0 +1,38 @@
...
+  layout?: string;
```
VWSCoronaDashboard22, line 8: Confusing parameter name. Think this is more about the url than the layout

packages/app/src/components/loader/loading-router.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const { children, layout } = props;
+  const router = useRouter()
+
+  const [isLoading, setIsLoading] = useState(false);
+  const [hasCompleted, setHasCompleted] = useState(false);
+
+  useEffect(() => {
+    router.events.on('routeChangeStart', (url) => {
+      if(url.startsWith("/" + layout)) {
+        setTimeout(() => {
```
VWSCoronaDashboard20, line 21: The way it is set up now, it will start after 700ms. If you set setIsLoading to true before the setTimeOut and in the setTimeOut set it to false, it will stop 700ms after, that's I think the intended behavior. See example below:

` if (url.startsWith('/' + layout)) {
      setIsLoading(true);
        setTimeout(() => {
          setIsLoading(false);
          setHasCompleted(!hasCompleted);
        }, 700);
      }
    });`

packages/app/src/components/loader/loader.tsx
VWSCoronaDashboard18: Because line 85 & line 99

packages/app/src/components/loader/loading-router.tsx
```diff
@@ -0,0 +1,38 @@
...
+          setIsLoading(hasCompleted)
+        }, 700);
+      }
+    });
+    router.events.on('routeChangeComplete', () => {
+      setIsLoading(false);
+      setHasCompleted(true);
+    });
+  });
+
```
VWSCoronaDashboard20, line 31: Do we want to do something with routeChangeError as well?

packages/app/src/components/loader/loading-router.tsx
```diff
@@ -0,0 +1,38 @@
+import React, { useEffect, useState }from 'react';
```
VWSCoronaDashboard20, line 1: loading-router doesn't describe what this component is doing. Can you think of something more descriptive? It will not only make it more readable but also helps you think about your component.

packages/app/src/components/loader/loading-router.tsx
VWSCoronaDashboard1: I think your assumption is incorrect. The requirement is that the loading indicator only is visible after a certain amount of milliseconds. So, the logic that sets the state to loading is correct in this case.

packages/app/src/components/typography.tsx
```diff
@@ -9,6 +9,7 @@ export interface TextProps {
   textTransform?: CSSProperties['textTransform'];
   textAlign?: CSSProperties['textAlign'];
   hyphens?: CSSProperties['hyphens'];
+  lineHeight?: CSSProperties['lineHeight'];
```
VWSCoronaDashboard1, line 12: do NOT add css properties to the Text component. Text styles are only allowed as variants, so no inline css props are allowed. Please refer to the docs: https://minvws.github.io/nl-covid19-data-dashboard/styled-system.html

packages/app/src/domain/layout/gm-layout.tsx
```diff
@@ -140,7 +141,9 @@ export function GmLayout(props: GmLayoutProps) {
           </>
         }
       >
-        <ErrorBoundary>{children}</ErrorBoundary>
+        <LoadingRouter previousUrl={"gemeente/" + code}>
+          <ErrorBoundary>{children}</ErrorBoundary>
+        </LoadingRouter>
```
VWSCoronaDashboard1: See my comment about rendering page as children of the loading indicator. I believe this makes more sense:
```
<ErrorBoundary>
  <LoadingRouter previousUrl={"gemeente/" + code}/>
  {children}
</ErrorBoundary>
```

packages/app/src/components/loader/loader.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  return (
+    <Box {...wrapperProps} {...toTopOverlay} position="absolute"
+      ref={el => {
+        if (!el) return;
+
+        const windowHeight = window.innerHeight;
+        const rect = el.getBoundingClientRect();
+        const elTop = (rect.top < 0 ? 0 : rect.top);
+        const elBottom = rect.bottom > windowHeight ? windowHeight : rect.bottom;
```
VWSCoronaDashboard1: As previously indicated, this is better solved using CCS.
Another problem with this code is that calling functions like `getBoundingClientRect` and even assigning `window.innerHeight` will trigger a reflow in the browser, so you potentially start layout thrashing here.

packages/app/src/components/loader/loading-router.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  const [isLoading, setIsLoading] = useState(false);
+  const [hasCompleted, setHasCompleted] = useState(false);
+
+  useEffect(() => {
+    router.events.on('routeChangeStart', (url) => {
+      if(url.startsWith("/" + previousUrl)) {
+        setTimeout(() => {
+          setIsLoading(hasCompleted)
+        }, 700);
```
VWSCoronaDashboard1, line 23: The requirement stated 700ms as a starting off point, but it was also mentioned that this value should be tweaked during development to find the sweet spot. Did you do you this and did 700ms turn out to be exactly the sweet spot?

packages/app/src/components/loader/loader.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+        const heightOfInView = elBottom - elTop + 'px';
+
+        setinViewHeight(heightOfInView);
+      }}
+    >
+      <Box {...wrapperProps} {...inView} position="absolute">
+        <Box spacing={3} m={3} display="flex" flexDirection="column" alignItems="center" justifyContent="center">
+          <Box spacingHorizontal={1} display={'flex'} alignItems="end">
+            {delayTiming.map((delayPercentage, i) => <MotionBox key={i} {...attributes} transition={{...transitionProps, delay: delayPercentage}} />)}
```
VWSCoronaDashboard1: nitpick: The requirements state that movement of the bars should happen at random, as far as I can see, currently they are deterministic. ;)

packages/app/src/components/loader/loader.tsx
VWSCoronaDashboard1: Its probably cleaner to just create two styled div instances and to give them a proper name that reflects their usage. That's sort of the point of styled components, imho.

packages/app/src/components/loader/loader.tsx
```diff
@@ -0,0 +1,122 @@
...
+          justifyContent="center"
+        >
+          <Box spacingHorizontal={1} display={'flex'} alignItems="end">
+            {delayTiming.map((delayPercentage, i) => (
+              <MotionBox
+                key={i}
+                {...attributes}
+                transition={{ ...transitionProps, delay: delayPercentage }}
+              />
+            ))}
```
VWSCoronaDashboard16, line 114: Using motion elements for this is quite expensive performance wise. I think the animation you are building here can also easily be defined as a css animation, no?

packages/app/src/components/loader/loading-router.tsx
```diff
@@ -0,0 +1,38 @@
...
+      }
+    });
+    router.events.on('routeChangeComplete', () => {
+      setIsLoading(false);
+      setHasCompleted(true);
+    });
+  });
+
+  return (
+    <Box position="relative">
```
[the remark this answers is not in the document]
VWSCoronaDashboard18: Then it needs an additional relative box on the parent of the siblings. Now it is a generic solution and works with any parent.

packages/app/src/components/loader/loading-router.tsx
VWSCoronaDashboard18: When testing 700ms felt right to me. I tried other values lower and higher, but this looked good.

packages/app/src/components/loader/loading-wrapper.tsx
```diff
@@ -0,0 +1,53 @@
...
+    const handleRouteChangeStart  = (url) => {
+      setRouterLoadState('idle');
+      window.scrollTo(0, 0);
+      console.log(routerLoadState);
+      if (url.startsWith('/' + previousUrl) && routerLoadState === 'idle' && currentRoute !== url) {
+        setTimeout(() => {
+          setRouterLoadState('loading');
+        }, 700);
+      }
+    }
```
VWSCoronaDashboard1, line 26:
```suggestion
    const handleRouteChangeStart  = (url, { shallow }: { shallow: boolean }) => {
      if (shallow) return;
      setRouterLoadState('idle');
      window.scrollTo(0, 0);
      setTimeout(() => {
        setRouterLoadState('loading');
      }, 700);
    }
```

packages/app/src/components/loader/loading-wrapper.tsx
```diff
@@ -0,0 +1,53 @@
...
+      router.events.off('routeChangeComplete', handleRouteChangeComplete);
+      router.events.off('routeChangeError', handleRouteChangeError);
+    }
+  }, [currentRoute, previousUrl, router.events, routerLoadState]);
+
+  return (
+    <>
+      {(routerLoadState === 'loading' || routerLoadState === 'idle') && <Loader showLoader={routerLoadState === 'loading'} />}
+    </>
+  );
```
VWSCoronaDashboard1, line 52:
```suggestion
return state === 'complete'
  ? null
  : <Loader showLoader={routerLoadState === 'loading'} />;
```

packages/app/src/components/loader/loading-wrapper.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+    router.events.on('routeChangeStart', handleRouteChangeStart);
+    router.events.on('routeChangeComplete', handleRouteChangeComplete);
+    router.events.on('routeChangeError', handleRouteChangeError);
+
+    return () => {
+      router.events.off('routeChangeStart', handleRouteChangeStart);
+      router.events.off('routeChangeComplete', handleRouteChangeComplete);
+      router.events.off('routeChangeError', handleRouteChangeError);
+    }
```
VWSCoronaDashboard1, line 45: the setTimeout method on line 23  doesn't get cleaned up here

packages/app/src/components/loader/loading-wrapper.tsx
VWSCoronaDashboard1: I think checking the `shallow` param probably negates the need for the `currentRoute` state and the `previousUrl` prop. It will need some testing and tweaking obviously, but I think its worth investigating.

VWSCoronaDashboard1 (changes requested):

packages/app/src/components/loader/loader.tsx
VWSCoronaDashboard18: Won't do

packages/app/src/components/loader/loader.tsx
VWSCoronaDashboard18: It needs this for the right CSS heigt

packages/app/src/components/loader/loader.tsx
```diff
@@ -0,0 +1,126 @@
...
+      ref={(el) => {
+        if (!el) return;
+
+        const windowHeight = window.innerHeight;
+        const rect = el.getBoundingClientRect();
+        const elTop = rect.top < 0 ? 0 : rect.top;
+        const elBottom =
+          rect.bottom > windowHeight ? windowHeight : rect.bottom;
+
+        const heightOfInView = elBottom - elTop + 'px';
```
VWSCoronaDashboard14, line 77: This isn't possible with regular CSS? I saw a comment from @VWSCoronaDashboard1 also mentioning this.

packages/app/src/style/preset.ts
```diff
@@ -88,6 +88,9 @@ export const preset = {
       textOverflow: 'ellipsis',
       mt: 0,
     },
+    loaderText: {
+      lineHeight: 1,
+    },
```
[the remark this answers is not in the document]
VWSCoronaDashboard18: https://minvws.github.io/nl-covid19-data-dashboard/styled-system.html#:~:text=Typography,body1%20preset%3C/Text%3E.

packages/app/src/components/loader/loader.tsx
VWSCoronaDashboard18: No it's not. Because css doesn't know the height of the top of the parent till the bottom of the parent OR the fold.

VWSCoronaDashboard16: What about 
```
height: 100%;
maxHeight: 100vh;
```?

Comment (VWSCoronaDashboard18) [reply packages/app/src/components/loader/loader.tsx:77]: Because 100vh is to high, the header and navigation height needs to be subtracted.

File: packages/app/src/components/loader/loading-wrapper.tsx [hunk 23]
```diff
@@ -0,0 +1,54 @@
...
+export function LoadingWrapper(props: LoadingRouterProps) {
+  const { previousUrl } = props;
+  const router = useRouter();
+
+  const [routerLoadState, setRouterLoadState] = useState<
+    'idle' | 'loading' | 'complete'
+  >('loading');
+  const [currentRoute, setCurrentRoute] = useState<string>('');
+  const handleLoadingTimeout = () => setRouterLoadState('loading');;
+  const loaderTimeout = useRef(() => setTimeout(handleLoadingTimeout, 700));
```
VWSCoronaDashboard1, line 18: This code doesn't make sense I'm afraid. The correct way to handle this is to create an empty ref and each time that `setTimeout()` is called assign the resulting id to it.
I.e. `const timeoutIdRef = useRef(0);`

Your `clearTimeout` calls in the rest of the file would otherwise always be called with a different id, so the original timeout would never get cleared. Did you run these code changes? Because I suspect that the loader would now always pop up, regardless of the timeout clearances.

packages/app/src/components/loader/loading-wrapper.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  useEffect(() => {
+    const handleRouteChangeStart = (url: string, { shallow }: { shallow: boolean }) => {
+      if (shallow) return;
+      setRouterLoadState('idle');
+      window.scrollTo(0, 0);
+      url.startsWith('/' + previousUrl) && currentRoute !== url && loaderTimeout.current();
```
VWSCoronaDashboard1, line 25:
```suggestion
      if (url.startsWith('/' + previousUrl) && currentRoute !== url) {
        timeoutIdRef.current = setTimeout(handleLoadingTimeout, 700);  
      }
```

packages/app/src/components/loader/loading-wrapper.tsx
```diff
@@ -0,0 +1,54 @@
...
+    }
+    
+    const handleRouteChangeComplete = (url: string) => {
+      clearTimeout(loaderTimeout.current());
```
VWSCoronaDashboard1, line 29:
```suggestion
      if (timeoutIdRef.current > 0) {
        clearTimeout(timeoutIdRef.current);
        timeoutIdRef.current = 0;
      }
```

packages/app/src/components/loader/loading-wrapper.tsx
```diff
@@ -0,0 +1,54 @@
...
+      setRouterLoadState('complete');
+      setCurrentRoute(url);
+    };
+
+    const handleRouteChangeError = () => {
+      clearTimeout(loaderTimeout.current());
```
VWSCoronaDashboard1, line 35:
```suggestion
      if (timeoutIdRef.current > 0) {
        clearTimeout(timeoutIdRef.current);
        timeoutIdRef.current = 0;
      }```

File: packages/app/src/components/loader/loading-wrapper.tsx [hunk 27]
```diff
@@ -0,0 +1,54 @@
...
+
+    router.events.on('routeChangeStart', handleRouteChangeStart);
+    router.events.on('routeChangeComplete', handleRouteChangeComplete);
+    router.events.on('routeChangeError', handleRouteChangeError);
+
+    return () => {
+      router.events.off('routeChangeStart', handleRouteChangeStart);
+      router.events.off('routeChangeComplete', handleRouteChangeComplete);
+      router.events.off('routeChangeError', handleRouteChangeError);
+      clearTimeout(loaderTimeout.current());
```
VWSCoronaDashboard1, line 47:
```suggestion
      if (timeoutIdRef.current > 0) {
        clearTimeout(timeoutIdRef.current);
        timeoutIdRef.current = 0;
      }```

File: packages/app/src/components/loader/loader.tsx [hunk 28]
```diff
@@ -0,0 +1,126 @@
...
+  zIndex: number;
+  bg: string;
+}
+
+interface InView {
+  display: string;
+  alignItems: string;
+  justifyContent: string;
+  height: string;
+}
```
VWSCoronaDashboard1, line 27: As mentioned before, instead of having all of these css property interfaces here, its better to just create a few styled divs with the appropriate styles configured there.
In the same way that LoaderBar is styled. You're mixing two styles of coding here, so for consistency sake its best to choose one approach and in the rest of the code base its preferred to use styled components.

VWSCoronaDashboard1 (dismissed):

VWSCoronaDashboard16:
I think we need to up some z-index somewhere 
image

VWSCoronaDashboard20 (approved):

packages/app/src/components/loader/loader.tsx
```diff
@@ -0,0 +1,128 @@
...
+  })
+);
+
+const BarWrapper = styled.div(() =>
+  css({
+    display: 'flex',
+    alignItems: 'start',
+    '& > *:not(:last-child)': {
+      marginRight: asResponsiveArray(1),
+    }
```
VWSCoronaDashboard14, line 86: I think the spacing prop on a box component works perfectly here.

packages/app/src/components/loader/loader.tsx
```diff
@@ -0,0 +1,128 @@
...
+const LoaderWrapper = styled.div(() =>
+  css({
+    margin: asResponsiveArray(3),
+    display: 'flex',
+    flexDirection: 'column',
+    alignItems: 'center',
+    justifyContent: 'center',
+    '& > *:not(:last-child)': {
+      marginBottom: asResponsiveArray(3),
+    },
```
VWSCoronaDashboard14, line 100: The same goes for here, and also the `asResponsiveArray` received now 1 argument (which is not an object nor an array), so I'm not sure if it even does the job. But maybe we should stronger type this function to avoid confusion.

VWSCoronaDashboard14 (commented):
If you're having a folder name that exports the same name for the component it might look a bit strange in the folder structure. We tend to use a barrel file for that, you can see an example in the folder for the page-information-block, and see what the `index.ts` does.

packages/app/src/components/loader/loader.tsx
```diff
@@ -0,0 +1,128 @@
...
+import css from '@styled-system/css';
+import { Text } from '~/components/typography';
+import { useIntl } from '~/intl';
+import { VisuallyHidden } from '../visually-hidden';
+import { asResponsiveArray } from '~/style/utils';
+
+interface LoaderProps {
+  showLoader?: boolean;
+}
+const duration = .7;
```
VWSCoronaDashboard14, line 13: When using constants we tend to use capital letters for the variable name `const DURATION = 0.7`.

packages/app/src/components/loader/loader.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+export function Loader(props: LoaderProps) {
+  const { showLoader = true } = props;
+  const { siteText } = useIntl();
+
+  const delayTiming = [0, 0.3, 0.5, 0.2, 0.4];
```
VWSCoronaDashboard14, line 19: Since these are almost the same as duration and won't change I would also put them above the component and make the variable in capitals letters.

packages/app/src/components/loader/loader.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+  const loadingText = siteText.common.loading_text;
```
VWSCoronaDashboard14, line 21: I don't think it's necessary to create a variable to hold this text that only gets used once. Just reference in the `<Text>` component directly.

packages/app/src/components/loader/loader.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+const LoaderOverlay = styled.div(() =>
+  css({
+    position: 'absolute',
+    top: '0',
+    right: '0',
+    bottom: '0',
+    left: '0',
+    zIndex: 999,
+    bg: colors.white,
```
VWSCoronaDashboard14, line 126: You can also use the string notation how we're using it in most places `'colors.white'`.

packages/app/src/style/preset.ts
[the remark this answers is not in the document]
VWSCoronaDashboard14: Is this resolved, I believe the variant is still there. :)

packages/app/src/components/loader/loading-wrapper.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export function LoadingWrapper(props: LoadingRouterProps) {
+  const { previousUrl } = props;
+  const router = useRouter();
+
+  const [routerLoadState, setRouterLoadState] = useState<
+    'idle' | 'loading' | 'complete'
+  >('loading');
+  const [currentRoute, setCurrentRoute] = useState<string>('');
+  const handleLoadingTimeout = () => setRouterLoadState('loading');;
```
VWSCoronaDashboard14, line 17: I see a double `;;`, are you sure the auto-linter is working on save?

packages/app/src/components/loader/loader.tsx
VWSCoronaDashboard1: 'white' will do, no need for the colors prefix

packages/app/src/components/loader/loader.tsx
VWSCoronaDashboard18: It's used on lines 51 & 52 too.

packages/app/src/components/loader/loader.tsx
```diff
@@ -0,0 +1,128 @@
...
+          <BarWrapper>
+            {delayTiming.map((delayTiming, i) => (
+              <LoaderBar
+                key={i}
+                delay={`-${delayTiming}s`}
+              />
+            ))}
+          </BarWrapper>
+          <VisuallyHidden>{loadingText}</VisuallyHidden>
+          <Text variant="loaderText" >{loadingText}</Text>
```
VWSCoronaDashboard14, line 52: We don't need the visually hidden text if it contains the same text as the regular text

packages/app/src/components/loader/loader.tsx
VWSCoronaDashboard14: Oops, you are right

packages/app/src/components/loader/loader.tsx
VWSCoronaDashboard18: https://github.com/minvws/nl-covid19-data-dashboard/pull/3890#discussion_r764127407

PR #3912, opened 2021-12-09
Min height fix for loader indicator

VWSCoronaDashboard18:
bugfix


VWSCoronaDashboard19 (approved):
Checked locally and looks good now

PR #3923, opened 2021-12-16
Get page lokalize texts via static props for deceased page

VWSCoronaDashboard22:
Added a function to extract lokalize texts from a lokalize export file and used in on the deceased page.

### DIFF
--- a/packages/app/src/static-props/get-data.ts
+++ b/packages/app/src/static-props/get-data.ts
@@ -384,11 +384,13 @@ export function getLokalizeTexts<T extends keyof SiteText>(
   keys: T[],
   locale: keyof Languages
 ) {
-  return keys.reduce(
-    (acc, key) => ({
-      ...acc,
-      [key]: languages[locale][key],
-    }),
-    {}
-  ) as Pick<SiteText, T>;
+  return {
+    pageText: keys.reduce(
+      (acc, key) => ({
+        ...acc,
+        [key]: languages[locale][key],
+      }),
+      {}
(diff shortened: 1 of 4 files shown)


packages/app/src/static-props/get-data.ts
```diff
@@ -369,3 +371,24 @@ function replaceInaccurateLastValue(data: any) {
...
+  locale: keyof Languages
+) {
+  return keys.reduce(
+    (acc, key) => ({
+      ...acc,
+      [key]: languages[locale][key],
+    }),
+    {}
+  ) as Pick<SiteText, T>;
+}
```
VWSCoronaDashboard1, line 394:
```suggestion
  return {
    pageText: keys.reduce(
    (acc, key) => ({
      ...acc,
      [key]: languages[locale][key],
    }),
    {}
  ) as Pick<SiteText, T>
  } as const;
}
```

VWSCoronaDashboard1 (changes requested):

VWSCoronaDashboard1 (changes requested):
One last comment for this: Don't forget to make the same changes on the VR and GM page as well

packages/app/src/static-props/get-data.ts
VWSCoronaDashboard22: Committed this

VWSCoronaDashboard1 (approved):

VWSCoronaDashboard20 (approved):

PR #3966, opened 2022-01-07
feat(page-lvl-banners): Add optional banners to each metric page

VWSCoronaDashboard19:
- Letting the user decide which metric page on GM and VR level should show a warning via Sanity, this is depended on the GM that is filled in Sanity.
- These warnings will show on the metric page of the GM and the VR that it is in.
- Add warning for sewer chart with the same philosophy
- New util function for returning correct text


packages/app/src/components/page-information-block/page-information-block.tsx
```diff
@@ -30,6 +35,8 @@ interface InformationBlockProps {
   id?: string;
   category?: string;
   screenReaderCategory?: string;
+  name?: string;
```
VWSCoronaDashboard24, line 38: Maybe rename to `vrNameOrGmName`? I think just `name` is a bit generic?

packages/app/src/components/page-information-block/page-information-block.tsx
```diff
@@ -43,7 +50,25 @@ export function PageInformationBlock({
...
+      gm.name === scopedGmName ||
+      (gm.searchTerms && gm.searchTerms.includes(scopedGmName))
+  );
+  const scopedVr = vrData.find((vr) => vr.code === scopedGm?.vrCode);
+
+  const scopedWarning = getScopedGmWarning(
+    scopedGmName,
+    name || '',
+    warning || '',
+    scopedVr?.name || ''
```
VWSCoronaDashboard24, line 69: Maybe you can move the `x || ''` to `getScopedGmWarning()`? Then you only need to do this in one spot?

VWSCoronaDashboard24 (approved):

packages/app/src/components/page-information-block/page-information-block.tsx
VWSCoronaDashboard19: Agreed, moving the whole logic above can be moved to the util function instead

VWSCoronaDashboard18 (approved):
Thumbs up

PR #3971, opened 2022-01-12
Feature/booster coverage

VWSCoronaDashboard19:
Hold for BE to be ready
- [x] Ready


packages/app/schema/nl/booster_coverage.json
```diff
@@ -0,0 +1,39 @@
...
+    },
+    "last_value": {
+      "$ref": "#/definitions/value"
+    }
+  },
+  "required": ["values", "last_value"],
+  "additionalProperties": false,
+  "definitions": {
+    "value": {
+      "title": "nl_booster_coverage",
```
VWSCoronaDashboard20, line 22: The key has the same title as line 3, this results in interface names with a 1 behind them when generated based on this file. Are you sure this is correct?

packages/common/src/types/data.ts
```diff
@@ -414,6 +414,15 @@ export interface NlBoosterShotPerAgeGroupValue {
   date_end_unix: number;
   birthyear_range: string;
 }
+export interface NlBoosterCoverage {
+  values: NlBoosterCoverage1[];
+  last_value: NlBoosterCoverage1;
+}
+export interface NlBoosterCoverage1 {
```
VWSCoronaDashboard20, line 421: The number 1 behind the interface usually indicates a duplicate key; see the previous comment.

VWSCoronaDashboard20 (changes requested):

packages/app/schema/nl/booster_coverage.json
VWSCoronaDashboard19: Sharp! that's correct, will double check it again and verify with be!

packages/app/schema/nl/booster_coverage.json
```diff
@@ -0,0 +1,39 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "nl_booster_coverage",
+  "type": "object",
+  "properties": {
+    "values": {
+      "type": "array",
+      "minItems": 1,
+      "maxItems": 1,
```
VWSCoronaDashboard24, line 9: Is this correct? Max 1 value in the array?

VWSCoronaDashboard18: Yes, there is no historical data in this entry point

VWSCoronaDashboard18 (approved):

VWSCoronaDashboard24 (approved):

VWSCoronaDashboard22 (approved):

packages/app/src/pages/index.tsx
```diff
@@ -559,7 +563,10 @@ const Home = (props: StaticProps<typeof getStaticProps>) => {
                               .booster_shots_administered_data_drive_text,
                             {
                               percentage: formatPercentage(
-                                boosterCoverageLastValue.percentage
+                                boosterCoverageFeature.isEnabled &&
```
VWSCoronaDashboard19, line 566: Added a feature flag here and also checking if the currently existing percentage is in `boosterCoverageEstimatedLastValue`. When `booster_coverage` data goes live, it won't show NaN on the dashboard, but switch over to the new data in the `else`

VWSCoronaDashboard19 (commented):
Small note added

PR #3981, opened 2022-01-18
fix: Remove feature flag and replace with fallback data

VWSCoronaDashboard19:
Safer approach for showing new data when old data has been removed by BE, so no `NaN` will be shown on the DB.


VWSCoronaDashboard22:
Can we also remove the feature from packages/common/src/feature-flags/features.ts?

VWSCoronaDashboard24 (approved):
This should be safe!
Agreed with @VWSCoronaDashboard22; let's see if the feature flag can be removed entirely.

VWSCoronaDashboard18 (approved):
Great!

PR #3982, opened 2022-01-18
Fix is touch check for android

VWSCoronaDashboard22:


packages/app/src/utils/use-is-touch-device.tsx
```diff
@@ -13,7 +13,11 @@ export function IsTouchDeviceContextProvider({
 }: {
   children: ReactNode;
 }) {
-  const isTouchDevice = useMediaQuery('(hover: none)');
+  let isTouchDevice = false;
+
+  if (typeof window !== 'undefined') {
+    isTouchDevice = isTouch();
+  }
```
VWSCoronaDashboard24, line 20: Maybe move this to the hook? We can just return `false` when no `window` is available. Now there are two places with logic.

packages/app/src/utils/is-touch.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Check wether a user is using a touch device.
+ * @returns boolean
+ */
+export function isTouch() {
```
VWSCoronaDashboard24, line 5: Maybe rename to `isTouchCapableDevice`? Or `isTouchDevice`?

VWSCoronaDashboard22: Renamed

packages/app/src/utils/use-is-touch-device.tsx
VWSCoronaDashboard22: Moved logic to isTouchDevice

VWSCoronaDashboard19 (approved):

VWSCoronaDashboard24 (approved):

PR #3994, opened 2022-01-25
Fix change in threshold per MapType logic

VWSCoronaDashboard18:
Fix for mismatch in treshold scale for the vr map

### DIFF
--- a/packages/app/src/pages/landelijk/ziekenhuis-opnames.tsx
+++ b/packages/app/src/pages/landelijk/ziekenhuis-opnames.tsx
@@ -425,6 +425,7 @@ const IntakeHospital = (props: StaticProps<typeof getStaticProps>) => {
                   key: 'hospital_admissions_region_choropleth',
                 }}
                 map="vr"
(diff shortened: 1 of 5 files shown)


packages/app/src/pages/landelijk/ziekenhuis-opnames.tsx
```diff
@@ -425,6 +425,7 @@ const IntakeHospital = (props: StaticProps<typeof getStaticProps>) => {
                   key: 'hospital_admissions_region_choropleth',
                 }}
                 map="vr"
+                thresholdMap="gm"
```
VWSCoronaDashboard19, line 428: Doesn't this need to be `vr`?

VWSCoronaDashboard18: Nope, it uses the GM scale because of the amount of cases

VWSCoronaDashboard19 (approved):
Looks good to me

PR #4000, opened 2022-01-31
Feature/match lokalize page names with page identifiers

VWSCoronaDashboard22:


packages/cms/src/lokalize/key-mutations.csv
```diff
@@ -816,3 +816,352 @@ timestamp,action,key,document_id,move_to
 2022-01-28T16:11:28.182Z,add,pages.vaccinations.nl.vaccination_grade_toggle_tile.age_18_plus.description_booster_grade_not_available,4RQC5Uo7Milcl0woBPcA6t,__
 2022-01-28T16:11:29.308Z,add,pages.vaccinations.nl.vaccination_grade_toggle_tile.age_18_plus.description_vaccination_one_shot_with_percentage,dEVJPM3ZTb5nede2drFRej,__
 >>>>>>> develop
```
VWSCoronaDashboard19, line 818: I think this shouldn't be here?

VWSCoronaDashboard19 (changes requested):

packages/cms/src/lokalize/key-mutations.csv
VWSCoronaDashboard22: Removed this

VWSCoronaDashboard19 (approved):

PR #4013, opened 2022-02-02
Fix/vaccination page adjustments

VWSCoronaDashboard19:


packages/app/src/pages/landelijk/vaccinaties.tsx
```diff
@@ -323,6 +335,39 @@ const VaccinationPage = (props: StaticProps<typeof getStaticProps>) => {
             }}
           />
 
+          {vaccineAdministeredGgdFeature.isEnabled &&
+            vaccineAdministeredHospitalsAndCareInstitutionsFeature.isEnabled &&
+            vaccineAdministeredDoctorsFeature.isEnabled &&
+            vaccineAdministeredGgdGhorFeature.isEnabled && (
```
VWSCoronaDashboard24, line 341: Do we actually need so many feature flags? Wouldn't just one suffice?

VWSCoronaDashboard24 (approved):

packages/app/src/pages/landelijk/vaccinaties.tsx
VWSCoronaDashboard19: I think that's because the component that is behind this flag needs to make sure all of these metric data (schema's) are available, so it won't break when one isn't.

PR #4020, opened 2022-02-04
Restore old method of defining is touch device

VWSCoronaDashboard22:


packages/app/src/utils/use-is-touch-device.tsx
```diff
@@ -1,5 +1,5 @@
 import { createContext, ReactNode, useContext } from 'react';
-import { isTouchDevice } from './is-touch-device';
```
VWSCoronaDashboard18, line 2: Is this file still being used? Otherwise maybe delete that file in that path?

VWSCoronaDashboard22: Not yet. We want to leave it there for now

VWSCoronaDashboard20 (approved):

VWSCoronaDashboard18 (approved):

PR #4021, opened 2022-02-07
Updated icons with correct viewbox and relative sizes

VWSCoronaDashboard18:
Update size of icons inside the viewbox for coherent looks


VWSCoronaDashboard24 (approved):
Hard to review, but I will check new behavior during QA.

PR #4038, opened 2022-02-15
Created a eslint base config and extend from this in each package

VWSCoronaDashboard22:


package.json
```diff
@@ -11,6 +11,7 @@
     "nuke": "yarn workspaces foreach run nuke && npx rimraf node_modules",
     "typecheck": "yarn workspaces foreach run typecheck",
     "lint": "yarn workspaces foreach run lint",
+    "lint:fix": "yarn workspaces foreach run lint --fix",
```
VWSCoronaDashboard24, line 14: 🙌

VWSCoronaDashboard24 (approved):
Very nice improvement! 👍

eslint-overrides.json
```diff
@@ -0,0 +1,38 @@
+{
```
VWSCoronaDashboard19, line 1: Why is there a separate file for the overrides? Can't we put these inside `eslint-config.json`?

VWSCoronaDashboard22: No, doesn't get picked up if we do it like that

VWSCoronaDashboard19 (approved):

PR #4042, opened 2022-02-16
bugfix: use correct key for the stap_1-horeca_pertafel.svg icon

VWSCoronaDashboard25:
Icon doesn't show without this fix.
Before:
image
After:
image


packages/app/src/domain/restrictions/restriction-icons.ts
```diff
@@ -35,7 +35,7 @@ export const restrictionIcons: Record<string, string | null> = {
   geenWedstrijden: 'geen-wedstrijden.svg',
   sporterMetZweetband: 'sporter-met-zweetband.svg',
   'stap_1-horeca_max': 'stap_1-horeca_max.svg',
-  'stap-1_horeca_per_tafel': 'stap_1-horeca_pertafel.svg',
+  'stap_1-horeca_per_tafel': 'stap_1-horeca_pertafel.svg',
```
VWSCoronaDashboard19, line 38: Sharp!

VWSCoronaDashboard19 (approved):

PR #4048, opened 2022-02-21
fix: align max display value with x-axis percentage

VWSCoronaDashboard19:
- Changed`maxDisplayValue` to max percentage, so it aligns better and shows correct colors on bars


VWSCoronaDashboard25 (approved):

VWSCoronaDashboard24 (approved):
Aha! 👍

PR #4052, opened 2022-02-22
feat(archived): toggle archived charts

VWSCoronaDashboard19:
### Summary

- Added button to toggle archived charts on every page
- Added new Sanity keys

### DIFF
--- a/packages/app/src/components/page-information-block/page-information-block.tsx
+++ b/packages/app/src/components/page-information-block/page-information-block.tsx
@@ -1,6 +1,6 @@
 import css from '@styled-system/css';
 import { Warning } from '@corona-dashboard/icons';
-import { isValidElement, ReactNode } from 'react';
+import { isValidElement, ReactNode, useState } from 'react';
 import styled from 'styled-components';
 import { ArticleSummary } from '~/components/article-teaser';
 import { Box } from '~/components/base';
@@ -15,6 +15,7 @@ import { Metadata, MetadataProps } from './components/metadata';
 import { PageLinks } from './components/page-links';
 import { WarningTile } from '~/components/warning-tile';
 import { useScopedWarning } from '~/utils/use-scoped-warning';
+import { useIntl } from '~/intl';
 
 interface InformationBlockProps {
   title?: string;
@@ -35,6 +36,8 @@ interface InformationBlockProps {
   screenReaderCategory?: string;
   vrNameOrGmName?: string;
   warning?: string;
+  hasButton?: boolean;
+  onArchivedClick?: any;
 }
 
 export function PageInformationBlock({
@@ -50,8 +53,16 @@ export function PageInformationBlock({
   screenReaderCategory,
   vrNameOrGmName,
   warning,
+  hasButton,
+  onArchivedClick,
 }: InformationBlockProps) {
   const scopedWarning = useScopedWarning(vrNameOrGmName || '', warning || '');
+  const { siteText } = useIntl();
+  const [isToggled, toggleArchived] = useState(false);
+  const handleClick = () => {
+    toggleArchived(!isToggled);
+    onArchivedClick(isToggled);
+  };
 
   const MetaDataBlock = metadata ? (
     <MetadataBox>
@@ -133,6 +144,15 @@ export function PageInformationBlock({
               </>
             )}
           </Box>
+          <Box my={4}>
+            {hasButton && (
+              <Button type="button" onClick={handleClick} isActive={isToggled}>
+                {!isToggled
+                  ? siteText.common.show_archived
+                  : siteText.common.hide_archived}
+              </Button>
+            )}
+          </Box>
         </Tile>
       )}
     </Box>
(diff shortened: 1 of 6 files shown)


packages/app/src/components/page-information-block/page-information-block.tsx
```diff
@@ -35,6 +36,8 @@ interface InformationBlockProps {
   screenReaderCategory?: string;
   vrNameOrGmName?: string;
   warning?: string;
+  hasButton?: boolean;
+  onArchivedClick?: any;
```
VWSCoronaDashboard24, line 40:
```suggestion
  onArchivedClick?: () => void;
```

packages/app/src/components/page-information-block/page-information-block.tsx
```diff
@@ -50,8 +53,16 @@ export function PageInformationBlock({
...
   warning,
+  hasButton,
+  onArchivedClick,
 }: InformationBlockProps) {
   const scopedWarning = useScopedWarning(vrNameOrGmName || '', warning || '');
+  const { siteText } = useIntl();
+  const [isToggled, toggleArchived] = useState(false);
+  const handleClick = () => {
+    toggleArchived(!isToggled);
+    onArchivedClick(isToggled);
```
VWSCoronaDashboard24, line 64: Since this prop is optional; would it make sense to check if it exists before calling it?

packages/app/src/pages/landelijk/positief-geteste-mensen.tsx
```diff
@@ -100,6 +100,10 @@ const PositivelyTestedPeople = (props: StaticProps<typeof getStaticProps>) => {
   const { siteText, formatNumber, formatPercentage, formatDateFromSeconds } =
     useIntl();
   const reverseRouter = useReverseRouter();
+  const [isToggled, toggleArchived] = useState(false);
```
VWSCoronaDashboard24, line 103: Maybe you can rename this to something like `showArchivedContent`? `isToggled` is maybe a bit generic here?

packages/app/src/components/page-information-block/page-information-block.tsx
VWSCoronaDashboard25, line 40: Typically a `onClick` function is typed `onClick: () => void` or `onClick: (event: MouseEvent) => void`.
Here you want a `(boolean) => void` I think, so it's better to name it something that represents that.
I'll try to think for something.

packages/app/src/components/page-information-block/page-information-block.tsx
```diff
@@ -50,8 +53,16 @@ export function PageInformationBlock({
   screenReaderCategory,
   vrNameOrGmName,
   warning,
+  hasButton,
```
VWSCoronaDashboard25, line 56: It's not just a button, it's specifically a "hide archived charts" button. I think I would name this `hasHideArchivedButton`.

Being so verbose, makes sure that on the page that passes this property, it's clear what is expected from this component. Especially if that page is expected to pass a `hideArchivedCharts` and `setHideArchivedCharts` if they flick this boolean to `true`.

packages/app/src/components/page-information-block/page-information-block.tsx
```diff
@@ -50,8 +54,12 @@ export function PageInformationBlock({
   screenReaderCategory,
   vrNameOrGmName,
   warning,
+  hasHideArchivedButton,
```
VWSCoronaDashboard18, line 57: Is this optional? Or could this be determined by the archived hidden prop? Because if that is set, it also needs the button. They seem dependent on each other.

VWSCoronaDashboard19: It's optional on line 39

VWSCoronaDashboard18: I understand that. I did not meant an optional prop, but if this logic is needed. My thought was: if the toggle is either true or false the button is needed, making the prop unnecessary. You only want that toggle of the state is true or false. En you don’t want the button if the hide or show state is null/not set.

VWSCoronaDashboard25: I think it's a matter of code reduction versus code readability.
Removing this prop and relying on `isArchivedHidden` and `onToggleArchived` being defined, means that further on the logic changes into:

```
{typeof isArchivedHidden !== 'undefined' && typeof onToggleArchived !== 'undefined' && (
  <Button ... />
)}
```

I personally prefer to tell the reader of the code what is going on, instead of having them to keep track of what it all means while they came to read the code to probably understand something else, possibly solving an important bug.

So instead of literally typing the logic as above, you could keep a descriptive variable, but only inside the component and remove the property:

```
const showArchivedToggleButton = typeof isArchivedHidden !== 'undefined' && typeof onToggleArchived !== 'undefined';

...

{showArchivedToggleButton && (
  <Button ... />
)}
```

VWSCoronaDashboard25 (approved):

packages/app/src/components/page-information-block/page-information-block.tsx
```diff
@@ -54,11 +54,13 @@ export function PageInformationBlock({
   screenReaderCategory,
   vrNameOrGmName,
   warning,
-  hasHideArchivedButton,
```
VWSCoronaDashboard25: Remove on line 39 as well

PR #4054, opened 2022-02-22
Positive tests graph peak

VWSCoronaDashboard22:


packages/app/src/components/time-series-chart/logic/series.ts
```diff
@@ -61,6 +63,7 @@ interface SeriesCommonDefinition {
    * legend, for example)
    */
   hideInLegend?: boolean;
+  exclude?: number[];
```
VWSCoronaDashboard25, line 66: Above is a lot of documentation on the properties, so for continuity, can you explain what `exclude` does?
I see from the code that the numbers are not the indices of the data that is going to be excluded, but the keys of the key-value pairs, so you can work that in the documentation.

And should the type be plain `number`? Because the values are (I think) `T extends TimestampedValue`, so maybe there is something like `Timestamp` as a type for the key of a `TimestampedValue` that can be used instead? (I don't know, I'm guessing)

packages/app/src/components/time-series-chart/components/bar-trend-out-of-bounds.tsx
```diff
@@ -0,0 +1,116 @@
...
+  getX: GetX;
+  getY: GetY;
+  bounds: Bounds;
+  bandPadding?: number;
+  id: string;
+  yScale: PositionScale;
+  outOfBoundsDates?: number[];
+};
+
+export function BarOutOfBoundsTrend({
```
VWSCoronaDashboard24, line 22: For consistency maybe rename this component to `BarTrendOutOfBounds` to match the filename? Or other way around?

packages/app/src/pages/landelijk/positief-geteste-mensen.tsx
```diff
@@ -240,6 +240,16 @@ const PositivelyTestedPeople = (props: StaticProps<typeof getStaticProps>) => {
...
+                    exclude: [1644318000],
+                  },
+                  {
+                    type: 'bar-out-of-bounds',
+                    metricProperty: 'infected',
+                    label:
+                      siteText.positief_geteste_personen.tooltip_labels
+                        .infected_out_of_bounds,
+                    color: colors.data.neutral,
+                    outOfBoundsDates: [1644318000],
```
VWSCoronaDashboard24, line 252: Maybe move `[1644318000]` to a `const` to make it easier to find these values in the future when needed?

VWSCoronaDashboard24 (approved):

packages/app/src/pages/landelijk/positief-geteste-mensen.tsx
VWSCoronaDashboard25: And then maybe make it `new Date('8 February 2022 12:00:00').getTime() / 1000` so that it's clear what it represents.

packages/app/src/components/time-series-chart/logic/series.ts
VWSCoronaDashboard22: Added documentation

packages/app/src/components/time-series-chart/components/bar-trend-out-of-bounds.tsx
VWSCoronaDashboard22: Renamed components

packages/app/src/pages/landelijk/positief-geteste-mensen.tsx
VWSCoronaDashboard22: Think the values are easier to find when they are added to the component. Changed the time to a more human readable format

VWSCoronaDashboard24 (approved):
LGTM!

PR #4063, opened 2022-02-25
Feature/grey areas in timeseries correcton

VWSCoronaDashboard18:
Cover the whole first bar-instance of the time-series-chart instead of 50%.


VWSCoronaDashboard24 (approved):
Just wondering; would it also be possible to change the timestamps to beginning of the day of the annotated periods to solve this issue? I've run into stuff like this before.

PR #4103, opened 2022-03-11
Remove (temporarily?) risk level information

VWSCoronaDashboard22:


VWSCoronaDashboard19:
Do we need to disable the route somewhere in next, so users can't directly go to the page via URL?

PR #4120, opened 2022-03-21
Remove nlRepeatingShotAdministered feature flag

VWSCoronaDashboard24:


VWSCoronaDashboard22:
Should we also remove keys in Sanity?

VWSCoronaDashboard19 (approved):

VWSCoronaDashboard24:
> Should we also remove keys in Sanity?

Nothing is removed actually. Only the feature flaf check around it is removed.

VWSCoronaDashboard22 (approved):

PR #4141, opened 2022-03-30
Update scripts in monorepo and subprojects

VWSCoronaDashboard19:
## Summary

- Updated scripts
- Added shortcut scripts
- Updated docs
- Updated linter


VWSCoronaDashboard18 (approved):
Import(end) PR

.github/workflows/main.yml
```diff
@@ -73,11 +73,11 @@ jobs:
 
       - name: Export texts (develop)
         if: ${{ env.BRANCH_NAME != 'master' }}
-        run: yarn workspace @corona-dashboard/cms lokalize:export --clean-json
+        run: yarn workspace @corona-dashboard/cms lokalize:import --clean-json
```
VWSCoronaDashboard24, line 76: Question: should we still use `lokalize` as it is a bit of legacy now we have Sanity in place?

packages/cms/src/data/data-structure.ts
```diff
@@ -74,7 +74,7 @@ export const dataStructure = {
       "birthyear_range",
       "age_group_total",
     ],
-    booster_coverage: ["percentage"],
+    booster_coverage: ["age_group", "percentage"],
```
VWSCoronaDashboard24, line 77: Just checking; is this change intentional?

VWSCoronaDashboard24 (approved):

PR #4156, opened 2022-04-06
Feature/update actueel sewer with data driven text

VWSCoronaDashboard18:
added support from sanity

### DIFF
--- a/packages/cms/src/data/data-structure.ts
+++ b/packages/cms/src/data/data-structure.ts
@@ -74,7 +74,7 @@ export const dataStructure = {
       "birthyear_range",
       "age_group_total",
     ],
-    booster_coverage: ["percentage"],
+    booster_coverage: ["age_group", "percentage"],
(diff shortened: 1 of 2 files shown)


packages/cms/src/data/data-structure.ts
```diff
@@ -317,6 +324,7 @@ export const dataStructure = {
       "cure_vac",
       "janssen",
       "sanofi",
+      "novavax",
       "total",
     ],
     vaccine_administered_estimate: [
```
VWSCoronaDashboard24, line 330: Are the changes in this file intentional?

VWSCoronaDashboard24 (approved):

packages/cms/src/data/data-structure.ts
VWSCoronaDashboard18: No, but are generated by a type generator. They are based on other backend changes which now appear in this file...

PR #4174, opened 2022-04-11
Build Docker on release branches

VWSCoronaDashboard24:
Proposal to also build the Docker Image when creating a release branch towards master. This will prevent the staging/production pipeline to get stuck when something goes wrong in the frontend. The (daily) update can then still proceed and other frontend changes can  be released later on when everything work again.

Downside of this is that it takes a bit more time to setup a new release. Upside is that you know know on forehand if there is still an error in the JSON validation or frontend build.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -82,6 +82,9 @@ jobs:
       - name: Typecheck all code
         run: yarn typecheck
 
+      - name: Build app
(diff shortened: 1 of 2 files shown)


VWSCoronaDashboard18 (approved):
Great job!

VWSCoronaDashboard19 (approved):

PR #4175, opened 2022-04-12
Feature/add week avarage instead of day

VWSCoronaDashboard18:


packages/app/src/pages/actueel/gemeente/[code].tsx
```diff
@@ -398,7 +398,7 @@ const TopicalMunicipality = (props: StaticProps<typeof getStaticProps>) => {
                     {
                       type: 'split-area',
                       metricProperty: 'average',
-                      label: commonTexts.common.daggemiddelde,
```
Comment (VWSCoronaDashboard19) [inline packages/app/src/pages/actueel/gemeente/[code].tsx:401]: Can we remove this key if we don't use it anymore in the cms?

Comment (VWSCoronaDashboard18) [reply packages/app/src/pages/actueel/gemeente/[code].tsx:401]: It's still being used on the VR and NL domain

mvdam (approved):

VWSCoronaDashboard24 (approved):

PR #4191, opened 2022-04-20
feat: improved handling of missing data for behavior_per_age_group

VWSCoronaDashboard24:
* Do not hide age group row when either of compliance or support is empty. Show message instead
* Show message instead of empty table when all values are `null` of all age groups for both compiance and support

Screenshot 2022-04-20 at 14 06 32


VWSCoronaDashboard18 (approved):
Nice

PR #4212, opened 2022-04-29
Fix/inconsistent dates on topical pages headers

VWSCoronaDashboard18:
Fix inconsistent metrics for date of insertion

### DIFF
--- a/packages/app/src/utils/get-last-insertion-date-of-page.ts
+++ b/packages/app/src/utils/get-last-insertion-date-of-page.ts
@@ -1,0 +1,13 @@
+/**
+ * This method gets the most recent insertion date from the metrics used in a page
+ *
+ */
+
+export function getLastInsertionDateOfPage(data: any, pageMetrics: string[]) {
+  let lastDate = 0;
+  pageMetrics.forEach((metric: any) => {
+    const metricDate = typeof data[metric].last_value?.date_of_insertion_unix === undefined ? 0 : data[metric].last_value?.date_of_insertion_unix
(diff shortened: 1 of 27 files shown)


packages/app/src/utils/get-last-insertion-date-of-page.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * This method gets the most recent insertion date from the metrics used in a page
+ *
+ */
+
+export function getLastInsertionDateOfPage(data: any, pageMetrics: string[]) {
```
VWSCoronaDashboard24, line 6: Instead of any it would be better to have something like `T extends object` & make pageMetrics something like `Array<keyof T>`.

packages/app/src/utils/get-last-insertion-date-of-page.ts
```diff
@@ -0,0 +1,13 @@
...
+  let lastDate = 0;
+  pageMetrics.forEach((metric: any) => {
+    const metricDate = typeof data[metric].last_value?.date_of_insertion_unix === undefined ? 0 : data[metric].last_value?.date_of_insertion_unix
```
VWSCoronaDashboard24, line 9: This looks very dangerous because in theory `data[metric]` could be undefined, resulting in runtime errors. Would at least add a `?` here; `data[metric]?.last_value?.....`

VWSCoronaDashboard24: Also applies to line 10 below

PR #4215, opened 2022-05-02
COR-604 - Use new hospital_nice_choropleth key for hospital choropleths

VWSCoronaDashboard24:
* Use new key `hospital_nice_choropleth` to display hospital data on Choropleths
* Use `date_unix` from new `hospital_nice_choropleth` key to display in metadata below Choropleths (T-5 days)
* Schema changes

hospital_nice_choropleth


packages/app/schema/gm/__index.json
```diff
@@ -12,6 +12,7 @@
     "tested_overall",
     "hospital_nice_sum",
     "hospital_nice",
+    "hospital_nice_choropleth",
```
VWSCoronaDashboard18, line 15: Does every GM page use this metric or does it only uses the gm_collection

packages/app/schema/vr/__index.json
```diff
@@ -9,6 +9,7 @@
     "code",
     "tested_overall",
     "hospital_nice",
+    "hospital_nice_choropleth",
```
VWSCoronaDashboard18, line 12: Does every VR page use this metric or does it only uses the vr_collection

packages/app/schema/gm/__index.json
VWSCoronaDashboard24: This needs some more investigation and we will pick it up separately. (COR-629)

packages/app/schema/vr/__index.json
VWSCoronaDashboard24: This needs some more investigation and we will pick it up separately. (COR-629)

VWSCoronaDashboard18 (approved):
Looks nice

PR #4224, opened 2022-05-05
Sanity: add ids to all listItems and disable Lokalize Document creation or deletion

VWSCoronaDashboard24:
All list items now have IDs. Solved:
Screenshot 2022-05-05 at 11 49 41

---

"Create new Lokalize key" is now removed. 

Before:
Screenshot 2022-05-09 at 14 02 40

After:
Screenshot 2022-05-09 at 16 07 24


packages/cms/src/data/data-structure.ts
```diff
@@ -58,21 +66,23 @@ export const dataStructure = {
   },
   in_collection: { tested_overall: ["infected", "infected_per_100k_average"] },
   nl: {
+    risk_level: [
```
VWSCoronaDashboard18, line 69: Why is this moved from line 194 to this position?

VWSCoronaDashboard24: This file is generated based on our JSON schema. But this file is not generated as part of our `generate-data-types` script but in this case only just before `sanity start`. It probably didn't run for some time so had some catching up to do.

VWSCoronaDashboard22 (approved):

PR #4231, opened 2022-05-09
Hospital nice choropleth metric data to collection

VWSCoronaDashboard19:


packages/app/schema/vr_collection/hospital_nice_choropleth.json
```diff
@@ -0,0 +1,31 @@
...
+    "date_of_insertion_unix"
+  ],
+  "properties": {
+    "date_unix": {
+      "type": "integer"
+    },
+    "gmcode": {
+      "type": "string",
+      "pattern": "^GM[0-9]+$"
+    },
```
VWSCoronaDashboard24, line 20: This doesn't look right; `vrcode` should be here I guess?

VWSCoronaDashboard19: That's correct, just a copy-paste mistake. Will delete this new schema because we can refer `hospital_nice_choropleth.json` to `hospital_nice.json` as it uses the same fields and values

PR #4241, opened 2022-05-12
fix: make selftest_visit_behavior optional in json schemas

VWSCoronaDashboard24:


VWSCoronaDashboard19 (approved):

packages/common/src/types/data.ts
```diff
@@ -607,75 +607,75 @@ export interface NlBehavior {
 }
 export interface NlBehaviorValue {
   number_of_participants: number;
-  curfew_compliance: number | null;
+  curfew_compliance?: number | null;
```
VWSCoronaDashboard25, line 610: By adding the `?`, I think you're adding the type `undefined` so this value can be both `null` as `undefined`. Is that the goal here?

VWSCoronaDashboard24: That is indeed correct behavior!

PR #4246, opened 2022-05-17
Ignore forced maximum when out of bounds value is outside time domain

VWSCoronaDashboard22:


packages/app/src/components/time-series-chart/time-series-chart.tsx
```diff
@@ -241,10 +241,14 @@ export function TimeSeriesChart<
...
-  const seriesMax = isDefined(forcedMaximumValue)
-    ? isFunction(forcedMaximumValue)
-      ? forcedMaximumValue(calculatedSeriesMax)
-      : forcedMaximumValue
+  const calculatedForcedMaximumValue = isFunction(forcedMaximumValue)
+    ? forcedMaximumValue(calculatedSeriesMax)
+    : forcedMaximumValue;
+
+  const seriesMax = calculatedForcedMaximumValue
+    ? calculatedForcedMaximumValue < calculatedSeriesMax
```
VWSCoronaDashboard25, line 249: My personal preference goes to `Math.min(calculatedForcedMaximumValue, calculatedSeriesMax)` because that tells a reader exactly what is happening here.

Also, best practise is not to use a simple boolean check on a number, because the number could be defined but `0` and still come through as falsy.

So unless that's the goal, I would make this
```
const seriesMax = typeof calculatedForcedMaximumValue === 'number'
    ? Math.min(calculatedForcedMaximumValue, calculatedSeriesMax)
    : calculatedSeriesMax;
```

packages/app/src/components/time-series-chart/time-series-chart.tsx
```diff
@@ -275,11 +279,13 @@ export function TimeSeriesChart<
...
     seriesConfig,
-    dataOptions
+    dataOptions,
+    dataOptions?.outOfBoundsConfig && seriesMax < calculatedSeriesMax
   );
 
   const timeDomain = useMemo(
-    () => getTimeDomain({ values, today: endDate ?? today, withPadding: false }),
+    () =>
+      getTimeDomain({ values, today: endDate ?? today, withPadding: false }),
```
VWSCoronaDashboard25, line 288: Could you tell your IDE to not do this 😅

packages/app/src/components/time-series-chart/logic/common.ts
```diff
@@ -20,7 +20,11 @@ export interface DataOptions {
 export type OutOfBoundsConfig = {
   label: string;
   tooltipLabel: string;
-  checkIsOutofBounds: (a: any, b: number) => boolean;
+  checkIsOutofBounds: (
+    a: any,
+    b: number,
+    c: [start: number, end: number]
```
VWSCoronaDashboard25, line 26: I know you are following suit here, but maybe a good opportunity to start naming the variables properly and see if you can remove the type `any` here.

packages/app/src/components/time-series-chart/time-series-chart.tsx
VWSCoronaDashboard22: Nice idea. Thanks!

packages/app/src/components/time-series-chart/logic/common.ts
VWSCoronaDashboard22: We'll fix that in another ticket

packages/app/src/components/time-series-chart/time-series-chart.tsx
VWSCoronaDashboard22: Just following eslint recommended settings :-)

VWSCoronaDashboard25 (approved):

packages/app/src/components/time-series-chart/time-series-chart.tsx
VWSCoronaDashboard25: Alright!

PR #4257, opened 2022-05-18
Feature/cor 615 fix two kpi tiles with one column

VWSCoronaDashboard18:
Fixing empty kpi tile in a two column tile

Before:
Screenshot 2022-05-19 at 00 04 58
Screenshot 2022-05-19 at 00 05 33

After:
Screenshot 2022-05-19 at 00 05 16
Screenshot 2022-05-19 at 00 05 46


VWSCoronaDashboard24 (approved):
👍

PR #4258, opened 2022-05-19
COR-736 Use fallback for null values in Behavior Choropleth Tooltips

VWSCoronaDashboard24:
Before:

Screenshot 2022-05-19 at 11 07 54

After:

Screenshot 2022-05-19 at 11 07 40

### DIFF
--- a/packages/app/src/domain/behavior/tooltip/vr-behavior-tooltip.tsx
+++ b/packages/app/src/domain/behavior/tooltip/vr-behavior-tooltip.tsx
@@ -15,8 +15,8 @@ import { useReverseRouter } from '~/utils/use-reverse-router';
 interface VrBehaviorTooltipProps {
   context: TooltipData<VrCollectionBehavior>;
   currentMetric: BehaviorIdentifier;
-  currentComplianceValue: number;
-  currentSupportValue: number;
+  currentComplianceValue: number | null;
+  currentSupportValue: number | null;
   behaviorType: 'compliance' | 'support';
   text: SiteText['pages']['behaviorPage'];
 }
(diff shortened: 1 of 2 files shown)


packages/app/src/domain/behavior/tooltip/vr-behavior-tooltip.tsx
```diff
@@ -29,33 +29,39 @@ export function VrBehaviorTooltip({
...
   const complianceFilteredThreshold = getThresholdValue(
     thresholds.vr[complianceThresholdKey],
-    currentComplianceValue
+    currentComplianceValue || 0
   );
 
   const supportFilteredThreshold = getThresholdValue(
     thresholds.vr[supportThresholdKey],
-    currentSupportValue
+    currentSupportValue || 0
```
VWSCoronaDashboard25, line 44: Almost perfect use-case for the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator). 😉

packages/app/src/domain/behavior/tooltip/vr-behavior-tooltip.tsx
```diff
@@ -29,33 +29,39 @@ export function VrBehaviorTooltip({
...
   );
 
   const complianceTooltipInfo = (
     <TooltipInfo
       title={text.nl.tooltip_labels.compliance}
-      value={currentComplianceValue}
+      value={
+        currentComplianceValue
+          ? `${formatPercentage(currentComplianceValue)}%`
```
VWSCoronaDashboard25, line 52: Wouldn't this make turn a well defined `0%` into  a `-`?

PR #4260, opened 2022-05-23
Archive page and add warning tile

VWSCoronaDashboard19:
### Summary of changes

**Before (NL)**
> Screenshot 2022-05-23 at 12 03 54

**After (NL)**

> Screenshot 2022-05-23 at 12 03 42

**Before (VR)**
> Screenshot 2022-05-23 at 12 01 22

**After (VR)**
> Screenshot 2022-05-23 at 12 01 33

### DIFF
--- a/packages/app/src/pages/landelijk/brononderzoek.tsx
+++ b/packages/app/src/pages/landelijk/brononderzoek.tsx
@@ -97,13 +97,14 @@ export default function BrononderzoekPage(
             articles={content.articles}
           />
 
-          {textShared.warning && !isEmpty(textShared.warning) && (
-            <WarningTile
-              isFullWidth
-              message={textShared.warning}
-              variant="emphasis"
-            />
-          )}
+          {textShared.belangrijk_bericht &&
+            !isEmpty(textShared.belangrijk_bericht) && (
+              <WarningTile
+                isFullWidth
+                message={textShared.belangrijk_bericht}
+                variant="emphasis"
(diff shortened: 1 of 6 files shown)


packages/app/src/pages/veiligheidsregio/[code]/brononderzoek.tsx
```diff
@@ -138,6 +142,14 @@ export default function BrononderzoekPage(
             warning={textShared.warning}
```
Comment (VWSCoronaDashboard25) [inline packages/app/src/pages/veiligheidsregio/[code]/brononderzoek.tsx:142]: If the message for the newly added `WarningTile` is the same as in the header, should it still be here in the header?

packages/app/src/pages/landelijk/brononderzoek.tsx
```diff
@@ -95,6 +97,14 @@ export default function BrononderzoekPage(
             articles={content.articles}
           />
 
+          {textShared.warning && !isEmpty(textShared.warning) && (
+            <WarningTile
+              isFullWidth
+              message={textShared.warning}
```
VWSCoronaDashboard25, line 103: The key for the other archived pages is `belangrijk_bericht` maybe it's better to use the same key in all the archived pages.

VWSCoronaDashboard19: Good to know, will do 👍

VWSCoronaDashboard25 (approved):

PR #4263, opened 2022-05-24
Reverted and adjusted empty headings

VWSCoronaDashboard19:
### Summary

- Reverted "one child" logic to it's previous version, because it was too buggy and error prone and lot of work was needed to get rid of a single h3 element on a second "empty" KPI Tile.
- - No need for complex logic for creating columns while the logic is already there that just needs a small tweak
- Pass optional props where a full-width border is needed
- Replace and empty KPI tile with a simple empty Box with no content, so the KPI content stay at it's current width.

### DIFF
--- a/packages/app/src/domain/vaccine/vaccinations-shot-kpi-section.tsx
+++ b/packages/app/src/domain/vaccine/vaccinations-shot-kpi-section.tsx
@@ -1,10 +1,13 @@
-import { KpiTile } from '~/components/kpi-tile';
-import { KpiValue } from '~/components/kpi-value';
-import { Markdown } from '~/components/markdown';
-import { TwoKpiSection } from '~/components/two-kpi-section';
 import { useIntl } from '~/intl';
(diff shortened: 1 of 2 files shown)


packages/app/src/domain/vaccine/vaccinations-shot-kpi-section.tsx
```diff
@@ -47,6 +50,7 @@ export function VaccinationsShotKpiSection({
           }}
         />
       </KpiTile>
+      <Box />
```
VWSCoronaDashboard24, line 53: This is required for the layout to work, I assume?

VWSCoronaDashboard24 (approved):

packages/app/src/domain/vaccine/vaccinations-shot-kpi-section.tsx
VWSCoronaDashboard18: It also prevents an empty h3 inside the twoKPI section.

PR #4264, opened 2022-05-24
Implemented redesign combobox and select list

VWSCoronaDashboard22:


VWSCoronaDashboard24 (changes requested):
Code looks good! But I have two functional findings;

1. The message when there are no search results is missing some padding. Both in desktop and mobile view
2. The width of the combobox should initially be the same as when expanded (on Actueel).

See screenshots below:
Screenshot 2022-05-25 at 09 24 16
Screenshot 2022-05-25 at 09 25 43
Screenshot 2022-05-25 at 09 29 05
Screenshot 2022-05-25 at 09 29 09

VWSCoronaDashboard24 (approved):
💯

PR #4268, opened 2022-05-25
Removed obsolete code for international pages

VWSCoronaDashboard22:


packages/app/src/static-props/get-data.ts
```diff
@@ -311,42 +304,20 @@ const NOOP = () => null;
 export function createGetChoroplethData<T1, T2, T3>(settings?: {
```
VWSCoronaDashboard25, line 304: Do we still need `T3` here?

VWSCoronaDashboard25 (approved):

PR #4269, opened 2022-05-30
Release/2.48.3

VWSCoronaDashboard18:
Fixing the last date per page not always showing the correct date

### DIFF
--- a/packages/app/src/utils/get-last-insertion-date-of-page.ts
+++ b/packages/app/src/utils/get-last-insertion-date-of-page.ts
@@ -11,12 +11,17 @@ export function getLastInsertionDateOfPage(
   return pageMetrics.reduce((lastDate, metricProperty) => {
     const metricOrUnixDate = get(data, metricProperty);
 
-    // TODO: extract all cases in seperate function. 1. date, 2. last_values, 3. values
+    let metricDate = 0;
 
-    const metricDate =
-      typeof metricOrUnixDate === 'number'
-        ? metricOrUnixDate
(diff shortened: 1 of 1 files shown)


packages/app/src/utils/get-last-insertion-date-of-page.ts
```diff
@@ -11,10 +11,17 @@ export function getLastInsertionDateOfPage(
...
+    let metricDate = 0;
+
+    if (typeof metricOrUnixDate === 'number') {
+      metricDate = metricOrUnixDate;
+    } else if (typeof metricOrUnixDate?.last_value?.date_of_insertion_unix !== 'undefined') {
+      metricDate = metricOrUnixDate?.last_value?.date_of_insertion_unix;
+    } else if (typeof metricOrUnixDate?.values !== 'undefined') {
+      metricDate = metricOrUnixDate?.values.reduce((max: number, value: any) => value.date_of_insertion_unix > max ? value.date_of_insertion_unix : max, 0);
+    } else {
+      metricDate = 0;
```
VWSCoronaDashboard25, line 23: you can do either
```
let metricDate: number;

if(..) metricDate = ..;
else if(..) metricDate = ..;
else metricDate = 0;
```
 or

```
let metricDate = 0;

if(..) metricDate = ..;
else if(..) metricDate = ..;
```
Giving a value of `0` before the `if` and in the `else` clause is redundant. My preference of these two is. the first, so you see that it's a value that has to be filled in, with the else clause the 'default' or 'fall back' value.

VWSCoronaDashboard25 (approved):

PR #4270, opened 2022-05-30
Bugfix/fix last date of insertion per page

VWSCoronaDashboard18:


packages/app/src/utils/get-last-insertion-date-of-page.ts
```diff
@@ -11,11 +11,18 @@ export function getLastInsertionDateOfPage(
...
+      metricDate = metricOrUnixDate;
+    } else if (typeof metricOrUnixDate?.last_value?.date_of_insertion_unix !== 'undefined') {
+      metricDate = metricOrUnixDate?.last_value?.date_of_insertion_unix;
+    } else if (typeof metricOrUnixDate?.values !== 'undefined') {
+      metricDate = metricOrUnixDate?.values.reduce((max: number, value: any) => value.date_of_insertion_unix > max ? value.date_of_insertion_unix : max, 0);
+    } else {
+      metricDate = 0;
+    }
+    
     return metricDate > lastDate ? metricDate : lastDate;
```
VWSCoronaDashboard25, line 26: I think this can get better with descriptive functions. This can help to break down the issue by breaking down the code. Afterwards you've done this, we can see what we want to keep and what we want to improve otherwise.

For example, in the extreme, you can create these little functions:
```
function isNumber(value): boolean {
  return typeof metricOrUnixDate === 'number';
}

function hasLastValue(metric): boolean {
  return typeof metric?.last_value?.date_of_insertion_unix !== 'undefined';
}

function getLastValue(metric): number {
  return metric?.last_value?.date_of_insertion_unix;
}

function hasValues(metric): boolean {
  return typeof metric?.values !== 'undefined';
}

function getLatestFromValues(metric): number {
  return metric?.values.reduce((max: number, value: any) => value.date_of_insertion_unix > max ? value.date_of_insertion_unix : max, 0);
  // Can this be just Math.max(metric.values) ?
  // Does it make sense to supply the metric if we only need the values?
}
```

So that this function can better be understandable:
```
function getMetricDate(metricOrUnixDate) {
  if (isNumber(metricOrUnixDate)) {
    return metricOrUnixDate;
  }
  if (hasLastValue(metricOrUnixDate)) {
    return getLastValue(metricOrUnixDate);
  }
  if (hasValues(metricOrUnixDate)) {
      return getLatestFromValues(metricOrUnixDate);
    }
  return 0;
}
```

Then you can use that here, to make it readable what this function does:
```
export function getLastInsertionDateOfPage(
  data: unknown,
  pageMetrics: string[]
) {
  return pageMetrics.reduce((lastDate, metricProperty) => {
    const metricOrUnixDate = get(data, metricProperty);
    const metricDate = getMetricDate(metricOrUnixDate);
    
    return metricDate > lastDate ? metricDate : lastDate;
    //This can probably be Math.max(metricDate, lastDate);
  }, 0);
}
```

You don't need all small functions from the first block, but I think it's nice to extract the second block at least. And check if all the `Math.max` occurrences do what you think they do.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max

VWSCoronaDashboard25, line 26: I think this can get better with descriptive functions. This can help to break down the issue by breaking down the code. Afterwards you've done this, we can see what we want to keep and what we want to improve otherwise.

For example, in the extreme, you can create these little functions:
```
function isNumber(value): boolean {
  return typeof metricOrUnixDate === 'number';
}

function hasLastValue(metric): boolean {
  return typeof metric?.last_value?.date_of_insertion_unix !== 'undefined';
}

function getLastValue(metric): number {
  return metric?.last_value?.date_of_insertion_unix;
}

function hasValues(metric): boolean {
  return typeof metric?.values !== 'undefined';
}

function getLatestFromValues(metric): number {
  return metric?.values.reduce((max: number, value: any) => value.date_of_insertion_unix > max ? value.date_of_insertion_unix : max, 0);
  // Can this be just Math.max(metric.values) ?
  // Does it make sense to supply the metric if we only need the values?
}
```

So that this function can better be understandable:
```
function getMetricDate(metricOrUnixDate) {
  if (isNumber(metricOrUnixDate)) {
    return metricOrUnixDate;
  }
  if (hasLastValue(metricOrUnixDate)) {
    return getLastValue(metricOrUnixDate);
  }
  if (hasValues(metricOrUnixDate)) {
      return getLatestFromValues(metricOrUnixDate);
    }
  return 0;
}
```

Then you can use that here, to make it readable what this function does:
```
export function getLastInsertionDateOfPage(
  data: unknown,
  pageMetrics: string[]
) {
  return pageMetrics.reduce((lastDate, metricProperty) => {
    const metricOrUnixDate = get(data, metricProperty);
    const metricDate = getMetricDate(metricOrUnixDate);
    
    return metricDate > lastDate ? metricDate : lastDate;
    //This can probably be Math.max(metricDate, lastDate);
  }, 0);
}
```

You don't need all small functions from the first block, but I think it's nice to extract the second block at least. And check if all the `Math.max` occurrences do what you think they do.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max

packages/app/src/utils/get-last-insertion-date-of-page.ts
```diff
@@ -3,19 +3,94 @@ import { get } from 'lodash';
  * This method gets the most recent insertion date from the metrics used in a page
  *
  */
+ const getLastInsertionDatePerMetric = (data: any, metricProperty: String): number => {
+   switch (metricProperty) {
```
VWSCoronaDashboard25, line 7: Hmmm.... let's not do this. This is a really big switch, and it assumes to know the correct structure of `data` for a given metric.

Although that assumption is probably correct now, it is important to keep it correct in the future. So if the data for a metric changes, or new metrics are added, that means we have to update this switch as well. But more importantly, we have to remember that this coupling is made here, so we won't forget to change it here as well.
This is a dangerous setup because we probably have forgotten this coupling in a few weeks from now.

This would be less bad if we'd have a TypeScript error if something is changed, so we get reminded to change this as well. But we can't do that here.

Best (for now at least) would be to assume no knowledge about the structure of `data` and instead just check if we can use the value that we want to use. So if we _can_ return `data.last_value.date_of_insertion_unix` just assume that we _want_ to return that. See my other comment for this.

If that results in wrong code for certain exceptions, you can handle them before you enter the if-else if- else if-else.

VWSCoronaDashboard18: Better now? 😉

VWSCoronaDashboard25:
Closed in favour of #4314

PR #4273, opened 2022-05-31
Style changes select component

VWSCoronaDashboard22:


packages/app/src/domain/topical/components/search/hit-list.tsx
```diff
@@ -94,17 +93,22 @@ const HitLink = forwardRef<HTMLAnchorElement, HitLinkProps>(
   }
 );
 
-const StyledHitLink = styled(Anchor)<{
-  hasFocus: boolean;
-}>((x) =>
+const paddingLeftNarrow = '50px';
```
VWSCoronaDashboard19, line 96: There is a file where you can re-use values for spaces and other stuff:`packages/app/src/style/theme.ts`. Also try using `rem` instead of `px` or `em` so it's aligned with the rest of the codebase.

packages/app/src/domain/topical/components/search/hit-list.tsx
```diff
@@ -154,6 +158,7 @@ const StyledHitList = styled.ol(
 
 const NoResultMessage = styled.div(
   css({
-    px: '1.75em',
+    ...paddedStyle,
+    py: '0em',
```
VWSCoronaDashboard19, line 162: Maybe a better idea to separately set the `py` above instead of trying to override here? Otherwise it will be double which looks ugly.

VWSCoronaDashboard19 (approved):

PR #4292, opened 2022-06-14
Refactored axes component

VWSCoronaDashboard22:


packages/app/src/components/time-series-chart/components/axes.tsx
```diff
@@ -118,122 +112,110 @@ export const Axes = memo(function Axes<
...
+          })
+        : dateUnix;
+
+      const style =
+        ((isDateSpanValue(tickValue)
+          ? startUnix === tickValue.date_start_unix ||
+            endUnix === tickValue.date_end_unix
+          : [startUnix, endUnix].includes(tickValue)) &&
+          startYear !== endYear) ||
+        previousYear !== currentYear
```
VWSCoronaDashboard25, line 200: line 195 to line 200 can be a separate function that describes what is going on with these checks. From the result I understand that it determines if the axis should be displayed with a year or not, but the code is too hard to understand, for me.

I should dive into this to judge this better, but I assume these lines themselves can be made better understandable as well by shifting some of this around. But let's start with putting it is a function that returns a boolean and we'll see from there what we can do.

packages/app/src/components/time-series-chart/components/axes.tsx
```diff
@@ -118,122 +112,110 @@ export const Axes = memo(function Axes<
...
+          ? 'axis-with-year'
+          : 'axis';
+
+      return isDateSpanValue(tickValue)
```
VWSCoronaDashboard25, line 204: `isDateSpanValue` and `isDateSpanValues` are recurring ternary initiators through this file.
Maybe we should split the whole thing into a `DateSpanAxes` and a `TimeStampAxes` or something, and choose what axes to use with `isDateSpanValues`.

VWSCoronaDashboard25 (changes requested):

VWSCoronaDashboard25 (approved):

PR #4293, opened 2022-06-14
COR-749 Fix responsive Choropleth sizing

VWSCoronaDashboard24:


packages/app/src/components/choropleth/components/canvas-choropleth-map.tsx
```diff
@@ -46,6 +46,7 @@ export type CanvasChoroplethMapProps = {
   fitExtent: FitExtent;
   getFeatureName: (code: string) => string;
   height: number;
+  useMinHeight?: boolean;
```
VWSCoronaDashboard19, line 49: Better option for prefixing booleans with: is, has or can.

Suggestion: `hasMinHeight`?

packages/app/src/components/choropleth/components/choropleth-map.tsx
```diff
@@ -136,6 +136,7 @@ export const ChoroplethMap: <T extends ChoroplethDataItem>(
             dataOptions={dataOptions}
             width={width}
             height={correctedHeight}
+            useMinHeight={minHeight > 0}
```
VWSCoronaDashboard25, line 139: Why should we add this logic here?
I think we should carry the the boolean `useMinHeight` through to the surface (full control on the page where we want the Choropleth).

packages/app/src/components/choropleth/components/canvas-choropleth-map.tsx
VWSCoronaDashboard24: Fixed!

packages/app/src/components/choropleth/components/choropleth-map.tsx
VWSCoronaDashboard24: Changed it to something better (I hope ;-) )

VWSCoronaDashboard19 (approved):

PR #4294, opened 2022-06-15
Scroll through list of results with keyboard

VWSCoronaDashboard19:
Fixed the issue where it was not possible to scroll through a list of results in a search bar. The bug was present in the index pages of GM and VR

### DIFF
--- a/packages/app/src/components/combo-box/combo-box.tsx
+++ b/packages/app/src/components/combo-box/combo-box.tsx
@@ -45,7 +45,7 @@ type TProps<Option extends TOption> = {
  * />
  * ```
  */
-export function ComboBox<Option extends TOption>(props: TProps<Option>) {
+export const ComboBox = <Option extends TOption>(props: TProps<Option>) => {
   const { options, placeholder, sorter, selectedOption } = props;
 
   const { commonTexts } = useIntl();
@@ -53,7 +53,7 @@ export function ComboBox<Option extends TOption>(props: TProps<Option>) {
   const router = useRouter();
   const { code } = router.query;
   const inputRef = useRef<HTMLInputElement>(null);
-  const containerRef = useRef<HTMLInputElement>(null);
+  const containerRef = useRef<HTMLUListElement>(null);
   const [inputValue, setInputValue] = useState<string>('');
   const results = useSearchedOptions<Option>(inputValue, options, sorter);
   const breakpoints = useBreakpoints();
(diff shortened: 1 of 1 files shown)


packages/app/src/components/combo-box/combo-box.tsx
```diff
@@ -53,12 +53,40 @@ export function ComboBox<Option extends TOption>(props: TProps<Option>) {
...
   const [inputValue, setInputValue] = useState<string>('');
   const results = useSearchedOptions<Option>(inputValue, options, sorter);
   const breakpoints = useBreakpoints();
   const isLargeScreen = breakpoints.md;
   const hasRegionSelected = !!code;
 
+  /**
+   * Allow keyboard interaction to scroll through a list of results.
+   */
+  const onKeyDown = (event: React.KeyboardEvent<HTMLInputElement>) => {
```
VWSCoronaDashboard25, line 66: The code standards say we should use `function doSomething(){}` instead of `const doSomething = () => {}`.
https://github.com/minvws/nl-covid19-data-dashboard/blob/develop/README.md#coding-standards

I'm happy if we remove that from the code standards, but then please update the other functions in this file as well, now you're adding function expressions to a file with named function declarations.

VWSCoronaDashboard25 (dismissed):

packages/app/src/components/combo-box/combo-box.tsx
```diff
@@ -45,25 +45,52 @@ type TProps<Option extends TOption> = {
...
   const results = useSearchedOptions<Option>(inputValue, options, sorter);
   const breakpoints = useBreakpoints();
   const isLargeScreen = breakpoints.md;
   const hasRegionSelected = !!code;
 
-  function handleInputChange(event: React.ChangeEvent<HTMLInputElement>): void {
+  /**
+   * Allow keyboard interaction to scroll through a list of results.
+   */
+  const onKeyDown = (event: React.KeyboardEvent<HTMLInputElement>) => {
```
VWSCoronaDashboard24, line 66: Should this be `handleKeyDown` according to our code standards?

VWSCoronaDashboard24 (approved):

packages/app/src/components/combo-box/combo-box.tsx
VWSCoronaDashboard19: That's a good one, will update the naming according to our coding standards!

PR #4295, opened 2022-06-15
feature/COR-497-split-next-config

VWSCoronaDashboard26:
feat: moved headers, redirects (and gmRedirects) and rewrites from `next.config.js` to separate files

### DIFF
--- a/packages/app/next.config.js
+++ b/packages/app/next.config.js
@@ -14,9 +14,9 @@ const withTranspileModules = require('next-transpile-modules')([
 const path = require('path');
 const { DuplicatesPlugin } = require('inspectpack/plugin');
 
-const headers = require('./src/next-config/headers');
-const redirects = require('./src/next-config/redirects/redirects');
-const rewrites = require('./src/next-config/rewrites');
+const { headers } = require('./src/next-config/headers');
+const { redirects } = require('./src/next-config/redirects/redirects');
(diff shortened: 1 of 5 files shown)


VWSCoronaDashboard24 (approved):
Very nice improvement!

packages/app/next.config.js
```diff
@@ -73,106 +51,9 @@ const nextConfig = {
   /**
    * More header management is done by the next.server.js for the HTML pages and JS/CSS assets.
```
VWSCoronaDashboard25, line 52: Have you thought about moving this comment to the `headers` file?
Reading the comment, it feels like it belongs there instead of here.

packages/app/src/next-config/headers.js
```diff
@@ -0,0 +1,14 @@
+export const headers = async () => {
```
VWSCoronaDashboard25, line 1: For now the code standards in this project are to use `function doSomething(){}` instead of `const doSomething = () => {}`.
That's against my personal preferences as well, but them are the code standards.
https://github.com/minvws/nl-covid19-data-dashboard/blob/develop/README.md#coding-standards

But maybe we can do a meeting with the developers to change these standards soon. We need to discuss other code style stuff as well.

Until that time, we should stick to these code standards.

VWSCoronaDashboard25 (approved):
LGTM!

PR #4302, opened 2022-06-27
fix: stop using archived data for unarchived chart

VWSCoronaDashboard25:
The graph was unarchived but still pointed to the archived dataset. This commit fixes that. And properly removes it from other calculations as well.


VWSCoronaDashboard24 (approved):
Sharp!!

PR #4306, opened 2022-07-01
fix(axes): can't calculate tickValue without values

VWSCoronaDashboard25:
Prevents a potential runtime error since #4292 by returning empty x-axis labels if the graph has no values.

Before:
image

After:
image

### DIFF
--- a/packages/app/src/components/time-series-chart/components/axes.tsx
+++ b/packages/app/src/components/time-series-chart/components/axes.tsx
@@ -172,6 +172,10 @@ export const Axes = memo(function Axes<T extends TimestampedValue>({
 
   const DateSpanTick = useCallback(
     (dateUnix: number, values: DateSpanValue[], index: number) => {
+      if (values.length === 0) {
+        return '';
+      }
+
       const previousYear = createDateFromUnixTimestamp(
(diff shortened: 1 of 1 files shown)


VWSCoronaDashboard26 (approved):
Approved, but minor question: is there a certain standard on handling things like
`if (values.length === 0) { ... }` vs `if (!values.length) { ... }`?

VWSCoronaDashboard25:
@VWSCoronaDashboard26 
> Approved, but minor question: is there a certain standard on handling things like
> `if (values.length === 0) { ... }` vs `if (!values.length) { ... }`?

No (at least not yet). Though I prefer the first, because for me that reads like "if the length of values is zero" which in my head becomes "if values is an array without elements".
The other reads for me as "if values doesn't have a length" which in my head becomes "if values is not an array" because `length` is a property of an array. So if something doesn't have a length, it can't be an array.

Of course I know this isn't meant like this. It makes use of the fact that `0` is falsy, while every other number is truthy, which is why `!values.length` works for all array lengths. However I believe by typing a few more characters, the code is more explicit and it asks for less thinking of the reader.

PR #4307, opened 2022-07-04
Renamed misused tooltip_labels

VWSCoronaDashboard22:


VWSCoronaDashboard19:
Graphs need to have a label and also a shortLabel property with two different keys attached to it. I can't see it in the changes. E.g in `packages/app/src/pages/landelijk/positief-geteste-mensen.tsx`

PR #4308, opened 2022-07-04
feature/COR-871-variant-page-table

VWSCoronaDashboard26:
**Before (on production)**
before

**After (on local)**
after


packages/app/src/domain/variants/static-props/get-variant-table-data.ts
```diff
@@ -83,15 +117,43 @@ export function getVariantTableData(
...
+      }
+
+      return VARIANT_TABLE_MAP.findIndex((variant) =>
+        variant.includes(variantA.name)
+      ) >
+        VARIANT_TABLE_MAP.findIndex((variant) =>
+          variant.includes(variantB.name)
+        )
+        ? 1
+        : -1;
```
VWSCoronaDashboard25, line 139: I think it's best to stick with the order that is provided by the backend. Because we can also get (sub) variants like `BA2.12` and `BA2.12.1`, which means we'll have to update `VARIANT_TABLE_MAP` with every new (sub) variant that the backend wants to show.

Otherwise, I think it's okay to just return the difference instead of `1` or `-1`.

Although historically you return either `-1`, `0` or `1` in a comparison, returning the difference is a common practise. JavaScript (as most other languages) just checks for `< 0`, `=== 0` or `> 0`. It is even encouraged in [the MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#examples) to subtract two numerical values.

VWSCoronaDashboard25 (changes requested):

VWSCoronaDashboard18 (approved):

PR #4316, opened 2022-07-07
Added alphabetical sorting

VWSCoronaDashboard18:
The order of variants in the graph is random when they have the same first occurrence. (See Omikron BA.4 & Omikron BA.2.12.1)

Before:
Screenshot 2022-07-08 at 01 16 01

After:
Screenshot 2022-07-08 at 01 15 40

### DIFF
--- a/packages/app/src/domain/variants/static-props/get-variant-chart-data.ts
+++ b/packages/app/src/domain/variants/static-props/get-variant-chart-data.ts
@@ -26,7 +26,7 @@ export function getVariantChartData(variants: NlVariants | undefined) {
     return EMPTY_VALUES;
   }
 
-  const firstOccurences = variants.values.reduce<Record<string, number>>(
+  const firstOccurences = variants.values.sort().reverse().reduce<Record<string, number>>(
     (acc, x) =>
       Object.assign(acc, {
         [x.name]: x.values.find((value) => value.percentage > 0)


packages/app/src/domain/variants/static-props/get-variant-chart-data.ts
```diff
@@ -26,7 +26,7 @@ export function getVariantChartData(variants: NlVariants | undefined) {
     return EMPTY_VALUES;
   }
 
-  const firstOccurences = variants.values.reduce<Record<string, number>>(
+  const firstOccurences = variants.values.sort().reverse().reduce<Record<string, number>>(
```
VWSCoronaDashboard26, line 29: I am a bit baffled by this. How I understand the code is that, from `packages/app/src/pages/landelijk/varianten.tsx:58`, we spread the values we receive from the `getVariantChartData` method (this functionality). Now we `sort()` and `reverse()` the values to get everything in a working alphabetical order (Sidenote #1). We then, in `packages/app/src/domain/variants/variants-stacked-area-tile/variants-stacked-area-tile.tsx:207`, do some more mapping, filtering and reversing again. I think I simply do not understand this flow exactly.

Code still works though, approved from my end.

Sidenote:
I still wonder how this works with new variants that start with a different Greek letter but is sorted before the current 'last' variant in the list (for example: `chi` (starting with a C) could be sorted before `omicron` (starting with an O.

VWSCoronaDashboard26 (approved):

packages/app/src/domain/variants/static-props/get-variant-chart-data.ts
VWSCoronaDashboard18: The Greek letter sorting for variants with the same first occurrence is postponed due to the rare occurrence. New ticket will be created.

PR #4319, opened 2022-07-08
bugfix/COR-918-remove-variant-description-and-country-of-origin

VWSCoronaDashboard26:
**Before (on production)**
before

**After (on local)**
after


VWSCoronaDashboard18 (commented):
This needs some more cleanup. This only removes it visually. But the functionality needs to be removed as well.

VWSCoronaDashboard25 (approved):
LGTM!
However, could you append the PR title with the ticket number? This helps to later understand why all of this was done.
At least, I assume that the PR title will be the commit name after squash-merge. If not, try to make sure the ticket number will end up in the eventual squash-commit in another way.

VWSCoronaDashboard25:
< deleted in favour of #4320 >

PR #4320, opened 2022-07-11
bugfix/COR-918-remove-variant-origin-and-description

VWSCoronaDashboard26:
**Before (on production)**
before

**After (on local)**
after


VWSCoronaDashboard25 (approved):
👍

PR #4305, opened 2022-06-30
COR-54 Refactor types

VWSCoronaDashboard24:

PR #4321, opened 2022-07-13
Remove international from codebase

VWSCoronaDashboard19:
## Summary

- Remove `international` from the entire codebase


packages/app/src/components/choropleth/index.tsx
```diff
@@ -62,14 +56,14 @@ export type DataOptions = {
...
    */
-  metricName: KeysOfType<InferedDataCollection<T>, T[]>;
+  metricName: string | number | boolean;
   /**
    * A property name of the object determined by the metric name. This value is used to determine the color
    * of the feature.
    */
-  metricProperty: KeysOfType<T, number | null | boolean | undefined, true>;
+  metricProperty: string | number | boolean;
   /**
```
VWSCoronaDashboard20, line 67: You are changing some logic here. Is that intended?

packages/app/src/utils/__tests__/use-reverse-router.spec.tsx
```diff
@@ -68,43 +63,38 @@ const TestBed = () => {
...
       <div data-testid="gm">{router.gm.index(gmCode)}</div>
-      <div data-testid="in">{router.in.index()}</div>
     </>
   );
 };
 
 UseReverseRouter(
   'indexes should link to the actual index on small screens',
-  (context) => {
+  () => {
```
VWSCoronaDashboard20, line 72: Just a check question. Why can context be removed as an argument?

VWSCoronaDashboard19: As it is an unused argument, therefore it can be removed

packages/app/src/components/choropleth/index.tsx
VWSCoronaDashboard19: Yes it was, but this is not working properly so I am reverting this back to original

packages/app/src/components/cms/inline-choropleth.tsx
```diff
@@ -58,7 +58,7 @@ export function InlineChoropleth(props: InlineChoroplethProps) {
   };
 
   const dataConfig: OptionalDataConfig<ChoroplethDataItem> = {
-    metricName: configuration.metricName,
+    metricName: configuration.metricName as any,
```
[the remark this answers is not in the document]
VWSCoronaDashboard24: As discussed; these `any`s will be addressed in PR #4305.

VWSCoronaDashboard24 (approved):
Nice cleanup! 💪

PR #4334, opened 2022-07-22
COR-949 Add vaccination campaign name localization

VWSCoronaDashboard21:


packages/app/src/domain/vaccine/vaccine-campaigns-tile/components/narrow-vaccine-campaign-table.tsx
```diff
@@ -62,6 +63,7 @@ const VaccineCampaignRow = ({
   const { formatNumber } = useIntl();
   const collapsible = useCollapsible({ isOpen: isFirst });
   const isOpen = collapsible.isOpen;
+  const { locale = 'nl' } = useRouter();
```
VWSCoronaDashboard19, line 66: Will `const { locale } = useRouter();` not be enough for this? because you are already checking if it is `nl` on line 76

VWSCoronaDashboard21: For consistency reasons, we will keep it this way.

VWSCoronaDashboard24 (approved):

PR #4335, opened 2022-07-25
Replace konva-node with konva

VWSCoronaDashboard19:
This PR was reverted at release do to some warnings in the console. These do not occur anymore in this PR as the value is always present in the `container`

Remove deprecated dependancy `konva-node` and replace it with the one that is supported `konva`
[See full conclusion of this fix](https://logius-prd.jira.odc-noord.nl/browse/COR-805)


VWSCoronaDashboard24 (approved):
Works as expected. The former runtime errors are now gone. 👍

PR #4338, opened 2022-07-25
feature/COR-938-vaccinations-per-supplier-over-last-week-tile

VWSCoronaDashboard21:
## Summary
- added new schema for `vaccine_administered_last_week` data coming from BE
- added `VaccinationsPerSupplierOverLastWeekTile` component
- updated `ChartTile` component to add layout variant
- updated `PieChart` component to fix layout issues and add/update some props for extension (`iconFill`, `hasHoverState`, renamed `paddingLeft` to `marginLeft`, `marginRight`)

#### Additional (OOS)
- updated `WideVaccineCampaignTable` and `NarrowVaccineCampaignTable` components to fix potential `.replaceAll()` bugs


packages/app/src/domain/vaccine/vaccinations-per-supplier-over-last-week-tile.tsx
```diff
@@ -0,0 +1,105 @@
...
+  'Moderna',
+  'astra_zeneca',
+  'AstraZeneca',
+  'janssen',
+  'Janssen',
+  'novavax',
+  'Novavax',
+  'cure_vac',
+  'sanofi',
+] as const;
```
VWSCoronaDashboard25, line 37: Why is it needed to cast this with `as const`?

VWSCoronaDashboard26: It is not needed/required to cast this with `as const`, yet I opted to do this for two reasons:
- firstly, yet this is not the main reason, to be consistent with other instances throughout the code base;
- secondly, and this is the main reason, is to mark `vaccines` as `readonly` in order to prevent mutations to this array at a lower level.

VWSCoronaDashboard25: Aight, I would expect `readonly` if something needs to be be readonly, but I see there is some stuff on the internet about [const assertions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions) vs [readonly](https://www.typescriptlang.org/docs/handbook/2/objects.html#readonly-properties), so I'll dive into that, some day. 👍

packages/app/src/components/chart-tile.tsx
```diff
@@ -37,40 +38,77 @@ export function ChartTile({
...
 }: ChartTileProps) {
   const [timeframe, setTimeframe] = useState<TimeframeOption>(
     timeframeInitialValue || TimeframeOption.ALL
   );
 
   return (
     <FullscreenChartTile metadata={metadata} disabled={disableFullscreen}>
-      <ChartTileHeader title={title} description={description}>
-        {timeframeOptions && timeframe && (
+      {hasSplitLayout ? (
```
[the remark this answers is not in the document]
VWSCoronaDashboard26: This has been resolved. `ChartTile` has been reset to its previous version and a new `ChartTileDoubleColumn` component has been introduced.

VWSCoronaDashboard25 (approved):

packages/app/src/components/chart-tile.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard25: Nice!

PR #4348, opened 2022-08-04
Hotfix/variant cahrt and table configuration fix

VWSCoronaDashboard18:
A hotfix from some issues

### DIFF
--- a/packages/app/src/domain/variants/static-props/get-variant-table-data.ts
+++ b/packages/app/src/domain/variants/static-props/get-variant-table-data.ts
@@ -67,7 +67,7 @@ export function getVariantTableData(
       (variant) =>
         variant.variant_code !== 'other_graph' && !variant.last_value.has_historical_significance
     )
-    .sort((a, b) => a.last_value.order - b.last_value.order)
+    .sort((a, b) =>b.last_value.order - a.last_value.order)
     .map<VariantRow>((variant) => {
       const color =
         variantColors.find(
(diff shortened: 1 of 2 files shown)


packages/app/src/domain/variants/static-props/get-variant-table-data.ts
```diff
@@ -67,7 +67,7 @@ export function getVariantTableData(
       (variant) =>
         variant.variant_code !== 'other_graph' && !variant.last_value.has_historical_significance
     )
-    .sort((a, b) => a.last_value.order - b.last_value.order)
+    .sort((a, b) =>b.last_value.order - a.last_value.order)
```
VWSCoronaDashboard19, line 70: I think your Prettier did not do its work properly, as there is not space here: `=>b` instead of `=> b`

VWSCoronaDashboard19 (approved):

VWSCoronaDashboard26 (commented):
Are variants with a value of 0% also filtered out, or is that done by BE?

packages/app/src/domain/variants/static-props/get-variant-table-data.ts
VWSCoronaDashboard24, line 70:
```suggestion
    .sort((a, b) => b.last_value.order - a.last_value.order)
```

VWSCoronaDashboard24 (approved):

PR #4353, opened 2022-08-09
bugfix/COR-973-time-series-chart-duplicate-x-axis-ticks

VWSCoronaDashboard26:
## Summary

* updated `createTimeTicks` method in Axes component to now also take a `valueCount` param into account, this prevents rendering too many or duplicate ticks in certain cases;

#### Before
Notice the triple ticks for range '25 - 31 jul' on desktop and the three ticks (despite having only two data points) on mobile. 
before_desktop_7-days

before_mobile_7-days

#### After
after_mobile_7-days
after_desktop_7-days

### DIFF
--- a/packages/app/src/components/time-series-chart/components/axes.tsx
+++ b/packages/app/src/components/time-series-chart/components/axes.tsx
@@ -75,7 +75,7 @@ function createTimeTicks(
   startTick: number,
   endTick: number,
   count: number,
-  valueCount: number | undefined
+  valuesCount: number | undefined
 ) {
   const start = middleOfDayInSeconds(startTick);
   const end = middleOfDayInSeconds(endTick);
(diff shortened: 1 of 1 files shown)


packages/app/src/components/time-series-chart/components/axes.tsx
```diff
@@ -71,7 +71,12 @@ export type AxesProps<T extends TimestampedValue> = {
   hasAllZeroValues?: boolean;
 };
 
-function createTimeTicks(startTick: number, endTick: number, count: number) {
+function createTimeTicks(
+  startTick: number,
+  endTick: number,
+  count: number,
+  valueCount: number | undefined
```
VWSCoronaDashboard19, line 78: Renaming this to `valuesCount` would be more consistent because you are passing a plural on line 179 `values.length`

VWSCoronaDashboard19 (approved):

VWSCoronaDashboard24 (approved):

PR #4355, opened 2022-08-11
COR-995 Topical page redesign schema

VWSCoronaDashboard24:


packages/app/src/pages/nieuw-actueel.tsx
```diff
@@ -78,6 +54,7 @@ const Home = (props: StaticProps<typeof getStaticProps>) => {
 
   return (
     <Layout {...metadata} lastGenerated={lastGenerated}>
+      <Box>{selectedTopicalData.title}</Box>
```
VWSCoronaDashboard24, line 57: Temporary for now; but `selectedTopicalData` contains all the (translated) data for the topical page.

packages/app/src/static-props/get-data.ts
```diff
@@ -70,6 +71,10 @@ const json = {
     loadJsonFromDataFile<GmCollection>('GM_COLLECTION.json'),
     'gm_collection'
   ),
+  topical: initializeFeatureFlaggedData<Topical>(
```
VWSCoronaDashboard19, line 74: Isn't topical also the same level as NL? or are you creating an extra key so we also keep the old one for now.

VWSCoronaDashboard19 (approved):

packages/common/src/types/data.ts
```diff
@@ -1028,6 +1028,247 @@ export interface NlVariantsVariantValue {
...
+}
+export interface TopicalTheme {
+  index: number;
+  title: MultilanguageString;
+  dynamicSubtitle: MultilanguageString;
+  icon: TopicalIcon;
+  themeTiles: TopicalThemeTile[];
+  moreLinks: {
+    label: {
+      DESKTOP: MultilanguageString;
```
VWSCoronaDashboard18, line 1235: Why are these in uppercase?

VWSCoronaDashboard18 (commented):
How does the new TOPICAL.json gets validated during the build/pipeline?

packages/common/src/types/data.ts
VWSCoronaDashboard24: Same as for the `NL` and `EN`, I was thinking that it would be nice to see the "dynamic" parts clearly in the structure. That's why these are uppercase and the rest isn't.

packages/app/src/static-props/get-data.ts
VWSCoronaDashboard24: Yeah for now we keep the former data until the new topical page is finished.

PR #4357, opened 2022-08-11
feature/COR-945-ignore-timeline-event-values-on-y-axis

VWSCoronaDashboard26:
## Summary

* ~~updated TimeSeriesChart component logic to add new `LeadingSeriessType` type definition~~
* ~~updated TimeSeriesChart Series component logic to update `calculateSeriesMaximum` method to override values if a certain series type is leading for calculating maximum values~~
* ~~updated all page components for positively tested people for different scopes of the dashboard~~
* removed all previous references to `leadingSeriesType` after discussion with team mates
* updated series logic to accept a new property, which takes an array of dates in `date_unix` or a combination of `date_start_unix` & `date_end_unix`
* updated series logic to filter out passed exception, resulting in clearer y-axis rendering;

### Motivation
~~Rather than filtering out outliers/anomalies/corrections, which would (arguably, negatively) influence actual data entries in order to properly read data visualisations, I opted to go with a slightly different approach. Instead, I added a new `LeadingSeriesType` definition/prop that determines which series in a TimeSeriesChart is leading out of the two/more series within that chart*. This leading series is the determinator on how to the calculate maximum value of all values (within that leading series) within a TimeSeriesChart.~~

~~_\* The downside here would be that outliers/anomalies/corrections in other visualisations might still give issues, but I am personally strongly against altering data (or making exceptions) within data visualisations for the sake of clearly visualising said data, but this is point for discussion._~~

Upon further discussion, I have implemented a different approach. This approach adds a new property called `yAxisExceptionValues` to a `SeriesCommonDefinition`, meaning it is applicable to all types of series. It accepts an array of `TimestampedValue`, in which an entry would either have a `date_unix` or a combination of `date_start_unix` and `date_end_unix`. The values of the series are then filtered out according to those dates, resulting in better y-axis scaling.

#### Before
National context, where the actual y-axis height (400.000) is influenced by the anomaly/correction on February 8th, 2022, which has a value of 391.379. This is prevented by applying a forced maximum value (`forceMaximumValue` prop) of 150.000, resulting in the y-axis of 150.000 (not in this screenshot). Other screenshots in 'Before' cover different contexts.

before_national_all
before_national_year

VR context.
before_vr_all
before_vr_year

#### After
National context, where the actual y-axis height (150.000) is now influenced by the values excluding the passed dates. Other screenshots in 'After' cover different contexts.

after_national_all
after_national_year

VR context.
after_vr_all
after_vr_year

### DIFF
--- a/packages/app/src/components/time-series-chart/logic/series.ts
+++ b/packages/app/src/components/time-series-chart/logic/series.ts
@@ -10,7 +10,11 @@ import { omit } from 'lodash';
 import { useMemo } from 'react';
 import { hasValueAtKey, isDefined, isPresent } from 'ts-is-present';
 import { useCurrentDate } from '~/utils/current-date-context';
-import { DataOptions, TimespanAnnotationConfig } from './common';
+import {
+  DataOptions,
+  LeadingSeriesType,
+  TimespanAnnotationConfig,
+} from './common';
 import { SplitPoint } from './split';
 
 type SeriesConfigSingle<T extends TimestampedValue> =
@@ -283,9 +287,10 @@ export function useValuesInTimeframe<T extends TimestampedValue>(
 export function calculateSeriesMaximum<T extends TimestampedValue>(
   seriesList: SeriesList,
   seriesConfig: SeriesConfig<T>,
-  benchmarkValue = -Infinity
+  benchmarkValue = -Infinity,
+  leadingSeriesType?: LeadingSeriesType
 ) {
-  const values = seriesList
+  let values = seriesList
     .filter((_, index) => isVisible(seriesConfig[index]))
     .flatMap((series) =>
       series.flatMap((x: SeriesSingleValue | SeriesDoubleValue) =>
@@ -293,6 +298,21 @@ export function calculateSeriesMaximum<T extends TimestampedValue>(
       )
     )
     .filter(isDefined);
+
+  if (leadingSeriesType) {
+    const leadingSeriesTypeIndex = seriesConfig.findIndex(
+      (seriesConfig) => seriesConfig.type === leadingSeriesType
+    );
+
+    if (leadingSeriesTypeIndex >= 0) {
+      const leadingSeries = seriesList[leadingSeriesTypeIndex];
+      values = leadingSeries
+        .flatMap((x: SeriesSingleValue | SeriesDoubleValue) =>
+          isSeriesSingleValue(x) ? x.__value : [x.__value_a, x.__value_b]
+        )
+        .filter(isDefined);
+    }
+  }
 
   const overallMaximum = Math.max(...values);
 
--- a/packages/app/src/components/time-series-chart/logic/series.ts
+++ b/packages/app/src/components/time-series-chart/logic/series.ts
@@ -2,6 +2,7 @@ import {
   getValuesInTimeframe,
   isDateSeries,
   isDateSpanSeries,
+  isDateSpanValue,
   TimeframeOption,
   TimestampedValue,
 } from '@corona-dashboard/common';
@@ -10,11 +11,7 @@ import { omit } from 'lodash';
 import { useMemo } from 'react';
 import { hasValueAtKey, isDefined, isPresent } from 'ts-is-present';
 import { useCurrentDate } from '~/utils/current-date-context';
-import {
-  DataOptions,
-  LeadingSeriesType,
-  TimespanAnnotationConfig,
-} from './common';
+import { DataOptions, TimespanAnnotationConfig } from './common';
 import { SplitPoint } from './split';
 
 type SeriesConfigSingle<T extends TimestampedValue> =
@@ -68,6 +65,13 @@ interface SeriesCommonDefinition {
    * legend, for example)
    */
   hideInLegend?: boolean;
+  /**
+   * Specifies different exception values (in either data_unix or date_start_unix & date_end_unix)
+   * notitation to be ignored when scaling the Y-axis based on the maximum values
+   * of a series. Values matching with the passed in date(s) will be ignored and
+   * are therefore not eligible as maximum values.
+   */
+  yAxisExceptionValues?: TimestampedValue[];
 }
 
 export interface GappedLineSeriesDefinition<T extends TimestampedValue>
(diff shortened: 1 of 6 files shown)


VWSCoronaDashboard19 (approved):

packages/app/src/components/time-series-chart/logic/series.ts
```diff
@@ -285,13 +293,37 @@ export function calculateSeriesMaximum<T extends TimestampedValue>(
...
+          });
+        } else {
+          if (!datesToFilter.includes(dateValue)) datesToFilter.push(dateValue);
+        }
+      });
+    }
+  });
+
   const values = seriesList
     .filter((_, index) => isVisible(seriesConfig[index]))
```
VWSCoronaDashboard25, line 315: Right now, you filter all series for a date given in a `yAxisExceptionValue`, instead of filtering it only in the corresponding series, which is not intended, otherwise it could we a graph config, instead of a series config.

On this line the assumption is made that `seriesConfig[0]` belongs to `seriesList[0]`, `...[1]` belongs to `...[1]`, etc.
This means you can zip them together with
```typescript
const zipped = seriesList.map(function(series, index) {
  return [series, seriesConfig[index]];
});
```
To create
```typescript
[
  [seriesList[0], seriesConfig[0]],
  [seriesList[1], seriesConfig[1]],
  etc.
]
```

With this zipped array, you can use anonymous functions in the form of `([series, config]) => doStuff()` for each `filter` and `map` etc. And always have the correct config with the series in your logic.

This way you can more easily assure a `yAxisExceptionValue` is only used for the given series and not for other series.

packages/app/src/components/time-series-chart/logic/series.ts
```diff
@@ -285,8 +293,38 @@ export function calculateSeriesMaximum<T extends TimestampedValue>(
...
+  const configsToFilterIndexes: number[] = [];
+  const datesToFilter: number[] = [];
+  seriesConfig.forEach((config, index) => {
+    if (config.yAxisExceptionValues) {
+      configsToFilterIndexes.push(index);
+      config.yAxisExceptionValues.forEach((value) => {
+        const dateValue = isDateSpanValue(value)
+          ? [value.date_start_unix, value.date_end_unix]
+          : value.date_unix;
+        if (Array.isArray(dateValue)) {
```
VWSCoronaDashboard25, line 305: Is`Array.isArray(dateValue)` not basically the same as `isDateSpanValue(value)`, because based on that you build either an array or a single value in `dateValue`?

I think this can be reworked to
```typescript
config.yAxisExceptionValues.forEach((value) => {
  if (isDateSpanValue(value) {
    datesToFilter.push(value.date_start_unix, value.date_end_unix); // TIL: push takes multiple arguments and adds them all
  } else {
    datesToFilter.push(value.date_unix);
  }
});
```

packages/app/src/components/time-series-chart/logic/series.ts
```diff
@@ -285,8 +293,38 @@ export function calculateSeriesMaximum<T extends TimestampedValue>(
   seriesConfig: SeriesConfig<T>,
   benchmarkValue = -Infinity
 ) {
+  const configsToFilterIndexes: number[] = [];
+  const datesToFilter: number[] = [];
+  seriesConfig.forEach((config, index) => {
+    if (config.yAxisExceptionValues) {
```
VWSCoronaDashboard25, line 299: An empty array is truthy in JS, which means that if `config.yAxisExceptionValues` is `[]`, this `if` will still be entered, which means the index of the series is added to `configsToFilterIndexes`, which has a side effect for the corresponding series (see other comment) even though the array is empty.

To check both for if it's an array and it's not empty, I think you can check for `config.yAxisExceptionValues?.length > 0`.

packages/app/src/components/time-series-chart/logic/series.ts
```diff
@@ -285,8 +293,38 @@ export function calculateSeriesMaximum<T extends TimestampedValue>(
...
+      configsToFilterIndexes.push(index);
+      config.yAxisExceptionValues.forEach((value) => {
+        const dateValue = isDateSpanValue(value)
+          ? [value.date_start_unix, value.date_end_unix]
+          : value.date_unix;
+        if (Array.isArray(dateValue)) {
+          dateValue.forEach((value) => {
+            if (!datesToFilter.includes(value)) datesToFilter.push(value);
```
VWSCoronaDashboard25, line 307: You can remove duplicates at the end by converting it to a `Set` and then spread that in an array:
```typescript
const uniqueDatesToFilter = [...new Set(datesToFilter)];
```

Although this is maybe less explicit than your solution, the reason to do keep duplicates out in the first way seems to be for a small optimisation later on, in `!datesToFilter.includes(series.__date_unix)`.
However, `includes` walks over the complete array, so you're trading walking less over `datesToFilter` later on, by walking more over `datesToFilter` early on.

packages/app/src/components/time-series-chart/logic/series.ts
```diff
@@ -285,8 +293,38 @@ export function calculateSeriesMaximum<T extends TimestampedValue>(
...
+          if (!datesToFilter.includes(dateValue)) datesToFilter.push(dateValue);
+        }
+      });
+    }
+  });
+
   const values = seriesList
     .filter((_, index) => isVisible(seriesConfig[index]))
+    .map((series, index) => {
+      if (configsToFilterIndexes.includes(index)) {
```
VWSCoronaDashboard25, line 319: If series A has date 1 to filter, and series B has date 2 to filter, `datesToFilter` will gain both date 1 and date 2, right?
And `configsToFilterIndexes` will both gain the indexes of series A and series B.

So in this part both date 1 and date 2 will be filtered from both series A and from series B. Because `configsToFilterIndexes.includes(index)` will yield `true` for series A, since it contributed date 1.
And `configsToFilterIndexes` wil also include the index for series B, since it contributed date 2.

Then `!datesToFilter.includes(series.__date_unix)` will be false (and thus excluded) for both date 1 and date 2 in series A and both date 1 and date 2 in series B.

Basically I think you want to achieve this filtering:

|   | A | B |
|---|---|---|
| 1 | X |   |
| 2 |   | X |

However I think this results in this filtering:

|   | A | B |
|---|---|---|
| 1 | X | X |
| 2 | X | X |

packages/app/src/components/time-series-chart/logic/series.ts
VWSCoronaDashboard26: Opted to not go with this route, hence marking this comment as resolved for now.

packages/app/src/components/time-series-chart/logic/series.ts
VWSCoronaDashboard26: Opted to not go with this route, hence marking this comment as resolved for now.

VWSCoronaDashboard25 (approved):
Nice! Thanks for taking the time to make this pretty! 😁

PR #4358, opened 2022-08-12
Dynamic tooltip variants chart

VWSCoronaDashboard19:
## Summary
- Tooltip has a dynamic width and columns layout
- Filtering out zero values from tooltip
- Show zero value when an option is selected

### DIFF
--- a/packages/app/src/domain/variants/variants-stacked-area-tile/variants-stacked-area-tile.tsx
+++ b/packages/app/src/domain/variants/variants-stacked-area-tile/variants-stacked-area-tile.tsx
@@ -150,17 +150,27 @@ function VariantStackedAreaTileWithData({
             }}
             formatTooltip={(context) => {
               /**
-               * In the chart the 'other_percentage' stack is rendered on top,
-               * but in the tooltip it needs to be displayed as the last item.
-               * (These are both design decisions)
+               * Filter out zero values in value object, so it will be invisible in the tooltip.
+               * When a selection has been made, the zero values will be shown in the tooltip.
                */
+              const metricCount = context.config.length;
+              const totalMetricCount = seriesConfig.length;
+
+              const filteredValues = Object.fromEntries(
+                Object.entries(context.value).filter(([key, value]) =>
+                  key.includes('percentage') ? value !== 0 : value
+                )
+              ) as VariantChartValue;
+
               const reorderContext = {
                 ...context,
                 config: [
                   ...context.config.filter(
                     (x) =>
                       !hasMetricProperty(x) ||
-                      x.metricProperty !== 'other_graph_percentage'
+                      filteredValues[x.metricProperty] ||
+                      (metricCount !== totalMetricCount &&
+                        x.metricProperty !== 'other_graph_percentage')
                   ),
                   context.config.find(
                     (x) =>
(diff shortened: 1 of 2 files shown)


packages/app/src/domain/variants/variants-stacked-area-tile/variants-stacked-area-tile.tsx
```diff
@@ -150,27 +150,55 @@ function VariantStackedAreaTileWithData({
...
-               * (These are both design decisions)
+               * Filter out zero values in value object, so it will be invisible in the tooltip.
+               * When a selection has been made, the zero values will be shown in the tooltip.
                */
+              const metricCount = context.config.length;
+              const totalMetricCount = seriesConfig.length;
+
+              const filteredValues = Object.fromEntries(
+                Object.entries(context.value).filter(([key, value]) =>
+                  key.includes('percentage') ? value !== 0 : value
```
VWSCoronaDashboard25, line 161: So if the key contains `percentage` it should only show if the value is not `0`.
If the key does not contain `percentage` it should only show if the value is truthy.

With `0` being already falsy, and other number values being truthy, the difference in the result for key with and keys without `percentage` is subtile: in the first `undefined`, `null`, and `NaN` are allowed because they are not `0`, in the second they are not allowed because they als falsy.

If this is intentional, maybe write it in such a way that it focuses on those edge cases instead? That way, the reader of this code in the future doesn't have to second guess whether or not this is intentional.

packages/app/src/domain/variants/variants-stacked-area-tile/variants-stacked-area-tile.tsx
```diff
@@ -150,27 +150,55 @@ function VariantStackedAreaTileWithData({
...
                   ),
                 ].filter(isDefined),
+                value:
+                  metricCount === totalMetricCount
+                    ? filteredValues
+                    : context.value,
               };
 
-              return <TooltipSeriesList data={reorderContext} hasTwoColumns />;
+              const percentageValues = Object.keys(reorderContext.value).filter(
```
VWSCoronaDashboard25, line 187: This is only created to take the length further on. Maybe take the length immediately and name the variable `percentageValuesCount` or `percentageValuesLength`.

Btw, I suggested 'count' because you've done that for the other variables as well. But personally I prefer 'amount' over 'count'. Because 'count' feels like something that can be updated further on in the code, for example as the internal state of a counter, which you +1 if something is added. Whereas 'amount' feels more as 'this is just how many there are now', which suits this situation better.
But this is just a personal preference, you do what you think is best.

packages/app/src/domain/variants/variants-stacked-area-tile/variants-stacked-area-tile.tsx
```diff
@@ -150,27 +150,57 @@ function VariantStackedAreaTileWithData({
...
               const reorderContext = {
                 ...context,
                 config: [
                   ...context.config.filter(
                     (x) =>
                       !hasMetricProperty(x) ||
-                      x.metricProperty !== 'other_graph_percentage'
+                      filteredValues[x.metricProperty] ||
+                      (metricAmount !== totalMetricAmount &&
+                        x.metricProperty !== 'other_graph_percentage')
```
VWSCoronaDashboard24, line 175: I find this a bit hard to read. Would it make sense to extract the statements into variables to make it a bit more explicit?

VWSCoronaDashboard24 (approved):

packages/app/src/domain/variants/variants-stacked-area-tile/variants-stacked-area-tile.tsx
VWSCoronaDashboard19: Agreed, will do

PR #4363, opened 2022-08-17
COR-990 the topical page component

VWSCoronaDashboard18:


packages/app/src/domain/topical/components/topical-tile/index.ts
```diff
@@ -0,0 +1,2 @@
+
+export * from './topical-tile';
```
VWSCoronaDashboard24, line 2:
```suggestion
export * from './topical-tile';
```

packages/app/src/domain/topical/components/topical-tile/topical-tile.tsx
```diff
@@ -0,0 +1,137 @@
...
+  label: string;
+  href: string;
+} | null;
+
+interface TopicalTileProps {
+  title: string;
+  tileIcon: TopicalIcon;
+  trendIcon?: TrendIcon;
+  dynamicDescription: string;
+  cta?: Cta;
```
VWSCoronaDashboard24, line 33:
```suggestion
  cta: Cta;
```
The call to action prop is always there, but it can be `null`. So it doesn't need to be optional :-)

packages/app/src/domain/topical/components/topical-tile/topical-tile.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+type TrendIcon = {
+  direction: 'UP' | 'DOWN';
+  color: 'GREEN' | 'RED';
+} | null;
+
+type Cta = {
+  label: string;
+  href: string;
+} | null;
```
VWSCoronaDashboard24, line 26: We probably can import this from `data.ts` but we can do it later

packages/app/src/domain/topical/components/topical-tile/topical-tile.tsx
```diff
@@ -0,0 +1,137 @@
...
+import { asResponsiveArray } from '~/style/utils';
+import { colors } from '@corona-dashboard/common';
+import DynamicIcon from '~/components/get-icon-by-name';
+import { Chevron, Down, Up } from '@corona-dashboard/icons';
+import { Markdown } from '~/components/markdown';
+import { TopicalIcon } from '@corona-dashboard/common/src/types';
+import { isDefined } from 'ts-is-present';
+
+interface IconWrapperProps {
+  iconColor: any;
```
VWSCoronaDashboard24, line 15:
```suggestion
  iconColor: string;
```

VWSCoronaDashboard24 (approved):

packages/app/src/domain/topical/components/topical-tile/topical-tile.tsx
VWSCoronaDashboard18: At the time of creating this part, this was not a distinct type in data.ts. So will fix this when structuring the page.

packages/app/src/domain/topical/components/topical-tile/topical-tile.tsx
VWSCoronaDashboard18: That's correct, but I imagine it could be optional in some cases when used outside the current JSON and therefore making it optional. How about the trendIcon? That one is optional as well.

PR #4372, opened 2022-08-22
Create consistent links on all pages

VWSCoronaDashboard19:
## Summary
- Add correct icons to internal and external links
- Add icon to download links
- Pass metadata from parent to create consistency

#### Small glimpse of the changes:

_Before:_

Screenshot 2022-08-22 at 11 08 23

--------

_After:_

Screenshot 2022-08-22 at 11 08 17


packages/app/src/components/anchor-tile.tsx
```diff
@@ -30,26 +30,34 @@ export function AnchorTile({
       </Content>
 
       <LinkContainer>
-        {!external && (
+        {!external ? (
```
VWSCoronaDashboard26, line 33: I would prefer to have an optimistic approach to our application flow, meaning we do not necessarily check if X 'is not' Y, but if X 'is' Y. This would result in changing `!external` to `external` and flipping the true/false values for this ternary.

packages/app/src/components/cms/rich-content.tsx
```diff
@@ -309,10 +322,20 @@ function InlineLinkMark(props: { children: ReactNode; mark: InlineLink }) {
   if (!mark.href) return <>{children}</>;
 
   return isAbsoluteUrl(mark.href) ? (
-    <ExternalLink href={mark.href}>{children}</ExternalLink>
+    <ExternalLink
+      href={mark.href}
+      display={'inline-flex'}
+      width={'max-content'}
```
VWSCoronaDashboard26, line 328: Please correct me if I am wrong here, but would `display="inline-flex"` and `width="max-content"` not suffice here? The values are static, so what is the reason to use expression symbols here?

packages/app/src/components/markdown.tsx
```diff
@@ -23,13 +24,17 @@ interface LinkProps {
 const renderers = {
   link: (props: LinkProps) =>
     isAbsoluteUrl(props.href) ? (
-      <ExternalLink href={props.href}>{props.children}</ExternalLink>
+      <ExternalLink href={props.href} display={'inline-flex'}>
```
VWSCoronaDashboard26, line 27: Ditto.

packages/app/src/components/anchor-tile.tsx
VWSCoronaDashboard19: Agreed

packages/app/src/components/cms/rich-content.tsx
VWSCoronaDashboard19: No specific reason but autocompletion of my text editor. Passing a string directly will suffice in this case.

VWSCoronaDashboard26 (approved):

PR #4373, opened 2022-08-22
bugfix/COR-973-time-series-chart-duplicate-x-axis-ticks

VWSCoronaDashboard26:
## Summary

* updated Axes component with new `getDx` and extended `getAnchor` functionality to 'resolve' alignment issues when data is incomplete and only one tick value is generated
* Associated data can be provided when needed for reviewing. I tested/reproduced the different scenarios (from the screenshots) by adjusting the date on my local machine.

## Drawbacks

Despite this fix, I think it would be wise to revise the generic Axes implementation when it comes to generating and setting tick values on the X-axis of different visualisations. There are some recurring issues with this implementation:

* tick values that do not align with the data entry points inside a visualisation (this is more common for visualisations with less entry points and less ticks);
* tick value hover states that do not align with the base tick value position;
* a hardcoded preferred tick value amount is set, regardless of the amount of data entries for the specific visualisation and the selected timeframe of that specific visualisation;

I opted to not go with this more extensive fix for now, as this is part 2 of a bug fix ticket that was part of an earlier sprint.

### Screenshots
Screenshots cover visualisations after changes with both complete and incomplete data sets, covering two different timeframes.
after_incomplete_data_week
Incomplete data set (relative to date), week timeframe.

after_incomplete_data_month
Incomplete data set (relative to date), month timeframe.

after_complete_data_week
Complete data set (relative to date), week timeframe.

after_complete_data_month
Complete data set (relative to date), month timeframe.


VWSCoronaDashboard19 (approved):

VWSCoronaDashboard24 (approved):
Looks good. Can you create a ticket for the refactor which you explained in this PR description?

VWSCoronaDashboard25 (approved):

PR #4374, opened 2022-08-23
Generate icons.md + add IntensiveCareOpnames icon

VWSCoronaDashboard24:


packages/icons/src/icon-name2filename.ts
```diff
@@ -189,6 +190,7 @@ export const iconName2filename: Record<IconName, string> = {
   HorecaEvenementen: 'horeca_evenementen.svg',
   Hospitality: 'hospitality.svg',
   Information: 'information.svg',
+  IntensiveCareOpnames: 'intensive care-opnames.svg',
```
VWSCoronaDashboard18, line 193: Use snake_case for the naming like al the other icons...

VWSCoronaDashboard18 (changes requested):

packages/icons/src/icon-name2filename.ts
VWSCoronaDashboard24: Done!

VWSCoronaDashboard18 (approved):

PR #4375, opened 2022-08-23
bugfix/COR-1005-undefined-time-series-chart-values

VWSCoronaDashboard26:
## Summary

* updated `getValueString` method in `useFormatSeriesValue` hook to correctly fall back to '-' when a value is either `null` or `undefined`

### Screenshots
Screenshots cover visualisations before and after changes, in which two scenarios are covered (data entries where a value is either `null` or `undefined`). It is important to note that these screenshots only cover the 'Number of confirmed cases among over-70s living at home over time' visualisation as both scenarios are applicable here, but the fix covers all other time series charts where either of the different scenarios occur.

#### Before
before_null_instance
Notice how the value is `null`.

before_undefined_instance
Notice how the value is `undefined`.

#### After
after_null_instance
Notice how the previous `null` value is now marked as '-'.

after_undefined_instance
Notice how the previous `undefined` value is now marked as '-'.

### DIFF
--- a/packages/app/src/components/time-series-chart/logic/use-format-series-value.ts
+++ b/packages/app/src/components/time-series-chart/logic/use-format-series-value.ts
@@ -19,12 +19,12 @@ export function useFormatSeriesValue<T extends TimestampedValue>(
       const formatter =
         metricPropertyFormatters[metricProperty] || intl.formatNumber;
       const numberValue = value[metricProperty];
+      if (!isPresent(numberValue)) return '-';
+
       const formattedValue =
         isPresent(numberValue) && typeof numberValue === 'number'
           ? formatter(numberValue)
-          : isPresent(numberValue)
-          ? String(numberValue)
-          : '-';
+          : String(numberValue);
 
       return isPresent(formattedValue)
         ? isPercentage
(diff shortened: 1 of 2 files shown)


packages/app/src/components/time-series-chart/logic/use-format-series-value.ts
```diff
@@ -22,7 +22,9 @@ export function useFormatSeriesValue<T extends TimestampedValue>(
       const formattedValue =
         isPresent(numberValue) && typeof numberValue === 'number'
           ? formatter(numberValue)
-          : String(numberValue);
+          : isPresent(numberValue)
+          ? String(numberValue)
+          : '-';
```
VWSCoronaDashboard24, line 27: Maybe we can introduce an early return to make the conditionals on line 23-27 en 30-33 a bit simpler?

```ts
if (!isPresent(numberValue)) {
  return '-';
}
```

VWSCoronaDashboard26: This indeed is much simpler and straight to the point. 😅 👍 

Marking as resolved as I pushed the fix.

VWSCoronaDashboard24 (approved):
Clean!

PR #4380, opened 2022-08-25
Removed unnecessary width for external links

VWSCoronaDashboard27:
## Summary

Removed `display: inline-block` and `width`, so the links became flexible again.

PR #4383, opened 2022-08-25
Added link property, removed redundancy and fixed size of icon

VWSCoronaDashboard19:
# Summary
- Fixed size of external icon
- Remove redundant property
- Added extra display prop so link and icon fits correctly

### DIFF
--- a/packages/app/src/components/markdown.tsx
+++ b/packages/app/src/components/markdown.tsx
@@ -33,7 +33,7 @@ const renderers = {
       </ExternalLink>
     ) : (
       <Link href={props.href} passHref>
-        <Anchor underline>
+        <Anchor underline display="inline-block">
           {props.children} <ChevronRight width={10} height={10} />
         </Anchor>
       </Link>
--- a/packages/app/src/components/typography.tsx
+++ b/packages/app/src/components/typography.tsx
@@ -74,8 +74,7 @@ export const Anchor = styled.a<AnchorProps>(
     props.display &&
     css({
       display: props.display,
-    }),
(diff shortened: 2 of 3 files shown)


packages/app/src/components/markdown.tsx
```diff
@@ -33,7 +33,7 @@ const renderers = {
       </ExternalLink>
     ) : (
       <Link href={props.href} passHref>
-        <Anchor underline>
+        <Anchor underline display="inline-block">
```
VWSCoronaDashboard18, line 36: At line very close recently "inline-block" is removed: https://github.com/minvws/nl-covid19-data-dashboard/pull/4380/files

It seems a different part but not entirely sure is that removal has the same backstory for this element NOT to have 'inline-block"

VWSCoronaDashboard19: This is for the markdown component only and it's also applied for the `<ExternalLink />` line 30

packages/app/src/components/typography.tsx
```diff
@@ -74,8 +74,7 @@ export const Anchor = styled.a<AnchorProps>(
     props.display &&
     css({
       display: props.display,
-    }),
-  (props) => props.width && css({ width: props.width })
+    })
```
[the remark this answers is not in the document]
VWSCoronaDashboard19: No, it was only applied for https://github.com/minvws/nl-covid19-data-dashboard/pull/4380/files. Therefore it can be removed.

VWSCoronaDashboard18 (approved):
Great! Thanks for the comment answers!

PR #4385, opened 2022-08-25
COR-1022 fixing CLS by removing nested Anchor tag

VWSCoronaDashboard18:
Before:

https://user-images.githubusercontent.com/93984341/186692839-466ef8fd-f3fa-49f2-ac3b-d90184718e50.mov

After:

https://user-images.githubusercontent.com/93984341/186692893-d7d0b9e3-744b-4bbd-9abf-4d9043f89241.mov

### DIFF
--- a/packages/app/src/domain/topical/components/text-with-chevron.tsx
+++ b/packages/app/src/domain/topical/components/text-with-chevron.tsx
@@ -9,11 +9,11 @@ type TextWithChevronProps = {
   label: string;
 };
 
-export function TextWithChevron({ label }: TextWithChevronProps) {
+export const TextWithChevron = ({ label }: TextWithChevronProps) => {
   return (
     <Box
       display="flex"
-      justifyContent={'center'}
+      justifyContent="center"
       alignItems="center"
       bg={colors.lightBlue}
       color={colors.blue}
--- a/packages/app/src/domain/topical/components/text-with-chevron.tsx
+++ b/packages/app/src/domain/topical/components/text-with-chevron.tsx
@@ -1,0 +1,44 @@
+import { Box } from '~/components/base';
+import css from '@styled-system/css';
+import styled from 'styled-components';
+import { ChevronRight } from '@corona-dashboard/icons';
+import { colors } from '@corona-dashboard/common';
+import { Text } from '~/components/typography';
+
+type TextWithChevronProps = {
+  label: string;
+};
+
+export function TextWithChevron({ label }: TextWithChevronProps) {
+  return (
+    <Box
+      display="flex"
+      justifyContent={'center'}
+      alignItems="center"
+      bg={colors.lightBlue}
+      color={colors.blue}
+      padding={3}
+      className="topical-tile-cta"
+      css={css({
+        transition: 'background .1s ease-in-out',
+      })}
+    >
+      <Text>{label}</Text>
+      <IconSmall>
+        <ChevronRight />
+      </IconSmall>
+    </Box>
+  );
+}
+
+const IconSmall = styled.div(
+  css({
+    display: 'inline',
+    textDecoration: 'inherit',
+    svg: {
+      width: 11,
+      height: 10,
+    },
+    marginLeft: 1,
+  })
+);
--- a/packages/app/src/domain/topical/components/topical-tile.tsx
+++ b/packages/app/src/domain/topical/components/topical-tile.tsx
@@ -111,7 +111,7 @@ export function TopicalTile({
           </Box>
         </Box>
 
-        {cta ? <TextWithChevron label={cta.label}></TextWithChevron> : null}
+        {cta && <TextWithChevron label={cta.label}></TextWithChevron>}
       </>
     </Box>
   );
--- a/packages/app/src/domain/topical/components/topical-tile.tsx
+++ b/packages/app/src/domain/topical/components/topical-tile.tsx
@@ -3,11 +3,11 @@ import { space } from '~/style/theme';
 import css from '@styled-system/css';
 import styled from 'styled-components';
 import { Heading } from '~/components/typography';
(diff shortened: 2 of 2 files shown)


packages/app/src/domain/topical/components/text-with-chevron.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+type TextWithChevronProps = {
+  label: string;
+};
+
+export function TextWithChevron({ label }: TextWithChevronProps) {
+  return (
+    <Box
+      display="flex"
+      justifyContent={'center'}
```
VWSCoronaDashboard26, line 16: I reviewed some of this same code yesterday and can imagine that you haven't gone through at all the comments yet as this has more priority, but you could also write this as `justifyContent="center"`, as the expression is not necessary because of the prop value being static.

packages/app/src/domain/topical/components/text-with-chevron.tsx
```diff
@@ -0,0 +1,44 @@
...
+      alignItems="center"
+      bg={colors.lightBlue}
+      color={colors.blue}
+      padding={3}
+      className="topical-tile-cta"
+      css={css({
+        transition: 'background .1s ease-in-out',
```
VWSCoronaDashboard26, line 23: Did design already make a decision about this? Something about two birds with one stone. 😜

packages/app/src/domain/topical/components/topical-tile.tsx
```diff
@@ -113,27 +111,7 @@ export function TopicalTile({
...
-            <LinkWithIcon
-              href={cta.href}
-              icon={<ChevronRight />}
-              iconPlacement="right"
-            >
-              {cta.label}
-            </LinkWithIcon>
-          </Box>
-        ) : null}
+        {cta ? <TextWithChevron label={cta.label}></TextWithChevron> : null}
```
VWSCoronaDashboard26, line 114: Would `{cta && <TextWithChevron label={cta.label}></TextWithChevron>}` not be sufficient?

It is not yet in our coding standards as the branch did not get merged yet, but I do want to point out [this line](https://github.com/minvws/nl-covid19-data-dashboard/blob/feature/update-code-conventions/README.md?plain=1#L83) of the WIP version.

packages/app/src/domain/topical/components/text-with-chevron.tsx
```diff
@@ -0,0 +1,44 @@
...
+import styled from 'styled-components';
+import { ChevronRight } from '@corona-dashboard/icons';
+import { colors } from '@corona-dashboard/common';
+import { Text } from '~/components/typography';
+
+type TextWithChevronProps = {
+  label: string;
+};
+
+export function TextWithChevron({ label }: TextWithChevronProps) {
```
VWSCoronaDashboard26, line 12: This is not yet in the coding standards (it is in the [WIP version](https://github.com/minvws/nl-covid19-data-dashboard/blob/feature/update-code-conventions/README.md?plain=1#L79)), but this should become a function expression.

packages/app/src/domain/topical/components/text-with-chevron.tsx
VWSCoronaDashboard18: Yeah, exactly. Therefor not changing the current state. (This is a component replacement for the topical <LinkWithIcon/>. So I kept the same look and functions, but only removed the a tag)

packages/app/src/domain/topical/components/topical-tile.tsx
VWSCoronaDashboard18: Yes, great addition

packages/app/src/domain/topical/components/text-with-chevron.tsx
VWSCoronaDashboard18: Fixed it by changing to a const with a arrow function

packages/app/src/domain/topical/components/text-with-chevron.tsx
```diff
@@ -0,0 +1,44 @@
...
+  css({
+    display: 'inline',
+    textDecoration: 'inherit',
+    svg: {
+      width: 11,
+      height: 10,
+    },
+    marginLeft: 1,
+  })
+);
```
VWSCoronaDashboard26, line 44: Sorry to write another comment after my initial review but this was actually part of my original review and somehow never made it in/got deleted.

Maybe I am taking this a bit overboard and overthinking this somewhat simple component, but would it make sense to do some minor adjustments to make it more reusable (also outside of the context of the new topical page)?
* The component now includes a `Box` parent component that basically defines that a relatively simple sounding component like this always has somewhat of a button-like appearance. What if you do not add the box here so that this component will be just what it is supposed to be: text with a chevron?
* The component now always assumes that it wants to display a chevron but, coming back to reusability, how about renaming this to somewhere along the lines of `TextWithIcon` where you would pass both the text (or label) and an Icon component as props to this component?

You can ignore the above if you would like, but I still wanted to mention this.

VWSCoronaDashboard18: Maybe it's good to know the backstory. This whole component is a stripped-down version of <LinkWithIcon> and it has no intention to be used somewhere else. I totally agree that the naming can be done better. But not at making it more general. For example: Yes, it will always have a Chevron, because of the specific purpose of this component. I can comply with the argument to use another name than 'text'. I think 'button' is a better descriptive name.

The reason for keeping this rather simple component in a separate file is because of the specific purpose of the code, not because of reusability.

VWSCoronaDashboard24: As discussed we address this feedback at a later moment.

VWSCoronaDashboard24 (approved):

PR #4387, opened 2022-08-29
documentation(choropleth): add documentation on the data structure that is used for choropleth maps (TopoJson/GeoJson) and what that means for the code base.

VWSCoronaDashboard25:


VWSCoronaDashboard18 (commented):
Screenshot 2022-08-30 at 08 46 29
grammarly.com returns some spelling errors. Some of the errors can be ignored. But some are relevant.

docs/choropleth-data.md
```diff
@@ -0,0 +1,96 @@
...
+This mapping is done in [use-projected-coordinates.ts](/packages/app/src/components/choropleth/logic/use-projected-coordinates.ts).
+Unfortunately, the author did not know about the holes in a polygon.
+Just before the code that extracts polygon holes and adds them as new separate polygons, it is commented:
+  > turf.flatten() doesn't properly flatten deeply nested Multipolygons,
+  so here we check if the coordinates consist of number tuples, if not,
+  we can assume it's a deeper nested polygon and extract the tuples.
+
+It's a shame that every multipolygon (a feature consisting of multiple polygons) was split up into separate polygons.
+It's even worse that every hole was extracted as its own polygon as well.
+The comment that `turf.flatten()` doesn't do its job properly is remarkable, considering that [Turf is an external library with over 200.000 weekly downloads](https://www.npmjs.com/package/@turf/turf), designed specifically to work with GeoJson data structures.
```
VWSCoronaDashboard24, line 71: Not sure about this section. I think it's good to document the "lessons learned". But the above sounds a little bit like a roast of the author and the decisions made. Which might be valid but is not that professional to do (imho). Feel free to leave it as is but maybe give it a second thought.

VWSCoronaDashboard24 (approved):
Very informative, important knowledge to document!

docs/choropleth-data.md
VWSCoronaDashboard25: I think it is important to try to understand why choices are made. At the moment, this is the current state of that understanding, which is, we don't understand why the choice was made.

If we're going to change how the choropleth is made, it is important we know what we are doing. So we need to feel comfortable with our understanding of it all. If we're going to revert the choice to use Konva, it is very important that we understand it to at least the level of our predecessors. Any hint on what our predecessors knew and what their level of understanding was, is helpful to gather in this process, because we can't contact them to ask.

For the above section, if you have a suggestion on how to put it differently, I'm happy to hear!

VWSCoronaDashboard24: Ok, for the sake of documenting the decisions made let's leave it like this then!

PR #4390, opened 2022-08-31
COR-941 removed access sidebar code

VWSCoronaDashboard18:


packages/app/src/static-props/situations/get-situations-sidebar-value.ts
```diff
@@ -1,20 +0,0 @@
-import { VrCollectionSituations } from '@corona-dashboard/common';
```
VWSCoronaDashboard19, line 1: Is the `situations` directory also being removed here? Because this file is the only file in there.

VWSCoronaDashboard18: Yes, It's removed as well. Github isn't sufficient enough to make that clear.

PR #4399, opened 2022-09-02
bugfix(combo-box): always use matchSorter to sort the options

VWSCoronaDashboard25:
Bug was introduced by the previous COR-748 commit

Closes COR-748

### DIFF
--- a/packages/app/src/components/combo-box/combo-box.tsx
+++ b/packages/app/src/components/combo-box/combo-box.tsx
@@ -54,7 +54,7 @@ export const ComboBox = <Option extends TOption>(props: TProps<Option>) => {
   const inputRef = useRef<HTMLInputElement>(null);
   const containerRef = useRef<HTMLUListElement>(null);
   const [inputValue, setInputValue] = useState<string>('');
-  const results = useSearchedOptions<Option>(inputValue, options);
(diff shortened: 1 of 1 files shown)


packages/app/src/components/combo-box/combo-box.tsx
```diff
@@ -150,19 +150,17 @@ export const ComboBox = <Option extends TOption>(props: TProps<Option>) => {
...
-const useSearchedOptions = <Option extends TOption>(
+const useSearchedAndSortedOptions = <Option extends TOption>(
   term: string,
   options: Option[]
 ): Option[] => {
   const throttledTerm = useThrottle(term, 100);
 
   return useMemo(
     () =>
-      throttledTerm.trim() === ''
```
VWSCoronaDashboard25, line 161: Because of this ternary, sorting on the full list (without anything entered in the search field) was different from sorting on the search results.
The earlier fix removed the sorting on the full list, and passed a smoke test because I looked at the sorting of the search results instead.

This fix sorts the full list the same as the search results by removing the ternary. It also updates the priority of sorting: sorting on displayName is most important.

VWSCoronaDashboard24 (approved):

VWSCoronaDashboard19 (approved):

PR #4400, opened 2022-09-06
Filter out props passed to the DOM

VWSCoronaDashboard19:
- Add Transient props to prevent props on the DOM: (https://styled-components.com/docs/api#transient-props)
- Change incorrect Aria rule to a valid one
- Different SVG structure for `varianten.svg` so an incorrect attribute will not be build which TypeScript does not understand. Visually nothing changes.


packages/icons/src/svg/varianten.svg
```diff
@@ -1,9 +1,18 @@
...
-      <path d="M34.73,42.36a14.75,14.75,0,0,0-.13-4.31l7.58-.12h.11a4.33,4.33,0,1,1,.17,4.27h0a.09.09,0,0,1,0,0A4.42,4.42,0,0,1,41.78,40v-.06c-.09,0-.4,0-.47.43l-.08,1.93Z" />
-      <path d="M28.09,50.59l-.91,1H5.52V24.41a18,18,0,0,1,5.8-1.95,15.88,15.88,0,0,1,4.34-.17,17.72,17.72,0,0,1,8.75,3.21,15.5,15.5,0,0,1,4.39,4.42,17.59,17.59,0,0,1,2.52,5.51,19.55,19.55,0,0,1,.06,8.48A17.5,17.5,0,0,1,28.09,50.59ZM21.4,34a2.93,2.93,0,0,0,1.67,3.81,3,3,0,0,0,3.86-1.61,2.92,2.92,0,0,0-1.66-3.8A3,3,0,0,0,21.4,34Zm-9-3.54A2.19,2.19,0,1,1,14,26.4a2.18,2.18,0,0,1,1.25,2.85A2.24,2.24,0,0,1,12.37,30.46Zm9.05,15.36a2.19,2.19,0,0,0,1.25,2.86,2.24,2.24,0,0,0,2.9-1.2,2.21,2.21,0,0,0-1.25-2.86A2.23,2.23,0,0,0,21.42,45.82Zm-7.22-2a4.39,4.39,0,0,1-2.5-5.71,4.44,4.44,0,0,1,5.79-2.4A4.37,4.37,0,0,1,20,41.38,4.47,4.47,0,0,1,14.2,43.79Z" fill-rule="evenodd" />
-   </g>
+    <path d="M11.8,12.7C11.8,12.7,11.8,12.8,11.8,12.7c0,0.1,0,0.1,0,0.1l0.1,7.2c1.1-0.2,2.7-0.3,4.2-0.1L16,13.7l-1.9-0.1
```
VWSCoronaDashboard25, line 3: > Different SVG structure for varianten.svg so an incorrect attribute will not be build which TypeScript does not understand. Visually nothing changes.

I see two things done here:
* remove the group tag around the paths
* use different values for the paths

In the result I see the icon is shifted a bit, what was the problem here and how are the above two things fixing that?

VWSCoronaDashboard25: Actually it's mostly a bit smaller:
image

VWSCoronaDashboard19: Checked with Design. The new version is the way to go so no reverting

VWSCoronaDashboard24 (approved):

PR #4401, opened 2022-09-07
Feature/refactor ChartTile

VWSCoronaDashboard21:


packages/app/src/components/chart-tile.tsx
```diff
@@ -10,38 +10,37 @@ import { Heading } from './typography';
...
+  timeframeOptions,
+  onSelectTimeframe,
+  children,
 }: ChartTileProps) {
   const [timeframe, setTimeframe] = useState<TimeframeOption>(
     timeframeInitialValue || TimeframeOption.ALL
   );
 
+  useEffect(() => {
+    if (typeof onSelectTimeframe !== 'undefined') {
```
VWSCoronaDashboard24, line 39:
```suggestion
    if (onSelectTimeframe) {
```
would also suffice I think :-)

packages/app/src/components/error-boundary/index.tsx
```diff
@@ -18,7 +18,7 @@ const PropsReportContext = createContext<
 >(() => undefined);
 
 type ErrorBoundaryProps = {
-  children: ReactNode;
+  children?: ReactNode;
```
VWSCoronaDashboard24, line 21: Without children there will never be an error to catch. Are you sure this prop should be optional?

packages/app/src/domain/sewer/sewer-chart/sewer-chart.tsx
```diff
@@ -137,91 +142,86 @@ export function SewerChart({
         source: text.source,
       }}
       description={text.description}
+      onSelectTimeframe={(timeframe) => setSewerTimeframe(timeframe)}
```
VWSCoronaDashboard24, line 145: Optional; but you could use the shorthand version:

```suggestion
      onSelectTimeframe={setSewerTimeframe}
```

packages/app/src/domain/tested/g-number-bar-chart-tile.tsx
```diff
@@ -49,51 +54,50 @@ export function GNumberBarChartTile({
         date: last_value.date_of_insertion_unix,
         source: text.bronnen,
       }}
+      onSelectTimeframe={(timeframe) => setGnumberTimeframe(timeframe)}
```
VWSCoronaDashboard24, line 57: Same as mentioned before; shorthand can be used.

packages/app/src/domain/tested/reproduction-chart-tile.tsx
```diff
@@ -49,30 +54,29 @@ export function ReproductionChartTile({
...
+            label: text.lineLegendLabel,
+            color: colors.data.primary,
+            minimumRange:
+              metricConfigs?.nl?.reproduction?.index_average?.minimumRange,
+          },
+        ]}
+        dataOptions={{
+          timelineEvents,
+        }}
+        numGridLines={reproductionTimeframe === 'all' ? 4 : 3}
```
VWSCoronaDashboard24, line 78:
```suggestion
        numGridLines={reproductionTimeframe === TimeframeOption.ALL ? 4 : 3}
```

packages/app/src/pages/gemeente/[code]/ziekenhuis-opnames.tsx
```diff
@@ -111,6 +113,12 @@ const IntakeHospital = (props: StaticProps<typeof getStaticProps>) => {
     content,
     lastGenerated,
   } = props;
+
+  const [
+    hospitalAdmissionsOverTimeTimeframe,
+    setHospitalAdmissionsOverTimeTimeframe,
+  ] = useState<TimeframeOption>(TimeframeOption.ALL);
```
Comment (VWSCoronaDashboard24) [inline packages/app/src/pages/gemeente/[code]/ziekenhuis-opnames.tsx:120]: When there is only one state property in a file you could argue that the name for the state could be simpler; `timeframe` & `setTimeframe` because you are already in the "ziekenhuis-opnames" context. But on the other hand: naming things explicit is also very nice

packages/app/src/pages/landelijk/positief-geteste-mensen.tsx
```diff
@@ -156,6 +157,24 @@ const PositivelyTestedPeople = (props: StaticProps<typeof getStaticProps>) => {
...
+
+  const [
+    confirmedCasesTestedOverTimeTimeframe,
+    setConfirmedCasesTestedOverTimeTimeframe,
+  ] = useState<TimeframeOption>(TimeframeOption.ALL);
+
+  const [
+    confirmedCasesInfectedPerAgeTimeframe,
+    setConfirmedCasesInfectedPerAgeTimeframe,
+  ] = useState<TimeframeOption>(TimeframeOption.ALL);
```
VWSCoronaDashboard24, line 176: Maybe we can convince design that we should synchronize the timeframe for graphs on page basis :-) Should clean up thing quite a bit

VWSCoronaDashboard24 (commented):
Nice refactoring! Great having it done in a more "React-way". Added a couple of minor comments.

packages/app/src/components/chart-tile.tsx
VWSCoronaDashboard21: Agreed and changed.

packages/app/src/pages/gemeente/[code]/ziekenhuis-opnames.tsx
Comment (VWSCoronaDashboard21) [reply packages/app/src/pages/gemeente/[code]/ziekenhuis-opnames.tsx:120]: I'll keep it like this to keep it in line with the naming in the other levels.

packages/app/src/domain/tested/reproduction-chart-tile.tsx
VWSCoronaDashboard21: Changed.

packages/app/src/pages/landelijk/positief-geteste-mensen.tsx
VWSCoronaDashboard21: I'll ask Design. :)

packages/app/src/domain/sewer/sewer-chart/sewer-chart.tsx
VWSCoronaDashboard21: Changed.

packages/app/src/domain/tested/g-number-bar-chart-tile.tsx
VWSCoronaDashboard21: Changed

packages/app/src/components/error-boundary/index.tsx
VWSCoronaDashboard25: on line 28 it gets a default value `null` so I think it makes sense to either have this as optional, or remove the default value if it is required. What do you think?

VWSCoronaDashboard24 (approved):

PR #4402, opened 2022-09-08
Add new keys for existing schemas regarding the new autumn shot

VWSCoronaDashboard19:
# Summary
- Add new keys
- Add new archived schemas


VWSCoronaDashboard24 (approved):
I think you should update / regenerate `data.ts` as well?

VWSCoronaDashboard19:
> I think you should update / regenerate `data.ts` as well?

Yes that's correct, I need to check Feature Flags aswell for validation!

PR #4403, opened 2022-09-12
use flexShrink instead of inline-flex

VWSCoronaDashboard21:


packages/app/src/domain/topical/components/topical-tile.tsx
```diff
@@ -121,7 +121,7 @@ const IconWrapper = styled.span<IconWrapperProps>((x) =>
   css({
     color: x.iconColor,
     display: 'inline-flex',
-    width: '20px',
+    minWidth: '20px',
```
VWSCoronaDashboard25, line 124: The `display` property is set on `inline-flex`, but there is only one element inside (the icon) and there are no other flex instructions in this css. So I'm curious why there would is a need for a flex container here.
I think you can remove the `display: 'inline-flex'` line, so you do not need to enforce a min with and can just use width.

VWSCoronaDashboard25: Sorry, correction, this being a flex container has nothing to do with how this is scaled, that's because the parent is a flex container.
By setting `flexShrink: 0` here, you'll let the parent know that the IconWrapper should not be shrunk.

VWSCoronaDashboard21: Thanks for the feedback, I have changed this.

VWSCoronaDashboard25 (approved):
Thanks!

PR #4405, opened 2022-09-12
Remove VR and GM topical links in header

VWSCoronaDashboard19:
## Summary
- Remove page link for VR and GM topical
- Remove redundant Sanity key for VR and GM topical

### DIFF
--- a/packages/app/src/components/layout/app-content.tsx
+++ b/packages/app/src/components/layout/app-content.tsx
@@ -63,6 +63,7 @@ export function AppContent({
         {backButtonUrl && (
           <>
             <BackButtonContainer
+              isMenuOpen={isMenuOpen}
               isVisible={!hideBackButton}
               css={css({
                 background: 'white',
@@ -94,7 +95,11 @@ export function AppContent({
             {children}
           </ResponsiveVisible>
           {backButtonUrl && (
-            <BackButtonContainer isVisible={!hideBackButton} mt={4}>
+            <BackButtonContainer
+              isVisible={!hideBackButton}
+              mt={4}
+              isMenuOpen={isMenuOpen}
+            >
               <LinkWithIcon icon={<ArrowIconLeft />} href={backButtonUrl}>
                 {backButtonText}
               </LinkWithIcon>
@@ -106,10 +111,17 @@ export function AppContent({
   );
 }
 
-const BackButtonContainer = styled(Box)<{ isVisible: boolean }>((x) =>
+const BackButtonContainer = styled(Box)<{
+  isVisible: boolean;
+  isMenuOpen: boolean;
+}>((x) =>
   css({
-    mx: [3, null, 0],
+    mx: x.isMenuOpen
+      ? asResponsiveArray({ _: 1, xs: 'auto' })
+      : asResponsiveArray({ _: 1, sm: 5 }),
     display: [x.isVisible ? 'block' : 'none', null, null, 'none'],
+    px: asResponsiveArray({ _: 1, sm: 1 }),
+    maxWidth: x.isMenuOpen ? '38rem' : undefined,
   })
 );
 
--- a/packages/app/src/components/layout/app-content.tsx
+++ b/packages/app/src/components/layout/app-content.tsx
@@ -1,4 +1,4 @@
-import { colors } from '@corona-dashboard/common';
+import { colors, ScopedData } from '@corona-dashboard/common';
 import css from '@styled-system/css';
 import { useRouter } from 'next/router';
 import styled from 'styled-components';
@@ -16,6 +16,11 @@ interface AppContentProps {
   sidebarComponent: React.ReactNode;
   searchComponent?: React.ReactNode;
   hideBackButton?: boolean;
+}
+
+interface BackButtonProps {
+  text: string;
+  url?: string;
 }
 
 export function AppContent({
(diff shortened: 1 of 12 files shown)


packages/app/src/components/layout/app-content.tsx
```diff
@@ -35,21 +35,25 @@ export function AppContent({
...
    * @TODO Open the menu purely client side without loading a new page
    */
   const backButtonUrl = currentPageScope
     ? isMenuOpen
-      ? reverseRouter.actueel[currentPageScope](currentCode)
+      ? isNational
+        ? reverseRouter.topical.nl
+        : undefined
       : reverseRouter[currentPageScope].index(currentCode)
     : undefined;
```
VWSCoronaDashboard25, line 50: It takes quite some brain power to understand this triple ternary correctly.
I feel like this can be a made better readable if it became a function.
You can write it out like this (add your own TypeScript): 
```typescript
const getBackButtonUrl = (currentPageScope, isMenuOpen) => {
  if (typeof currentPageScope === 'undefined') { // or check with === '', I don't know
    return undefined;
  }
  if (!isMenuOpen) {
    return reverseRouter[currentPageScope].index(currentCode);
  }
  if (currentPageScope === 'nl') {
    return reverseRouter.topical.nl;
  }
  return undefined;
}
```

And then you see that prior `isMenuOpen` led to `reverseRouter.actueel[currentPageScope](currentCode)` where now you need `!isMenuOpen` instead.
Is that intended?

Same for the back button text

packages/app/src/components/layout/app-content.tsx
```diff
@@ -35,21 +35,25 @@ export function AppContent({
     router.query.menu === '1';
 
   const currentPageScope = getCurrentPageScope(router);
-
   const currentCode = router.query.code as string | undefined;
+  const isNational = currentPageScope === 'nl';
```
VWSCoronaDashboard25, line 39: Because this is only used in two almost identical pieces of code, where `currentPageScope` itself is already evaluated, I think it might be nicer to not make a separate variable for this check.
But having the variable is also okay.

packages/app/src/components/layout/app-content.tsx
```diff
@@ -102,10 +111,17 @@ export function AppContent({
   );
 }
 
-const BackButtonContainer = styled(Box)<{ isVisible: boolean }>((x) =>
+const BackButtonContainer = styled(Box)<{
+  isVisible: boolean;
+  isMenuOpen: boolean;
```
VWSCoronaDashboard25, line 116: This property is named after what happens in the parent, instead of the effect it has in this component.
Can you explain what `mx: 'auto'` and `maxWidth: 38rem` do for the `BackButtonContainer`?

And from that answer, think of a name for this property that describes what happens to the component if this boolean is `true`?

(For the usage on line 101 you can still use `isMenuOpen` to set the property)

packages/app/src/components/layout/app-content.tsx
```diff
@@ -102,10 +111,17 @@ export function AppContent({
...
+}>((x) =>
   css({
-    mx: [3, null, 0],
+    mx: x.isMenuOpen
+      ? asResponsiveArray({ _: 1, xs: 'auto' })
```
VWSCoronaDashboard25, line 120: I think this means `1` for each size smaller than `xs` and `'auto'` for size `xs` and larger.
That's basically every size, because `xs` is really small and I don't think we actually need to worry about sizes smaller than `xs`.

So if the above is correct (I'm always in doubt how these things work), I think it's better to just put it at `'auto'` regardless of size.

packages/app/src/domain/topical/vaccination-coverage-choropleth.tsx
```diff
@@ -199,7 +199,7 @@ export function VaccinationCoverageChoropleth(
               }}
               dataOptions={{
                 isPercentage: true,
-                getLink: (vrcode) => reverseRouter.actueel.vr(vrcode),
+                getLink: (vrcode) => reverseRouter.vr.rioolwater(vrcode),
```
VWSCoronaDashboard25, line 202: How about we add a function somewhere:
```typscript
const getVrDefaultLink = (vrCode) => {
  return reverseRouter.vr.rioolwater(vrcode);
}
```
And use that everywhere. Next time, we only have to change the default link in 1 place.

packages/app/src/components/layout/app-content.tsx
VWSCoronaDashboard19: Agreed, it's a bit unnecessary to create an extra variable in this case.

packages/app/src/components/layout/app-content.tsx
VWSCoronaDashboard19: When the sidebar menu is visible on mobile, the width of the page is slightly different compared to the width of the page on page level (where the sidebar menu is not visible). That's is why there is a `maxWidth` and `mx: auto` to align the width and position with the sidebar. The current property name `isMenuOpen` is currently describing the behavior of the sidebar menu, open or closed.

packages/app/src/components/layout/app-content.tsx
VWSCoronaDashboard19: The `1` adds a small margin when the screen gets smaller then `xs`. But I agree that `auto` will do, because of the size of `xs` is already really small

packages/app/src/domain/topical/vaccination-coverage-choropleth.tsx
VWSCoronaDashboard19: Good idea, will do 👍

VWSCoronaDashboard24 (approved):

PR #4406, opened 2022-09-12
change border color

VWSCoronaDashboard21:


packages/common/src/theme/colors.ts
```diff
@@ -54,6 +54,7 @@ export const colors = {
   link: '#01689b',
   shadow: '#e5e5e5',
   gray: '#808080',
+  gray3: 'CCCCCC',
```
VWSCoronaDashboard25, line 57: You forgot a `#`. And maybe we can agree to not use capital letters for readability.
So if you ask me, either `#cccccc` or `#ccc`. (I prefer the first though, for consistency.)

VWSCoronaDashboard25: Ooh look how fancy, github likes it written out as well 😁

VWSCoronaDashboard21: Thanks, I have changed this to lower case.

VWSCoronaDashboard25 (approved):

PR #4411, opened 2022-09-14
archived booster shot per age group and add autumn 2022 shot per age …

VWSCoronaDashboard21:


VWSCoronaDashboard18 (approved):
Maybe check some naming before merging.
Sometimes you use CamelCase with an '_' before a number. And in the component, you use no '-' between 'autumn' and '2022'.

PR #4413, opened 2022-09-14
Add new keys for existing schemas regarding the n…

VWSCoronaDashboard19:
# Summary
- Add new keys
- Add new archived schemas
- Remove "has_one_shot" in new schemas


packages/app/schema/vr_collection/vaccine_coverage_per_age_group.json
```diff
@@ -25,10 +25,10 @@
       "type": "string",
       "enum": ["12+", "18+"]
```
VWSCoronaDashboard24, line 26: Shouldn't this also have `60+`?

packages/app/schema/gm_collection/vaccine_coverage_per_age_group.json
```diff
@@ -25,10 +25,10 @@
       "type": "string",
       "enum": ["12+", "18+"]
```
VWSCoronaDashboard24, line 26: Shouldn't this also have `60+`?

packages/app/schema/vr_collection/vaccine_coverage_per_age_group.json
VWSCoronaDashboard19: Correct, will add that

PR #4423, opened 2022-09-23
COR-1036 chore(deps): update uvu and ts-node

VWSCoronaDashboard24:
Update `c8`, `uvu`, `ts-node` and take all files into account when running `yarn test:coverage`


VWSCoronaDashboard26 (approved):
It's quite the shame that our testing tools now give an actually correct representation of the code coverage because they are in fact much lower than previously presented. Good thing to take into account for future work.

PR #4426, opened 2022-09-26
Feature/cor 1061 choropleth changes for autumn 2022 vaccine

VWSCoronaDashboard18:
Updated the NL choropleth to the new vaccine campaign.
- Changed the tooltip
- Changed the vaccine campaign dropdown
- Changed the age group accordingly.
- removed the same choropleth from VR and GM

### DIFF
--- a/packages/app/src/domain/vaccine/components/age-group-select.tsx
+++ b/packages/app/src/domain/vaccine/components/age-group-select.tsx
@@ -8,7 +8,7 @@ import { useIntl } from '~/intl';
 import { replaceVariablesInText } from '~/utils/replace-variables-in-text';
 import { parseBirthyearRange } from '../logic/parse-birthyear-range';
 
-export type AgeGroup = '12+' | '18+';
+export type AgeGroup = '12+' | '18+' | '60+';
 
 const AGE_GROUPS = [
   {
@@ -18,6 +18,10 @@ const AGE_GROUPS = [
   {
     ageGroup: '18+',
     birthyearRange: '-2003',
+  },
+  {
+    ageGroup: '60+',
+    birthyearRange: '-1961',
   },
 ] as const;
 
--- a/packages/app/src/domain/vaccine/components/age-group-select.tsx
+++ b/packages/app/src/domain/vaccine/components/age-group-select.tsx
@@ -28,10 +28,11 @@ const AGE_GROUPS = [
 type AgeGroupSelectProps = {
   onChange: (value: AgeGroup) => void;
   initialValue?: AgeGroup;
+  shownAgeGroups?: AgeGroup[];
 };
 
 export function AgeGroupSelect(props: AgeGroupSelectProps) {
-  const { onChange, initialValue = '18+' } = props;
+  const { onChange, initialValue = '18+', shownAgeGroups } = props;
 
   const { commonTexts } = useIntl();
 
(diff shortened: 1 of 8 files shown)


packages/app/src/domain/vaccine/components/age-group-select.tsx
```diff
@@ -19,42 +19,52 @@ const AGE_GROUPS = [
...
+  const { onChange, initialValue = '18+', shownAgeGroups } = props;
 
   const { commonTexts } = useIntl();
 
   const options: Option<AgeGroup>[] = useMemo(
     () =>
       AGE_GROUPS.map((el) => {
         const birthyearRange = parseBirthyearRange(el.birthyearRange);
-
         if (isPresent(birthyearRange)) {
```
VWSCoronaDashboard24, line 43: Maybe convert this into an early return for improved readability?

```
if (!isPresent(birthyearRange)) {
  return;
}
```

packages/app/src/domain/vaccine/vaccine-coverage-choropleth-per-gm.tsx
```diff
@@ -142,17 +158,15 @@ export function VaccineCoverageChoroplethPerGm({
...
           formatTooltip={(context) => (
             <ChoroplethTooltip
               data={context}
-              percentageProps={[
-                'booster_shot_percentage',
-                'fully_vaccinated_percentage',
-              ]}
+              mapData={data.gm.filter(
+                (singleGM) => singleGM.gmcode === context.code
+              )}
```
VWSCoronaDashboard24, line 167: It feels like the performance is degraded with this change. Perhaps we can make a lookup table instead? So you can access the desired data directly by using the `context.code`. We could use `data.gm` to create an object with the `context.code` as key. So a lookup could look like this?

```
mapData={dataGmLookup[context.code]}
```

packages/app/src/domain/vaccine/vaccine-coverage-choropleth-per-gm.tsx
```diff
@@ -170,17 +184,15 @@ export function VaccineCoverageChoroplethPerGm({
...
           formatTooltip={(context) => (
             <ChoroplethTooltip
               data={context}
-              percentageProps={[
-                'booster_shot_percentage',
-                'fully_vaccinated_percentage',
-              ]}
+              mapData={data.vr.filter(
+                (singleVR) => singleVR.vrcode === context.code
+              )}
```
VWSCoronaDashboard24, line 193: Same as previous comment but then for `vr` offc

packages/app/src/domain/vaccine/components/age-group-select.tsx
VWSCoronaDashboard18: It's needed for typescript:
Screenshot 2022-09-26 at 11 52 18
Screenshot 2022-09-26 at 11 53 57

packages/app/src/domain/vaccine/vaccine-coverage-choropleth-per-gm.tsx
VWSCoronaDashboard18: It's more complicated looking into it after this is merged. If we find an optimization we can create e new Pull Request.

packages/app/src/domain/vaccine/vaccine-coverage-choropleth-per-gm.tsx
VWSCoronaDashboard18: Same as the previous reply

PR #4431, opened 2022-09-27
fix: add support for BA_4_6

VWSCoronaDashboard24:


packages/common/src/types/data.ts
```diff
@@ -1031,6 +1033,7 @@ export interface NlVariantsVariantValue {
 export type TopicalIcon =
   | "AfstandSporten"
   | "AlcoholVerkoop"
+  | "Archive"
```
VWSCoronaDashboard18, line 1036: Why is archived added?

VWSCoronaDashboard18 (approved):

packages/common/src/types/data.ts
VWSCoronaDashboard18: I understand, just wanted to check if that was the case. Because I wasn't sure it was changed inside the generated file.

VWSCoronaDashboard18 (approved):

PR #4439, opened 2022-10-03
COR-1037 Sanity language toggle

VWSCoronaDashboard24:
Fixes the NL/EN toggle in Sanity Debug UI. It now shows the EN texts correctly when enabled.

Before:
Screenshot 2022-10-03 at 15 22 42

After:
Screenshot 2022-10-03 at 13 59 15

### DIFF
--- a/packages/app/src/utils/cms/use-dynamic-lokalize-texts.ts
+++ b/packages/app/src/utils/cms/use-dynamic-lokalize-texts.ts
@@ -26,8 +26,8 @@ export const useDynamicLokalizeTexts = <T extends Record<string, unknown>>(
         .then((texts) => mapSiteTextValuesToKeys(texts))
         .then((texts) => setTexts(selector(texts)));
     }
-    // when locale is set to 'en' with the debug toggle we fetch the en texts and show them instead of the default nl texts
-    else if (locale === 'en') {
+    // when selected locale is not the default we fetch the texts again and show those instead
+    else if (locale !== 'nl') {
       fetchLokalizeTexts(environment)
         .catch(handleSanityError)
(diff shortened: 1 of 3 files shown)


VWSCoronaDashboard25 (approved):
Looks good and simpel. Nicely done!

packages/app/src/utils/cms/use-dynamic-lokalize-texts.ts
```diff
@@ -15,20 +16,35 @@ export const useDynamicLokalizeTexts = <T extends Record<string, unknown>>(
...
-          );
-        })
+      fetchLokalizeTexts(environment)
+        .catch(handleSanityError)
         .then((texts) => texts[locale] as unknown as SiteText)
         .then((texts) => mapSiteTextValuesToKeys(texts))
         .then((texts) => setTexts(selector(texts)));
+    }
+    // when locale is set to 'en' with the debug toggle we fetch the en texts and show them instead of the default nl texts
+    else if (locale === 'en') {
```
VWSCoronaDashboard26, line 30: We now only facilitate only two locales on the dashboard (and I reckon we will not introduce a new locale at any time soon), but would we want to change this approach ever so slightly to also facilitate potentially newly added locales directly out of the box? If so, would it make sense to change this to `else if (locale !== 'nl') { ... }`?

VWSCoronaDashboard26 (approved):
Approved but with a minor comment that might not need revision.

packages/app/src/utils/cms/use-dynamic-lokalize-texts.ts
VWSCoronaDashboard24: Good idea, I pushed some changed :-)

PR #4441, opened 2022-10-04
COR-1066 Rewrite absolute urls on topical page

VWSCoronaDashboard24:
Removes `https://coronadashboard.rijksoverheid.nl` from links in TOPICAL.json to be able to show the correct link icon. This also fixes the other links on the topical page to always point to production (even if on dev or staging).

Before:
Screenshot 2022-10-04 at 14 59 35

After:
Screenshot 2022-10-04 at 14 59 17


VWSCoronaDashboard24:
Lets tackle this problem at the source

PR #4444, opened 2022-10-05
feature/COR-1105-severity-indicator-component

VWSCoronaDashboard26:
## Summary

* Introduced new severity indicator component (and associated subcomponents) to be used as a 'thermometer'.
* Something to improve/investigate if possible would be dynamic `SeverityLevel` and `SEVERITY_STEPS` definitions, so that a SeverityIndicator can be used in different contexts.

COR-1105_mobile
COR-1105_desktop

## Notes

* Final texts and levels should come from Sanity and are dependant on the position/context of the severity indicator, hence they are not included in this pull request. A separate pull request should cover the implementation of the component.


packages/app/src/components/severity-indicator/components/severity-indicator-step.tsx
```diff
@@ -0,0 +1,42 @@
...
+}: SeverityIndicatorStepProps) => {
+  return (
+    <Box
+      backgroundColor={step.level <= level ? color : colors.gray3}
+      borderRadius={4}
+      height={step.height}
+      position="relative"
+      width={`${100 / SEVERITY_STEPS.length}%`}
+    >
+      {level === step.level && (
```
VWSCoronaDashboard18, line 25: Why use this logic instead of keeping it a 'dumb' component? Maybe give it an active state via props. And not letting this component decide by itself if it is active or not. And you can even extract the whole arrow to a separate const and have that rendered conditionally based on the active state.

packages/app/src/components/severity-indicator/components/severity-indicator-step.tsx
```diff
@@ -0,0 +1,42 @@
...
+        <Box
+          left="50%"
+          position="absolute"
+          top="100%"
+          transform="translate3d(-50%, 0, 0)"
```
VWSCoronaDashboard18, line 30: Why use 3d transformations? Are we going to use a transition or animation later on? Or does it have a best practice? translateX(-50%) is a bit more descriptive in my opinion.

packages/app/src/components/severity-indicator/severity-indicator.tsx
```diff
@@ -0,0 +1,86 @@
...
+        >
+          <SeverityIndicatorLabel color={color} label={label} level={level} />
+        </Box>
+
+        <Box
+          alignItems="flex-end"
+          display="flex"
+          css={css({ gap: `0 ${space[1]}` })}
+          height={space[4]}
+          justifyContent="space-between"
```
VWSCoronaDashboard18, line 63: And a gap and space-between seem a bit duplicate. Because the <SeverityIndicatorStep> has a width set relative and not absolute/fixed. (if I'm correct 25%. (100/steps)) So I can imagine the Gap will take care of the proper spacing?

packages/app/src/components/severity-indicator/severity-indicator.tsx
```diff
@@ -0,0 +1,86 @@
...
+      alignItems="flex-start"
+      border={`1px solid ${colors.gray3}`}
+      borderLeft={`${space[2]} solid ${color}`}
+      css={css({ gap: `0 ${space[4]}` })}
+      display="flex"
+      flexDirection="row"
+      flexWrap="wrap"
+      justifyContent="space-between"
+      p={4}
+    >
```
VWSCoronaDashboard24, line 40: Might be unrelevant for now since we're doing this in a lot of places. But would it be nice to make a dedicated styled component instead of using inline styles? (Like <SeverityIndicatorLevel />)

packages/app/src/components/severity-indicator/types.ts
```diff
@@ -0,0 +1,34 @@
...
+};
+
+export const SEVERITY_COLORS: string[] = [
+  colors.blue8,
+  colors.scale.magenta[1],
+  colors.scale.magenta[2],
+  colors.scale.magenta[3],
+];
+
+export const SEVERITY_STEPS: SeverityStep[] = [
```
VWSCoronaDashboard18, line 17: Can this change? for example is the height/step gets distributed between 50% and 100%. That logic can be implemented right? why hard coded it like here? (I can understand the 4 or 16 part) but that can be solved rather easy the use of a modulus (% 4) with the increments of the step level. 

I don't really prefer hardcoded values instead of hardcoded logic. If it occurs we want a 5th level we need to change this. But when using coded logic that won't be necessary. When the logic changes from 4 to 4x4 (the 16 version), you would only have to change the logic implementation.

packages/app/src/components/severity-indicator/severity-indicator.tsx
```diff
@@ -0,0 +1,86 @@
...
+      <Box
+        flexGrow={1}
+        minWidth="200px"
```
VWSCoronaDashboard24, line 43: I see a lot of `200px` in this file. Would it make sense to put it in a variable and use that instead? It should help with maintaining the code later on.

packages/app/src/components/severity-indicator/types.ts
VWSCoronaDashboard24, line 17: Would it be nice to add the color in this step as well? Then you have one definition of all the config you need for a step?

VWSCoronaDashboard24 (approved):
Code looks really neat! Added a couple of suggestions.

packages/app/src/components/severity-indicator/components/severity-indicator-label.tsx
```diff
@@ -0,0 +1,38 @@
...
+}
+
+export const SeverityIndicatorLabel = ({
+  color,
+  label,
+  level,
+}: SeverityIndicatorLabelProps) => {
+  return (
+    <>
+      <SeverityIndicatorLevel color={color}>{level}</SeverityIndicatorLevel>
```
VWSCoronaDashboard18, line 20: It feels weird to have the prop color being used for the background color. And the CSS property 'color' uses colors.white. Maybe be a bit more descriptive about this prop and value

VWSCoronaDashboard18 (approved):
I approve because my comments were not on code that did not work but on improvements. Nevertheless, you could still have a look at some of my comments.

packages/app/src/components/severity-indicator/severity-indicator.tsx
VWSCoronaDashboard26: Good call! Adjusted at line 27 of `packages/app/src/components/severity-indicator-tile/severity-indicator-tile.tsx`.

packages/app/src/components/severity-indicator/components/severity-indicator-label.tsx
```diff
@@ -0,0 +1,38 @@
...
+const SeverityIndicatorLevel = styled.div`
+  align-items: center;
+  background-color: ${({ color }: { color: string }) => color};
+  border-radius: 50%;
+  color: ${colors.white};
+  display: flex;
+  font-weight: ${fontWeights.heavy};
+  height: ${space[4]};
+  justify-content: center;
+  margin-right: ${space[2]};
```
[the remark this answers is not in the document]
VWSCoronaDashboard26: Opted to indeed use `gap` as this was a much better alternative! `margin-right` would not be applied to a 'last div' as this only covers the level (coloured circle with level value inside it).

packages/app/src/components/severity-indicator/components/severity-indicator-label.tsx
```diff
@@ -0,0 +1,38 @@
...
+export const SeverityIndicatorLabel = ({
+  color,
+  label,
+  level,
+}: SeverityIndicatorLabelProps) => {
+  return (
+    <>
+      <SeverityIndicatorLevel color={color}>{level}</SeverityIndicatorLevel>
+
+      <Text variant="h3">{label}</Text>
```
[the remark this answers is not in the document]
VWSCoronaDashboard26: Converted the component to a `figure` with `figcaption`.

packages/app/src/components/severity-indicator/components/severity-indicator-label.tsx
VWSCoronaDashboard26: This is all slightly different now due to some new logic being introduced. Rather than passing in a color, I simply pass in the level which gets used to dynamically get the associated color.

packages/app/src/components/severity-indicator/types.ts
VWSCoronaDashboard26: Setup has slightly changed, hence I will mark this conversation as resolved.

packages/app/src/components/severity-indicator/severity-indicator.tsx
VWSCoronaDashboard26: I agree, but then I would very much like to come up with a standard approach for this:
* When do we opt for inline styles and when for a styled component?
* How do we write styled components (`styled.div'...'` (consider the `'` backticks) or `styled.div(css({ ... }))`

Resolving this conversation for now.

packages/app/src/components/severity-indicator/components/severity-indicator-label.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard18: A mistook this for the four bars. But nevertheless thanks for applying the gap to the parent.

packages/app/src/components/severity-indicator/components/severity-indicator-label.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard18: Looking nice!

PR #4447, opened 2022-10-07
Add choropleth autumn options for VR and GM

VWSCoronaDashboard19:
## Summary
- Added autumn dropdown options for VR and GM choropleth maps.
- Added extra space so the choropleth map will not hit the dropdown, only for VR and GM by prop
- Renamed regular choropleth component to a more generic name for NL

----
### **---- Before ----**
Schermafbeelding 2022-10-07 om 13 21 03

### **---- After ----** 
Schermafbeelding 2022-10-07 om 13 49 00

### DIFF
--- a/packages/app/src/components/choropleth-tile.tsx
+++ b/packages/app/src/components/choropleth-tile.tsx
@@ -23,6 +23,7 @@ type ChoroplethTileProps = DataProps & {
   };
   metadata?: MetadataProps;
   valueAnnotation?: string;
+  hasPadding?: boolean;
 } & (
     | {
         onChartRegionChange: (v: RegionControlOption) => void;
@@ -43,6 +44,7 @@ export function ChoroplethTile({
   children,
   metadata,
   valueAnnotation,
+  hasPadding,
   ...dataProps
 }: ChoroplethTileProps) {
   const breakpoints = useBreakpoints(true);
@@ -104,7 +106,11 @@ export function ChoroplethTile({
           height="100%"
           spacing={3}
         >
-          <Box height="100%" marginTop={4}>
+          <Box
+            height="100%"
+            mt={4}
+            pl={hasPadding && breakpoints.lg ? 4 : undefined}
+          >
             <ErrorBoundary>{children}</ErrorBoundary>
           </Box>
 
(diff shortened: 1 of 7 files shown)


packages/app/src/pages/veiligheidsregio/[code]/vaccinaties.tsx
```diff
@@ -137,16 +144,31 @@ export const VaccinationsVrPage = (
...
+   */
+  const setSelectedCoverageKindAndAge = (
+    coverageKind: CoverageKindProperty
+  ) => {
+    if (coverageKind === selectedCoverageKind) return;
+    if (selectedAgeGroup !== '12+') {
+      setSelectedAgeGroup(selectedAgeGroup === '18+' ? '60+' : '18+');
+    }
+    setSelectedCoverageKind(coverageKind);
+  };
```
Comment (VWSCoronaDashboard25) [inline packages/app/src/pages/veiligheidsregio/[code]/vaccinaties.tsx:170]: This logic is also featured in the other versions of this page. I was hoping we could create a helper function in the `domain/vaccine` part of the code and reuse that on all pages. But after thinking on it for a while, with the different page specific setters going on, I don't see how to do that in a way without overcomplicating everything.

Comment (VWSCoronaDashboard19) [reply packages/app/src/pages/veiligheidsregio/[code]/vaccinaties.tsx:170]: That was exactly what I was thinking, just didn't quite work out for me.

packages/app/src/components/choropleth-tile.tsx
```diff
@@ -23,6 +23,7 @@ type ChoroplethTileProps = DataProps & {
   };
   metadata?: MetadataProps;
   valueAnnotation?: string;
+  hasPadding?: boolean;
```
VWSCoronaDashboard25, line 26: The padding is quite specific I think, so I'd hope for a name that better describes where the padding is added. 

But instead of adding padding to the _wrapper_ of the DynamicChoropleth (the `children`) so the dropdown doesn't overlap, would it make sense to pass a smaller DynamicChoropleth instead?

I'm not sure why only the gemeente and veiligheidsregio pages need the padding, though, so I can't really argue about what's best.

VWSCoronaDashboard19: Even when passing a width to the Choropleth to make it smaller, we will still need padding on the left side (or add more css/props to not align it to the left side of the grid). By adding a padding, it resizes and adds space on the left size.

The reason why only GM and VR needs this is because there is already enough space on NL page.

PR #4448, opened 2022-10-10
COR-483 [2/2] Format all files

VWSCoronaDashboard24:
- Set `printWidth` to 500
- Ran `yarn format` from the root of the project to format all files following the Prettier config.


VWSCoronaDashboard18 (approved):
LG2M

VWSCoronaDashboard25:
Although the idea is good to set a high `printWidth` in the file `.prettierrc.js`, this many changes are impossible to review properly even though all changes are generated automatically.
In this PR, several bad practices came to light that we'd like to address properly. So I am closing this PR and instead will open a new one frequently to increase the `printWidth` with small amounts over time so we can address all issues that arise individually.

PR #4451, opened 2022-10-11
SVG icons states for screenreader

VWSCoronaDashboard27:
## Summary

- added new text for icons to the Sanity
- hided unnecessary icons from screenreader


VWSCoronaDashboard19 (approved):

VWSCoronaDashboard25:
There are several `aria-hidden="true"` and some `aria-hidden='true'` added, so with double quotes and with single quotes.
Could you make them all `aria-hidden="true"`?

Basically we want strings in JavaScript to be single quotes. However in the HTML-like part of JSX we use double quotes, because in plain HTML you _must_ use double quotes.

VWSCoronaDashboard18 (changes requested):
I'm missing the aria-label at the trend icons on the topical theme tiles.

Are the external links and warning icons the only ones missing context for the visually impaired?

packages/app/src/components/difference-indicator/tile-difference.tsx
```diff
@@ -13,45 +13,43 @@ export function TileDifference({
...
   showOldDateUnix?: boolean;
   isAmount: boolean;
+  ariaLabel?: string;
 }) {
   const { commonTexts, formatNumber, formatDateFromSeconds } = useIntl();
   const text = commonTexts.toe_en_afname;
   const { difference } = value;
+  const TrendLabelUp = ariaLabel || commonTexts.accessibility.visual_context_labels.up_trend_label;
+  const TrendLabelDown = ariaLabel || commonTexts.accessibility.visual_context_labels.down_trend_label;
+  const TrendLabelNeutral = ariaLabel || commonTexts.accessibility.visual_context_labels.neutral_trend_label;
```
VWSCoronaDashboard25, line 30: I see this returning in three components now:
- `components/difference-indicator/tile-difference.tsx`
- `components/behaviour-trend.tsx`
- `components/topical-tile.tsx`
Each uses the `<Up />`, `<Down />` or maybe `<Dot />` icon and each result in the same aria-label.

I think we should invest in a `TrendIcon` component instead, and use that in all three places.

Something like:

```typescript
import { Down, Dot, Up } from '@corona-dashboard/icons';

export enum TrendDirection {
  UP,
  DOWN,
  NEUTRAL
};

interface TrendIconProps {
  trendDirection: TrendDirection;
  ariaLabel?: string;
}

export const TrendIcon = ({ trendDirection, ariaLabel }: TrendIconProps) => {

  const TrendLabelUp = ariaLabel || commonTexts.accessibility.visual_context_labels.up_trend_label;
  const TrendLabelDown = ariaLabel || commonTexts.accessibility.visual_context_labels.down_trend_label;
  const TrendLabelNeutral = ariaLabel || commonTexts.accessibility.visual_context_labels.neutral_trend_label;

  switch(trendDirection) {
    case TrendDirection.UP: return <Up aria-label={TrendLabelUp} />;
    case TrendDirection.DOWN: return <Down aria-label={TrendLabelDown} />;
    case TrendDirection.NEUTRAL: return <Dot aria-label={TrendLabelNeutral} />;
    default:
      const exhaustiveCheck: never = trendDirection;
      throw new Error(`Unhandled TrendDirection case: ${exhaustiveCheck}`);
  }
};
```

VWSCoronaDashboard25: But we can create a separate ticket for it, off course.

packages/app/src/components/difference-indicator/tile-difference.tsx
```diff
@@ -37,19 +37,19 @@ export function TileDifference({
   if (difference > 0) {
```
VWSCoronaDashboard28, line 37: I think this part could also be optimised a little bit to get rid of some if conditions, which are essentially doing the same thing, i.e. setting `content` and `containerWithIcon`. Of course, this is not a blocking comment and merely a suggestion for optimisation. I understand if you just want to merge this PR and not implement this optimisation for now. I would suggest something like the following:

```
let content: string = text.waarde_gelijk;
let containerWithIcon: React.ReactNode = <ContainerWithIcon icon={<TrendIcon trendDirection={TrendDirection.NEUTRAL} />} color="neutral" />;

const containerWithIconMapping = [
  {
    condition: difference > 0,
    content: isAmount ? text.waarde_meer : text.waarde_hoger,
    icon: <ContainerWithIcon icon={<TrendIcon trendDirection={TrendDirection.UP} />} color="red2" />,
  },
  {
    condition: difference < 0,
    content: isAmount ? text.waarde_minder : text.waarde_lager,
    icon: <ContainerWithIcon icon={<TrendIcon trendDirection={TrendDirection.DOWN} />} color="primary" />,
  }
];

containerWithIconMapping.forEach(mapping => {
  const { condition, content: mappingContent, icon } = mapping;
  if (condition) {
    content = mappingContent;
    containerWithIcon = icon;
  }
});
```

I tested it out by converting everything to a string to see what would log to the console for `content` and `containerWithIcon` in different scenarios and it looks like it works. But again, not blocking at all.

packages/app/src/domain/topical/components/topical-tile.tsx
```diff
@@ -30,15 +31,19 @@ interface TopicalTileProps {
...
 
-export function TopicalTile({ title, tileIcon, trendIcon, dynamicDescription, kpiValue, cta, ariaLabel }: TopicalTileProps) {
-  const { formatNumber, commonTexts } = useIntl();
+export function TopicalTile({ title, tileIcon, trendIcon, dynamicDescription, kpiValue, cta }: TopicalTileProps) {
+  const { formatNumber } = useIntl();
 
   const formatedKpiValue = typeof kpiValue === 'number' ? formatNumber(kpiValue) : typeof kpiValue === 'string' ? kpiValue : false;
-  const TrendLabelUp = ariaLabel || commonTexts.accessibility.visual_context_labels.up_trend_label;
-  const TrendLabelDown = ariaLabel || commonTexts.accessibility.visual_context_labels.down_trend_label;
+  // const TrendLabelUp = ariaLabel || commonTexts.accessibility.visual_context_labels.up_trend_label;
```
VWSCoronaDashboard28, line 40: I'm not sure why the labels are removed but I would assume if they are not needed then these commented lines can also be removed?

packages/app/src/domain/variants/variants-table-tile/components/variant-difference.tsx
```diff
@@ -1,43 +1,45 @@
...
-import { Down, Dot, Up } from '@corona-dashboard/icons';
 import css from '@styled-system/css';
 import styled from 'styled-components';
 import { useIntl } from '~/intl';
 import { TableText } from '../types';
+import { TrendDirection, TrendIcon } from '~/components/trend-icon';
 
-export function VariantDifference({ value, text, ariaLabel }: { value: DifferenceDecimal; text: TableText; ariaLabel?: string }) {
-  const { formatPercentage, commonTexts } = useIntl();
+type TrendIcon = {
```
VWSCoronaDashboard28, line 8: I notice this type declaration in a number of files in this commit. If i'm not wrong, it should be possible to also define this type in the scope of `packages/app/src/components/trend-icon.tsx` and then import it where required? I'm no typescript expert so I might be wrong, but seems like it should be possible.

packages/app/src/domain/variants/variants-table-tile/components/variant-difference.tsx
```diff
@@ -1,43 +1,45 @@
...
 
   const options = {
     minimumFractionDigits: 1,
     maximumFractionDigits: 1,
   };
-  const TrendLabelUp = ariaLabel || commonTexts.accessibility.visual_context_labels.up_trend_label;
-  const TrendLabelDown = ariaLabel || commonTexts.accessibility.visual_context_labels.down_trend_label;
-  const TrendLabelNeutral = ariaLabel || commonTexts.accessibility.visual_context_labels.neutral_trend_label;
 
   if (value === undefined) {
```
VWSCoronaDashboard28, line 21: Same as my first comment, this is not blocking and merely a suggestion for optimisation. I tested it with some console logs and it looks like it works.

```
let returnValue: React.ReactNode = (
  <Difference color={colors.neutral}>
    <Dot color={colors.neutral} aria-label={TrendLabelNeutral} />
    <TrendIcon trendDirection={TrendDirection.NEUTRAL} />
    {text.verschil.gelijk}
  </Difference>
);

const renderingConditionMapping = [
  {
    condition: value === undefined,
    renderingValue: <>-</>
  },
  {
    condition: value?.difference > 0,
    renderingValue: (
      <Difference color={colors.black}>
        <Up aria-label={TrendLabelUp} />
        <TrendIcon trendDirection={TrendDirection.UP} />
        {formatPercentage(value.difference, options)} {text.verschil.meer}
      </Difference>
    )
  },
  {
    condition: value?.difference < 0,
    renderingValue: (
      <Difference color={colors.black}>
        <Down aria-label={TrendLabelDown} />
        <TrendIcon trendDirection={TrendDirection.DOWN} />
        {formatPercentage(-value.difference, options)} {text.verschil.minder}
      </Difference>
    )
  }
]

renderingConditionMapping.forEach(mapping => {
  const { condition, renderingValue } = mapping;
  if (condition) {
    returnValue = renderingValue;
  }
});

return {returnValue};
```

VWSCoronaDashboard28 (approved):

PR #4457, opened 2022-10-13
Font size of home page/ COR-1110

VWSCoronaDashboard27:
## Summary

Changed font-size for descriptions at Home Page.

### DIFF
--- a/packages/app/src/domain/topical/components/topical-theme-header.tsx
+++ b/packages/app/src/domain/topical/components/topical-theme-header.tsx
@@ -27,8 +27,9 @@ export const TopicalThemeHeader = ({
         )}
         <Heading level={2}>{title}</Heading>
       </Box>
-
+      <Box fontSize={3}>
       <Markdown content={dynamicSubtitle} />
+      </Box>
     </Box>
   );
 };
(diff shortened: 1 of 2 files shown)


VWSCoronaDashboard25 (approved):

packages/app/src/domain/topical/components/topical-theme-header.tsx
```diff
@@ -27,8 +27,9 @@ export const TopicalThemeHeader = ({
         )}
         <Heading level={2}>{title}</Heading>
       </Box>
-
+      <Box fontSize={3}>
```
VWSCoronaDashboard18, line 30: Why is this box created? Cant the font size be set on the markdown element?

VWSCoronaDashboard18 (changes requested):

packages/app/src/domain/topical/components/topical-theme-header.tsx
VWSCoronaDashboard19: It is not possible to set a `font-size` to a `Markdown` component, therefore a `Box` element an easier solution.

VWSCoronaDashboard18: I did not know about the markdown not accepting extra CSS. (Not sure why though it doesn't)
It makes sense to have a box element for the font-size instead.

VWSCoronaDashboard18 (approved):

PR #4462, opened 2022-10-19
task/COR-1151-safety-region-behaviour-page-archiving

VWSCoronaDashboard26:
## Summary

* renamed `behavior.json` to `behavior_archived_20221019.json` (including references inside the schema, schema index and associated files)
* adjusted the menu for VR layouts to archive the behaviour page
* adjusted the page category of the behaviour page

#### After
COR-1151_after

### DIFF
--- a/packages/common/src/types/data.ts
+++ b/packages/common/src/types/data.ts
@@ -605,13 +605,17 @@ export interface NlBehaviorValue {
   avoid_crowds_compliance?: number | null;
   avoid_crowds_compliance_trend: ('up' | 'down' | 'equal') | null;
   symptoms_stay_home_if_mandatory_compliance?: number | null;
-  symptoms_stay_home_if_mandatory_compliance_trend?: ('up' | 'down' | 'equal') | null;
+  symptoms_stay_home_if_mandatory_compliance_trend?:
+    | ('up' | 'down' | 'equal')
+    | null;
   symptoms_get_tested_compliance?: number | null;
   symptoms_get_tested_compliance_trend?: ('up' | 'down' | 'equal') | null;
   wear_mask_public_indoors_compliance?: number | null;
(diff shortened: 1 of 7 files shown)


packages/common/src/types/data.ts
```diff
@@ -1272,7 +1305,7 @@ export interface TopicalTheme {
 }
 export interface TopicalThemeTile {
   index: number;
-  kpiValue?: number | null | string;
+  kpiValue: number | null | string;
```
VWSCoronaDashboard19, line 1308: Why is this key required after your changes?

VWSCoronaDashboard19:
What we usually did until this day was adding the newly archived sidebar menu above the existing one, instead of under in (like in your screenshot on the left side)

packages/common/src/types/data.ts
VWSCoronaDashboard26: I am not entirely sure, but I want to make an unpopular assumption that this might have changed due to [this pull request](https://github.com/minvws/nl-covid19-data-dashboard/pull/4452). When I execute the `yarn generate-data-types` command on the `develop` branch (with the code of the aforementioned PR), I see the same results. Would this make sense? 🤔

VWSCoronaDashboard26:
> What we usually did until this day was adding the newly archived item in the sidebar menu above the existing one, instead of under it (like in your screenshot on the left side)

Fair point! I have adjust the code and the attached screenshot in the description.

VWSCoronaDashboard28 (approved):

packages/common/src/types/data.ts
VWSCoronaDashboard19: That definitely would make sense.

VWSCoronaDashboard19 (approved):

VWSCoronaDashboard18 (approved):
Looks good 2 me

PR #4464, opened 2022-10-19
COR-1135: dynamic-thermometer

VWSCoronaDashboard27:
## Summary

- Change title to a bold font;
- Moved date from tile footer below the tile title;
- Add description about increase/decrease/neutral;
- Add arrow for interpretation description;
- Add Sanity keys for description & arrow for interpretation;
- Customised icon for Corona Thermometer section;


packages/app/src/components/severity-indicator-tile/severity-indicator-tile.tsx
```diff
@@ -9,16 +9,23 @@ import { getSeverityColor } from './logic/get-severity-color';
...
 interface SeverityIndicatorTileProps {
   description: string;
   label: string;
   level: SeverityLevels;
-  footerText: string;
   title: string;
+  dates_label: string;
+  source_label: string;
+  level_description: string;
+  trend_icon: TrendIcon | null;
```
VWSCoronaDashboard26, line 25: Any particular reason that the props passed into this component follow `snake_case` rather than the more widely used `camelCase`? We traditionally use `snake_case` formats for either Sanity keys or data entries, so using it for props might cause mixups.

packages/app/src/components/severity-indicator-tile/severity-indicator-tile.tsx
```diff
@@ -43,10 +53,33 @@ export const SeverityIndicatorTile = ({ description, label, level, title, footer
...
       <Box flexGrow={1} width={`min(${SEVERITY_INDICATOR_TILE_COLUMN_MIN_WIDTH}px, 50%)`} as="figcaption">
         <Markdown content={description} />
+        <Box
+          display="flex"
+          alignItems="center"
+          mt={3}
+          css={css({
+            gap: 2,
+          })}
+        >
```
VWSCoronaDashboard26, line 63: This should all be put on all be put on a single line according to the new Prettier rules (`printWidth` of 180). Is this not forced to one line because of the `css()` method used on line 60? 🤔

packages/app/src/domain/topical/types.ts
```diff
@@ -0,0 +1,4 @@
+export type TrendIcon = {
+  direction: 'UP' | 'DOWN';
+  color: string;
+};
```
VWSCoronaDashboard26, line 4: 👏

packages/app/src/components/severity-indicator-tile/severity-indicator-tile.tsx
```diff
@@ -9,16 +9,23 @@ import { getSeverityColor } from './logic/get-severity-color';
 import { SeverityLevels } from './types';
 import { InlineText } from '../typography';
 import { SEVERITY_INDICATOR_TILE_COLUMN_MIN_WIDTH } from '~/components/severity-indicator-tile/constants';
+import { BoldText } from '~/components/typography';
+import { Down, Up } from '@corona-dashboard/icons';
+import styled from 'styled-components';
+import { TrendIcon } from '~/domain/topical/types';
```
VWSCoronaDashboard26, line 15: We have not yet discussed a convention for grouping imports yet (at least, not that I am aware of), but I would propose to group the newly added lines at slightly different places so they are grouped with their 'siblings' a bit nicer. You'd end up with the below:
```
  import { Down, Up } from '@corona-dashboard/icons';
import { space } from '~/style/theme';
import { Markdown } from '~/components/markdown';
  import { BoldText, InlineText } from '~/components/typography';
  import { TrendIcon } from '~/domain/topical/types';
import { SeverityIndicatorLabel } from './components/severity-indicator-label';
import { SeverityIndicator } from './components/severity-indicator';
  import { SEVERITY_INDICATOR_TILE_COLUMN_MIN_WIDTH } from './constants';
import { getSeverityColor } from './logic/get-severity-color';
import { SeverityLevels } from './types';
```

The adjusted lines are indented ever so slightly, grouped with similar paths and sorted alphabetically. Having a consistent order while looking at the imports of a file makes searching, by simply glancing over them, a bit easier in my experience.

packages/app/src/pages/index.tsx
```diff
@@ -80,7 +80,10 @@ const Home = (props: StaticProps<typeof getStaticProps>) => {
                 })}
                 title={textNl.thermometer.indicator.title}
                 label={textNl.thermometer.indicator.label}
-                footerText={textNl.thermometer.indicator.footerText}
+                source_label={textNl.thermometer.indicator.source_label}
+                dates_label={textNl.thermometer.indicator.dates_label}
+                level_description={textNl.thermometer.indicator.level_description}
+                trend_icon={textNl.thermometer.indicator.trend_icon}
```
VWSCoronaDashboard26, line 86: This line actually causes the GitHub compile action to fail and should probably also give you a warning in the IDE. You can solve this by passing the prop with `as TrendIcon` behind it, so it actually registers the type of both color and direction (which is literal string type rather than a string type) correctly.

VWSCoronaDashboard26 (changes requested):

packages/app/src/components/severity-indicator-tile/severity-indicator-tile.tsx
VWSCoronaDashboard19: That is because Sanity keys are being passed, therefore it needs to match that key name.

packages/app/src/pages/index.tsx
VWSCoronaDashboard19: Correct, this still causes an error, also the color/size of the icon needs to be fixed 🙂

packages/app/src/components/severity-indicator-tile/severity-indicator-tile.tsx
VWSCoronaDashboard19: Good point, we can also discuss this with the rest of the team during our Chapter/code conventions

packages/app/src/components/severity-indicator-tile/severity-indicator-tile.tsx
VWSCoronaDashboard18: The props don't need to match the case formats of the lokalize keys.

VWSCoronaDashboard19: Yeah you are right, I was wrong about this one! will change it also.

VWSCoronaDashboard26: I might have missed this convention earlier. Does this mean that React component props are following `snake_case` format when the value for that prop comes from Sanity **and** its value is a key following the `snake_case` format in Sanity, but any other 'normal' (non-Sanity or Sanity **not** following a `snake_case` format) props follow a `camelCase` format?

VWSCoronaDashboard19: Yes, regular React props should always be `camelCase` and keys in Sanity should be `snake_case`. In this example, the props should be `camelCase`

packages/app/src/components/severity-indicator-tile/severity-indicator-tile.tsx
VWSCoronaDashboard19: Unfortunately no, but it should be in the next change.

packages/app/src/components/severity-indicator-tile/severity-indicator-tile.tsx
```diff
@@ -43,10 +58,26 @@ export const SeverityIndicatorTile = ({ description, label, level, title, footer
 
       <Box flexGrow={1} width={`min(${SEVERITY_INDICATOR_TILE_COLUMN_MIN_WIDTH}px, 50%)`} as="figcaption">
         <Markdown content={description} />
+        <Box display={hasIconProps ? 'flex' : 'block'} alignItems="center" mt={3} css={css({ gap: 2 })}>
+          {hasIconProps && (
+            <TrendIconWrapper color={mapStringToColors(iconColor)}>
+              {iconDirection === ICON_DIRECTION_DOWN && <Down />}
+              {iconDirection === ICON_DIRECTION_UP && <Up />}
```
VWSCoronaDashboard26, line 65: Not blocking at all and merely a suggestion, but these two lines are also repeated in `packages/app/src/domain/topical/components/topical-tile.tsx`. Would it make sense to extract this to a separate function in which you'd pass the `direction` as a parameter which in turn returns a component (either `<Down />` or `<Up />`) based on the value of `direction` you've passed in. Seeing as we do not have any 'equal' icon, you do not necessarily have to check for both 'down' and 'up', but simply a single check for either one would suffice.

packages/app/src/domain/topical/types.ts
```diff
@@ -0,0 +1,4 @@
+export type TrendIcon = {
+  direction: 'UP' | 'DOWN';
```
VWSCoronaDashboard26, line 2: We now reference both `ICON_DIRECTION_DOWN` and `ICON_DIRECTION_UP` in `packages/app/src/domain/topical/components/topical-tile.tsx` and `packages/app/src/components/severity-indicator-tile/severity-indicator-tile.tsx` which I think is a good thing (as their values are consistent throughout these different components). Could we also apply these constants here rather than the string equivalent (`'UP'` or `'DOWN'`) to make sure we only have to touch a single file when applying changes?

VWSCoronaDashboard26 (approved):
Approved because non of my (new) pending comments are blocking, but you can pick these up if you'd like.

PR #4466, opened 2022-10-20
task/COR-1153-behaviour-page-indicator

VWSCoronaDashboard26:
## Summary

* updated `behavior.json`, `behavior_annotations.json` and `behavior_per_age_group.json` schemas for NL to add new keys
    * `posttest_isolation_compliance`
    * `posttest_isolation_compliance_trend`
    * `posttest_isolation_support`
    * `posttest_isolation_support_trend`
* updated icons and ran yarn build:icons script for automated icon component generation;


packages/app/src/domain/behavior/components/behavior-icon.tsx
```diff
@@ -27,6 +28,7 @@ const icons: Record<BehaviorIdentifier, typeof Avondklok> = {
   max_visitors: MaxVisitors,
   ventilate_home: FrisseLucht,
   selftest_visit: BasisregelsTesten,
+  posttest_isolation: BasisregelsBlijfThuis,
```
VWSCoronaDashboard28, line 31: Didn't you change this to be the same icon as `symptoms_stay_home_if_mandatory`? I think there was a size difference?

VWSCoronaDashboard28 (approved):
I left a small comment but pre-approving.

packages/app/src/domain/behavior/components/behavior-icon.tsx
VWSCoronaDashboard26: I indeed originally opted to use the same icon as for `symptoms_get_tested`, yet I worked around this by adding another icon. I personally think that the name of the component better covers its usage in combination with `posttest_isolation`.

VWSCoronaDashboard19 (approved):

packages/icons/src/svg/basisregels_blijf-thuis.svg
```diff
@@ -0,0 +1 @@
+<svg id="Layer_1" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 56 56" fill="currentColor"><path d="M51.54,26.35,29.19,5A1.57,1.57,0,0,0,28,4.42a1.54,1.54,0,0,0-1.19.6L4.45,26.35a3.28,3.28,0,0,0-.89,2.1V30c0,.9.89,1.5,2.09,1.5h3V51a.64.64,0,0,0,.59.6H47.08a.65.65,0,0,0,.59-.61V31.75h3A1.93,1.93,0,0,0,52.44,30v-1.5A2.72,2.72,0,0,0,51.54,26.35ZM32,36.44a.68.68,0,0,1-.51.78,2.48,2.48,0,0,1-1,.13s.11.06.32.42.53,1.25.12,1.46-.55.1-.77.54c-.39.79-.91,1.61-2.11,1.07s-4.57-2.35-4.57-2.35a.23.23,0,0,0,0,.3l2.2,2.4V43h-11A73.51,73.51,0,0,0,18,36c.59-2.65-.89-6.78-1-9.44C17,24.25,17.75,20,22.41,19c4.41-1,9.56.7,10.71,2.08a2.92,2.92,0,0,1-.3,4.15,18.55,18.55,0,0,1,0,3.58c0,.32-.36.95-.39,1.34l1.67,3.64a.87.87,0,0,1-.57,1.08L32,35.14A1.73,1.73,0,0,0,32,36.44Zm.66,2.21a1,1,0,1,1,.95,1A1,1,0,0,1,32.63,38.65Zm1.09,3.29a.95.95,0,1,1,.94,1A.95.95,0,0,1,33.72,41.94Zm2,4.24a1,1,0,1,1,1-1A1,1,0,0,1,35.75,46.18Zm.68-8.62a.95.95,0,1,1,1,.95A1,1,0,0,1,36.43,37.56Zm2,4.33a1,1,0,1,1,1-1A.95.95,0,0,1,38.43,41.89Zm2.75-4.47a1,1,0,1,1,1-1A1,1,0,0,1,41.18,37.42Z"/></svg>
```
VWSCoronaDashboard25, line 1: According to the design team, this should've been an already existing icon. So I'd rather not have a duplicate icon with a different name.

If this file is here to stay:
- Please check if this needs an `id`. Depending on that, either remove it or rename it to something unique. On a page with multiple icons, generic generated ids can create issues.
- The `data-name` can probably be removed anyway.
- Please spread the nested tags over multiple lines as you would do with an html file (and maybe add a EOF new line for my pleasure? 😉)

packages/common/src/types/data.ts
```diff
@@ -1272,7 +1280,7 @@ export interface TopicalTheme {
 }
 export interface TopicalThemeTile {
   index: number;
-  kpiValue?: number | null | string;
+  kpiValue: number | null | string;
```
VWSCoronaDashboard25, line 1283: I know this is generated (I encountered it my self as well), but maybe check with the original author if this is okay.
Because I think this was maybe done manually to prevent a bug.

VWSCoronaDashboard25:
Is `posttest` short for positive test? If so, please don't abbreviate.

packages/icons/src/svg/basisregels_blijf-thuis.svg
VWSCoronaDashboard25: Ah, I see the other one is part of the specific restrictions icons and is still hardcoded to be 36px by 36px.
Still weird for me that we have icons in separate places.
I also noticed that the other icons in this directory are single line coded 😥. So this is just following suit, that's okay. The `Layer_1` stuff is not okay though.

Eventually I'd hope we can change the icon names to what they show instead of what we're using them for (and have them in one place), that would make much more sense. But I guess it's better to overhaul all of them in one go, especially because this icon has a brother with the same name in the restrictions section.

PR #4468, opened 2022-10-24
Feature/cor 1159 collapsible thermometer description

VWSCoronaDashboard18:
Created the thermometer collapsible.
This is a new element and looks like this:

**COLLAPSED**
Screenshot 2022-10-24 at 09 59 19

**EXPANDED**
Screenshot 2022-10-24 at 09 59 15

### DIFF
--- a/packages/app/src/components/collapsible/collapsible-section.tsx
+++ b/packages/app/src/components/collapsible/collapsible-section.tsx
@@ -6,20 +6,18 @@ import { isAtBottomOfPage } from '~/utils/is-at-bottom-of-page';
 import { isElementAtTopOfViewport } from '~/utils/is-element-at-top-of-viewport';
 import { useCollapsible } from '~/utils/use-collapsible';
 import { Anchor } from '../typography';
+import { colors } from '@corona-dashboard/common';
 
 interface CollapsibleSectionProps extends BoxProps {
   summary: string;
   children: ReactNode;
   id?: string;
   hideBorder?: boolean;
+  textColorOverride?: string;
+  borderColorOverride?: string;
 }
 
-export const CollapsibleSection = ({
-  summary,
-  children,
-  id,
-  hideBorder,
-}: CollapsibleSectionProps) => {
+export const CollapsibleSection = ({ summary, children, id, hideBorder, textColorOverride = colors.blue8, borderColorOverride = colors.gray2 }: CollapsibleSectionProps) => {
   const section = useRef<HTMLElement>(null);
 
   const collapsible = useCollapsible();
@@ -53,24 +51,13 @@ export const CollapsibleSection = ({
   }, [toggle, id]);
 
   return (
-    <Box
-      as="section"
-      borderTop={hideBorder ? undefined : '1px solid'}
-      borderTopColor={hideBorder ? undefined : 'gray2'}
-      id={id}
-      ref={section}
-    >
+    <Box as="section" borderTop={hideBorder ? undefined : '1px solid'} borderTopColor={hideBorder ? undefined : borderColorOverride} id={id} ref={section}>
       {collapsible.button(
-        <Summary>
+        <Summary textColor={textColorOverride}>
           <Box width="100%">
             {summary}
             {id && (
-              <StyledAnchor
-                aria-hidden="true"
-                tabIndex={-1}
-                onClick={(e) => e.stopPropagation()}
-                href={`#${id}`}
-              >
+              <StyledAnchor aria-hidden="true" tabIndex={-1} onClick={(e) => e.stopPropagation()} href={`#${id}`}>
                 #
               </StyledAnchor>
             )}
@@ -86,7 +73,7 @@ export const CollapsibleSection = ({
 
 const StyledAnchor = styled(Anchor)(
   css({
-    color: 'gray2',
+    color: colors.gray2,
     px: 3,
     py: 1,
     width: 0,
@@ -94,12 +81,15 @@ const StyledAnchor = styled(Anchor)(
     position: 'absolute',
     right: '100%',
     '&:hover, &:focus': {
-      color: 'blue',
+      color: colors.blue1,
     },
   })
 );
 
-const Summary = styled.button(
+interface SummaryPropsType {
+  textColor: string;
+}
+const Summary = styled.button((summaryProps: SummaryPropsType) =>
   css({
     display: 'flex',
     alignItems: 'flex-start',
@@ -110,7 +100,7 @@ const Summary = styled.button(
     p: 3,
     bg: 'transparent',
     border: 'none',
-    color: 'blue8',
+    color: summaryProps.textColor,
     fontFamily: 'body',
     fontWeight: 'bold',
     fontSize: '1.25rem',
@@ -121,7 +111,7 @@ const Summary = styled.button(
     '&:focus': {
       outlineWidth: '1px',
       outlineStyle: 'dashed',
-      outlineColor: 'blue8',
+      outlineColor: colors.blue8,
     },
 
     [StyledAnchor]: { opacity: 0 },
--- a/packages/app/src/components/collapsible/collapsible-section.tsx
+++ b/packages/app/src/components/collapsible/collapsible-section.tsx
@@ -86,7 +86,7 @@ const StyledAnchor = styled(Anchor)(
   })
 );
 
-interface SummaryPropsType {
+interface SummaryProps {
   textColor: string;
 }
 const Summary = styled.button((summaryProps: SummaryPropsType) =>
--- a/packages/app/src/components/collapsible/collapsible-section.tsx
+++ b/packages/app/src/components/collapsible/collapsible-section.tsx
@@ -13,11 +13,11 @@ interface CollapsibleSectionProps extends BoxProps {
   children: ReactNode;
   id?: string;
   hideBorder?: boolean;
-  textColorOverride?: string;
-  borderColorOverride?: string;
+  textColor?: string;
+  borderColor?: string;
 }
 
-export const CollapsibleSection = ({ summary, children, id, hideBorder, textColorOverride = colors.blue8, borderColorOverride = colors.gray2 }: CollapsibleSectionProps) => {
+export const CollapsibleSection = ({ summary, children, id, hideBorder, textColor = colors.blue8, borderColor = colors.gray2 }: CollapsibleSectionProps) => {
   const section = useRef<HTMLElement>(null);
 
   const collapsible = useCollapsible();
@@ -51,9 +51,9 @@ export const CollapsibleSection = ({ summary, children, id, hideBorder, textColo
   }, [toggle, id]);
 
   return (
-    <Box as="section" borderTop={hideBorder ? undefined : '1px solid'} borderTopColor={hideBorder ? undefined : borderColorOverride} id={id} ref={section}>
+    <Box as="section" borderTop={hideBorder ? undefined : '1px solid'} borderTopColor={hideBorder ? undefined : borderColor} id={id} ref={section}>
       {collapsible.button(
-        <Summary textColor={textColorOverride}>
+        <Summary textColor={textColor}>
           <Box width="100%">
             {summary}
             {id && (
@@ -89,15 +89,15 @@ const StyledAnchor = styled(Anchor)(
 interface SummaryProps {
   textColor: string;
 }
-const Summary = styled.button((summaryProps: SummaryPropsType) =>
+const Summary = styled.button((summaryProps: SummaryProps) =>
   css({
     display: 'flex',
     alignItems: 'flex-start',
     justifyContent: 'space-between',
     overflow: 'visible',
     width: '100%',
-    m: 0,
-    p: 3,
+    margin: 0,
+    padding: 3,
     bg: 'transparent',
     border: 'none',
     color: summaryProps.textColor,
--- a/packages/app/src/components/collapsible/collapsible-section.tsx
+++ b/packages/app/src/components/collapsible/collapsible-section.tsx
@@ -57,7 +57,7 @@ export const CollapsibleSection = ({ summary, children, id, hideBorder, textColo
           <Box width="100%">
             {summary}
             {id && (
-              <StyledAnchor aria-hidden="true" tabIndex={-1} onClick={(e) => e.stopPropagation()} href={`#${id}`}>
+              <StyledAnchor aria-hidden="true" tabIndex={-1} onClick={(event) => event.stopPropagation()} href={`#${id}`}>
                 #
               </StyledAnchor>
             )}
--- a/packages/app/src/components/severity-indicator-tile/components/severity-indicator-label.tsx
+++ b/packages/app/src/components/severity-indicator-tile/components/severity-indicator-label.tsx
@@ -12,18 +12,9 @@ interface SeverityIndicatorLabelProps {
   level: SeverityLevel;
 }
 
-export const SeverityIndicatorLabel = ({
-  label,
-  level,
-}: SeverityIndicatorLabelProps) => {
+export const SeverityIndicatorLabel = ({ label, level }: SeverityIndicatorLabelProps) => {
   return (
-    <Box
-      alignItems="center"
-      display="flex"
-      justifyContent="flex-start"
-      my={space[3]}
-      css={css({ gap: `0 ${space[2]}` })}
-    >
+    <Box alignItems="center" display="flex" justifyContent="flex-start" my={space[3]} css={css({ gap: `0 ${space[2]}` })}>
       <SeverityIndicatorLevel level={level}>{level}</SeverityIndicatorLevel>
 
       <Text variant="h3">{label}</Text>
@@ -31,10 +22,9 @@ export const SeverityIndicatorLabel = ({
   );
 };
 
-const SeverityIndicatorLevel = styled.div`
+export const SeverityIndicatorLevel = styled.div`
   align-items: center;
-  background-color: ${({ level }: { level: SeverityLevel }) =>
-    getSeverityColor(level as SeverityLevels)};
+  background-color: ${({ level }: { level: SeverityLevel }) => getSeverityColor(level as SeverityLevels)};
   border-radius: 50%;
   color: ${colors.white};
   display: flex;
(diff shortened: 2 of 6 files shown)


packages/app/src/components/collapsible/collapsible-section.tsx
```diff
@@ -110,7 +100,7 @@ const Summary = styled.button(
     p: 3,
     bg: 'transparent',
     border: 'none',
-    color: 'blue8',
+    color: summaryProps.textColor,
```
VWSCoronaDashboard19, line 103: What happens when the prop `textColor` is not being passed? because `textColorOverride` is optional.

VWSCoronaDashboard19: Also shouldn't this be conditional, when the prop is not being passed, then color should fallback to `blue8`?

packages/app/src/components/collapsible/collapsible-section.tsx
```diff
@@ -6,20 +6,18 @@ import { isAtBottomOfPage } from '~/utils/is-at-bottom-of-page';
...
 import { useCollapsible } from '~/utils/use-collapsible';
 import { Anchor } from '../typography';
+import { colors } from '@corona-dashboard/common';
 
 interface CollapsibleSectionProps extends BoxProps {
   summary: string;
   children: ReactNode;
   id?: string;
   hideBorder?: boolean;
+  textColorOverride?: string;
```
VWSCoronaDashboard19, line 16: Why did you call these two override, instead of just regular `textColor` and `borderColor`, setting a default color for this component that is being used on other pages, then passing custom optional colors via these props.

packages/app/src/components/severity-indicator-tile/components/severity-indicator-label.tsx
```diff
@@ -12,29 +12,19 @@ interface SeverityIndicatorLabelProps {
...
-  label,
-  level,
-}: SeverityIndicatorLabelProps) => {
+export const SeverityIndicatorLabel = ({ label, level }: SeverityIndicatorLabelProps) => {
   return (
-    <Box
-      alignItems="center"
-      display="flex"
-      justifyContent="flex-start"
-      my={space[3]}
```
VWSCoronaDashboard19, line 24: Not blocking but just a suggestion, passing a `{3}` instead of `space[3]` would also suffice.

packages/app/src/domain/topical/components/indicator-level-description.tsx
```diff
@@ -0,0 +1,35 @@
...
+    <Box
+      display="grid"
+      gridTemplateRows="auto"
+      gridTemplateColumns={`${space[4]} auto`}
+      alignItems="center"
+      mb={4}
+      css={css({ columnGap: 3, rowGap: asResponsiveArray({ _: 3, sm: 1 }) })}
+    >
+      <SeverityIndicatorLevel level={level}>{level}</SeverityIndicatorLevel>
+      <Text fontWeight="bold">{label}</Text>
```
VWSCoronaDashboard19, line 27: Please use the `<BoldText>` instead of making the text bold via CSS, so the `<strong>` tag will be used that can be useful for screen readers.

packages/app/src/pages/index.tsx
```diff
@@ -82,6 +93,17 @@ const Home = (props: StaticProps<typeof getStaticProps>) => {
                 label={textNl.thermometer.indicator.label}
                 footerText={textNl.thermometer.indicator.footerText}
               />
+              <Box my={{ _: 3, md: 4 }} borderBottom={'1px solid'} borderBottomColor={colors.gray3}>
+                <CollapsibleSection summary={textNl.thermometer.collapsible_title} textColorOverride={colors.black} borderColorOverride={colors.gray3}>
+                  <Box my={3}>
+                    {Object.values(SeverityLevels).map((severityLevel, index) => {
```
VWSCoronaDashboard19, line 99: Please move this part above the return or somewhere at the top, because you are defining and returning inside the HTML part.

packages/app/src/components/collapsible/collapsible-section.tsx
```diff
@@ -86,20 +73,23 @@ export const CollapsibleSection = ({
...
     right: '100%',
     '&:hover, &:focus': {
-      color: 'blue',
+      color: colors.blue1,
     },
   })
 );
 
-const Summary = styled.button(
+interface SummaryPropsType {
```
VWSCoronaDashboard25, line 89: Please refrain from "Hungarian notation"-like suffices.
```suggestion
interface SummaryProps {
```

packages/app/src/components/collapsible/collapsible-section.tsx
VWSCoronaDashboard25: I actually think it's okay like this, because `Summary` is a sub-component of the `CollapsibleSection` and not used anywhere else.

(For the question, what would happen? I hope that Typescript complains on line 56, but I'm not sure.)

packages/app/src/components/severity-indicator-tile/components/severity-indicator-label.tsx
VWSCoronaDashboard25: I actually prefer if we don't do that anymore, because `3` is pretty meaningless on its own. In my opinion it's a weird construction to have a file somewhere deeply tucked away where this `3` is looked up in the `space` part of the theme file.
With `space[3]` it's more clear what's happening.

packages/app/src/domain/topical/components/indicator-level-description.tsx
```diff
@@ -0,0 +1,35 @@
+import { SeverityIndicatorLevel } from '~/components/severity-indicator-tile/components/severity-indicator-label';
+import { SeverityLevel } from '~/components/severity-indicator-tile/types';
+import { replaceVariablesInText } from '~/utils';
+import { asResponsiveArray } from '~/style/utils';
+import { Text } from '~/components/typography';
+import { Box } from '~/components/base';
+import css from '@styled-system/css';
+import { space } from '~/style/theme';
+
+interface indicatorLevelDescriptionProps {
```
VWSCoronaDashboard25, line 10: Type and interface definitions always `PascalCase`
```suggestion
interface IndicatorLevelDescriptionProps {
```

packages/app/src/domain/topical/components/indicator-level-description.tsx
```diff
@@ -0,0 +1,35 @@
...
+      gridTemplateRows="auto"
+      gridTemplateColumns={`${space[4]} auto`}
+      alignItems="center"
+      mb={4}
+      css={css({ columnGap: 3, rowGap: asResponsiveArray({ _: 3, sm: 1 }) })}
+    >
+      <SeverityIndicatorLevel level={level}>{level}</SeverityIndicatorLevel>
+      <Text fontWeight="bold">{label}</Text>
+      <Text css={css({ gridColumnStart: asResponsiveArray({ _: 1, sm: 2 }), gridColumnEnd: 3 })}>
+        {replaceVariablesInText(description.split('**').join(''), {
```
VWSCoronaDashboard25, line 29: > `description.split('**').join('')`
Why are `**` not replaced with spaces and why is it even necessary to do this in the first place?
Wouldn't it be better to just not put `**` in the description if we're going to ignore whatever the purpose is anyway?

packages/cms/src/lokalize/key-mutations.csv
```diff
@@ -30,4 +30,20 @@ timestamp,action,key,document_id,move_to
 2022-10-13T12:16:30.506Z,add,pages.vaccinations_page.vr.vaccination_coverage.top_level_information_block.dates,Hbf18OLBqHEYSTRqYPuVtR,__
 2022-10-13T12:16:31.512Z,add,pages.vaccinations_page.vr.vaccination_coverage.top_level_information_block.reference.href,Hbf18OLBqHEYSTRqYPuWBB,__
 2022-10-13T12:16:32.533Z,add,pages.vaccinations_page.vr.vaccination_coverage.top_level_information_block.title,hisQZuyoPo60cR6z1YNuqw,__
+2022-10-21T08:27:05.448Z,add,pages.topical_page.nl.thermometer.current_indicator_level,K48I3DmxlzVCAHe3YoBCIQ,__
+2022-10-21T08:27:06.474Z,add,pages.topical_page.nl.thermometer.indicator_for_level_1.title,i1eZmWfHjT5bLGKFlhIEjW,__
```
VWSCoronaDashboard25, line 34: Maybe instead just have this named `level_1` instead of `indicator_for_level_1` because it's already scoped to `thermometer` and (severity) indicator is just another/descriptive name for that.

packages/app/src/components/collapsible/collapsible-section.tsx
VWSCoronaDashboard25: Agreed, these can just be named `textColor` and `borderColor` without the `Override` part. The fact they are optional already implies the component has something in mind to use if you do not provide it.

packages/app/src/components/severity-indicator-tile/components/severity-indicator-label.tsx
VWSCoronaDashboard26: > I actually prefer if we don't do that anymore, because `3` is pretty meaningless on its own. In my opinion it's a weird construction to have a file somewhere deeply tucked away where this `3` is looked up in the `space` part of the theme file. With `space[3]` it's more clear what's happening.

I wholeheartedly agree with this. Using 'magic numbers' for these sort of things could have a tendency to be unclear. I would argue that we should revise this in our coding standards and come up with a consistent way of using these sort of theme variables (such as colours, spacing, fonts, etc.).

packages/app/src/components/collapsible/collapsible-section.tsx
```diff
@@ -86,20 +73,23 @@ export const CollapsibleSection = ({
...
+  textColor: string;
+}
+const Summary = styled.button((summaryProps: SummaryPropsType) =>
```
VWSCoronaDashboard25, line 92: I can't find anything like that after 5 minutes on Google, but I hoped there was a more React way to do this, like
```const Summary = styled.button<SummaryProps>({ textColor }) =>```

File: packages/app/src/pages/index.tsx [hunk 11]
```diff
@@ -82,6 +93,17 @@ const Home = (props: StaticProps<typeof getStaticProps>) => {
                 label={textNl.thermometer.indicator.label}
                 footerText={textNl.thermometer.indicator.footerText}
               />
+              <Box my={{ _: 3, md: 4 }} borderBottom={'1px solid'} borderBottomColor={colors.gray3}>
+                <CollapsibleSection summary={textNl.thermometer.collapsible_title} textColorOverride={colors.black} borderColorOverride={colors.gray3}>
+                  <Box my={3}>
```
VWSCoronaDashboard26, line 98: How about using a more semantic `ol` element here (in which each `IndicatorLevelDescription` component is represented by an `li` element with a `value` attrobite) instead of a generic `div` element (the default element used for a `Box` component)? The dropdown represents a list of items (severity levels) that are intentionally ordered (their order has impact on how severities are interpreted).

packages/app/src/components/collapsible/collapsible-section.tsx
VWSCoronaDashboard18: Couldn't find it either...

packages/app/src/components/collapsible/collapsible-section.tsx
VWSCoronaDashboard18: See line 20. there the defaults are set. And when optional this gets overridden by the defaults on line 20

packages/app/src/components/collapsible/collapsible-section.tsx
VWSCoronaDashboard18: This component has a default color for the text and borders.
For this one instance, they needed to be overwritten. On line 20 the default is set if the override is not used. I agree that the 'Override' part in the text is a bit excessive. But makes it really clear what happens. I will remove 'Override' for clearity.

packages/app/src/domain/topical/components/indicator-level-description.tsx
VWSCoronaDashboard18: This is because the description is the same as the thermometer itself. there the variable is bold. Here the variable has no bold. That's why we only have one key to work with and must remove the ```**```.
And why replace them with spaces? those will be merged into a single space character in the end. Because HTML doesn't allow for multiple space characters.

packages/app/src/pages/index.tsx
VWSCoronaDashboard18: Not sure how you would have this outside the HTML part? The return is because of the object iteration. And the const is to make it more readable inside the iteration and not having to get the same text index twice.

packages/cms/src/lokalize/key-mutations.csv
VWSCoronaDashboard18: FYI: Sanity gets a refactor inside the merged branch.

packages/app/src/pages/index.tsx
VWSCoronaDashboard19: Creating a function in this file which accepts `severity` as a parameter and return the `<IndicatorLevelDescription />` component might help with seperation.

VWSCoronaDashboard18: As discussed we take this to the next developer meeting.

packages/app/src/pages/index.tsx
VWSCoronaDashboard25: I am a bit on the fence about this. The order we're showing these severity level descriptions in, is a human decision. The decision could also be(come): active level first, then the rest (in reverse order, why not?).
The levels could also become different. Now we have 3 which is worse than 2 but not as bad as 4. But what if they decide to get another level that's just as bad as 3, but _in a different way_? 

I'd say safe ordered lists for natural orders such as sequential instructions, where the order is emphasised and use a `ul` instead. But that's just my opinion. I'll leave it to the developer.

Do make sure to test this properly on different browsers and systems though, because browsers add a number/bullet to `li`s natively, so making sure they don't show up in none of them can be tricky.

packages/app/src/domain/topical/components/indicator-level-description.tsx
```diff
@@ -0,0 +1,37 @@
...
+      <Box
+        display="grid"
+        gridTemplateRows="auto"
+        gridTemplateColumns={`${space[4]} auto`}
+        alignItems="center"
+        marginBottom={4}
+        css={css({ columnGap: 3, rowGap: asResponsiveArray({ _: 3, sm: 1 }) })}
+      >
+        <SeverityIndicatorLevel level={level}>{level}</SeverityIndicatorLevel>
+        <BoldText fontWeight="bold">{label}</BoldText>
```
VWSCoronaDashboard26, line 28: Nitpicky comment, but `fontWeight="bold"` seems to be legacy after refactoring from `Text` to `BoldText`. Having this redundant prop on the component could cause some confusion (`BoldText` is rendered as `strong` with a default `font-weight` of 700).

packages/app/src/pages/index.tsx
VWSCoronaDashboard18: The first part of your question shows me it's the right thing to have it in an ordered list. Because as stated by you and previously the list has an order. And it's not random. It's the same every time, based on a human decision and executed manually. And not something unordered. And the part about sequential instructions? Who decides the order? It's a human decision, getting really abstract really fast.

The second part of your question: I agree testing is an important part of this list behaviour.

packages/app/src/domain/topical/components/indicator-level-description.tsx
VWSCoronaDashboard18: Nice one, removed this part.

packages/app/src/pages/index.tsx
VWSCoronaDashboard26: > I am a bit on the fence about this. The order we're showing these severity level descriptions in, is a human decision. The decision could also be(come): active level first, then the rest (in reverse order, why not?). The levels could also become different. Now we have 3 which is worse than 2 but not as bad as 4. But what if they decide to get another level that's just as bad as 3, but _in a different way_?
>
> I'd say safe ordered lists for natural orders such as sequential instructions, where the order is emphasised and use a `ul` instead. But that's just my opinion. I'll leave it to the developer.

I get where you are coming from, but this is currently not supported by the implementation of the severity indicator. As of right now, it assumes there is a sequential order of severities based on what types are defined in code. If we would like to prepare for these potential changes like completely shuffling the order of severities (and possibly having the ability to do that on the fly) or having multiple severities of the same level but _different_, I agree that it would make sense to go with an unordered list as the order of severities could seem completely off (especially if an active severity is are sorted at the top, for example).

I mainly suggested using a semantic element over a `div` because we recently discussed to apply more semantic elements where possible.

VWSCoronaDashboard25: My wording was maybe not on the spot, every website is formed on a human decision, so by continuing that logic in this manner, `ul` is obsolete and every list on a website should be a `ol`.

MDN has a nice hint that maybe can help to decide when to use an `ol` and when an `ul`
> To determine which list to use, try changing the order of the list items; if the meaning changes, use the <ol> element — otherwise you can use <ul>.

In my opinion, changing the order doesn't change the meaning, because we have our own indicator inside that represents the severity level.

VWSCoronaDashboard25: @VWSCoronaDashboard26 
> I mainly suggested using a semantic element over a div because we recently discussed to apply more semantic elements where possible.

This, I fully support! I doubt though if we have forced our hand by the way it is coded, but it's fine either way. I just don't fully agree 😜

packages/app/src/components/collapsible/collapsible-section.tsx
```diff
@@ -53,24 +51,13 @@ export const CollapsibleSection = ({
...
           <Box width="100%">
             {summary}
             {id && (
-              <StyledAnchor
-                aria-hidden="true"
-                tabIndex={-1}
-                onClick={(e) => e.stopPropagation()}
-                href={`#${id}`}
-              >
+              <StyledAnchor aria-hidden="true" tabIndex={-1} onClick={(e) => e.stopPropagation()} href={`#${id}`}>
```
VWSCoronaDashboard28, line 60: A bit of nitpicking here and not blocking at all, but according to this line of the coding conventions:

`We avoid unnecessary short-hand variable names like **arr** for array or i for index or **acc** for a reduce accumulator.`

It would appear that `e` should be `event`.

Additionally, is there a convention for wrapping single params in parentheses for arrow functions?

VWSCoronaDashboard28 (approved):
Approved with one small non-blocking comment

VWSCoronaDashboard26 (approved):

PR #4472, opened 2022-11-01
feature/COR-1144-severity-indicator-timeline

VWSCoronaDashboard26:
## Summary

This branch covers all subtasks of COR-1144, meaning COR-1161, COR-1162 and COR-1168.
* COR-1161 describes the addition of new Timeline Event Collections to the CMS, yet these types were not extensive enough for the implementation of the timeline for the severity indicator (also known as thermometer). Instead, a new type called 'Thermometer Event Collections' has been introduced accordingly.
* COR-1162 describes the development of the new timeline for the severity indicator component.
* COR-1168 describes the inclusion of the timeline on the topical page.

## Motivation

Initially, the idea was to make the new timeline component as loosely coupled to the severity indicator as possible. However, due to time and complexity constraints (the fact that at some point nearly everything would become 'configurable' through React props), we decided to not take this approach. The new timeline component is in ways different than the existing timeline component often associated with data visualisations over time, in the fact that the layout is different, its bar contains parts (which are highlighted by the severity level of the associated timeline event), and so on.

#### After
|after_desktop|
|:--:|
| *Desktop without tooltip* |

|after_desktop_tooltip|
|:--:|
| *Desktop with tooltip* |

|after_mobile|
|:--:|
| *Mobile without tooltip* |

|after_mobile_tooltip|
|:--:|
| *Mobile with tooltip* |


packages/app/src/pages/index.tsx
```diff
@@ -97,6 +122,27 @@ const Home = (props: StaticProps<typeof getStaticProps>) => {
                 levelDescription={textNl.thermometer.indicator.level_description}
                 trendIcon={textNl.thermometer.indicator.trend_icon as TrendIcon}
               />
+
+              {thermometerEvents && thermometerEvents.length > 0 && (
```
VWSCoronaDashboard21, line 126: To make it more verbose, you could change it to `thermometerEvents.length !== 0`

packages/app/src/components/severity-indicator-tile/components/timeline/timeline.tsx
```diff
@@ -0,0 +1,156 @@
...
+  start: number;
+  end: number;
+  level: number;
+}
+
+interface TimelineProps {
+  startDate: number;
+  endDate: number;
+  timelineEvents?: TimelineEventConfig[];
+  size?: number;
```
VWSCoronaDashboard21, line 30: In the interface 'TimelineProps', 'size' is optional but in the component 'Timeline', 'size' is '10'. I therefor think you can make 'size' required in the interface 'TimelineProps'.

packages/app/src/components/severity-indicator-tile/components/timeline/timeline.tsx
```diff
@@ -0,0 +1,156 @@
...
+  description: string;
+  start: number;
+  end: number;
+  level: number;
+}
+
+interface TimelineProps {
+  startDate: number;
+  endDate: number;
+  timelineEvents?: TimelineEventConfig[];
```
VWSCoronaDashboard18, line 29: Why is this optional? It would make no sense to have a timeline without events, right?

packages/app/src/components/time-series-chart/components/timeline/components/timeline-marker.tsx
```diff
@@ -4,10 +4,14 @@ import { m } from 'framer-motion';
 import styled from 'styled-components';
 
 export function TimelineMarker({
+  color = colors.primary,
   isHighlighted,
+  hasPadding = true,
   size = 10,
 }: {
```
VWSCoronaDashboard18, line 11: Why not convert it to a sperate prop type while already changing this file?

packages/app/src/components/severity-indicator-tile/components/timeline/components/timeline-bar-part.tsx
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export const TimelineBarPart = ({ children, color, size = 10, width, isFirst = false, isLast = false }: TimelineBarPartsProps) => {
+  const borderRadius = isFirst ? `${size / 2}px 0 0 ${size / 2}px` : null;
+
+  return (
+    <Box
+      as="li"
+      alignItems="center"
+      backgroundColor={transparentize(0.7, color)}
```
VWSCoronaDashboard18, line 22: this would make it also 70% transparent and would get rid of the transparentize functionality
backgroundColor={`${color}B3`)}
```suggestion
      backgroundColor={`${color}B3`)}
```

packages/app/src/locale/use-lokalize-text.tsx
```diff
@@ -39,7 +39,7 @@ export function useLokalizeText(initialLocale: LanguageKey) {
   const [locale, setLocale] = useState(initialLocale);
   const [text, setText] = useState<SiteText>(languages[locale]);
   const lokalizeTextsRef = useRef<SanityDocument<LokalizeText>[]>([]);
-  const showSanityDebugToggle = enableHotReload || IS_STAGING_ENV;
+  const showSanityDebugToggle = true || enableHotReload || IS_STAGING_ENV;
```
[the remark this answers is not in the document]
VWSCoronaDashboard26: I must've accidentally included this in my commit. I reverted the change but now that Prettier rules have changed there's obviously a lot of changes to the actual file which could theoretically be ignored.

packages/app/src/components/severity-indicator-tile/components/timeline/components/timeline-bar-part.tsx
VWSCoronaDashboard26: While I also looked into this when working on this feature, I argued to not go with solution as it might not as clear as using a method with a straight forward name such as 'transparentize'. Using an 8-digit HEX color will require you to look up the HEX code value for the opacity, while something like `transparentize` is already available in the project and makes this code more readable in my opinion.

The tricky part about `transparentize` is that it actually deducts an opacity of `0.7`, leaving an actual opacity of `0.3` on the color. 🤔

packages/app/src/components/severity-indicator-tile/components/timeline/timeline.tsx
```diff
@@ -0,0 +1,156 @@
...
+  size?: number;
+  labels: {
+    [key: string]: string;
+  };
+  disableLegend?: boolean;
+  disableTimelineArrow?: boolean;
+  disableTimelineTimestamps?: boolean;
```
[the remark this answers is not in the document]
VWSCoronaDashboard26: I had initially added these props to make the timeline component somewhat generic, yet this did not actually work due to time and complexity restraints (I have also dedicated some of the PR description to this matter). I have opted to remove these for now as they go unused in the single instance of the timeline.

packages/app/src/components/severity-indicator-tile/components/timeline/timeline.tsx
VWSCoronaDashboard26: You are correct that `size` is marked as an optional prop and that `size` gets a default value of `10` when destructuring the prop from all props, but I would argue against making `size` a required prop. This way, we make sure the value of `size` always falls back to `10` unless an actual `size` value is set on the component's props. Otherwise, you'd always have to include `size` when using this `Timeline` component somewhere.

VWSCoronaDashboard18: Ah, I was mixing up some things. With my question, I intended to mean it can be set with a const with size 10. Because it's always 10 right now. Only when we are going to use this component more often.

packages/app/src/components/severity-indicator-tile/components/timeline/components/timeline-bar-part.tsx
VWSCoronaDashboard18: Good point, that makes sense. The inverted opacity would also be a no-go for me. I think we can discuss this or an alternative solution at another time and another place.

packages/app/src/components/severity-indicator-tile/components/timeline/components/tooltip-content.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  return (
+    <Box
+      color="black"
+      maxWidth="100%"
+      onMouseLeave={stopEventPropagation}
+      onMouseMove={stopEventPropagation}
+      onTouchMove={stopEventPropagation}
+      onTouchStart={stopEventPropagation}
+      px={18}
```
VWSCoronaDashboard28, line 40: Based on yesterdays discussion, this should be referencing the space array, right?

packages/app/src/components/severity-indicator-tile/components/timeline/logic/get-timeline-range-dates.ts
```diff
@@ -0,0 +1,24 @@
...
+    };
+  }
+
+  const timelineEventDates = timelineEvents
+    .flatMap((timelineEvent: SeverityIndicatorTimelineEventConfig) => [timelineEvent.start, timelineEvent.end])
+    .map((timelineEventDate: number) => {
+      return createDateFromUnixTimestamp(timelineEventDate).getTime();
+    });
+
+  const sortedTimelineEventDates = timelineEventDates.sort();
```
VWSCoronaDashboard28, line 18: Do we actually need to sort the array if we are calling `Math.min` and `Math.max` on it? 🤔

VWSCoronaDashboard28 (approved):
Looks good. I left a few non-blocking comments

packages/app/src/components/severity-indicator-tile/components/timeline/components/tooltip-content.tsx
VWSCoronaDashboard26: You are totally right! I've adjust this instance and all other instances in this pull request.

VWSCoronaDashboard19 (approved):

PR #4473, opened 2022-11-01
Task/COR-1182-nieuwe-varianten-toevoegen

VWSCoronaDashboard28:
### Summary
This branch covers the implementation of new variants as described by COR-1182.

- New variants added to schemas
- Corresponding labels for the new variants

The changes in the before and after are to illustrate that the newly added variants show up where required. Look for BQ.1. For the sake of brevity I assumed that if one worked the rest should work too (i.e. BQ.1.1, XBB).

Before (Graph and List)

1182-before-graph
1182-before-list

After (Graph and List)

1182-after-graph
1182-after-list


VWSCoronaDashboard18 (approved):
Nicely done!

PR #4479, opened 2022-11-04
bugfix/COR-1201-vandaag-is-not-positioned-proportionally-to-the-timeline

VWSCoronaDashboard28:
# Summary

This PR addresses the issue described in COR-1201.

The fix changes the logic for how the offset for the `vandaag` label is calculated. 

NOTE: In the images below, today refers to the 4th of November 2022.

## Before:
Screenshot 2022-11-04 at 17 11 44

## After:
Screenshot 2022-11-04 at 17 15 39

This fix also allows future events to be entered (in case that is a possibility):
Screenshot 2022-11-04 at 16 48 46

### DIFF
--- a/packages/app/src/components/severity-indicator-tile/components/timeline/components/timeline-bar-part.tsx
+++ b/packages/app/src/components/severity-indicator-tile/components/timeline/components/timeline-bar-part.tsx
@@ -12,6 +12,10 @@ interface TimelineBarPartsProps {
   isLast?: boolean;
 }
 
+/**
+ * TODO: An improvement would be to divide this bar part further into parts
+ * where each part refers to a day in the date range this bar part represents.
+ */
 export const TimelineBarPart = ({ children, color, size = 10, width, isFirst = false, isLast = false }: TimelineBarPartsProps) => {
   const borderRadius = isFirst ? `${size / 2}px 0 0 ${size / 2}px` : null;
 
--- a/packages/app/src/components/severity-indicator-tile/components/timeline/logic/get-timeline-bar-arrow-offset.ts
+++ b/packages/app/src/components/severity-indicator-tile/components/timeline/logic/get-timeline-bar-arrow-offset.ts
@@ -1,0 +1,17 @@
+import { createDateFromUnixTimestamp } from '~/utils/create-date-from-unix-timestamp';
+
+// Returns the difference betweeen two dates in days.
+const getDifferenceInDays = (start: Date, end: Date): number => {
+  return Math.ceil((end.getTime() - start.getTime()) / (1000 * 3600 * 24));
+};
+
+// Determines where to position the 'Vandaag' label on the timeline.
+export const getTimelineBarArrowOffset = (today: Date, startDate: number, endDate: number): number => {
+  const startOfIntervalDate = createDateFromUnixTimestamp(startDate);
+  const endOfIntervalDate = createDateFromUnixTimestamp(endDate);
+  const noOfDaysInInterval = getDifferenceInDays(startOfIntervalDate, endOfIntervalDate);
+  const noOfDaysFromStartToToday = getDifferenceInDays(startOfIntervalDate, today);
+  const arrowOffset = (noOfDaysFromStartToToday / noOfDaysInInterval) * 100;
+
+  return arrowOffset;
+};
--- a/packages/app/src/components/severity-indicator-tile/components/timeline/timeline.tsx
+++ b/packages/app/src/components/severity-indicator-tile/components/timeline/timeline.tsx
@@ -1,4 +1,4 @@
-import { colors, middleOfDayInSeconds } from '@corona-dashboard/common';
+import { colors } from '@corona-dashboard/common';
 import { ReactNode, useCallback, useRef, useState } from 'react';
 import styled from 'styled-components';
 import { useIntl } from '~/intl';
@@ -13,6 +13,7 @@ import { TimelineBar } from './components/timeline-bar';
 import { TimelineBarPart } from './components/timeline-bar-part';
 import { TimelineTooltipContent } from './components/tooltip-content';
 import { getSeverityColor } from '../../logic/get-severity-color';
+import { getTimelineBarArrowOffset } from './logic/get-timeline-bar-arrow-offset';
 import { SeverityLevels } from '../../types';
 
 export interface SeverityIndicatorTimelineEventConfig {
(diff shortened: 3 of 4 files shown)


packages/app/src/components/severity-indicator-tile/components/timeline/components/timeline-bar-part.tsx
```diff
@@ -12,6 +12,10 @@ interface TimelineBarPartsProps {
   isLast?: boolean;
 }
 
+/**
+ * TODO: An improvement would be to divide this bar part further into parts
```
VWSCoronaDashboard25, line 16: If we add TODO's, I think it's best to add a corresponding Jira ticket as well. This way we ensure that it's also on the radar of the product owner (or can be, at least).

However, I think this is a suggestion to improve the feature, not the code, and therefore should not be a TODO in the code (can still be a Jira ticket). Because a feature change can become a Won't Do for a product owner, while a refactoring of code is something we as developers can demand to be able to continue doing our job.

packages/app/src/components/severity-indicator-tile/components/timeline/logic/get-timeline-bar-arrow-offset.ts
```diff
@@ -0,0 +1,17 @@
...
+// Determines where to position the 'Vandaag' label on the timeline.
+export const getTimelineBarArrowOffset = (today: Date, startDate: number, endDate: number): number => {
+  const startOfIntervalDate = createDateFromUnixTimestamp(startDate);
+  const endOfIntervalDate = createDateFromUnixTimestamp(endDate);
+  const noOfDaysInInterval = getDifferenceInDays(startOfIntervalDate, endOfIntervalDate);
+  const noOfDaysFromStartToToday = getDifferenceInDays(startOfIntervalDate, today);
+  const arrowOffset = (noOfDaysFromStartToToday / noOfDaysInInterval) * 100;
+
+  return arrowOffset;
+};
```
VWSCoronaDashboard25, line 17: Why calculate it through the amount of days and round off in between?
A simpler version would be this:

```typescript
// Determines where to position the 'Vandaag' label on the timeline.
export const getTimelineBarArrowOffset = (today: Date, startDate: number, endDate: number): number => {
  const todayDate = today.valueOf();
  const progressOfToday = (todayDate - startDate) / (endDate - startDate);
  const arrowOffset = progressOfToday * 100;

  return arrowOffset;
};
```

packages/app/src/components/severity-indicator-tile/components/timeline/timeline.tsx
```diff
@@ -103,31 +104,22 @@ export const Timeline = ({ labels, startDate, endDate, legendItems, size = 10, t
   );
 };
 
-const getTimelineBarArrowOffset = (today: Date, startDate: number, endDate: number) => {
-  const todayInSeconds = middleOfDayInSeconds(today.getTime() / 1000);
-  return todayInSeconds / (endDate - startDate) / 100;
-};
+const TimelineBarArrow = ({ children, today, startDate, endDate }: { children: ReactNode; today: Date; startDate: number; endDate: number }) => {
+  const arrowLeftOffset = getTimelineBarArrowOffset(today, startDate, endDate) ?? 0;
```
VWSCoronaDashboard25, line 108: Maybe default to `undefined` and not show the arrow if it can't calculate the offset?

What would be a scenario that it can't calculate the offset?

packages/app/src/components/severity-indicator-tile/components/timeline/components/timeline-bar-part.tsx
VWSCoronaDashboard28: Good point, as I can't create a Jira ticket for this, would you mind creating one?

packages/app/src/components/severity-indicator-tile/components/timeline/logic/get-timeline-bar-arrow-offset.ts
```diff
@@ -0,0 +1,17 @@
+import { createDateFromUnixTimestamp } from '~/utils/create-date-from-unix-timestamp';
+
+// Returns the difference betweeen two dates in days.
```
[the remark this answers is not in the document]
VWSCoronaDashboard28: Overly excited between 😛

packages/app/src/components/severity-indicator-tile/components/timeline/timeline.tsx
VWSCoronaDashboard28: I'm not entirely sure, this code was left untouched in the changes I made here. I just moved it from the `Box` components `left` prop to the outer scope for the sake of readability. But I can make this change

packages/app/src/components/severity-indicator-tile/components/timeline/logic/get-timeline-bar-arrow-offset.ts
VWSCoronaDashboard25: The difference is because your version converts both 'now' and 'end' to the amount of whole days since 'start'.
So if start and end are 5 days apart and today is in-between, the offset will be either be 20%, 40%, 60%, 80% or 100% (one day after start, two days after start, etc.).
My code doesn't round to the whole days, so a 'today' in the morning can be 24% while the same 'today' in the evening can be 38%.

So basically, with your code in will take 24 hours for the marker to move, and then it will jump to the new position for 24 hours, while my versions creeps slowly to the right after every refresh of the page to account for the time that has elapsed.

Now I think about it, I maybe like your result better. Or not. I don't know. You decide. Or we could ask a designer.

packages/app/src/components/severity-indicator-tile/components/timeline/components/timeline-bar-part.tsx
VWSCoronaDashboard25: Yes, but remind me Monday again.

VWSCoronaDashboard28: Reminding you again as its monday 😄

packages/app/src/components/severity-indicator-tile/components/timeline/logic/get-timeline-bar-arrow-offset.ts
```diff
@@ -0,0 +1,17 @@
+import { createDateFromUnixTimestamp } from '~/utils/create-date-from-unix-timestamp';
+
+// Returns the difference between two dates in days.
+const getDifferenceInDays = (start: Date, end: Date): number => {
+  return Math.ceil((end.getTime() - start.getTime()) / (1000 * 3600 * 24));
```
VWSCoronaDashboard26, line 5: Would you not argue that using `Math.floor()` is more desirable than `Math.ceil()`? `Math.ceil()` always rounds up, meaning that, while still on a previous day, we might be rounded up to a next day. I think we should consider days as something that can not be rounded off but is always an absolute value. As per example: a Monday morning and evening are both still Monday and said evening should not be rounded up to Tuesday.

I came across this issue because of the current date (November 7th, 2022) and the current end date of the last data entry (November 8th, 2022). Using `Math.ceil()` in this case would cause the arrow to get an offset of `100%`, which is not desirable because November 7th is not the same as November 8th. 

image

However, applying `Math.floor()` keeps today at 'today' (with a left offset of `92.85714285714286%`). The difference between `100%` and the first percentage is `7.142857142857139%`, which is 1/14th of the total amount of days (there's assumably 14 days in each data entry). Multiply this percentage by 14 and you end up with something that is _roughly_ `100%`.

image

packages/app/src/components/severity-indicator-tile/components/timeline/timeline.tsx
```diff
@@ -103,31 +104,24 @@ export const Timeline = ({ labels, startDate, endDate, legendItems, size = 10, t
   );
 };
 
-const getTimelineBarArrowOffset = (today: Date, startDate: number, endDate: number) => {
-  const todayInSeconds = middleOfDayInSeconds(today.getTime() / 1000);
-  return todayInSeconds / (endDate - startDate) / 100;
-};
+const TimelineBarArrow = ({ children, today, startDate, endDate }: { children: ReactNode; today: Date; startDate: number; endDate: number }) => {
```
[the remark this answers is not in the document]
VWSCoronaDashboard28: I would have also thought that is the preference, but I'm pretty sure I didn't change anything there from the original implementation 🤔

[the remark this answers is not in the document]
VWSCoronaDashboard28: I can change it though

packages/app/src/components/severity-indicator-tile/components/timeline/logic/get-timeline-bar-arrow-offset.ts
VWSCoronaDashboard28: I've updated this part.

VWSCoronaDashboard26 (approved):

packages/app/src/components/severity-indicator-tile/components/timeline/timeline.tsx
```diff
@@ -103,31 +103,33 @@ export const Timeline = ({ labels, startDate, endDate, legendItems, size = 10, t
...
-};
+interface TimelineBarArrowProps {
+  children: ReactNode;
+  startDate: number;
+  endDate: number;
+}
+
+const TimelineBarArrow = ({ children, startDate, endDate }: TimelineBarArrowProps) => {
+  const todayDate = useCurrentDate();
+  const middleOfTodaysDateInSeconds = middleOfDayInSeconds(todayDate.getTime() / 1000);
```
VWSCoronaDashboard28, line 114: @VWSCoronaDashboard25 Would you mind reviewing this part again? This is regarding the issue Luuk mentioned to you during the demo.

VWSCoronaDashboard28, line 114: @VWSCoronaDashboard25 Would you mind reviewing this part again? This is regarding the issue @VWSCoronaDashboard26 mentioned to you during the demo.

VWSCoronaDashboard25 (approved):
Looks good!

PR #4492, opened 2022-11-10
Feature/cor 1174 implementation of the epidemiological situation

VWSCoronaDashboard27:
## Summary

- Created a new component `TopicalWeeklySummaryTile.tsx`
- Created new Sanity schemas for adding weekly summary items.

### DIFF
--- a/packages/app/src/components/weekly-summary/TopicalWeeklySummaryTile.tsx
+++ b/packages/app/src/components/weekly-summary/TopicalWeeklySummaryTile.tsx
@@ -1,0 +1,113 @@
+import css from '@styled-system/css';
+import { colors } from '@corona-dashboard/common';
+import { Box } from '~/components/base';
+import styled from 'styled-components';
+// import DynamicIcon from '~/components/get-icon-by-name';
+import theme from '~/style/theme';
+import { Markdown } from '~/components/markdown';
+import { space } from '~/style/theme';
+import { BoldText } from '~/components/typography';
+// import { ICONS_LIST } from '../severity-indicator-tile/constants';
+import { SeverityIndicatorLabel } from '../severity-indicator-tile/components/severity-indicator-label';
+import { SeverityLevels } from '../severity-indicator-tile/types';
+
+interface TopicalWeeklySummaryProps {
+  label: string | undefined;
+  level: SeverityLevels;
+  // name: IconName;
+
+  // title: string;
+  // dynamicSubtitle: string;
+  // icon: TopicalIcon;
+}
+
+export const TopicalWeeklySummaryTile = ({ label, level }: TopicalWeeklySummaryProps) => {
+  const HeadText = 'Weeksamenvatting 10 t/m 16 oktober';
+  const WeeklySummaryDescription = [
+    {
+      icon: 'Eye',
+      text: 'De verspreiding van het virus zet zich verder door. We zitten in de najaarsgolf.',
+    },
+    {
+      icon: 'MedischeScreening',
+      text: 'Het aantal COVID-patiënten dat in het ziekenhuis wordt opgenomen blijft stijgen. Het aantal bezette bedden op de verpleegafdeling is op dit moment zo goed als gelijk aan het niveau van de piek van de zomergolf.',
+    },
+    {
+      icon: 'Coronathermometer',
+      text: 'Op basis van bovenstaande ontwikkelingen staat de Coronathermometer op stand',
+    },
+  ];
+  // console.log('RRR', WeeklySummaryDescription);
+  // const ICONS_LIST = ['Eye', 'MedischeScreening', 'Coronathermometer'];
+  // const SummaryText = ['De verspreiding van het virus zet zich verder door. We zitten in de najaarsgolf.', 'gj;sjg;skgl;skgl;sd', 'efljalfjldf'];
+  // const TopicalWeeklySummaryText = SummaryText.map((text, index) => <InlineText key={index}>{text}</InlineText>);
+  // console.log(TopicalWeeklySummaryText, 'DDD');
+  // const DynamicIcon = icons[name];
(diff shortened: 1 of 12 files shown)


VWSCoronaDashboard19:
- [ ] Fix merge conflicts
- [ ] Create new Sanity fields and replace static texts

packages/app/src/components/weekly-summary/TopicalWeeklySummaryTile.tsx
```diff
@@ -0,0 +1,68 @@
...
+          {WEEKLY_SUMMARY_CONTENT.map((iconText, index) => {
+            return (
+              <>
+                <Box key={index} display="flex" pt={2} css={css({ gap: `0 ${space[3]}` })} alignItems="center" as="li">
+                  <Box minWidth={25} height={25}>
+                    <DynamicIcon width={25} name={iconText.icon as IconName} />
+                  </Box>
+                  <Box>
+                    <InlineText>{iconText.text}</InlineText>
+                    {iconText.isThermometerMetric ? (
```
VWSCoronaDashboard19, line 38: This should also suffice instead of adding an empty string for the other items
```suggestion
                    {iconText.isThermometerMetric &&
```

packages/app/src/components/weekly-summary/TopicalWeeklySummaryTile.tsx
```diff
@@ -0,0 +1,68 @@
+import css from '@styled-system/css';
```
VWSCoronaDashboard28, line 1: Little bit of nitpicking here but most components are named in lower kebab case as far as i can see so if we are to stay consisted this file should be `topical-weekly-summary-tile`.

packages/app/src/components/weekly-summary/TopicalWeeklySummaryTile.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { Box } from '~/components/base';
+import { BoldText, InlineText } from '~/components/typography';
+import DynamicIcon from '~/components/get-icon-by-name';
+import { SeverityLevel, SeverityLevels } from '../severity-indicator-tile/types';
+import { getSeverityColor } from '../severity-indicator-tile/logic/get-severity-color';
+import { asResponsiveArray } from '~/style/utils';
+import { BaseTile } from '~/queries/query-types';
+import { getFilenameToIconName } from '~/utils';
+
+interface TopicalWeeklySummaryProps {
```
VWSCoronaDashboard28, line 15: I would argue that the name here should be the same as the component with the addition of `Props` so `TopicalWeeklySummaryTileProps`.

packages/app/src/components/weekly-summary/TopicalWeeklySummaryTile.tsx
```diff
@@ -0,0 +1,68 @@
...
+  label: string | undefined;
+  level: SeverityLevels;
+  title: string;
+  summaryItems: BaseTile[];
+}
+
+export const TopicalWeeklySummaryTile = ({ label, level, title, summaryItems }: TopicalWeeklySummaryProps) => {
+  return (
+    <Box px={3} maxWidth={930}>
```
VWSCoronaDashboard28, line 24: `px` should be `paddingX` according to the latest discussions and `3` should be `space[3]`. This comment also applied to other uses of the same props in this file. so `padding={space[2]}` instead of `p={2}` on the line below etc.

packages/app/src/components/weekly-summary/TopicalWeeklySummaryTile.tsx
```diff
@@ -0,0 +1,68 @@
...
+      <Box border={`1px solid ${colors.gray3}`} p={4} m={asResponsiveArray({ _: undefined, sm: 3 })}>
+        <BoldText css={css({ fontSize: [3] })}>
+          <Markdown content={title} />
+        </BoldText>
+
+        <Box as="ul">
+          {summaryItems.map((summaryItem, index) => {
+            return (
+              <>
```
VWSCoronaDashboard26, line 33: These couple of lines can be simplified/reduced a little bit, as you do not have to include a `return` statement when rendering elements using `map()`. That would mean you would leverage parentheses instead of curly brackets after the arrow function `{summaryItems.map((summaryItem, index) => ( ... )}`. 

Another thing that could be simplified/reduced is the use of fragments (`<>...</>`). They are mainly useful for wrapping two or more top-level parent components/elements inside a map (which should only return a single parent) and will allow you group them and not render an element in the DOM. This line only returns a single parent (`Box`), hence the fragment can be ignored.

packages/app/src/components/weekly-summary/TopicalWeeklySummaryTile.tsx
```diff
@@ -0,0 +1,68 @@
...
+                <Box key={index} display="flex" pt={2} css={css({ gap: `0 ${space[3]}` })} alignItems="center" as="li">
+                  <Box minWidth={25} height={25}>
```
VWSCoronaDashboard26, line 35: Are these `px` values? If so, I would propose to write this as `propName="25px"` instead, as these feel like magic numbers.

packages/app/src/components/weekly-summary/TopicalWeeklySummaryTile.tsx
VWSCoronaDashboard26, line 24: Could we leverage `TOPICAL_SEVERITY_INDICATOR_TILE_MAX_WIDTH` for the `maxWidth` of this tile, as that value is already defined? You could opt that the name could then become a bit more generic as well.

VWSCoronaDashboard28 (approved):

VWSCoronaDashboard26 (approved):

VWSCoronaDashboard28 (approved):
I assume this ticket can be merged, right?

PR #4497, opened 2022-11-16
Feature/cor 1213 choropleth sorting age position

VWSCoronaDashboard27:
## Summary

- Changed ages sorting order of Choropleth Legenda in dropdown and tooltipVaccination campaigns.

### DIFF
--- a/packages/app/src/components/choropleth/components/canvas-choropleth-map.tsx
+++ b/packages/app/src/components/choropleth/components/canvas-choropleth-map.tsx
@@ -1,16 +1,7 @@
 import { colors } from '@corona-dashboard/common';
 import type { GeoProjection } from 'd3-geo';
 import Konva from 'konva';
-import {
-  memo,
-  MouseEvent,
-  MutableRefObject,
-  RefObject,
-  useCallback,
-  useEffect,
-  useRef,
-  useState,
-} from 'react';
+import { memo, MouseEvent, MutableRefObject, RefObject, useCallback, useEffect, useRef, useState } from 'react';
 import { Group, Layer, Line, Stage } from 'react-konva';
 import { isDefined, isPresent } from 'ts-is-present';
 import { isIOSDevice } from '~/utils/is-ios-device';
@@ -18,22 +9,12 @@ import { AccessibilityAnnotations } from '~/utils/use-accessibility-annotations'
 import { useIsTouchDevice } from '~/utils/use-is-touch-device';
 import { useUniqueId } from '~/utils/use-unique-id';
 import type { DataOptions } from '..';
-import type {
-  ChoroplethFeatures,
-  ChoroplethTooltipHandlers,
-  CodeProp,
-  FeatureProps,
-  FitExtent,
-} from '../logic';
+import type { ChoroplethFeatures, ChoroplethTooltipHandlers, CodeProp, FeatureProps, FitExtent } from '../logic';
 import { useHighlightedFeature } from '../logic/use-highlighted-feature';
-import {
-  ProjectedGeoInfo,
-  useProjectedCoordinates,
-} from '../logic/use-projected-coordinates';
+import { ProjectedGeoInfo, useProjectedCoordinates } from '../logic/use-projected-coordinates';
 import type { AnchorEventHandler } from './choropleth-map';
 
-Konva.pixelRatio =
-  typeof window !== 'undefined' ? Math.min(window.devicePixelRatio, 2) : 1;
+Konva.pixelRatio = typeof window !== 'undefined' ? Math.min(window.devicePixelRatio, 2) : 1;
 
 export type CanvasChoroplethMapProps = {
   anchorEventHandlers: AnchorEventHandler;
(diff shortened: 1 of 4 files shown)


VWSCoronaDashboard26 (commented):
As discussed during refinement, I am not sure if we would want to change the order of the ages directly, as done [here](https://github.com/minvws/nl-covid19-data-dashboard/pull/4497/files#diff-0824ef8ca798e447b6171f20b829836085827ccb847214a618bb455f2db31c1eL11), [here](https://github.com/minvws/nl-covid19-data-dashboard/pull/4497/files#diff-0824ef8ca798e447b6171f20b829836085827ccb847214a618bb455f2db31c1eL13) and [here](https://github.com/minvws/nl-covid19-data-dashboard/pull/4497/files#diff-682a0b5666a3b5a4d56ff6f94bf6174f38abd387c77eff6d36177fb6ab1372afL232). I am not sure if these changes introduce any side effects, but I think it might be less desirable to take this approach.

Instead, I would propose to use the initial value of the selected kind/campaign of vaccination and sort the ages based on that. I have prepared an alternate solution that takes this approach, so we can have a look at it together if you'd like.

VWSCoronaDashboard18:
> As discussed during refinement, I am not sure if we would want to change the order of the ages directly, as done [here](https://github.com/minvws/nl-covid19-data-dashboard/pull/4497/files#diff-0824ef8ca798e447b6171f20b829836085827ccb847214a618bb455f2db31c1eL11), [here](https://github.com/minvws/nl-covid19-data-dashboard/pull/4497/files#diff-0824ef8ca798e447b6171f20b829836085827ccb847214a618bb455f2db31c1eL13) and [here](https://github.com/minvws/nl-covid19-data-dashboard/pull/4497/files#diff-682a0b5666a3b5a4d56ff6f94bf6174f38abd387c77eff6d36177fb6ab1372afL232). I am not sure if these changes introduce any side effects, but I think it might be less desirable to take this approach.
> 
> Instead, I would propose to use the initial value of the selected kind/campaign of vaccination and sort the ages based on that. I have prepared an alternate solution that takes this approach, so we can have a look at it together if you'd like.

@VWSCoronaDashboard26 As discussed outside GitHub we talked about it in the refinement but decided that a dynamic solution would be too much effort and involves other fields as well. So for the sake of this ticket, we adjust the sort function..

packages/app/src/domain/vaccine/components/age-group-select.tsx
```diff
@@ -8,20 +8,20 @@ import { useIntl } from '~/intl';
 import { replaceVariablesInText } from '~/utils/replace-variables-in-text';
 import { parseBirthyearRange } from '../logic/parse-birthyear-range';
 
-export type AgeGroup = '12+' | '18+' | '60+';
+export type AgeGroup = '60+' | '18+' | '12+';
```
VWSCoronaDashboard25, line 11: Same as for the other type change

packages/app/src/domain/vaccine/vaccine-coverage-choropleth.tsx
```diff
@@ -200,41 +143,29 @@ export const VaccineCoverageChoropleth = ({
...
-    commonTexts.common.age_groups;
+  const ageGroupsText: SiteText['common']['common']['age_groups'] = commonTexts.common.age_groups;
   const formatCoveragePercentage = useVaccineCoveragePercentageFormatter();
 
   const secondaryContent = mapData
     .sort((a, b) => {
       const age1 = Number(a.age_group_range.replace(/\D/g, ''));
       const age2 = Number(b.age_group_range.replace(/\D/g, ''));
-      return age1 - age2;
+      return age2 - age1;
```
VWSCoronaDashboard25, line 165: Could you add a comment that this logic only coincidentally matches the business logic?
Something like this

```suggestion
      return age2 - age1; // This logic only coincidentally matches the business logic because for both campaigns the older age group happens to be the primary focus group, which might not be the same for a new campaign
```

VWSCoronaDashboard25:
> @VWSCoronaDashboard26 As discussed outside GitHub we talked about it in the refinement but decided that a dynamic solution would be too much effort and involves other fields as well. So for the sake of this ticket, we adjust the sort function..

This is how I remember it as well, and I still think it makes sense, especially seeing how little effort it is. So I propose we continue this as the PR is right now.

The only thing that's bothering me right now is that we have a sorting function that happens to match the business logic right now, so I've asked @VWSCoronaDashboard27 to add a comment on this.

VWSCoronaDashboard25 (approved):

packages/app/src/components/choropleth/components/canvas-choropleth-map.tsx
```diff
@@ -1,39 +1,20 @@
 import { colors } from '@corona-dashboard/common';
```
VWSCoronaDashboard28, line 1: Are there actually changes in this file or just formatting?

VWSCoronaDashboard28 (approved):

VWSCoronaDashboard26 (approved):

PR #4500, opened 2022-11-18
bugfix/COR-1222-vaccination-page-table

VWSCoronaDashboard26:
## Summary

* fixed issue where each expanded row would get a blue background, as only the first (highlighted) row should receive a blue background (regardless of its collapsed/expanded state)
* rewrote the styled components and their references to meet updated coding conventions

#### Before
image

image

#### After
image

image


packages/app/src/domain/vaccine/vaccine-campaigns-tile/components/narrow-vaccine-campaign-table.tsx
```diff
@@ -1,44 +1,30 @@
...
 import { Markdown } from '~/components';
-import {
-  VaccineCampaign,
-  VaccineCampaignDescriptions,
-  VaccineCampaignHeaders,
-} from '../types';
+import { BoldText } from '~/components/typography';
+import { space } from '~/style/theme';
+import { useCollapsible } from '~/utils/use-collapsible';
+import { StyledCell as Cell, StyledHeaderCell as HeaderCell, StyledRow as Row, StyledTable } from '.';
```
VWSCoronaDashboard28, line 7: What made you want to Alias the styled components here to a name which omits the `Styled` prefix? I thought it was discussed and agreed that it is preferred to have this prefix.

packages/app/src/domain/vaccine/vaccine-campaigns-tile/components/narrow-vaccine-campaign-table.tsx
```diff
@@ -1,44 +1,30 @@
...
-  campaignDescriptions,
-  headers,
-}: NarrowVaccineCampaignTableProps) => {
+export const NarrowVaccineCampaignTable = ({ campaigns, campaignDescriptions, headers }: NarrowVaccineCampaignTableProps) => {
   return (
     <StyledTable>
       <thead>
         <tr>
-          <HeaderCell mobile>{headers.vaccine}</HeaderCell>
+          <HeaderCell isMobile>{headers.vaccine}</HeaderCell>
```
VWSCoronaDashboard28, line 21: I'm wondering if its better to be more verbose here and go with `isMobile={true}` here versus presence of the prop. Do we have a convention for that?

packages/app/src/domain/vaccine/vaccine-campaigns-tile/components/shared-table-components.tsx
```diff
@@ -1,49 +1,49 @@
...
+  isMobile?: boolean;
+}
+
+export const StyledHeaderCell = styled.th<StyledHeaderCellProps>`
+  font-size: ${fontSizes[1]};
+  fontweight: ${fontWeights.bold};
+  padding: ${({ isMobile }) => (isMobile ? space[3] : space[2])} ${space[3]};
+  text-align: left;
+  vertical-align: top;
+  width: ${({ isMobile }) => (!isMobile ? '33%' : undefined)};
```
VWSCoronaDashboard28, line 21: What is the width set to if this defaults to `undefined`? Is it possible to change it to `auto` or something else in the case that it is not a mobile screen?

packages/app/src/domain/vaccine/vaccine-campaigns-tile/components/shared-table-components.tsx
```diff
@@ -1,49 +1,49 @@
...
+`;
+
+interface StyledCellProps extends PaddingProps {
```
VWSCoronaDashboard28, line 24: Does paddingProps simply add a prop for all padding values?

packages/app/src/domain/vaccine/vaccine-campaigns-tile/components/shared-table-components.tsx
```diff
@@ -1,49 +1,49 @@
...
-  })
-);
+}
+
+export const StyledCell = styled.td<StyledCellProps>`
+  float: ${({ alignRight }) => (alignRight ? 'right' : undefined)};
+  padding: ${({ isMobile }) => (isMobile ? space[2] : space[4])} ${space[3]};
+  vertical-align: top;
+  width: ${({ isMobile }) => (!isMobile ? '33%' : undefined)};
+  ${compose(padding)}
```
VWSCoronaDashboard28, line 35: From my understanding, `compose` is used to combine multiple style functions into one, so im wondering how much value it adds here for just one style function (i.e. padding)?

packages/app/src/domain/vaccine/vaccine-campaigns-tile/components/shared-table-components.tsx
```diff
@@ -1,49 +1,49 @@
...
+`;
+
+interface StyledRowProps {
+  isFirst: boolean;
+}
+
+const getBorderColor = (isFirst: boolean) => (isFirst ? colors.blue8 : colors.gray2);
+
+export const StyledRow = styled.tr<StyledRowProps>`
+  background: ${({ isFirst }) => (isFirst ? colors.primaryOpacity : undefined)};
```
VWSCoronaDashboard28, line 45: Is it maybe more "css-like" to default background to `none`?

packages/app/src/domain/vaccine/vaccine-campaigns-tile/components/wide-vaccine-campaign-table.tsx
```diff
@@ -1,27 +1,20 @@
...
-import {
-  VaccineCampaign,
-  VaccineCampaignDescriptions,
-  VaccineCampaignHeaders,
-} from '../types';
+import { Box } from '~/components/base';
+import { BoldText } from '~/components/typography';
+import { space } from '~/style/theme';
+import { useCollapsible } from '~/utils/use-collapsible';
+import { StyledCell as Cell, StyledHeaderCell as HeaderCell, StyledRow as Row, StyledTable } from '.';
```
VWSCoronaDashboard28, line 8: Same comment as the other file regarding the aliasing here

VWSCoronaDashboard28 (approved):
I've left a few comments, but there is nothing blocking the PR from being merged, so ill go ahead and pre-approve it.

packages/app/src/domain/vaccine/vaccine-campaigns-tile/components/shared-table-components.tsx
VWSCoronaDashboard26: `undefined` basically resolves to not setting the property, and allows the element to figure out the width of itself. Adding `auto` still works and is more verbose, but does not change anything about the actual styles and adds a 'redundant' CSS property to the element.

packages/app/src/domain/vaccine/vaccine-campaigns-tile/components/shared-table-components.tsx
VWSCoronaDashboard26: Correct: you're now able to use props like `padding`, `paddingRight` or `px` directly on a `StyledCellProps` component and use autocompletion through your IDE.

packages/app/src/domain/vaccine/vaccine-campaigns-tile/components/shared-table-components.tsx
VWSCoronaDashboard26: Similar answer to your question as [here](https://github.com/minvws/nl-covid19-data-dashboard/pull/4500#discussion_r1028224684): if there is no need for a background (and we can leave the row white), we simply do not set a background using `undefined`. I personally do not really see the value of adding a CSS property that could also be left out, but maybe we can discuss this some time.

packages/app/src/domain/vaccine/vaccine-campaigns-tile/components/shared-table-components.tsx
VWSCoronaDashboard26: This was my solution for simplifying overriding very specific padding styles per individual `StyledCell` without going overboard with adding too much logic. Please let me know if you have a different idea about this.

packages/app/src/domain/vaccine/vaccine-campaigns-tile/components/narrow-vaccine-campaign-table.tsx
VWSCoronaDashboard26: We do not have a convention for this as far as I am personally aware of.

VWSCoronaDashboard19 (approved):

PR #4505, opened 2022-11-21
bugfix/delete-undefined-sanity-key

VWSCoronaDashboard26:
## Summary

* added new manual script to delete key
* executed Sanity commands to delete rogue key

### DIFF
--- a/packages/cms/src/migrations/sprint59/index.ts
+++ b/packages/cms/src/migrations/sprint59/index.ts
@@ -1,0 +1,20 @@
+/**
+ * THIS SCRIPT DELETES DATA!
+ *
+ * To use this script:
+ * 1. Run `sanity dataset export` to backup your dataset before deleting a bunch of documents
+ * 2. Run `sanity exec src/migrations/sprint59 --with-user-token` to delete the documents
+ *
+ * NOTE: For the time being you should not delete more than ~1000 documents in one transaction. This will change in the future.
+ * See docs:https://www.sanity.io/docs/http-api/http-mutations#deleting-multiple-documents-by-query
+ */
+
+import { getClient } from '../../client';
+const client = getClient('development');
+
+client
+  .delete({
+    query: `*[_type == "lokalizeText" && [_id] match "c39981a1-bfa4-4aec-bf65-f0143820e150"]`,
+  })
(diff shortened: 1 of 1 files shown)


VWSCoronaDashboard28 (commented):
Correct me if I'm wrong, but this is a one off script specifically to delete keys pertaining to sprint 59? 

If so, would it not make sense to make it more generic?

Also, is this executing on both development and production environments of sanity?

VWSCoronaDashboard28 (approved):
I'm going to pre-approve this as I don't see anything wrong, but I did leave an earlier comment with a couple of questions.

VWSCoronaDashboard26:
> Correct me if I'm wrong, but this is a one off script specifically to delete keys pertaining to sprint 59?
>
> If so, would it not make sense to make it more generic?

You are correct, but I am not entirely sure if it would make sense to make this generic. We have [several other folders](https://github.com/minvws/nl-covid19-data-dashboard/tree/develop/packages/cms/src/migrations) containing files (going back as far as sprint 18) that do execute different one-off actions that were relevant to that sprint.

This script in particular deletes a single lokalize 'key' which actually did not have a `key` property (we are still not sure how this sprung into existence), so you could argue that it actually is no key.
 
> Also, is this executing on both development and production environments of sanity?

I must admit that I am not 100% sure about this, though I have not seen or experienced any issues on the production environment of Sanity. Could you shine your light on this, @VWSCoronaDashboard19?

VWSCoronaDashboard28:
> You are correct, but I am not entirely sure if it would make sense to make this generic. We have [several other folders](https://github.com/minvws/nl-covid19-data-dashboard/tree/develop/packages/cms/src/migrations) containing files (going back as far as sprint 18) that do execute different one-off actions that were relevant to that sprint.

Hmm, do we need to have these files in the codebase in the long term? Seems a bit odd, personally, to have files/folders from previous sprints which would most likely not be used again? Perhaps this is something to discuss outside of this PR.  

> I must admit that I am not 100% sure about this, though I have not seen or experienced any issues on the production environment of Sanity. Could you shine your light on this, @VWSCoronaDashboard19?

Alright, but the `key` for which this PR was created existed in both the development and production datasets or just development?

PR #4507, opened 2022-11-21
bugfix/COR-1069-inconsistent-link-hover-styles

VWSCoronaDashboard26:
## Summary

* updated LinkWithIcon and HeadingLinkWithIcon components to be more generic and prevent faulty hover states

## Notes
These changes are based upon the design proposals and work done as part of COR-937, for which the earlier pull request can be found [here](https://github.com/minvws/nl-covid19-data-dashboard/pull/4372/files).
I clicked through the entire dashboard looking for any instances of `LinkWithIcon` and `HeadingLinkWithIcon` to be sure that the simplified changes did not introduce any issues and meet the requirements specified.

### Screenshots
Both screenshots show a hover state, but no cursor.

#### Before
image

#### After
image

### DIFF
--- a/packages/app/src/components/link-with-icon.tsx
+++ b/packages/app/src/components/link-with-icon.tsx
@@ -9,100 +9,48 @@ import { Box } from './base';
 import { Anchor } from './typography';
 
 interface LinkWithIconProps {
-  href: UrlObject | string;
   children: string;
+  href: UrlObject | string;
   icon: ReactNode;
-  iconPlacement?: 'left' | 'right';
-  underline?: boolean;
   fontWeight?: 'normal' | 'bold';
+  iconPlacement?: 'left' | 'right';
   showAsButton?: boolean;
+  underline?: boolean;
 }
 
-interface IconProps {
-  icon: ReactNode;
-  isSingleWord?: boolean;
-  width: number;
-  height: number;
-  mr?: number | string;
-}
-
-interface LinkContainerProps {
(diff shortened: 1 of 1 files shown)


packages/app/src/components/link-with-icon.tsx
```diff
@@ -9,127 +9,74 @@ import { Box } from './base';
...
-            fontWeight,
-          })}
-        >
-          {iconPlacement === 'right' && (
-            <>
-              {!words.length ? children : firstWords}
-              <IconWrapper>
-                {words[words.length - 1]}
-                <IconSmall icon={icon} width={11} height={10} />
-              </IconWrapper>
```
VWSCoronaDashboard26, line 59: Perhaps someone could elaborate on this functionality and why it was written like this initially. I do not understand why we would

- split all words only to correctly place the icon after the last word?
- adjust positioning of the icon based on whether there is only a single word or not?
- only use this logic when the `iconPlacement` evaluates to `right`, and not also `left`?

I removed all this logic as I personally did not see any differences between the before and after versions of this component, and I feel my new solution is more straightforward and simple.

VWSCoronaDashboard28: I cannot think of why this would have been done this way in the past 🤷

VWSCoronaDashboard28 (commented):
Correct me if im wrong, but from your 'After' screenshot, I understand that the entire link should not be underlined when hovering? 🤔

PR #4508, opened 2022-11-22
Feature/COR-1088-archive-button-styling

VWSCoronaDashboard21:
## Summary

This PR addresses the changes described in COR-1088.
In this PR, the styling of the archived graphs button has been changed.

**Before**

_Default state_
Schermafbeelding 2022-11-22 om 15 56 49

_Active state_
Schermafbeelding 2022-11-22 om 15 56 54

_Active state on focus_

old_archived_button_active_focus

**After:**

_Default state_
Schermafbeelding 2022-11-22 om 15 57 30

_Default state on hover_
Schermafbeelding 2022-11-22 om 15 57 34

_Default state on focus_

archived_button_default_focus

_Default state on hover with focus_
archived_button_default_hover_focus

_Active state_
Schermafbeelding 2022-11-22 om 15 57 39

_Active state on hover_
Schermafbeelding 2022-11-22 om 15 57 42

_Active state on focus_
archived_button_active_focus

_Active state on hover with focus_
archived_button_active_hover_focus

### DIFF
--- a/packages/app/src/components/page-information-block/page-information-block.tsx
+++ b/packages/app/src/components/page-information-block/page-information-block.tsx
@@ -152,23 +152,35 @@ const Button = styled.button<{
   isActive?: boolean;
 }>`
   cursor: pointer;
-  background: ${({ isActive }) => (isActive ? colors.blue8 : colors.white)};
+  background: ${({ isActive }) => (isActive ? colors.blue1 : colors.white)};
   border: ${({ isActive }) => (isActive ? colors.transparent : colors.gray3)};
   border-radius: 5px;
-  color: ${({ isActive }) => (isActive ? colors.white : colors.blue8)};
-  padding: ${space[3]} 12px;
+  color: ${({ isActive }) => (isActive ? colors.blue8 : colors.blue8)};
+  padding: 12px ${space[3]};
   border-style: solid;
   min-height: 36px;
   border-width: 1px;
   transition: 0.1s background-color;
 
   ${({ isActive }) =>
+    isActive &&
+    `
+    '&:focus': {
+      position: 'absolute',
+      outline: '2px dotted white',
+      outlineOffset: '-2px',
+      top: 2,
+      left: 2,
+    }
+  `}
+
+  ${({ isActive }) =>
     !isActive &&
     `
     &:hover {
-      background: ${colors.gray1};
-      color: ${colors.blue8};
-      border-color: ${colors.blue8};
+      background: ${colors.blue8};
+      color: ${colors.white};
+      border-color: ${colors.transparent};
     }
   `}
 `;
--- a/packages/app/src/components/page-information-block/page-information-block.tsx
+++ b/packages/app/src/components/page-information-block/page-information-block.tsx
@@ -121,9 +121,9 @@ export function PageInformationBlock({
           </Box>
           <Box my={3}>
             {showArchivedToggleButton && (
-              <StyledButton type="button" onClick={onToggleArchived} isActive={isArchivedHidden}>
+              <StyledArchiveButton type="button" onClick={onToggleArchived} isActive={isArchivedHidden}>
                 {!isArchivedHidden ? commonTexts.common.show_archived : commonTexts.common.hide_archived}
-              </StyledButton>
+              </StyledArchiveButton>
             )}
           </Box>
         </Tile>
@@ -148,9 +148,11 @@ const MetadataBox = styled.div(
   })
 );
 
-const StyledButton = styled.button<{
+interface StyledArchiveButtonProps {
   isActive?: boolean;
-}>`
+}
+
+const StyledArchiveButton = styled.button<StyledArchiveButtonProps>`
   background: ${({ isActive }) => (isActive ? colors.blue1 : colors.white)};
   border: ${({ isActive }) => (isActive ? colors.transparent : colors.gray3)};
   border-radius: 5px;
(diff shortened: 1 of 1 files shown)


packages/app/src/components/page-information-block/page-information-block.tsx
```diff
@@ -172,14 +148,29 @@ const MetadataBox = styled.div(
...
+  &:hover {
+    background: ${colors.blue8};
+    color: ${colors.white};
+    border-color: ${colors.transparent};
+  }
+
+  &:focus {
+    outline: 2px dotted white;
+    top: 2;
+    left: 2;
```
VWSCoronaDashboard26, line 174: What do the `top` and `left` properties achieve here? These sort of properties are usually associated with the `position` property, but since that is not present/set here, I think these are redundant?

Aside from the above, the value of `2` is not valid, unless it would be a reference to a `space` value, in which case it should be referenced like on line 159.

packages/app/src/components/page-information-block/page-information-block.tsx
```diff
@@ -172,14 +148,29 @@ const MetadataBox = styled.div(
...
+  cursor: pointer;
+  background: ${({ isActive }) => (isActive ? colors.blue1 : colors.white)};
+  border: ${({ isActive }) => (isActive ? colors.transparent : colors.gray3)};
+  border-radius: 5px;
+  color: ${({ isActive }) => (isActive ? colors.blue8 : colors.blue8)};
+  padding: 12px ${space[3]};
+  border-style: solid;
+  min-height: 36px;
+  border-width: 1px;
+  transition: 0.1s background-color;
```
VWSCoronaDashboard26, line 163: Nitpicky comment and nothing blocking nor something we have agreed upon as a team, but I personally prefer to alphabetically sort CSS properties. If you use VS Code, for example, you can easily achieve this by selecting the lines, press `cmd + shift + p` and then use `Sort Lines Ascending`.

The same applies to the other properties below, but it is not an absolute requirement to pick this up as this has not been discussed.

packages/app/src/components/page-information-block/page-information-block.tsx
```diff
@@ -172,14 +148,29 @@ const MetadataBox = styled.div(
...
-    bg: !isActive ? 'blue8' : 'transparent',
-    border: 'none',
-    borderRadius: '5px',
-    color: !isActive ? 'white' : 'blue8',
-    px: !isActive ? 3 : 0,
-    py: !isActive ? 12 : 0,
-    cursor: 'pointer',
-  })
-);
+const Button = styled.button<{
```
VWSCoronaDashboard26, line 151: As per the recently updated coding conventions, we would want to prefix styled components with `Styled` to easily distinguish styled components from non-styled components. In this case, that would become `StyledButton`.

packages/app/src/components/page-information-block/page-information-block.tsx
VWSCoronaDashboard21: I have removed these.

packages/app/src/components/page-information-block/page-information-block.tsx
VWSCoronaDashboard21: Nice, I have sorted the lines.

VWSCoronaDashboard26 (approved):

packages/app/src/components/page-information-block/page-information-block.tsx
```diff
@@ -172,14 +148,27 @@ const MetadataBox = styled.div(
...
+  transition: 0.1s background-color;
+
+  &:hover {
+    background: ${colors.blue8};
+    color: ${colors.white};
+    border-color: ${colors.transparent};
+  }
+
+  &:focus {
+    outline: 2px dotted white;
```
VWSCoronaDashboard19, line 172: can you also use `${colors.white}` here for the white color?

VWSCoronaDashboard19 (approved):

packages/app/src/components/page-information-block/page-information-block.tsx
VWSCoronaDashboard21: Done.

VWSCoronaDashboard19:
Is the keyboard focus state visible for the light-blue and the dark-blue button?

packages/app/src/components/page-information-block/page-information-block.tsx
```diff
@@ -172,14 +148,27 @@ const MetadataBox = styled.div(
...
-    bg: !isActive ? 'blue8' : 'transparent',
-    border: 'none',
-    borderRadius: '5px',
-    color: !isActive ? 'white' : 'blue8',
-    px: !isActive ? 3 : 0,
-    py: !isActive ? 12 : 0,
-    cursor: 'pointer',
-  })
-);
+const StyledButton = styled.button<{
```
VWSCoronaDashboard28, line 151: I think there could be some confusion using the name `StyledButton` as it is possible that this can be created or already exists elsewhere. Something like `StyledArchiveButton` would be more clear and less likely to be duplicated.

Additionally, If I'm not wrong it is preferred to have interfaces for props, so an interface with the name `StyledArchiveButtonProps`.

VWSCoronaDashboard21: Agreed, I have changed it.

VWSCoronaDashboard27 (approved):

PR #4510, opened 2022-11-23
Bugfix/cor 1067 fullscreen modal close button is not aligned properly

VWSCoronaDashboard28:
### Summary

The fullscreen `expand` button and the `close` button were not aligned properly. This PR addresses that.

Apart from that, this PR also introduces some changes to the code which have been made to align with the updated conventions.

Before - Expand button
BEFORE - EXPAND

After - Expand button
AFTER - EXPAND

Before - Close button
BEFORE - CLOSE

After - Close button
AFTER - CLOSE

### DIFF
--- a/packages/app/src/components/fullscreen-chart-tile.tsx
+++ b/packages/app/src/components/fullscreen-chart-tile.tsx
@@ -1,6 +1,7 @@
+import { colors } from '@corona-dashboard/common';
 import { Close, Expand } from '@corona-dashboard/icons';
-import css from '@styled-system/css';
 import { useEffect, useRef, useState } from 'react';
+import styled from 'styled-components';
 import { Tile } from '~/components/tile';
 import { useIntl } from '~/intl';
 import { replaceVariablesInText } from '~/utils/replace-variables-in-text';
@@ -12,15 +13,13 @@ import { IconButton } from './icon-button';
 import { Metadata, MetadataProps } from './metadata';
 import { Modal } from './modal';
 
-export function FullscreenChartTile({
-  children,
-  metadata,
-  disabled,
-}: {
+interface FullscreenChartTileProps {
   children: React.ReactNode;
   metadata?: MetadataProps;
   disabled?: boolean;
-}) {
+}
+
+export function FullscreenChartTile({ children, metadata, disabled }: FullscreenChartTileProps) {
   const [isFullscreen, setIsFullscreen] = useState(false);
   const wasFullscreen = usePrevious(isFullscreen);
   const breakpoints = useBreakpoints();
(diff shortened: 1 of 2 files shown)


packages/app/src/components/fullscreen-chart-tile.tsx
```diff
@@ -59,47 +47,42 @@ export function FullscreenChartTile({
...
-            })}
-          >
-            <IconButton
-              ref={isFullscreen ? undefined : buttonRef}
-              title={label}
-              onClick={() => setIsFullscreen((x) => !x)}
-              size={16}
-            >
+          <StyledModalCloseButtonWrapper isFullscreen={isFullscreen}>
+            <IconButton ref={isFullscreen ? undefined : buttonRef} title={label} onClick={() => setIsFullscreen((x) => !x)} size={16}>
```
VWSCoronaDashboard26, line 51: Could we rename `x` to something more descriptive like `currentFullscreenState`?

VWSCoronaDashboard26 (approved):
Pre-approving as none on my comments are blocking.

packages/app/src/components/fullscreen-chart-tile.tsx
VWSCoronaDashboard28: Checkity check

VWSCoronaDashboard28: Opted to go with `previousValue` because thats actually what it is

VWSCoronaDashboard21 (approved):
Looks good to me!

packages/app/src/components/fullscreen-chart-tile.tsx
```diff
@@ -53,53 +48,48 @@ export function FullscreenChartTile({
...
 }
+
+interface StyledModalCloseButtonWrapperProps {
+  isFullscreen: boolean;
+}
+
+const StyledModalCloseButtonWrapper = styled.div<StyledModalCloseButtonWrapperProps>`
+  color: ${colors.gray3};
+  position: absolute;
+  right: ${({ isFullscreen }) => (isFullscreen ? '25px' : '0')};
```
VWSCoronaDashboard19, line 85: Off-topic and non-blocking: We should seriously consider adding `24px` or `25px` to the `spaces[]`, because we use it a LOT lately but not reusing it while other code parts do `like space[x]`. This keeps our code more consistent and we would be more aligned with the design team. This one will be a good subject for our next chapter to discuss how to approach this change.

packages/app/src/components/icon-button.tsx
```diff
@@ -12,54 +11,33 @@ interface IconButtonProps {
...
+}
+
+const StyledIconButton = styled.button<StyledIconButtonProps>`
+  background: transparent;
+  border: none;
+  color: ${({ color }) => color};
+  cursor: pointer;
+  display: block;
+  margin: 0;
+  padding: ${({ padding }) => (padding ? `${padding}px` : '0')};
```
VWSCoronaDashboard19, line 36: Nit-picky comment and non-blocking, but I believe that `0` will also suffice without being a string

packages/app/src/components/fullscreen-chart-tile.tsx
```diff
@@ -53,53 +48,48 @@ export function FullscreenChartTile({
...
-            })}
-          >
-            <IconButton
-              ref={isFullscreen ? undefined : buttonRef}
-              title={label}
-              onClick={() => setIsFullscreen((x) => !x)}
-              size={16}
-            >
+          <StyledModalCloseButtonWrapper isFullscreen={isFullscreen}>
+            <IconButton ref={isFullscreen ? undefined : buttonRef} title={label} onClick={() => setIsFullscreen((previousValue) => !previousValue)} size={16}>
```
VWSCoronaDashboard19, line 58: Can you use the `space[]` for spaces or `fontSizes[]` for the sizes for this component? If it's a `font-size` you want, please use `fontSizes[2]`

packages/app/src/components/fullscreen-chart-tile.tsx
VWSCoronaDashboard28: Yeah I agree, its a good idea to have a chat about how we can do that without too many side effects,

packages/app/src/components/fullscreen-chart-tile.tsx
VWSCoronaDashboard28: Yeah I looked into this when working on the ticket, but the size prop is expecting a number, and changing this right now would have an impact on a few components unrelated to this feature and would increase the testing scope. 

Do you still want me to change it?

VWSCoronaDashboard19: No, you can ignore my comment because of the type of the size property being a number.

PR #4514, opened 2022-11-24
Release/2.62.0

VWSCoronaDashboard19:
## Summary
Release branch for v2.62.0


VWSCoronaDashboard25 (approved):

.github/workflows/infra.yml
```diff
@@ -6,7 +6,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  build:
+  old-build:
```
VWSCoronaDashboard18, line 9: Is this intended? As it's been added quite recently?

VWSCoronaDashboard18: And I'm not sure if this is preferable to do for the first time while releasing other features.

VWSCoronaDashboard19: Yes. For the new environment and the current one, we need to trigger two actions. That's why an additional job has been added here.

VWSCoronaDashboard18 (approved):

PR #4516, opened 2022-11-28
bugfix/COR-1069-inconsistent-link-hover-styles

VWSCoronaDashboard26:
## Summary

* updated general typography styles to always make anchor elements underlined when associated props are supplied
* rewrote (pars of) the styled components for typography elements

## Notes
The original pull request for bugfix can be found [here](https://github.com/minvws/nl-covid19-data-dashboard/pull/4507), yet I opted to close it as it gave the wrong impression about the bug and took a wrong approach to resolve it.

I left a question in said pull request to try and figure out why certain logic was applied to JSX, resulting in complex JSX and a large redundancy of `span` elements in the DOM. However, during testing I realised that these changes were there because of certain sentences being cut off and the fact that we need to group the last word and the arrow icon together. I do still think that this logic should be removed, yet I opted to not pick it up as part of this ticket, mainly due to the fact that this type of logic can be found in multiple places throughout the code base and does not result in inconsistent link styles.

As per agreement with design: all links should get an underline when hovered. The below screenshot only covers the link mentioned in the associated ticket, but all links have been checked. Other links that are covered by this pull request are

* links in the footer
* navigational links on an article page
* navigational links on a GM page

Links that already existed but have a distinct `text-decoration` value of `none` have been left out for now.

As mentioned above: there are some more things to generally fix for the various flavours of links, but I think it makes sense pick this up seperately.

### Screenshots
Both screenshots show a hover state, but no cursor.

#### Before
image

#### After
image

### DIFF
--- a/packages/app/src/components/typography.tsx
+++ b/packages/app/src/components/typography.tsx
@@ -1,4 +1,4 @@
-import { Color } from '@corona-dashboard/common';
+import { Color, colors } from '@corona-dashboard/common';
 import css, { CSSProperties } from '@styled-system/css';
 import styled, { DefaultTheme } from 'styled-components';
 import { Preset, preset } from '~/style/preset';
@@ -13,20 +13,7 @@ export interface TextProps {
   ariaLabel?: string;
 }
 
-export interface AnchorProps extends TextProps {
-  underline?: boolean | 'hover';
-  hoverColor?: TextProps['color'];
-  display?: string;
-  width?: string | number;
-}
-
-export interface HeadingProps extends TextProps {
-  level: HeadingLevel;
-}
-
-export type HeadingLevel = 1 | 2 | 3 | 4 | 5;
-
-function textStyle(props: TextProps & { as?: string }) {
+export const textStyle = (props: TextProps & { as?: string }) => {
   return css({
     ...(props.as === 'button'
       ? {
@@ -40,16 +27,13 @@ function textStyle(props: TextProps & { as?: string }) {
       : undefined),
 
     ...(props.variant ? preset.typography[props.variant] : undefined),
-
     ...(props.fontWeight ? { fontWeight: props.fontWeight } : undefined),
     ...(props.color ? { color: props.color } : undefined),
-    ...(props.textTransform
-      ? { textTransform: props.textTransform }
-      : undefined),
+    ...(props.textTransform ? { textTransform: props.textTransform } : undefined),
     ...(props.textAlign ? { textAlign: props.textAlign } : undefined),
     ...(props.hyphens ? { hyphens: props.hyphens } : undefined),
   });
-}
+};
 
 export const Text = styled.p<TextProps>(textStyle);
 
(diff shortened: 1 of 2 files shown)


packages/app/src/components/typography.tsx
```diff
@@ -40,57 +27,76 @@ function textStyle(props: TextProps & { as?: string }) {
...
+`;
 
-export function styledTextVariant(variant: string, as?: string) {
+export interface HeadingProps extends TextProps {
+  level: HeadingLevel;
+}
+
+export type HeadingLevel = 1 | 2 | 3 | 4 | 5;
+
+export const Heading = styled.h1.attrs((props: HeadingProps & { as?: string }) => ({
```
VWSCoronaDashboard28, line 92: I'm a little unsure whats going on here? Could you explain a little?

VWSCoronaDashboard28 (approved):
Approving because I have no blocking comments, but I did ask a question in a comment

packages/app/src/components/typography.tsx
```diff
@@ -40,57 +27,76 @@ function textStyle(props: TextProps & { as?: string }) {
...
-    })
-);
+export interface AnchorProps extends TextProps {
+  underline?: boolean | 'hover';
+  hoverColor?: TextProps['color'];
+  display?: string;
+  width?: string | number;
+}
+
+export const Anchor = styled.a<AnchorProps>`
```
VWSCoronaDashboard21, line 51: Should this become StyledAnchor, based on the latest code conventions?

packages/app/src/components/typography.tsx
VWSCoronaDashboard21, line 92: Should this become 'StyledHeading' based on the latest code conventions?

packages/app/src/components/typography.tsx
```diff
@@ -40,57 +27,76 @@ function textStyle(props: TextProps & { as?: string }) {
...
+}
+
+export type HeadingLevel = 1 | 2 | 3 | 4 | 5;
+
+export const Heading = styled.h1.attrs((props: HeadingProps & { as?: string }) => ({
+  as: props.as ?? (`h${props.level}` as const),
+  variant: props.variant ?? (`h${props.level}` as const),
+}))<HeadingProps>(textStyle);
+
+export const styledTextVariant = (variant: string, as?: string) => {
```
VWSCoronaDashboard21, line 97: Should this be 'StyledTextVariant' instead of 'styledTextVariant' based on the latest code conventions?

packages/app/src/components/typography.tsx
VWSCoronaDashboard19: I was going to ask the same, but after I saw that `<Heading />` needs to change to `StyledHeading` in 48 different files and `<Anchor />` in 15 different files, I thought twice about it because it might be out of scope. 

But when it comes to the agreements regarding boy scouting, it should actually be done. Not sure if it should happen in this PR.

VWSCoronaDashboard26: > Should this become 'StyledHeading' based on the latest code conventions?

While I do agree with what you are saying, I do have the tendency to not pick this up for now. As @VWSCoronaDashboard19 also points out: changing these names would have quite the impact on the entire code base and is, in my opinion, something we should rather pick up outside of this pull request. If I were to pick this up I would have to change roughly 300 instances of either `Heading`, `Anchor`, `styledTextVariant`, `Text`, `InlineText` and `BoldText`.

VWSCoronaDashboard26: `.attrs()` is another styled components method that allows you to attach some props to the component. This line allows us to create a new styled `Heading` component and add props to it (such as anything that can be found in `HeadingProps`, `TextProps` and `as`). These props then determine how the component is eventually rendered. You can find the official docs for `.attrs()` [here](https://styled-components.com/docs/api#attrs).

VWSCoronaDashboard28: Aha, thats pretty neat

PR #4519, opened 2022-11-29
feature/COR-694-choropleth-clickable-tooltip

VWSCoronaDashboard26:
## Summary

* updated TooltipContent component so that the entire tooltip becomes a clickable anchor element if a link is supplied, but remains a div element when no link is present
* rewrote the TooltipContent component to leverage/satisfy updated conventions

### Screenshots
Screenshots contain both examples of a tooltip with (first image) and without (second image) a link to a page.

#### Before
COR-694_before_anchor

COR-694_before_box

#### After
COR-694_after_anchor

COR-694_after_box

### DIFF
--- a/packages/app/src/components/choropleth/tooltips/tooltip-content.tsx
+++ b/packages/app/src/components/choropleth/tooltips/tooltip-content.tsx
@@ -19,7 +19,7 @@ export const TooltipContent = (props: TooltipContentProps) => {
   const isTouch = useIsTouchDevice();
 
   return (
-    <StyledTooltipContent as={link ? 'a' : 'div'} href={link} isInteractive={isTouch} onClick={onSelect} aria-live="polite">
+    <StyledTooltipContent as={link ? 'a' : 'div'} href={link ? link : undefined} isInteractive={isTouch} onClick={onSelect} aria-live="polite">
       <StyledTooltipHeader>
         <Text variant="choroplethTooltipHeader">
           <StyledLocationIcon>
--- a/packages/app/src/components/choropleth/tooltips/tooltip-content.tsx
+++ b/packages/app/src/components/choropleth/tooltips/tooltip-content.tsx
@@ -9,17 +9,15 @@ import { useIsTouchDevice } from '~/utils/use-is-touch-device';
 
 interface TooltipContentProps {
   title: string;
-  onSelect?: (event: React.MouseEvent<HTMLElement>) => void;
   link?: string;
   children?: ReactNode;
 }
 
-export const TooltipContent = (props: TooltipContentProps) => {
-  const { title, onSelect, link, children } = props;
+export const TooltipContent = ({ title, link, children }: TooltipContentProps) => {
   const isTouch = useIsTouchDevice();
 
   return (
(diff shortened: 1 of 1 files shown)


packages/app/src/components/choropleth/tooltips/tooltip-content.tsx
```diff
@@ -1,107 +1,82 @@
...
 
-export function TooltipContent(props: IProps) {
+export const TooltipContent = (props: TooltipContentProps) => {
   const { title, onSelect, link, children } = props;
   const isTouch = useIsTouchDevice();
 
   return (
-    <StyledTooltipContent isInteractive={isTouch} onClick={onSelect} aria-live="polite">
-      <TooltipHeader href={link}>
+    <StyledTooltipContent as={link ? 'a' : 'div'} href={link} isInteractive={isTouch} onClick={onSelect} aria-live="polite">
```
VWSCoronaDashboard28, line 22: Shouldn't the `href` also be conditionally rendered? It would be weird having an `href` on a `div` 🤔

VWSCoronaDashboard28 (approved):
Looks good so I'm approving, but please see my comment

packages/app/src/components/choropleth/tooltips/tooltip-content.tsx
VWSCoronaDashboard26: The `href` would not be rendered if `link`(which is an optional prop) has no value (`undefined`). However, I do agree that this could be misread as if the `href` is always present, so I will adjust it accordingly.

packages/app/src/components/choropleth/tooltips/tooltip-content.tsx
```diff
@@ -1,107 +1,82 @@
...
-    px: 3,
-    display: 'flex',
-    justifyContent: 'space-between',
-    alignItems: 'center',
-    textDecoration: 'none!important',
-  })
-);
+const StyledTooltipContent = styled(Box)<StyledTooltipContentProps>`
+  color: ${colors.black};
+  cursor: ${({ onClick }) => (onClick ? 'pointer' : 'default')};
```
VWSCoronaDashboard18, line 45: How would you see this work? As there is no cursor on the tooltip? I think this line can be removed.

packages/app/src/components/choropleth/tooltips/tooltip-content.tsx
```diff
@@ -1,107 +1,82 @@
...
-import { Text } from '~/components/typography';
+import { Box } from '~/components/base';
+import { InlineText, Text } from '~/components/typography';
 import { space } from '~/style/theme';
 import { useIsTouchDevice } from '~/utils/use-is-touch-device';
 
-interface IProps {
+interface TooltipContentProps {
   title: string;
   onSelect?: (event: React.MouseEvent<HTMLElement>) => void;
```
VWSCoronaDashboard18, line 12: I think the whole onSelect prop can be removed. As it's never passed as a property where TooltipContent is used.

VWSCoronaDashboard18 (approved):
Good stuff!

PR #4521, opened 2022-11-30
feat: Add new Sanity field for collection title

VWSCoronaDashboard19:
## Summary

- Add a new Sanity field for the title above the measures list, so it can be edited independently.

**Before**
- Title above measures list (table) was the same as the title of the Measures page.

**After**
- Title above measures list (table) is not the same as the title of the Measures page and can be edited in Sanity as a seperate field.


VWSCoronaDashboard26 (approved):

VWSCoronaDashboard18 (approved):
Looks good

PR #4522, opened 2022-12-01
Feature/COR-1149 adjust hoverstate styling of choropleth

VWSCoronaDashboard28:
# Summary

This ticket required a change in the hover/tab state of the Choropleth maps on the dashboard. Previously, on hover a thick white border was present. The border, however, was 50% inside the Hovered region and 50% outside the hovered region. To make it clearer what is being hovered, this was changed to 100% outside border with a drop shadow.

The tab focus state changes from a black outline to a white outline.

FYI: The cursor is missing from the screenshots below.

BEFORE HOVER
BEFORE-Hover

BEFORE TAB
BEFORE-Tab State

AFTER HOVER
AFTER-Hover

AFTER TAB
AFTER-Tab state


packages/app/src/components/choropleth/components/canvas-choropleth-map.tsx
```diff
@@ -257,50 +212,78 @@ const HighlightedFeature = memo((props: HighlightedFeatureProps) => {
...
-            points={x.flat()}
-            strokeWidth={featureProps.hover.strokeWidth(hoverCode, true)}
-            closed
-            stroke={featureProps.hover.stroke(
-              hoverCode,
-              true,
-              isKeyboardActive
-            )}
-          />
+        {landCoords.map((x, i) => (
```
VWSCoronaDashboard26, line 253: While we are slightly optimising this code (switch from `type` to `interface` for example), let's also change variables such as `x` and `i` to something more meaningful like `coord` and `index`.

Same comment applies for the `waterCoords` map.

packages/app/src/components/choropleth/components/canvas-choropleth-map.tsx
```diff
@@ -257,50 +212,78 @@ const HighlightedFeature = memo((props: HighlightedFeatureProps) => {
...
+              shadowOpacity={0.5}
+              shadowBlur={6}
+              shadowOffset={{ x: 0, y: 3 }}
+            />
+            {/* The additional line is used as an overlay on the original to make it seem like the stroke on the original line is on the outside */}
+            <Line key={`hover-${i}`} x={0} y={0} points={x.flat()} closed fill={featureProps.hover.fill(hoverCode, true)} />
+          </>
+        ))}
+        {waterCoords?.map((x, i) => (
+          <>
```
VWSCoronaDashboard26, line 274: There's no need to use a fragment if the map only returns a single (parent) component.

packages/app/src/components/choropleth/components/canvas-choropleth-map.tsx
VWSCoronaDashboard28: Good one

VWSCoronaDashboard18 (approved):
Looks great

packages/app/src/components/choropleth/components/canvas-choropleth-map.tsx
```diff
@@ -257,50 +212,78 @@ const HighlightedFeature = memo((props: HighlightedFeatureProps) => {
...
+   */
+  let landCoords = hover;
+  let waterCoords;
+  if (hoverCode === 'VR19') {
+    // Represents the coordinates to create the shape of landmasses in Zeeland
+    landCoords = hover.filter((_, index) => index === 0 || index === 5);
+
+    // Represents the coordinates to create the shape of water bodies in Zeeland
+    waterCoords = hover.filter((_, index) => !(index === 0 || index === 5));
+  }
```
VWSCoronaDashboard25, line 248: I think you can omit the comments now because that's already explained above. Besides, these comments now just comment "what" and not "why", which can be considered a bad practice.

Furthermore, I'm not a big fan of using `let` in general. It adds cognitive load for the reader. (keeping in mind that the variable can change later on, forcing to read all the code to make sure such a change is not missed). Especially when the variable is not typed.
So in general, I'd advise to avoid it as much as possible.

Here you can create empty arrays and fill them conditionally. This way the reader can assume they will be filled directly after and don't have to remember the initial value that later on might be removed. I think it's a bit more foreboding.
Having them typed helps as well. The reader already how the content will look like and only has to dive into the logic if they want to know what exactly is being.
If `hover` was named similarly, like `coords` or something, the reader might even already guess `landCoords` and `waterCoords` will become a certain split of `coords` without looking at the logic. 

```suggestion
  const landCoords: [number, number][][] = [];
  const waterCoords: [number, number][][] = [];
  hover.forEach((coord, index) => {
    if (hoverCode !== 'VR19') landCoords.push(coord);
    if (hoverCode === 'VR19' && (index === 0 || index === 5)) landCoords.push(coord);
    if (hoverCode === 'VR19' && index !== 0 && index !== 5) waterCoords.push(coord);
  }
```
By putting the `if`s on a single line, you put the focus on the `if` conditions and they are easy to compare.

But that's just an opinion. I think this is a good subject for a chapter meeting.

packages/app/src/components/choropleth/components/canvas-choropleth-map.tsx
```diff
@@ -257,50 +212,78 @@ const HighlightedFeature = memo((props: HighlightedFeatureProps) => {
...
+interface HoveredFeatureProps {
   hoveredRef: RefObject<Konva.Group>;
   hover: [number, number][][] | undefined;
   hoverCode: string | undefined;
   featureProps: FeatureProps;
   isKeyboardActive?: boolean;
-};
+}
 
 const HoveredFeature = memo((props: HoveredFeatureProps) => {
```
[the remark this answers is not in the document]
VWSCoronaDashboard28: We have numerous instances where its done either of these two ways. Ill add this as a topic to the chapter meeting document and leave it as is for now.

packages/app/src/components/choropleth/components/canvas-choropleth-map.tsx
VWSCoronaDashboard28: What about this?

```
  let landCoords: [number, number][][] = [...hover];
  let waterCoords: [number, number][][];
  if (hoverCode === 'VR19') {
    landCoords = hover.filter((_, index) => index === 0 || index === 5);
    waterCoords = hover.filter((_, index) => !(index === 0 || index === 5));
  }
```

Like this you know that `waterCoords` is going to be an array of the given type at some point. I agree that having `let` declarations which are then filled much later can possibly add cognitive load, however given that `waterCoords` is assigned a value just below, I dont think it's an issue here.

VWSCoronaDashboard26 (approved):

packages/app/src/components/choropleth/components/canvas-choropleth-map.tsx
VWSCoronaDashboard25: It was quite a workaround to use `const` instead of `let` in this case, so I think this is fine as well. By typing the undefined variables, you remove a lot of the cognitive load, arguably just as well or better compared to my solution. 👍

PR #4523, opened 2022-12-02
bugfix/COR-1001-scaling-tooltip-width

VWSCoronaDashboard21:
## Summary

Corrected the tooltip width on desktop by changing max-width.

**Before:**
variants_tooltip_width_before

**After:**
variants_tooltip_width_after

### DIFF
--- a/packages/app/src/components/time-series-chart/components/tooltip/tooltip-wrapper.tsx
+++ b/packages/app/src/components/time-series-chart/components/tooltip/tooltip-wrapper.tsx
@@ -77,6 +77,17 @@ export function TooltipWrapper({ title, children, left, top: _top, bounds, paddi
   );
 }
 
+const StyledTooltipContainer = styled.div`
+  position: absolute;
+  background: ${colors.white};
+  box-shadow: ${shadows.tooltip};
+  pointer-events: none;
+  z-index: 1000;
+  border-radius: 1px;
+  top: 0;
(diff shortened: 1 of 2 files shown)


packages/app/src/components/time-series-chart/components/tooltip/tooltip-wrapper.tsx
```diff
@@ -159,50 +157,52 @@ function TooltipContent(props: TooltipContentProps) {
...
+interface StyledTooltipContentProps {
+  onClick?: (event: React.MouseEvent<HTMLElement>) => void;
+}
+
+const StyledTooltipContent = styled.div<StyledTooltipContentProps>`
+  border-radius: 1px;
+  color: ${colors.black};
+  cursor: ${(props) => (props.onClick ? 'pointer' : 'default')};
+  font-size: ${fontSizes[1]};
+  max-width: 440px;
```
VWSCoronaDashboard18, line 182: So this is just an increase in the max-width? How about a next longer name? Shouldn't this be a generic solution? with a responsive width? As I'm not mistaken the ticket says it should resize according to the width of the name...

VWSCoronaDashboard18 (commented):
Nice adjustment

VWSCoronaDashboard18 (approved):
Nice job, and good to have an follow up ticket!

PR #4524, opened 2022-12-02
Added middelOfDay integration

VWSCoronaDashboard18:
The input from the JSON files mismatched the timeline scope in some time zones.
The solution is to convert all times to the middle of the day. As this is the value that gets used by other functionality as well.

Before output DateTime:
Screenshot 2022-12-02 at 14 06 53

Current output DateTime:
Screenshot 2022-12-02 at 14 06 57


VWSCoronaDashboard21 (approved):
Looks good to me!

VWSCoronaDashboard28 (approved):

PR #4527, opened 2022-12-06
Feature/COR-1178: Change trend icons kpi tile

VWSCoronaDashboard28:
# Summary

This PR implements changes as described in COR-1178. The focal point of the changes is the introduction of a new trend icon for the theme tiles (KPI tiles) on the homepage. The new icons represent the intensity of a change in a measurement. To achieve this, a new sanity field has also been created to allow content managers to select an intensity level, along with the colour and direction, which was already possible. 

### Before:
BEFORE

### After:
AFTER


packages/cms/src/schemas/topical/trend-icon.ts
```diff
@@ -31,5 +31,20 @@ export const trendIcon = {
...
       validation: REQUIRED,
     },
+    {
+      title: 'Intensiteit',
+      name: 'intensity',
+      description: 'Beschrijft de intensiteit van relatieve verandering ten opzichte van de vorige meeting.',
+      type: 'number',
+      options: {
+        list: [
+          { value: 1, title: '1' },
```
VWSCoronaDashboard18, line 41: Maybe have instead of title: x, go with title: 'X pijltje(s) gekleurd'

packages/app/src/components/trend-icon.tsx
```diff
@@ -1,33 +1,83 @@
...
+interface TrendIconWithIntensityProps {
+  color: string;
+  direction: TrendDirection;
+  intensity: number;
+}
+
+const TrendIconWithIntensity = styled(ArrowWithIntensity)<TrendIconWithIntensityProps>`
+  flex-shrink: 0;
+  margin-left: ${space[2]};
+  transform: ${({ direction }) => (direction === TrendDirection.DOWN ? 'scaleY(-1)' : undefined)};
```
VWSCoronaDashboard26, line 74: I realised this just now, yet I do not think it is blocking or something to really worry about at this point because the `TrendIconWithIntensity` is only used on the homepage. Regardless, I do want to point out that a wrapper (like the deleted `TrendIconWrapper`) allows for simpler components in regards to styling, allowing styles in the actual component itself (`TrendIconWithIntensity`) to be more generic as it can be used throughout the code base. Given this, I would argue that wrapping components is not a bad practice, but props such as `colour` should go on the component directly rather than the wrapper.

packages/app/src/components/trend-icon.tsx
```diff
@@ -1,33 +1,83 @@
...
+    fill: '.one-arrow',
+    stroke: '.two-stroke, .three-stroke',
+  },
+  2: {
+    fill: '.one-arrow, .two-arrows',
+    stroke: '.three-stroke',
+  },
+  3: {
+    fill: '.one-arrow, .two-arrows, .three-arrows',
+    stroke: '.no-stroke',
```
VWSCoronaDashboard26, line 61: Should we include a comment about this `no-stroke` class selector?

packages/app/src/components/trend-icon.tsx
VWSCoronaDashboard28: I do agree, and you're right, wrappers aren't a bad thing. However stuff like colours, for example, I believe, as you mentioned, should be controlled from the component which is actually using that colour. Which is what we did in this case. Maybe the `flex-shrink` could still come from the wrapper which we removed but things like `margin-left` makes sense to me to come from the component as you don't want it to be present in case the icon isn't used. A wrapper would help there, but i feel like it's overkill to have a wrapper (an extra component) for 1 or 2 styling properties. 

In the future if `TrendIconWithIntensity` needs to be used elsewhere, you could pass in the `margin` as a prop and it becomes entirely flexible.

packages/app/src/components/trend-icon.tsx
VWSCoronaDashboard28: Good one, ill add this.

packages/app/src/components/trend-icon.tsx
```diff
@@ -1,33 +1,88 @@
...
+  direction: TrendDirection;
+  intensity: number;
+}
+
+const TrendIconWithIntensity = styled(ArrowWithIntensity)<TrendIconWithIntensityProps>`
+  flex-shrink: 0;
+  margin-left: ${space[2]};
+  transform: ${({ direction }) => (direction === TrendDirection.DOWN ? 'scaleY(-1)' : undefined)};
+
+  ${({ intensity, color }): string =>
```
VWSCoronaDashboard28, line 81: @VWSCoronaDashboard26 Just an FYI, I refactored the two separate functions for the `fill` and `stroke` into one.

VWSCoronaDashboard26 (approved):

packages/app/src/components/trend-icon.tsx
```diff
@@ -1,33 +1,88 @@
...
-  const TrendLabelNeutral = ariaLabel || commonTexts.accessibility.visual_context_labels.neutral_trend_label;
+  const trendLabels: { [key: string]: string } = {
+    up: ariaLabel || commonTexts.accessibility.visual_context_labels.up_trend_label,
+    down: ariaLabel || commonTexts.accessibility.visual_context_labels.down_trend_label,
+    neutral: ariaLabel || commonTexts.accessibility.visual_context_labels.neutral_trend_label,
+  };
+
+  const ariaLabelText = trendLabels[TrendDirection[trendDirection].toLowerCase()];
+
+  // Icon with intensity is used only on the homepage at the moment, for all other trend icons the default (below) are used.
```
VWSCoronaDashboard18, line 31: Why though? Solve it by having a condition on line 85

packages/app/src/components/trend-icon.tsx
```diff
@@ -1,33 +1,88 @@
...
+    fill: '.one-arrow, .two-arrows',
+    stroke: '.three-stroke',
+  },
+  3: {
+    fill: '.one-arrow, .two-arrows, .three-arrows',
+    /**
+     * The .no-stroke class does not actually exist on the ArrowWithIntensity component, however, because
+     * of the way we are using these classes in the styled TrendIconWithIntensity, it cannot be empty ('')
+     * or undefined, as this breaks all the icons on the page.
+     */
```
VWSCoronaDashboard18, line 65: Why though? Solve it by having a condition on line 85. Don't return a stroke over there if there is no stroke set...

VWSCoronaDashboard28: That did not work when we tried it yesterday. However, after your comment I was trying again but now I see that I can remove the `stroke` property from `3` and it still works. I'm not sure if thats because of the way I changed the styling or not. 🤔

VWSCoronaDashboard18 (approved):
Nice work!

packages/app/src/domain/topical/components/topical-tile.tsx
```diff
@@ -135,13 +131,6 @@ export function TopicalTile({ title, tileIcon, trendIcon, description, kpiValue,
...
 }
 
-const TrendIconWrapper = styled.span`
-  color: ${({ color }) => color};
-  flex-shrink: 0;
-  margin-left: ${space[2]};
-  width: 20px;
-`;
-
 const TileIcon = styled.span`
```
VWSCoronaDashboard27, line 134: Since you was touching this file, maybe it is good to rename this
style component to `styleTileIcon`

packages/app/src/components/trend-icon.tsx
```diff
@@ -1,33 +1,82 @@
...
-  const TrendLabelUp = ariaLabel || commonTexts.accessibility.visual_context_labels.up_trend_label;
-  const TrendLabelDown = ariaLabel || commonTexts.accessibility.visual_context_labels.down_trend_label;
-  const TrendLabelNeutral = ariaLabel || commonTexts.accessibility.visual_context_labels.neutral_trend_label;
+  const trendLabels: { [key: string]: string } = {
+    up: ariaLabel || commonTexts.accessibility.visual_context_labels.up_trend_label,
+    down: ariaLabel || commonTexts.accessibility.visual_context_labels.down_trend_label,
+    neutral: ariaLabel || commonTexts.accessibility.visual_context_labels.neutral_trend_label,
+  };
+
+  const ariaLabelText = trendLabels[TrendDirection[trendDirection].toLowerCase()];
```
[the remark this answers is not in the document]
VWSCoronaDashboard28: If you look carefully, the label is not lowercased.

Any and all code could potentially add cognitive load depending on who is looking at it and when. Usually that load is experienced when you first see something which requires more "working memory" from you, however thereafter it's drastically reduced, in my opinion. 

I could just as easily place `trendLabels[TrendDirection[trendDirection].toLowerCase()]` in the switch statement below, however then it would just be repeated 3 times. Which is why it's extracted here. 

I'm not sure if i understand the last point.

packages/icons/src/svg/arrow_with_intensity.svg
```diff
@@ -0,0 +1,27 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="44.96" height="27.8" viewBox="0 0 44.96 27.8">
+    <g>
+        <g>
+            <path
+                d="M34.67,2.74l7.9,10.92a2.09,2.09,0,0,1,0,2.48,1.68,1.68,0,0,1-2.24.42L35.9,12.74v11c0,1.56-1.24,2-2.63,2s-2.63-.46-2.63-2v-11s-3.85,3.49-4.38,3.82A1.69,1.69,0,0,1,24,16.14a2,2,0,0,1-.43-1.27A2.11,2.11,0,0,1,24,13.66l7.9-10.92a1.69,1.69,0,0,1,2.8,0Z"
+                fill="#fff"
```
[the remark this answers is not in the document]
VWSCoronaDashboard28: Maybe we can have a chat about this on Thursday, I guess this won't make it into the release with these pending points.

PR #4532, opened 2022-12-09
bugfix/COR-1146-choropleth-issues-on-android

VWSCoronaDashboard26:
## Summary

* added `onClick` event to selected image map areas to properly show tooltips for Android devices;
* adjusted determinators for checking if the current device is a touch device;
* fixed tooltip notification styling issues likely introduced by COR-694

## Notes

Disclaimer: let me start off by saying that I am not entirely sure if any of these changes will actually work 100% of the time, as I was only able to test this on Android emulators, not physical devices. I invite other developers that own Android devices to pull this branch locally and test the bugfix accordingly.

Here is a list of my findings

* I initially noticed that only regions with holes in them (Midden- en West-Brabant), or regions containing detached parts (Fryslân, Groningen, Flevoland, and so on) run into issues with showing the tooltip on Android devices;
* it appears that these regions can not be focussed upon (the `onFocus` event listener does not do anything for those regions), though I am not entirely sure why not;
* iOS devices seem to handle the above just fine;
* I added a new `onClick` on the areas to be handled for touch devices; oddly enough, `onTouchStart` does end up showing the tooltip for broken regions, but also instantly navigates to the pages of the selected region; the same behaviour occurs on iOS devices (both emulated as well as physical);
* before the changes to `packages/app/src/utils/is-touch-device.ts` the Android emulator tends to first recognise a device as non-touch at first; as soon as I use Chrome DevTools to inspect the emulator, it is recognised as a touch device; I am not sure if this is a hiccup of the emulator or something else; testing using physical devices (with and without this change) could help with insights on this;

I personally also believe that this fix also covers COR-51, or that COR-51 is not even applicable anymore. There is no popover as described in the ticket, unless a user long-presses the choropleth map. Long-presses act as 'right mouse clicks' in the context of most touch devices. In case if long-pressing, a popover shows information about an image. This image is used to facilitate the image map areas to make the choropleth interactive and allows for the popup to show and it therefore required.

### DIFF
--- a/packages/app/src/components/choropleth/components/choropleth-map.tsx
+++ b/packages/app/src/components/choropleth/components/choropleth-map.tsx
@@ -1,5 +1,5 @@
 import { geoConicConformal, geoMercator } from 'd3-geo';
-import { FocusEvent, memo, useMemo } from 'react';
+import { FocusEvent, MouseEvent, memo, useMemo } from 'react';
 import { isDefined } from 'ts-is-present';
 import { Box } from '~/components/base';
 import { useAccessibilityAnnotations } from '~/utils/use-accessibility-annotations';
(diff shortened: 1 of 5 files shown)


packages/app/src/components/choropleth/components/choropleth-map.tsx
```diff
@@ -21,22 +21,17 @@ import { TooltipSettings } from '../tooltips/types';
 import { CanvasChoroplethMap } from './canvas-choropleth-map';
 
 export type AnchorEventHandler = {
-  onFocus: (evt: FocusEvent<HTMLElement>) => void;
+  onFocus: (evt: FocusEvent<HTMLElement> | MouseEvent<HTMLElement>) => void;
```
VWSCoronaDashboard26, line 24: Lines 2 and 24 are the only changes to this file. The rest is auto-formatted.

packages/app/src/components/choropleth/tooltips/tooltip-notification.tsx
```diff
@@ -23,6 +23,7 @@ export const TooltipNotification = (props: TooltipNotificationProps) => {
 // Negative margin is used her to offset the padding from the parent component and to not alter styles for tooltips without a notification.
 const StyledTooltipNotification = styled.div`
   background-color: ${colors.yellow1};
+  border-radius: 0 0 ${radii[1]}px ${radii[1]}px;
```
VWSCoronaDashboard28, line 26: Nice that you took care of this as well!

VWSCoronaDashboard25 (approved):

PR #4534, opened 2022-12-14
bugfix/COR-1050-removed-timeframeOption-one-week-safety-region-municipality

VWSCoronaDashboard21:
## Summary

Removed timeframeOption for Last 7 days on the Safety Region and Municipality pages the dropdown function.

**Before**
National level
before_national

Safety region
before_safety_region

Municipality
before_municipality

**After**
National level
after_national

Safety region 
after_safety_region

Municipality
after_municipality

### DIFF
--- a/packages/app/src/domain/sewer/sewer-chart/sewer-chart.tsx
+++ b/packages/app/src/domain/sewer/sewer-chart/sewer-chart.tsx
@@ -12,6 +12,7 @@ import { LocationTooltip } from './components/location-tooltip';
 import { WarningTile } from '~/components/warning-tile';
 import { mergeData, useSewerStationSelectPropsSimplified } from './logic';
 import { useIntl } from '~/intl';
+import { space } from '~/style/theme';
 import { useScopedWarning } from '~/utils/use-scoped-warning';
 import { TimelineEventConfig } from '~/components/time-series-chart/components/timeline';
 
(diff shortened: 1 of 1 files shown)


packages/app/src/domain/sewer/sewer-chart/sewer-chart.tsx
```diff
@@ -134,9 +113,13 @@ export function SewerChart({
     [options]
   );
 
+  const timeframeOptionsWithoutOneWeek = TimeframeOptionsList.filter((option) => option !== TimeframeOption.ONE_WEEK);
+
+  const timeframeOptionsToUse = !vrNameOrGmName ? TimeframeOptionsList : timeframeOptionsWithoutOneWeek;
```
VWSCoronaDashboard26, line 118: Nitpicky comment and not blocking at all, but I would prefer to take more positive approaches to these sort of conditions. How about the below instead?

```
const timeframeOptionsVrOrGm = TimeframeOptionsList.filter((timeframeOption) => timeframeOption !== TimeframeOption.ONE_WEEK);
const timeframeOptions = vrNameOrGmName ? timeframeOptionsVrOrGm : TimeframeOptionsList;
```

The above says more about the timeframe options used for VR or GM instead of what it contains (or does not contain). The name is in a sense more generic yet is still very straight about what its intended use is, and does not require change if we, at some point, also want to include other timeframe options from VR or GM pages.

VWSCoronaDashboard26 (approved):

VWSCoronaDashboard28 (approved):
Looks good

PR #4536, opened 2022-12-14
bugfix/COR-1240-Timeseries-chart-label-goes-off-screen-on-mobile

VWSCoronaDashboard28:
# Summary

This PR fixes the issue where x-axis end labels were falling outside of the viewport on mobile AND falling out of grid on desktop. The issue was caused due to the end label being `centered` when it is a short label (without year). Because of this, the label exceeded the grid. This has been changed so that end labels are always aligned to the `end` which means to the left of the `tick`.

### BEFORE
BEFORE

### AFTER
AFTER

### DIFF
--- a/packages/app/src/components/time-series-chart/components/axes.tsx
+++ b/packages/app/src/components/time-series-chart/components/axes.tsx
@@ -14,6 +14,7 @@ import { NumberValue, ScaleBand, ScaleLinear } from 'd3-scale';
 import { memo, Ref, useCallback, useMemo } from 'react';
 import { isPresent } from 'ts-is-present';
 import { useIntl } from '~/intl';
+import { fontSizes } from '~/style/theme';
 import { createDateFromUnixTimestamp } from '~/utils/create-date-from-unix-timestamp';
 import { useBreakpoints } from '~/utils/use-breakpoints';
 import { Bounds } from '../logic';
@@ -197,10 +198,7 @@ export const Axes = memo(function Axes<T extends TimestampedValue>({
 
   /**
    * Using anchor middle the line marker label will fall nicely on top
-   * of the axis label.
-   *
-   * The only times at which we can not use middle is if we are
-   * rendering a year in the label, because it becomes too long.
+   * of the axis label. This will only happen for labels which are not the first or last label.
    */
   const getAnchor = (x: NumberValue) => {
     return x === tickValues[0] && isLongStartLabel ? (tickValues.length === 1 ? 'middle' : 'start') : x === tickValues[tickValues.length - 1] ? 'end' : 'middle';
@@ -241,7 +239,7 @@ export const Axes = memo(function Axes<T extends TimestampedValue>({
         rangePadding={xRangePadding}
         tickLabelProps={(x) => ({
           fill: colors.gray6,
-          fontSize: 12,
+          fontSize: fontSizes[0],
           /**
            * Applying a dx of -50%, when there's only a single tick value, prevents
            * the tick to go out of bounds and centers the tick value relative to the graph.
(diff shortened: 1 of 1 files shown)


packages/app/src/components/time-series-chart/components/axes.tsx
```diff
@@ -119,143 +108,70 @@ export const Axes = memo(function Axes<T extends TimestampedValue>({
...
-        ? hasDatesAsRange
-          ? 6
-          : 4
-        : hasDatesAsRange
-        ? 5
-        : 3
-      : hasDatesAsRange
-      ? 3
-      : 2;
+    const preferredDateTicks = breakpoints.sm ? (timeframe === 'all' ? (hasDatesAsRange ? 6 : 4) : hasDatesAsRange ? 5 : 3) : hasDatesAsRange ? 3 : 2;
```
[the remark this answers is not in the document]
VWSCoronaDashboard28: Yeah Its not very clear at all, however I did not touch this piece of code, its formatting. If i try to put it back, then the formatter will put it on one line again. If you want, I can try to simplify it.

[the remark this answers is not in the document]
VWSCoronaDashboard19: Yes please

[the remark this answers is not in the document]
VWSCoronaDashboard25: In regards of the axes component, I like that the details of how many date ticks are preferred are put away. I'd argue this makes it easier to understanding the axes component, because those details are irrelevant.

In regards of the amount of preferred date ticks, I agree the logic is (still) hard to follow, but it's hopefully rare that we as developers need to dive into this logic, but it's doable. So I'd argue to keep it on one line.

I would like the variable to be typed though, so that it's more easy to see it's just stuff to get a `number`.

```suggestion
    const preferredDateTicks: number = breakpoints.sm ? (timeframe === 'all' ? (hasDatesAsRange ? 6 : 4) : hasDatesAsRange ? 5 : 3) : hasDatesAsRange ? 3 : 2;
```

packages/app/src/components/time-series-chart/components/axes.tsx
```diff
@@ -283,19 +198,10 @@ export const Axes = memo(function Axes<T extends TimestampedValue>({
 
   /**
    * Using anchor middle the line marker label will fall nicely on top
-   * of the axis label.
-   *
-   * The only times at which we can not use middle is if we are
-   * rendering a year in the label, because it becomes too long.
+   * of the axis label. This will only happen for labels which are not the first or last label.
```
VWSCoronaDashboard25, line 201: No need to split the comment mid-sentence anymore.

```typescript
/**
 * Using anchor middle the line marker label will fall nicely on top of the axis label.
 * This will only happen for labels which are not the first or last label.
 */
```

packages/app/src/components/time-series-chart/components/axes.tsx
```diff
@@ -265,7 +181,6 @@ export const Axes = memo(function Axes<T extends TimestampedValue>({
    * a space plus a three character month, which makes 6.
    */
   const isLongStartLabel = xTicks[0].length > 6;
```
VWSCoronaDashboard25, line 183: This variable is not needed in the next section, so I think it's best to move it down just above the first section where it's used.

VWSCoronaDashboard25: It's even only used anywhere else I think, so best to stuff it inside the function, I guess.

```typescript
const getAnchor = (x: NumberValue) => {
  const isLongStartLabel = xTicks[0].length > 6;
  return x === tickValues[0] && isLongStartLabel ? (tickValues.length === 1 ? 'middle' : 'start') : x === tickValues[tickValues.length - 1] ? 'end' : 'middle';
```

VWSCoronaDashboard19 (approved):

packages/app/src/components/time-series-chart/components/axes.tsx
```diff
@@ -201,7 +194,19 @@ export const Axes = memo(function Axes<T extends TimestampedValue>({
...
+
+    if (isFirstTick && isLongStartLabel && tickValues.length !== 1) {
+      return 'start';
+    }
+
+    if (isLastTick) {
+      return 'end';
+    }
+
+    return 'middle';
```
VWSCoronaDashboard25, line 209: Thanks! This is much better!

VWSCoronaDashboard25 (approved):
Well done!

PR #4537, opened 2022-12-15
feature/COR-1260-variants-colour-updates

VWSCoronaDashboard26:
## Summary
* updated colour order for variant colours;
* updated `getVariantOrderColors` function to filter out variants with an 'other'-prefix;
* updated `VariantStackedAreaTileWithData` component to set the default timeframe selection to three months;

### Screenshots
Both screenshots show the complete timeframe and not the updated default selection of three months. This is done to
* show that the order of colours has been updated;
* the bug with skipping a first colour has been fixed (`#D360E5` is the new first colour, but this used to be `#FFC000`, which was skipped);

#### Before
image

#### After
image


packages/app/src/domain/variants/variants-stacked-area-tile/variants-stacked-area-tile.tsx
```diff
@@ -34,37 +30,22 @@ type VariantsStackedAreaTileProps = {
   noDataMessage?: ReactNode;
 };
 
-export function VariantsStackedAreaTile({
```
VWSCoronaDashboard28, line 37: The props type up top could and should be an interface

packages/app/src/domain/variants/variants-stacked-area-tile/variants-stacked-area-tile.tsx
```diff
@@ -124,20 +85,14 @@ function VariantStackedAreaTileWithData({
       description={text.toelichting}
       metadata={metadata}
       timeframeOptions={TimeframeOptionsList}
-      timeframeInitialValue={TimeframeOption.SIX_MONTHS}
+      timeframeInitialValue={TimeframeOption.THREE_MONTHS}
```
VWSCoronaDashboard28, line 88: This is out of scope here, but an idea for making this future ready would be to allow these kinds of options to be selected from Sanity.

packages/app/src/domain/variants/variants-stacked-area-tile/variants-stacked-area-tile.tsx
VWSCoronaDashboard19: Agreed

VWSCoronaDashboard28 (approved):
Looks good, I've left a small comment about type vs interface

packages/app/src/domain/variants/variants-stacked-area-tile/variants-stacked-area-tile.tsx
VWSCoronaDashboard26: I have updated both lines 24 and 52 to use `interface` rather than `type`.

packages/app/src/domain/variants/variants-stacked-area-tile/variants-stacked-area-tile.tsx
```diff
@@ -282,21 +202,14 @@ function useSeriesConfig(
...
-  css({
-    width: '100%',
-    display: 'flex',
-    height: '8em',
-    color: 'gray5',
-    justifyContent: 'center',
-    alignItems: 'center',
-  })
-);
+const NoDataBox = styled(Box)`
```
VWSCoronaDashboard19, line 208: Is there a specific reason why there is a custom styled-component of `<Box />`? Why not pass all these attributes directly to a `<Box />` component?

packages/app/src/domain/variants/variants-stacked-area-tile/variants-stacked-area-tile.tsx
```diff
@@ -34,37 +30,22 @@ type VariantsStackedAreaTileProps = {
...
   }
+
   return (
-    <VariantStackedAreaTileWithData
-      text={text}
-      values={values}
-      metadata={metadata}
-      variantColors={variantColors}
-    >
+    <VariantStackedAreaTileWithData text={text} values={values} metadata={metadata} variantColors={variantColors}>
```
VWSCoronaDashboard19, line 44: I know it's formatting, but I am not really super happy with the name of this component because it sounds weird, as if `VariantsStackedAreaTile` component has no data. Also not sure why we need a separate component for this one. Why not directly put the contents from `VariantStackedAreaTileWithData` in here?

packages/app/src/domain/variants/variants-stacked-area-tile/variants-stacked-area-tile.tsx
VWSCoronaDashboard26: I have thought about this and while I have used `<Box />` with inline props, I do think that it messes with the flow of the JSX, or more so the 'flow of reading' the JSX a bit. We have not really agreed on which approach should be used, so I am also completely fine with converting it to a `<Box />` and adding inline props for styling purposes, but do note that not all styles are applicable through props. In that case, you'd still have to use either a `style={ ... }` prop (for inline styles) or a `css={css({ ... })}` (using the `css` method from styles system) for 'unsupported' styles, which is not the case when using styled components.

packages/app/src/domain/variants/variants-stacked-area-tile/variants-stacked-area-tile.tsx
VWSCoronaDashboard26: Great idea! Simplified this component.

VWSCoronaDashboard19 (approved):

VWSCoronaDashboard18 (approved):
Nice PR

packages/app/src/domain/variants/variants-stacked-area-tile/variants-stacked-area-tile.tsx
```diff
@@ -20,95 +15,36 @@ import { SiteText } from '~/locale';
...
-    </VariantStackedAreaTileWithData>
-  );
-}
-
 const alwaysEnabled: (keyof VariantChartValue)[] = [];
 
-type VariantStackedAreaTileWithDataProps = {
+interface VariantsStackedAreaTileProps {
   text: VariantsStackedAreaTileText;
   values: VariantChartValue[];
```
VWSCoronaDashboard18, line 28: I see a typescript error. Maybe have this checked in a early return and have it also possible null here as well

VWSCoronaDashboard18 (changes requested):

packages/app/src/domain/variants/variants-stacked-area-tile/variants-stacked-area-tile.tsx
VWSCoronaDashboard26: I've updated the entire `VariantsStackedAreaTile` component to not leverage `children` and `noDataMessage` and render the component conditionally instead.

VWSCoronaDashboard18 (approved):
Great!

PR #4539, opened 2022-12-15
Cor-1235: make some fields larger, so long texts will be readable

VWSCoronaDashboard19:
## Summary

Some texts were not readable in Sanity, because the text field was too small and not sizing.

- Changed `localString` fields to `localeText` fields, so it becomes a field with more rows.

#### Before

Schermafbeelding 2022-12-15 om 11 34 38

#### After
Schermafbeelding 2022-12-15 om 11 34 00

### DIFF
--- a/packages/cms/src/schemas/topical/thermometer.ts
+++ b/packages/cms/src/schemas/topical/thermometer.ts
@@ -58,8 +58,8 @@ export const thermometer = {
       type: 'localeString',
     },
     {
-      title: 'Huidige stand omschrijvig',
-      description: 'De omschrijving spcifiek voor de huidige thermometer stand bij de trendIcon',
+      title: 'Huidige stand omschrijving',
+      description: 'De omschrijving specifiek voor de huidige thermometer stand bij de trendIcon',
       name: 'levelDescription',
       type: 'localeText',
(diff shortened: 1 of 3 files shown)


packages/cms/src/schemas/topical/thermometer.ts
```diff
@@ -59,9 +59,9 @@ export const thermometer = {
     },
     {
       title: 'Huidige stand omschrijvig',
-      description: 'De omschrijving spcifiek voor de huidige themrmometer stand bij de trendIcon',
+      description: 'De omschrijving spcifiek voor de huidige thermometer stand bij de trendIcon',
```
VWSCoronaDashboard26, line 62: If you are correcting typos in this file, I would propose to also change `omschrijvig` to `omschrijving` (line 61) and `spcifiek` to `specifiek` (line 62). Do not that there is also a separate ticket (COR-1288) for this, though.

VWSCoronaDashboard26 (approved):

packages/cms/src/schemas/topical/thermometer.ts
VWSCoronaDashboard19: Will do.

two flies in one slap.

PR #4540, opened 2022-12-15
Feature/cor 1285 sanity thermometer timeline events

VWSCoronaDashboard27:
### DONE
**- Remove the thermometerTimelineEvent's title & description fields from the Sanity schema in Tooltip Label**

*Before:*sanity_thermometer_tijdlijn_gebeurtenissen
*After:* Screenshot 2022-12-15 at 14 28 37

**- Added a Description text: 'Extra beschrijving voor in de laatste gebeurtenis in de tijdlijn' to the timeline event;**
Screenshot 2022-12-15 at 14 41 57

**- Convert 'level' field of 'thermometerTimelineEvent' to a dropdown;**

_Before:_sanity_thermometer_standen
_After:_Screenshot 2022-12-15 at 14 33 37

**-Adjusted connection between thermometerLevel's title & description and a thermometerTimelineEvent's title & description;**
_Before:_
thermometer-tooltop-en-uitleg-stand
_After:_
Screenshot 2022-12-15 at 14 49 35

### DIFF
--- a/packages/app/src/components/severity-indicator-tile/components/timeline/components/tooltip-content.tsx
+++ b/packages/app/src/components/severity-indicator-tile/components/timeline/components/tooltip-content.tsx
@@ -73,7 +73,11 @@ export const TimelineTooltipContent = ({ config, hasMultipleEvents, onNext, onPr
           <BoldText>{config.title}</BoldText>
         </Box>
 
-        <Text variant="label1">{config.description}</Text>
+        <Text variant="label1">
+          {replaceVariablesInText(config.description.split('**').join(''), {
+            label: config.title.toLowerCase(),
+          })}
+        </Text>
 
         {currentEstimationLabel && (
           <Box textVariant="label1">
--- a/packages/app/src/queries/get-topical-structure-query.ts
+++ b/packages/app/src/queries/get-topical-structure-query.ts
@@ -108,17 +108,13 @@ export function getTopicalStructureQuery(locale: string) {
 }
 
 export const getThermometerEvents = (thermometerEvents: ThermometerTimelineEvent[], thermometerLevels: ThermometerLevel[]) => {
-  // 1. USE thermometerEvent.level TO FIND level IN thermometerLevels
-  // 2. USE level TO REFERENCE level.title AND level.description
-
   return thermometerEvents.map<SeverityIndicatorTimelineEventConfig>((thermometerEvent) => {
-    // STEP 1 GOES HERE
-    const currentLevel = thermometerLevels?.find((thermometerLevel) => thermometerLevel.level === (thermometerEvent.level as SeverityLevel));
+    const levelDetails = thermometerLevels.find((thermometerLevel) => thermometerLevel.level === (thermometerEvent.level as SeverityLevel)) as ThermometerLevel;
 
     return {
-      title: currentLevel?.label ?? thermometerEvent.title, // STEP 2.1 - title
-      description: currentLevel?.description ?? thermometerEvent.description, // STEP 2.2 - description
-      level: thermometerEvent.level,
+      title: levelDetails.label,
+      description: levelDetails.description,
+      level: levelDetails.level,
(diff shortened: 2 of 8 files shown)


VWSCoronaDashboard28 (approved):

VWSCoronaDashboard19:
By removing the title and description as stated in the first bullet point, the title of the timeline annotation tiles became "Untitled" which looks weird IMO. Can these tiles have the level title?

packages/app/src/components/severity-indicator-tile/components/timeline/components/tooltip-content.tsx
```diff
@@ -73,7 +73,11 @@ export const TimelineTooltipContent = ({ config, hasMultipleEvents, onNext, onPr
           <BoldText>{config.title}</BoldText>
         </Box>
 
-        <Text variant="label1">{config.description}</Text>
+        <Text variant="label1">
+          {replaceVariablesInText(config.description.split('**').join(''), {
```
VWSCoronaDashboard19, line 77: I do not fully understand why we want to remove the bold text here `**example**`. I can't really find a motivation to do this in the ticket.

packages/app/src/queries/get-topical-structure-query.ts
```diff
@@ -106,12 +107,14 @@ export function getTopicalStructureQuery(locale: string) {
   return query;
 }
 
-export const getThermometerEvents = (thermometerEvents: ThermometerTimelineEvent[]) => {
+export const getThermometerEvents = (thermometerEvents: ThermometerTimelineEvent[], thermometerLevels: ThermometerLevel[]) => {
   return thermometerEvents.map<SeverityIndicatorTimelineEventConfig>((thermometerEvent) => {
```
VWSCoronaDashboard19, line 111: Is it necessary to return this map? I think returning the object on line 114 is sufficient.

packages/app/src/components/severity-indicator-tile/components/timeline/components/tooltip-content.tsx
VWSCoronaDashboard25: I've seen this before somewhere and I think it's because the description was defined for the severity level indicator description. There it should have a small part bold (see screenshot), but in the tooltip and explaination of the levels it should not be bold (see other screenshot).
image
image

This is an easy fix to remove the bold parts, but it's also highly specific for the current use case.
We might want to consider to invert the logic: make it default none-bold and put effort in making it bold in the one place where that emphasis is needed.

But I'm okay to do that separately, or not even if it's too complex.

VWSCoronaDashboard19: Why not add/remove the bold text in Sanity instead of forcing it to be (non)-bold in the code?

VWSCoronaDashboard25: We could do that as well, which means the description of a level is stored twice: one with the last word bold, and one without bold.
I'm not sure what the best approach is, because defining it twice also means that it's harder for the user to keep track of which is which and it creates room for mistakes in making sure they keep saying the same thing.

After consideration, I think removing the markdown indicators for bold (as it is now) is the easiest and safest way to make sure all three places have the text as they should have.

VWSCoronaDashboard25:
> By removing the title and description as stated in the first bullet point, the title of the timeline annotation tiles became "Untitled" which looks weird IMO. Can these tiles have the level title or something else that is more meaningful?

I like this idea. It would be nice if the events wouldn't just become a list of `Laag`, `Laag`, `Laag`, `Laag`, etc. because that's confusing for the users (and I think using the word `Laag` is difficult because that's also in the cms somewhere).
Maybe construct a string based on the startdate as well? If that's possible.
Something like `6/12: level 1`, `22/11: level 1`, `8/11: level 2`, `25/10: level 2`, etc.

VWSCoronaDashboard19 (approved):

PR #4541, opened 2022-12-15
feature/COR-1284-Improve-readability-Sanity-shrink-or-hide-fields

VWSCoronaDashboard28:
# Summary

This ticket wraps rich text fields for the homepage (Themes) and thermometer sections in sanity in field-sets to make the UX better.

The adjustments are for rich text fields mentioned in the ticket and one additional one, `Artikel Referentie`.

The before and after only shows one instance, but it's the same for all.

### Before
BEFORE

### After
AFTER

### DIFF
--- a/packages/cms/src/schemas/topical/thermometer.ts
+++ b/packages/cms/src/schemas/topical/thermometer.ts
@@ -10,6 +10,35 @@ export const thermometer = {
   type: 'object',
   title: 'Thermometer',
   name: 'thermometer',
+  fieldsets: [
+    {
+      title: 'De beschrijving boven de thermometer',
+      name: 'description',
+      description: 'Klik op het label om de velden te tonen.',
+      options: {
+        collapsible: true,
+        collapsed: true,
+      },
+    },
+    {
+      title: 'Artikel referentie',
+      name: 'artikel-referentie',
+      description: 'Klik op het label om de velden te tonen.',
+      options: {
+        collapsible: true,
+        collapsed: true,
+      },
+    },
+    {
+      title: 'Titel van standen informatie',
+      name: 'level-information',
+      description: 'Klik op het label om de velden te tonen.',
+      options: {
+        collapsible: true,
+        collapsed: true,
+      },
+    },
+  ],
   fields: [
     {
       title: 'Thermometer icoon',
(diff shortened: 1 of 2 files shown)


packages/cms/src/schemas/topical/thermometer.ts
```diff
@@ -84,6 +115,7 @@ export const thermometer = {
       title: 'Titel van standen informatie',
       name: 'trendIcon',
       type: 'trendIcon',
```
VWSCoronaDashboard26, line 117: I actually only just realised this: the thermometer uses the `trendIcon` type, which also introduces the new `intensity` for the thermometer, which I think is undesirable. However, this is currently a non-issue, out of scope for this ticket and non-blocking as we opted to never use the icon on the thermometer as it causes too much confusion about the meaning in combination with the text and the current level.

VWSCoronaDashboard26 (approved):

packages/cms/src/schemas/topical/thermometer.ts
```diff
@@ -10,6 +10,35 @@ export const thermometer = {
   type: 'object',
   title: 'Thermometer',
   name: 'thermometer',
+  fieldsets: [
```
VWSCoronaDashboard18, line 13: Not sure if it would make sense, but why three fieldsets and not one?

VWSCoronaDashboard28: Because there are three parts of that section which needs to be wrapped in a collapsible. Having just one collapsible with a vague name would not indicate well what is inside it.

packages/cms/src/schemas/topical/thermometer.ts
VWSCoronaDashboard28: Why does it actually use the `trendIcon` type? 🤔

VWSCoronaDashboard26: Because the thermometer used to have the same 'old' arrow as KPI tiles, using a colour and direction. But again, this is not something we have to change right now.

VWSCoronaDashboard28: Ah right, I got it

VWSCoronaDashboard18 (approved):
I hope this will help fill in these fields

PR #4542, opened 2022-12-16
feat: add lines to graphs and add legends

VWSCoronaDashboard19:
## Summary

Currently, four charts are missing legends and/or need to remove the fill below the line.

- Added `forceLegend` prop to force a legend to be shown, even if it has one data point for the legend.
- Change area charts to line charts

### DIFF
--- a/packages/app/src/domain/sewer/sewer-chart/sewer-chart.tsx
+++ b/packages/app/src/domain/sewer/sewer-chart/sewer-chart.tsx
@@ -185,13 +185,14 @@ export function SewerChart({ accessibility, dataAverages, dataPerInstallation, t
             timeframe={sewerTimeframe}
             seriesConfig={[
               {
-                type: 'area',
+                type: 'line',
                 metricProperty: 'average',
                 label: text.averagesDataLabel,
                 color: colors.scale.blue[3],
               },
             ]}
             dataOptions={dataOptions}
+            forceLegend
           />
         )
       }
--- a/packages/app/src/domain/sewer/sewer-chart/sewer-chart.tsx
+++ b/packages/app/src/domain/sewer/sewer-chart/sewer-chart.tsx
@@ -161,7 +161,7 @@ export function SewerChart({ accessibility, dataAverages, dataPerInstallation, t
                 style: 'dashed',
               },
               {
-                type: 'area',
+                type: 'line',
                 metricProperty: 'average',
                 label: text.averagesDataLabel,
                 color: colors.scale.blue[3],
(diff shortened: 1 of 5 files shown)


packages/app/src/domain/sewer/sewer-chart/sewer-chart.tsx
```diff
@@ -185,13 +185,14 @@ export function SewerChart({ accessibility, dataAverages, dataPerInstallation, t
...
               {
-                type: 'area',
+                type: 'line',
                 metricProperty: 'average',
                 label: text.averagesDataLabel,
                 color: colors.scale.blue[3],
               },
             ]}
             dataOptions={dataOptions}
+            forceLegend
```
VWSCoronaDashboard26, line 195: It is impossible for me to add a comment to [line 164](https://github.com/minvws/nl-covid19-data-dashboard/pull/4542/files#diff-ba2f12c8cba10df698242039d52240b392f95f57e028880122096ec790cfd1e5R164), but is this also something that needs to be changed? Currently the sewer chart is still shown as an area graph when on VR or GM level and selection an installation.

packages/app/src/pages/landelijk/positief-geteste-mensen.tsx
```diff
@@ -1,9 +1,4 @@
-import {
-  colors,
-  NlTestedOverallValue,
-  TimeframeOption,
-  TimeframeOptionsList,
-} from '@corona-dashboard/common';
+import { colors, NlTestedOverallValue, TimeframeOption, TimeframeOptionsList } from '@corona-dashboard/common';
```
VWSCoronaDashboard26, line 1: Am I correct to assume that this file does not include any changes nor that it is in scope for this feature?

VWSCoronaDashboard26 (changes requested):
Please have a look at my comment, as I think you might have missed something. Other than that, it looks good!

packages/app/src/pages/landelijk/positief-geteste-mensen.tsx
VWSCoronaDashboard19: Correct, it looks like I accidentally saved this file and caused a lot a formatting changes. This file can be ignored.

VWSCoronaDashboard26 (approved):

packages/app/src/domain/sewer/sewer-chart/sewer-chart.tsx
VWSCoronaDashboard19: Thanks for the heads up, didn't check that one before. I changed it in my last commit.

VWSCoronaDashboard25 (approved):

PR #4544, opened 2022-12-16
bugfix/COR-1241-Age-demographic-chart-labels-are-not-sizing-correctly-on-EN-dashboard

VWSCoronaDashboard28:
# Summary

The bug in this ticket was regarding labels being cut off in the age demographic chart. This PR fixes that issue and also refactors some stuff and adjusts styling based on updated conventions.

### Before
BEFORE

### After
AFTER

### DIFF
--- a/packages/app/src/components/age-demographic/age-demographic-chart.tsx
+++ b/packages/app/src/components/age-demographic/age-demographic-chart.tsx
@@ -1,5 +1,4 @@
 import { Color, colors } from '@corona-dashboard/common';
-import css from '@styled-system/css';
 import { AxisBottom, TickRendererProps } from '@visx/axis';
 import { GridColumns } from '@visx/grid';
 import { Group } from '@visx/group';
@@ -10,16 +9,22 @@ import { KeyboardEvent, MouseEvent } from 'react';
 import styled from 'styled-components';
 import { Box } from '~/components/base';
 import { Text } from '~/components/typography';
+import { fontSizes, fontWeights, space } from '~/style/theme';
 import { replaceVariablesInText } from '~/utils';
-import {
-  AccessibilityDefinition,
-  useAccessibilityAnnotations,
-} from '~/utils/use-accessibility-annotations';
+import { AccessibilityDefinition, useAccessibilityAnnotations } from '~/utils/use-accessibility-annotations';
 import { AgeDemographicCoordinates } from './age-demographic-coordinates';
 import { AgeDemographicChartText, AgeDemographicDefaultValue } from './types';
 import { formatAgeGroupRange } from './utils';
 
 export const AGE_GROUP_TOOLTIP_WIDTH = 340;
+
+const TickValue = ({ x, y, formattedValue }: TickRendererProps) => {
+  return (
+    <VisxText x={x} y={y} fill={colors.gray7} fontSize={fontSizes[2]} textAnchor="middle">
+      {formattedValue}
+    </VisxText>
+  );
+};
 
 interface AgeDemographicChartProps<T extends AgeDemographicDefaultValue> {
   /**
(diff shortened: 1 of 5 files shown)


packages/app/src/components/age-demographic/age-demographic-coordinates.ts
```diff
@@ -93,15 +68,15 @@ function calculateAgeDemographicCoordinates<
   const singleBarHeight = 25;
 
   // Define the graph dimensions and margins
-  const axisWidth = isSmallScreen ? 60 : 100;
+  const axisWidth = 100;
```
VWSCoronaDashboard28, line 71: This is the fix for the issue in the ticket, the rest is refactoring.

VWSCoronaDashboard25: How does this now look on a 'small screen'?

VWSCoronaDashboard25: Oh never mind, I get it now. It was a problem on small screens, so it's not a problem anymore.

VWSCoronaDashboard25 (approved):
216 lines of code less! 🙌

packages/app/src/components/age-demographic/age-demographic-chart.tsx
```diff
@@ -293,23 +195,48 @@ export function AgeDemographicChart<T extends AgeDemographicDefaultValue>({
...
 const StyledGroup = styled(Group)({});
-const StyledHoverBar = styled(Bar)(
-  css({
-    fill: 'transparent',
-    // transparent stroke is to capture mouse movements in between bars for the tooltip
-    stroke: 'transparent',
-    strokeWidth: 12,
+const StyledHoverBar = styled(Bar)`
+  fill: transparent;
+  stroke: transparent;
```
VWSCoronaDashboard26, line 223: Nitpicky comment and not blocking at all, but you could reference `colors.transparent` here. You could argue that a value like `transparent` never changes, but for consistency purposes I think it would be nice.

packages/app/src/components/age-demographic/age-demographic-coordinates.ts
```diff
@@ -93,15 +68,15 @@ function calculateAgeDemographicCoordinates<
...
   const width = parentWidth;
 
   // Height and top margin are higher for small screens to fit the heading texts
   const isNarrowScreen = parentWidth < 400;
```
VWSCoronaDashboard26, line 75: Could we use something like `useBreakpoints` here rather than a hardcoded numeric value? Breakpoints are already defined in the theme, so I think it makes sense to reuse those values.

packages/app/src/components/age-demographic/age-demographic-coordinates.ts
```diff
@@ -93,15 +68,15 @@ function calculateAgeDemographicCoordinates<
...
   // Set height height according to amount of bars in chart.
-  const height = isNarrowScreen
-    ? 234 + singleBarHeight * barCount
-    : 214 + singleBarHeight * barCount;
+  const narrowScreenHeight = 234 + singleBarHeight * barCount;
+  const normalScreenHeight = 214 + singleBarHeight * barCount;
```
VWSCoronaDashboard26, line 78: Just a question: what are these magic numbers? They feel a bit random, though they are fairly important for calculating the height of the entire graph. To be honest, I do not see too much a difference using either `234` and `214` on either mobile or desktop, so I wonder how important these twenty pixels are.

VWSCoronaDashboard26 (approved):
Pre-approved, but have a look at my comments.

packages/app/src/components/age-demographic/age-demographic-coordinates.ts
VWSCoronaDashboard28: You mean for `isNarrowScreen`?

packages/app/src/components/age-demographic/age-demographic-coordinates.ts
VWSCoronaDashboard28: I know just about as much about them as you do 😅 Diving into figuring out how this all comes together is another ballgame altogether.

packages/app/src/components/age-demographic/age-demographic-coordinates.ts
VWSCoronaDashboard28: I'm not entirely sure if I want to change this to breakpoints. `isNarrowScreen` is being determined by `parentWidth` which is determined by `useResizeObserver` with a default value of `840`. So I will leave it as is for now.

PR #4545, opened 2022-12-19
feat(github): Delete old and redundant workflow from actions

VWSCoronaDashboard19:
## Summary

- Delete old workflow where our code got deployed to the old environment, which is not relevant anymore. The new workflow to the new environment should only matter.

### DIFF
--- a/.github/workflows/infra-develop.yml
+++ b/.github/workflows/infra-develop.yml
@@ -5,17 +5,6 @@ on:
       - develop
 
 jobs:
-  trigger-old-develop-build:
-    environment: Development
-    name: Call Azure Pipeline
-    runs-on: ubuntu-latest
-    steps:
-      - name: Azure Pipelines Action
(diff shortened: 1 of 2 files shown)


VWSCoronaDashboard28 (commented):
I don't see anything about removing this in the ticket COR-1167. Did I miss something?

VWSCoronaDashboard19:
> I don't see anything about removing this in the ticket COR-1167. Did I miss something?

That's right, the ticket doesn't describe the exact steps in detail (which it should). The intention was to remove the old workflow and keep the new workflow in the new environment. The migration to the new environment has already been done by BE and I have also made the necessary adjustments in the FE. In any case, I will briefly add these steps to the relevant ticket.

VWSCoronaDashboard28 (approved):

VWSCoronaDashboard26 (approved):

VWSCoronaDashboard25 (approved):

PR #4550, opened 2022-12-20
feat: Move all legend's to vertical designs

VWSCoronaDashboard18:
Before:
Screenshot 2022-12-20 at 11 27 42

Screenshot 2022-12-20 at 11 27 51

Screenshot 2022-12-20 at 11 28 44

After:
Screenshot 2022-12-20 at 11 27 28

Screenshot 2022-12-20 at 11 28 10

Screenshot 2022-12-20 at 11 28 33

### DIFF
--- a/packages/app/src/components/choropleth-legenda.tsx
+++ b/packages/app/src/components/choropleth-legenda.tsx
@@ -5,6 +5,7 @@ import { ValueAnnotation } from '~/components/value-annotation';
 import { useIntl } from '~/intl';
 import { space, fontSizes } from '~/style/theme';
 import { replaceVariablesInText } from '~/utils';
+import { useBreakpoints } from '~/utils/use-breakpoints';
 import { Box } from './base';
 import { Legend, LegendItem } from './legend';
 import { InlineText, Text } from './typography';
@@ -13,13 +14,14 @@ interface ChoroplethLegendaProps {
   title: string;
   thresholds: ChoroplethThresholdsValue[];
   valueAnnotation?: string;
-  type?: 'default' | 'bar';
   pageType?: string;
   outdatedDataLabel?: string;
 }
 
 export function ChoroplethLegenda({ title, thresholds, valueAnnotation, pageType, outdatedDataLabel }: ChoroplethLegendaProps) {
   const { commonTexts } = useIntl();
+  const breakpoints = useBreakpoints(true);
+
   const legendItems = thresholds.map(
     (x: ChoroplethThresholdsValue, i) =>
       ({
@@ -40,7 +42,7 @@ export function ChoroplethLegenda({ title, thresholds, valueAnnotation, pageType
     <Box width="100%" spacing={2} aria-hidden="true">
       {title && <Text variant="subtitle1">{title}</Text>}
 
-      <Legend items={legendItems} columns={2} />
+      <Legend items={legendItems} columns={breakpoints.lg ? 1 : 2} />
 
       {pageType === 'sewer' && (
         <Box display="flex" alignItems="center" paddingBottom={space[2]}>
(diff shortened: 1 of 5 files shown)


packages/app/src/components/choropleth-legenda.tsx
```diff
@@ -53,54 +50,11 @@ export function ChoroplethLegenda({ title, thresholds, valueAnnotation, type = '
           </Box>
         </Box>
       )}
-
-      {'bar' === type ? (
```
VWSCoronaDashboard26, line 57: Since we are completely removing this, do we still need to pass in a `type` prop to this component as defined on line [18](https://github.com/minvws/nl-covid19-data-dashboard/pull/4550/files#diff-99dd104e084a2cd2a59e0c4a078e1b6e883c4f32b635f901113c59a4a1fe643bL18)?

packages/app/src/components/choropleth-legenda.tsx
```diff
@@ -42,9 +37,11 @@ export function ChoroplethLegenda({ title, thresholds, valueAnnotation, type = '
   );
 
   return (
-    <Box width="100%" paddingRight={`${endLabelSize.width ?? 0 / 2}px`} spacing={2} aria-hidden="true">
+    <Box width="100%" spacing={2} aria-hidden="true">
       {title && <Text variant="subtitle1">{title}</Text>}
 
+      <Legend items={legendItems} columns={2} />
+
       {pageType === 'sewer' && (
```
VWSCoronaDashboard26, line 45: This item is not in the design language of the other legend items, which I think it should. You probably also want to make some adjustments to this as well as making sure that it is actually part of the legend and not an individual item.

packages/app/src/components/choropleth-legenda.tsx
```diff
@@ -53,54 +50,11 @@ export function ChoroplethLegenda({ title, thresholds, valueAnnotation, type = '
...
-  width: string;
-}>(({ width }) =>
-  css({
-    flex: 1,
-    flexBasis: width,
-    position: 'relative',
-  })
-);
-
 const LegendaColor = styled.div<{
```
VWSCoronaDashboard26, line 58: I think we should rewrite these things to be 'vanilla' styled components instead of using the styled systems.

packages/app/src/components/choropleth-tile.tsx
```diff
@@ -70,13 +70,13 @@ export function ChoroplethTile({
           {typeof description === 'string' ? <Text>{description}</Text> : description}
 
           {onChartRegionChange && chartRegion && (
-            <Box display="flex" justifyContent={{ _: 'center', lg: 'flex-start' }} pt={4}>
+            <Box display="flex" justifyContent={'flex-start'} pt={4}>
```
VWSCoronaDashboard26, line 73: I think we can rewrite it like this.

```suggestion
<Box display="flex" justifyContent="flex-start" paddingTop={space[4]}>
```

VWSCoronaDashboard26 (changes requested):

packages/app/src/components/choropleth-legenda.tsx
VWSCoronaDashboard18: This is already removed

packages/app/src/components/choropleth-legenda.tsx
VWSCoronaDashboard18: I've added it to the list with legends

VWSCoronaDashboard18: Can you resolve this if you're oke with the solution?

packages/app/src/components/choropleth-legenda.tsx
VWSCoronaDashboard18: This is not there anymore because of another change

VWSCoronaDashboard18:
There are new 'after' screenshots:

Screenshot 2022-12-21 at 17 21 11
Screenshot 2022-12-21 at 17 22 03

VWSCoronaDashboard26 (approved):
Very elegant solution and a proper clean up! :shipit:

PR #4555, opened 2022-12-22
feature/COR-1278-infection-radar-implementation

VWSCoronaDashboard28:
# Summary

This PR implements the new infection radar line chart along with a timeline. There is no "before" scenario, so the screenshot below only illustrates the "after" situation. 

Regarding the timeline, an element has been created in Sanity for the new `nl__self_test_overall__time_series` time series chart. If you search for that key, you can find the area where the timeline events need to be created. This already exists on production.

Additionally, this PR points to COR-1262-infection-radar-main. This was initially created as the infection radar implementation had a few additional tickets. However now it is just what is covered in COR-1278. 

### After
image

### New timeline in sanity
Screenshot 2022-12-22 at 16 04 18

### DIFF
--- a/packages/app/src/components/time-series-chart/logic/legend.tsx
+++ b/packages/app/src/components/time-series-chart/logic/legend.tsx
@@ -4,12 +4,7 @@ import { useMemo } from 'react';
 import { isDefined } from 'ts-is-present';
 import { LegendItem } from '~/components/legend';
 import { useIntl } from '~/intl';
-import {
-  HatchedTimespanAnnotationIcon,
-  OutOfBoundsIcon,
-  SeriesIcon,
-  SolidTimespanAnnotationIcon,
-} from '../components';
+import { HatchedTimespanAnnotationIcon, OutOfBoundsIcon, SeriesIcon, SolidTimespanAnnotationIcon } from '../components';
 import { TimelineMarker } from '../components/timeline';
 import { isVisibleEvent } from '../components/timeline/logic';
 import { DataOptions } from './common';
@@ -21,11 +16,10 @@ export function useLegendItems<T extends TimestampedValue>(
   domain: number[],
   config: SeriesConfig<T>,
   dataOptions?: DataOptions,
-  hasOutofBoudsValues = false,
+  hasOutofBoundsValues = false,
   forceLegend = false
 ) {
-  const { timelineEvents, timespanAnnotations, outOfBoundsConfig } =
-    dataOptions || {};
+  const { timelineEvents, timespanAnnotations, outOfBoundsConfig } = dataOptions || {};
   const { commonTexts } = useIntl();
 
   return useMemo(() => {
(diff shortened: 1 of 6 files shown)


packages/app/src/components/time-series-chart/logic/legend.tsx
```diff
@@ -21,11 +16,10 @@ export function useLegendItems<T extends TimestampedValue>(
   domain: number[],
   config: SeriesConfig<T>,
   dataOptions?: DataOptions,
-  hasOutofBoudsValues = false,
+  hasOutofBoundsValues = false,
```
VWSCoronaDashboard26, line 19: Nitpicky comment, but let's also camelcase the `Of`, resulting in `hasOutOfBoundsValues`. Line 22 also refers to an `outOfBoundsConfig` (with an camelcased `Of`).

packages/app/src/pages/landelijk/positief-geteste-mensen.tsx
```diff
@@ -142,6 +145,40 @@ function PositivelyTestedPeople(props: StaticProps<typeof getStaticProps>) {
             articles={content.articles}
           />
 
+          {/* Infection radar */}
```
VWSCoronaDashboard26, line 148: I think this line, and line 180, are mainly meant for development purposes? I do like the idea of these sort of pointers of where certain components start (I can deduce where it ends by seeing the start of a new component). Are these comments rendered to the DOM?

VWSCoronaDashboard26 (approved):
Pre-approving, but have a look at my comment and question.

packages/app/src/components/time-series-chart/logic/legend.tsx
```diff
@@ -66,20 +60,13 @@ export function useLegendItems<T extends TimestampedValue>(
      */
     if (timespanAnnotations) {
       for (const annotation of timespanAnnotations) {
-        const isAnnotationVisible =
-          (first(domain) as number) <= annotation.end &&
-          annotation.start <= (last(domain) as number);
+        const isAnnotationVisible = (first(domain) as number) <= annotation.end && annotation.start <= (last(domain) as number);
```
VWSCoronaDashboard25, line 63: The reason we didn't blindly merge a PR that fixed all line lengths earlier, is because it gives us an opportunity to look at weird stuff in our code and see if we need to do something with it.

Casting stuff to primitives is something I'd consider weird stuff and I don't think we should ignore this. But given the timing, I'll create a new ticket for it to take the time to look into this.

VWSCoronaDashboard25 (approved):

packages/app/src/pages/landelijk/positief-geteste-mensen.tsx
VWSCoronaDashboard25: At first glance this seems like a fine idea, especially if this isn't rendered in the DOM.
But if we came to the point where we think comments that state where a certain part of the code starts and ends are a good idea, I'd say it's a clear sign that we should make our files shorter by extracting components to their own files.

VWSCoronaDashboard28: I was actually considering extracting the two charts to `InfectionRadarChart` and `PositiveTestsChart`. I decided against it because it would result in more prop drilling. However, I could still do that. What do you think?

VWSCoronaDashboard28: Also, no, they are not rendered to the DOM (as far as I can see).

PR #4579, opened 2022-12-30
Magic number cleanup part 4

VWSCoronaDashboard18:


packages/app/src/components/markdown.tsx
```diff
@@ -26,13 +26,13 @@ const renderers = {
     isInternalUrl(props.href) ? (
       <Link href={props.href} passHref>
         <Anchor underline display="inline-block">
-          {props.children} <ChevronRight width={10} height={10} />
+          {props.children} <ChevronRight width={'10px'} height={'10px'} />
```
VWSCoronaDashboard26, line 29: Nitpicky comment and non-blocking, but width="10px" would also be sufficient and might be easier to interpret as a static value (the { ... } can imply that this is actually a variable property). The same applies for the other with and height examples in this file.

packages/app/src/components/metadata.tsx
```diff
@@ -70,7 +62,7 @@ export function Metadata({
          * @TODO split up the `isTileFooter` vs non `isTileFooter` implementations,
          * should be separate components.
          */
-        <Box as="footer" mt={3} mb={mb || { _: 0, sm: -3 }} gridArea="metadata">
+        <Box as="footer" marginTop={space[3]} marginBottom={mb || { _: space[0], sm: '-3px' }} gridArea="metadata">
```
VWSCoronaDashboard26, line 65: A bit out-of-scope for this pull request, but what do you think about renaming the `mb` prop to `marginBottom` as well? It might be confusing to see `mb` in the codebase and it not being a Styled System prop.

packages/app/src/components/page-information-block/components/articles.tsx
```diff
@@ -47,14 +43,14 @@ function ArticleItem({ slug, cover, title }: ArticleItemProps) {
   return (
     <Link passHref href={`/artikelen/${slug}`}>
       <StyledAnchor>
-        <Box width={100} minWidth={100} maxHeight={66} overflow="hidden">
+        <Box width={'100px'} minWidth={100} maxHeight={66} overflow="hidden">
```
VWSCoronaDashboard26, line 46: See my earlier remarks. `minWidth` and `maxHeight` are also supposed to change.

packages/app/src/components/page-information-block/components/metadata.tsx
```diff
@@ -236,7 +194,7 @@ function MetadataReference({ icon, referenceLink }: metadataReferenceProps) {
                   })}
                 >
                   {word}&nbsp;
-                  <ChevronRight width={10} height={10} />
+                  <ChevronRight width={'10px'} height={'10px'} />
```
VWSCoronaDashboard26, line 197: See my earlier remarks.

packages/app/src/components/page-information-block/components/articles.tsx
```diff
@@ -47,14 +43,14 @@ function ArticleItem({ slug, cover, title }: ArticleItemProps) {
...
           <SanityImage
             {...getImageProps(cover, {
               defaultWidth: 122,
```
VWSCoronaDashboard28, line 49: Should `defaultWidth` also be a pixel value?

VWSCoronaDashboard21 (approved):
Looks good to me

packages/app/src/components/markdown.tsx
VWSCoronaDashboard18: adjusted accordingly, please resolve.

packages/app/src/components/metadata.tsx
VWSCoronaDashboard18: adjusted accordingly, please resolve.

packages/app/src/components/page-information-block/components/articles.tsx
VWSCoronaDashboard18: adjusted accordingly, please resolve.

packages/app/src/components/page-information-block/components/articles.tsx
VWSCoronaDashboard18: No, it does some calculation on the number

packages/app/src/components/metadata.tsx
```diff
@@ -70,7 +62,7 @@ export function Metadata({
          * @TODO split up the `isTileFooter` vs non `isTileFooter` implementations,
          * should be separate components.
          */
-        <Box as="footer" mt={3} mb={mb || { _: 0, sm: -3 }} gridArea="metadata">
+        <Box as="footer" marginTop={space[3]} marginBottom={marginBottom || { _: '0', sm: '-3px' }} gridArea="metadata">
```
VWSCoronaDashboard26, line 65: I believe I had already mentioned this before on a different pull request regarding magic numbers, but `sm: -3` evaluates to `sm: -${space[3]}` (with backticks for interpolation) or `sm: -1rem`, and should not be replaced by `sm: '-3px'`.

You can validate this in the current context of this component.

VWSCoronaDashboard26 (changes requested):
Simply decided to not pre-approve due to a difference in how a component should work. Other than that, all my earlier comments have been marked as resolved.

VWSCoronaDashboard28 (approved):

packages/app/src/components/metadata.tsx
VWSCoronaDashboard18: This is solved now

packages/app/src/components/max-width.tsx
```diff
@@ -5,6 +5,6 @@ import { Box } from '~/components/base';
 export const MaxWidth = styled(Box)(
   css({
     maxWidth: 'maxWidth',
-    margin: '0 auto',
+    margin: `0 auto`,
```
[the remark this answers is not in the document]
VWSCoronaDashboard18: Yes, it is... But adjusted it to clarify, please resolve this.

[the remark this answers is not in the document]
VWSCoronaDashboard26: What exactly needs to be clarified here? `'0 auto'` is as clear as with backtick-notation. I would even argue that the former example is more straight forward as backticks are often used when doing string interpolation using template literals. That is not the case here.

VWSCoronaDashboard26 (approved):
Pre-approved as my last pending comment is technically not blocking.

packages/app/src/components/max-width.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard18: I clarified the maxWidth string... Updated it to theme.sizes.maxWidth

And I am definitely sure I have updated those backticks, but do not know why there are still backticks here... I probably adjusted them in another branch...

PR #4589, opened 2023-01-02
task/COR-1321-restructure-and-add-graphs

VWSCoronaDashboard26:
## Summary

* updated `HospitalsAndCarePage` (or 'Ziekenhuizen en de zorg' as it is called on the dashboard) page component to add new covid patient influx graphs for both hospitals and icus;
* do note: this has been developed using a dataset meant for testing purposes; graphs might not represent actual data;
* do note: I have also created two additional timelines in the CMS to be shown for these metrics (`influx_covid_patients` for both `hospital_lcps` and `intensive_care_lcps`);
* do note: earlier changes (merging/moving graphs had been picked up in this branch but was part of a different pull request as to not introduce many conflicts);

#### Before
There is no before scenario for this graph as it is newly introduced.

#### After
COR-1321_after_1

COR-1321_after_2

### Sanity
Sanity keys for the new graphs can be found in the following path inside Sanity studio: `Dashboard Pagina's > Ziekenhuizen en de zorg > nl`. 
Sanity timelines for the new graphs can be found in the following path inside Sanity studio: `Datagerelateerde content > Elements > nl > hospital_lcps - Grafiek OR intensive_care_lcps - Grafiek` (depending on which graph's timeline is relevant). New timelines are called 'Annotaties patiëteninstroom ziekenhuizen' and 'Annotaties patiënteninstroom intensive care'. An example is as below.

image

PR #4559, opened 2022-12-28
Feature/cor 1322 restructure pages

VWSCoronaDashboard19:
## Summary

- Created two new pages named `Ziekenhuizen en de zorg` and `Patienten in beeld`
- Create Sanity keys for the new pages, also delete old ones that are unused
- Restructure pages in the sidebar and delete old structure
- Redirect old URLs to new pages.
- Combine existing graphs

**Before**
Schermafbeelding 2023-01-02 om 10 32 37

**After**
Schermafbeelding 2023-01-02 om 10 31 26

PR #4554, opened 2022-12-22
task/COR-1323-lcps-nice-schema-updates

VWSCoronaDashboard26:
## Summary

* enriched `hospital_lcps` schema with new keys for patient influx
* enriched `intensive_care_lcps` schema with new keys for patient influx
* removed `beds_occupied_non_covid` key from `intensive_care_lcps` schema as it is not actively used
* generated new data types

PR #4596, opened 2023-01-04
feature/COR-1275-hospital-icu-pages

VWSCoronaDashboard26:
## Summary

Most of these changes have already previously been reviewed as individual pull requests because they were picked up as subtasks for the larger COR-1275 feature. However, this feature should still be reviewed accordingly. This feature contains work previously reviewed in the below pull requests:

* https://github.com/minvws/nl-covid19-data-dashboard/pull/4554
* https://github.com/minvws/nl-covid19-data-dashboard/pull/4559
* https://github.com/minvws/nl-covid19-data-dashboard/pull/4589
* https://github.com/minvws/nl-covid19-data-dashboard/pull/4590

This feature contains the following significant changes:

* enriched `hospital_lcps` and `intensive_care_lcps` schemas with new keys for covid patient influx
* generated new data types for both keys and CMS
* created two new pages named 'Ziekenhuizen en de zorg' and 'Patiënten in beeld'
  * created new KPI tiles covering patient influx and occupied beds for both hospitals and ICUs
  * created new graph covering patient influx for both hospitals and ICUs
  * add new Sanity keys
* merged existing graphs from retired 'Ziekenhuisopnames' and 'IC-opnames' pages
  * merged occupied beds graphs for both hospitals and ICUs into single toggle-able graph
  * merged admissions over time graph for both hospitals and ICUs into single toggle-able graph
  * merged admissions per age group over time graph for both hospitals and ICUs into single toggle-able graph
* retired old 'Ziekenhuisopnames' and 'IC-opnames' pages
  * updated the sidebar
  * set up redirects
  * remove old Sanity keys

### Screenshots
More detailed screenshots can be found in the aforementioned pull requests.

#### Before
Before screenshots cover the retired 'Ziekenhuisopnames' and 'IC-opnames' pages.
<details>

COR-1275_before_1
COR-1275_before_2
</details>

#### After
After screenshots cover the new 'Ziekenhuizen en de zorg' and 'Patiënten in beeld' pages.
<details>

COR-1275_after_1
COR-1275_after_2
</details>


packages/app/src/components/chart-tile.tsx
```diff
@@ -9,6 +9,7 @@ import { FullscreenChartTile } from './fullscreen-chart-tile';
 import { Heading } from './typography';
 import { Markdown } from './markdown';
 import { MetadataProps } from './metadata';
+import { ChartTileToggle, ChartTileToggleProps } from './chart-tile-toggle';
 
 interface ChartTileProps {
   title: string;
```
VWSCoronaDashboard28, line 15: Could we group the optional and mandatory props together?

VWSCoronaDashboard28 (approved):
Looks good, one small comment

packages/app/src/components/chart-tile.tsx
```diff
@@ -74,20 +61,19 @@ interface ChartTileHeaderProps {
...
-function ChartTileHeader({
-  title,
-  description,
-  children,
-}: ChartTileHeaderProps) {
+function ChartTileHeader({ title, description, children, toggle }: ChartTileHeaderProps) {
   return (
     <Box spacing={3}>
       {/* padding-right to make sure the title doesn't touch/overlap the full screen button */}
       <Heading level={3} css={css({ pr: asResponsiveArray({ md: 5 }) })}>
```
VWSCoronaDashboard28, line 71: I believe you can get rid of the `asResponsiveArray` here. At least thats what we said in one of the chapter meetings.

packages/app/src/components/chart-tile.tsx
VWSCoronaDashboard26: Good suggestion. All mandatory and optional props are now grouped together and sorted alphabetically (aside from function props, which I personally like to keep at the back of their respective group (mandatory or optional)).

packages/app/src/components/chart-tile.tsx
VWSCoronaDashboard26: Rewritten this to a `ChartTileHeading` component. Also simplified the `Box` component on line 39.

VWSCoronaDashboard28 (approved):
Looking good

VWSCoronaDashboard28 (approved):

VWSCoronaDashboard18 (approved):
Nice! Great job!

packages/app/src/pages/landelijk/ziekenhuizen-en-zorg.tsx
```diff
@@ -0,0 +1,385 @@
...
+  const intensiveCarePatientInfluxUnderReportedRange = getBoundaryDateStartUnix(
+    data.intensive_care_lcps.values,
+    countTrailingNullValues(data.intensive_care_lcps.values, 'influx_covid_patients')
+  );
+
+  const hospitalLastValue = getLastFilledValue(data.hospital_lcps);
+  const icuLastValue = getLastFilledValue(data.intensive_care_lcps);
+
+  const lastInsertionDateOfPage = getLastInsertionDateOfPage(data, pageMetrics);
+  const mostRecentDateUnix = Math.max(hospitalLastValue.date_unix, icuLastValue.date_unix);
```
VWSCoronaDashboard25, line 112: I understand from #4590 [why this is needed](https://github.com/minvws/nl-covid19-data-dashboard/pull/4590#discussion_r1060482392).
Has it in the meanwhile been checked:
- that it theoretically indeed can be two different dates? It might for example be that backend is generating this from the same source and just provide us with the same date twice.
- that if that the case, this solution is accepted by the higher ups? Using `max` over `min` might be misleading, but the same argument can be made for using `min` over `max`. This reasoning can lead to a demand for redesign to prevent misinformation of the public.

I figure that if the dates technically can be different, it's not likely to happen. And if demand for changes come to light, the changes can be added later and will not be a show stopper. So please consider this a non-blocking comment.

packages/app/src/components/chart-tile.tsx
```diff
@@ -55,35 +47,40 @@ export function ChartTile({ title, metadata, description, timeframeInitialValue,
...
           {children}
         </Box>
       )}
     </Box>
   );
-}
+};
+
+const ChartTileHeading = styled(Heading)`
+  // padding-right to make sure the title doesn't touch/overlap the full screen button
```
VWSCoronaDashboard25, line 82: Out of scope and not directed at you, @VWSCoronaDashboard26, because this is just rearranging the code (thanks for that!) so I just want to comment on this way of solving stuff in general.

If someone encounters such a problem like this, try to solve it from the perspective of the full screen button instead. Prevent the possibility of overlap by using same level structuring css, like flexbox.

This solution is sub-optimal because it copied the logic for when the full screen button is visible, resulting in having to change the logic in (at least) two locations if we want to alter the logic for the full screen button.

packages/common/src/data/reverse-router.ts
```diff
@@ -23,6 +23,8 @@ export function getReverseRouter(isMobile: boolean) {
       sterfte: () => `/landelijk/sterfte`,
       ziekenhuisopnames: () => `/landelijk/ziekenhuis-opnames`,
       intensiveCareOpnames: () => `/landelijk/intensive-care-opnames`,
```
VWSCoronaDashboard25, line 25: These two (`ziekenhuisopnames` and `intensiveCareOpnames`) [have been re-added in the last commit](https://github.com/minvws/nl-covid19-data-dashboard/pull/4596/commits/703920683875f7c6ea4c7d4c39f493530b157cb5#diff-5a7dc0de2e0a011d53abf1dc127ef787061b7a2018625c186fef761b59bad24a), but the pages have been deleted.

Might be valuable to check other code changes from the commit(s) where these where removed initially from this file, to be sure nothing else came back.

VWSCoronaDashboard26: I intentionally brought these back as to satisfy the compiler. Essentially, `ziekenhuisopnames` still exists on VR and GM levels, meaning we ought to keep that in here. We _could_ get rid of `intensiveCareOpnames` but I was pretty sure that was still giving me _some_ issues.

VWSCoronaDashboard25: > ziekenhuisopnames still exists on VR and GM levels, meaning we ought to keep that in here

Well no, not here on line 24. Because this section is for `nl`. For VR and GM they are on (new) lines 46 and 59.

VWSCoronaDashboard25: But for the issues part, I think that is something that needs to be investigated further. But we can do that in a new ticket as well.

packages/app/src/components/chart-tile-toggle.tsx
```diff
@@ -0,0 +1,21 @@
+import { Box } from './base';
+import { RadioGroup } from './radio-group';
+
+export type ChartTileToggleItem = {
```
VWSCoronaDashboard21, line 4: Could this be a interface as well?

VWSCoronaDashboard21 (approved):
Looks good! I have some small nitpicky things/questions.

packages/common/src/data/reverse-router.ts
VWSCoronaDashboard26: The compiler starts complaining at `packages/app/src/domain/layout/logic/use-sidebar.tsx:56` when `ziekenhuisopnames` is left out of this file, because `ziekenhuisopnames` is used as a lookup key within that file.

I fully agree that we should resolve this, but due to time constraints I opted to not do this now.

VWSCoronaDashboard25 (approved):
Only non-blocking comments left, so good to go!

PR #4597, opened 2023-01-05
task/COR-1349-Show sanity debugging tool on staging

VWSCoronaDashboard28:
# Summary

The motivation behind this task is to have the sanity debugging tool on the staging environment ASAP. 

- Added a new environment variable for the staging environment URL
- Removed hard coded reference to an old staging URL in `useLokalizeText.tsx`
- Updated the `DockerFile` and the example `env` file


VWSCoronaDashboard18 (approved):
Nice collaboration of the team!

VWSCoronaDashboard26 (approved):
Good one on adding a `_STG` suffix to the URL as well. Makes more sense to use that rather than just `NEXT_HOST_URL`. 👌

VWSCoronaDashboard19 (approved):

PR #4601, opened 2023-01-16
feature/COR-1347-add-new-variant

VWSCoronaDashboard26:
## Summary

* added new XBB.1.5 (XBB_1_5) variant
* added new CH.1.1 (CH_1_1) variant

Due to change of scope (CH.1.1 'suddenly' popping up), I have also included that variant. Basically, this feature now includes the following tickets:

* COR-1347
* COR-1371

#### Before
COR-1347_before

#### After
After screenshots cover a two copied variant data entries and give a visual representation of the the new variants (and their colours) working accordingly. The actual position and colour is to be determined by data once that becomes available.
COR_1347_after

### DIFF
--- a/packages/app/schema/nl/__named_difference.json
+++ b/packages/app/schema/nl/__named_difference.json
@@ -42,6 +42,7 @@
             "BQ_1_1",
             "XBB",
             "XBB_1_5",
+            "CH_1_1",
             "other_table",
             "other_graph"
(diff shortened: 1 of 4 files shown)


VWSCoronaDashboard18 (approved):
Nice job! We can already test it by locally copying a current variant and see how that looks

PR #4618, opened 2023-01-24
Fix: Solved (double) quote issue on generating data types

VWSCoronaDashboard18:
Because we don't want to edit auto-generated files every time


VWSCoronaDashboard28 (approved):
Good stuff!

VWSCoronaDashboard26 (approved):

PR #4622, opened 2023-01-25
Release/2.66.2

VWSCoronaDashboard18:


VWSCoronaDashboard21 (approved):
Only thing I noticed but not blocking, I'll create a ticket for this:

Kwetsbare groepen en 70 plussers > add Divider below PageInformationBlock (like National level)

PR #4623, opened 2023-01-25
Fix/release fixes

VWSCoronaDashboard18:
Fixing warning type and missing sanity keys


VWSCoronaDashboard18 (commented):
The test fails for certain sanity keys. But that is fixed within the CSV file inside this PR.

VWSCoronaDashboard21 (approved):
Only thing I noticed but not blocking, I'll create a ticket for this:

- Kwetsbare groepen en 70 plussers > add Divider below PageInformationBlock (like National level)

PR #4636, opened 2023-01-31
fix/COR-1263-fix-incorrect-dates

VWSCoronaDashboard28:
## Summary

The two adjusted tables on the behavior page (NL and VR) had the wrong date values. This has been fixed in this PR.


VWSCoronaDashboard21 (approved):
LGTM!

PR #4638, opened 2023-01-31
feature/COR-1275-hospital-icu-pages

VWSCoronaDashboard26:
## Summary

* cleaned up unwanted Sanity keys
* adjusted sources for both new pages
  * 'Ziekenhuizen en de zorg' only contains LCPS data (was both)
  * 'Patiënten in beeld' only contains NICE data (was both)
* adjusted icon to be used on the 'Ziekenhuizen en de zorg' page

### Screenshots
Differences might not be too obvious, but basically the sources of both pages weren't completely correct.
#### Before
COR-1275_before_1
COR-1275_before_2

#### After
COR-1275_after_1
COR_1275_after_2


VWSCoronaDashboard28 (approved):
G2G

VWSCoronaDashboard25 (approved):

PR #4641, opened 2023-02-02
feature/COR-1255-create-workflow-for-acceptance-environment

VWSCoronaDashboard25:


.github/workflows/infra-acceptance.yml
```diff
@@ -0,0 +1,19 @@
+name: CI-acceptance
```
VWSCoronaDashboard28, line 1: What happens to the acceptance environment in the time between merging for example `release/2.68` into master and creating `release/2.69`? Or do we accept that the creation of a new release branch is part of the release procedure now? 

If that is the case, I would also like to see the release documentation updated, mentioning the above.

Additionally, what happens when we have `release/2.69` created for the sprint (after merging the previous release), but there is an intermediate release of  release/2.68.1 which suddenly needs to happen. We will then have two `release/*` branches on the remote. How does the environment decide which one will be live? I think this is something that should be considered as well, and its outcome should also be added to the release documentation.

I'm also curious how we would handle `hotfix/...` branches. Does this mean that a `hotfix/...` branch will require us to create a release branch, merge it in there, and then move the `hotfix/...` to master? I remember previous discussions where we said as soon as a `hotfix` branch exists, it will be made live on the acceptance environment. 

So in conclusion, it feels like we need to account for some more scenarios.

VWSCoronaDashboard18: Would you like a hotfix going through the acceptance first??

And if so, you can just add `release/hotfix/...` instead of `hotfix/...`

VWSCoronaDashboard18 (approved):
Great stuff! Hope this helps the process

.github/workflows/infra-acceptance.yml
VWSCoronaDashboard18: I think it's good to know this is for the acceptance environment so that things can be tested more early on as an addition. And if things need to go to staging fast, like an intermediate release or hotfix. Do you want that to go through acceptance? Or directly on staging?

VWSCoronaDashboard28: I do know this is for the acceptance environment. Regarding your first reply, I think `release/hotfix/...` is rather undesirable as we always follow the convention of `release/{version}`.  

To your second comment, I guess a hotfix does not necessarily have to go to the acceptance environment. But then as my comment states, these kinds of things should be documented in the release procedure once we make this change.

VWSCoronaDashboard28 (commented):
Also, if this is merged now, then until COR-1397 is worked on, we have to merge the development branch into the release every time a feature is merged into it, but only after it has been tested and approved on development. Which can be quite cumbersome. 

In the future, when features are merged into dev, and immediately then into the release, I think it could lead to some confusion. For example, if something is merged to dev and the release before it has been tested and it contains something that doesn't work as expected or causes a breaking issue for some reason, the the acceptance environment is impacted as well and we could potentially get feedback from the client that stuff is breaking and that they are unable to test other stuff on acceptance because of it.

So do we need to ensure that only features tested on development and approved make it to acceptance?

.github/workflows/infra-acceptance.yml
VWSCoronaDashboard25: When the pipeline is 'doing a build', our code is used once to create an artefact (the actual build) and that artefact is deployed and hosted on the specific url. So if a build is deployed on acceptance (or any other environment) it will stay there and be accessible until it is removed because of a new deployment.

So after merging and deleting `release/2.68.0`, that build will still be accessible on the acceptance environment. Which is okay, because then that means `master` is updated and the acceptance environment will just be the same as the staging environment, and after approval, the production environment.
When we then create (and push to) `release/2.69.0`, a new build is triggered and the `release/2.69.0` branch will be deployed to the acceptance environment. So if `release/2.69.0` is at that point still a clean copy of `master`, there should be no differences after the 2.69.0 build.

When we have multiple `release/*` branches, each of them can trigger a build. But the build will always be based on the branch that triggered the build.
This means that if branch `release/2.69.0` is created and pushed to, a build is triggered and version 2.69.0 will be deployed to the acceptance environment.
If then branch `release/2.68.1` is created and pushed to, a new build is triggered and version 2.68.1 will be deployed to the acceptance environment.
So it will always build and deploy the branch that triggers the last. But we'll always have the option to manually re-trigger anything, so if by accident 2.69.0 is deployed while we are still working on 2.68.1, we can trigger a new build for 2.68.1 and have that one deployed instead.

For a `hotfix/` you are right. The idea of a hotfix is that the PR can go straight to `master` so it would defeat the purpose if it needs to go to a release-branch first. We're not using that workflow yet, but I can add it of course. Then it will just as well trigger a build and the same goes for that branch as for having two release branches.

VWSCoronaDashboard25: > Would you like a hotfix going through the acceptance first??
> 
> And if so, you can just add `release/hotfix/...` instead of `hotfix/...`

It doesn't need to, but it might be a good idea to check the hotfix in a build similar to production before merging it to `master`. So it would be nice if we can use the acceptance environment for that.
But instead of changing the branch name, to make it start with `release/` we can make the trigger also work for `hotfix/*`.

VWSCoronaDashboard28: Very well explained, thank you! However, I would like to see the documentation updated. Is that something that can be picked up in this branch?

VWSCoronaDashboard25:
> Also, if this is merged now, then until COR-1397 is worked on, we have to merge the development branch into the release every time a feature is merged into it, but only after it has been tested and approved on development. Which can be quite cumbersome.

Actually we can just proceed the way we are working now. That just means that we can't fully utilise the new acceptance environment, because we'd create a release branch just before we intend to merge it to `master`, so a build is done for acceptance parallel of us just going through to `master`.

The main objective of the acceptance environment is to decouple that. So we can already see the build with the production Sanity database and the latest proto's, without having to go through the whole release process and creating release branches up to `x.xx.6`.


> In the future, [...].
> 
> So do we need to ensure that only features tested on development and approved make it to acceptance?

Yes, that is the idea. The acceptance environment should represent what we intend to release. If it would automatically receive the features that are merged to `develop` that means the acceptance environment will always just be a copy of the development environment, because as soon as something is merged to `develop`, a build is triggered to deploy on the development environment, from the develop branch (in another workflow file).

My first idea is to try to automatically create a PR, but not auto-merge it to the release branch as well.

.github/workflows/infra-acceptance.yml
VWSCoronaDashboard25: Yes... :(

VWSCoronaDashboard28 (approved):

VWSCoronaDashboard27 (approved):

PR #4645, opened 2023-02-05
Magic number cleanup part 13

VWSCoronaDashboard18:
A new set of files that were not part of the first 12 initial PRs about the magic numbers.
But when applying the PR feedback, these files had the same issues as the magic number PRs 1 through 12.
So the changes also need to be merged into develop.

### DIFF
--- a/packages/app/src/style/preset.ts
+++ b/packages/app/src/style/preset.ts
@@ -96,7 +96,7 @@ export const preset = {
 
 export const nestedHtml = {
   ...spacingStyle(3),
-  /** padding: inherit p-styles from the container */
+  /** p: inherit p-styles from the container */
 
   h1: preset.typography.h1,
   h2: preset.typography.h2,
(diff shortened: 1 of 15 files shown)


packages/app/src/style/preset.ts
```diff
@@ -96,7 +96,7 @@ export const preset = {
 
 export const nestedHtml = {
   ...spacingStyle(3),
-  /** p: inherit p-styles from the container */
+  /** padding: inherit p-styles from the container */
```
VWSCoronaDashboard26, line 99: You might want te revert this change, again. 😜

VWSCoronaDashboard26 (approved):
Pre-approved, but have a look at my single non-blocking comment.

packages/app/src/style/preset.ts
VWSCoronaDashboard18: Reverted. Too many PRs will solve some issues but creates other issues.

VWSCoronaDashboard27 (approved):

PR #4649, opened 2023-02-07
feature/COR-1349-show-Sanity-debugging-tool-on-acceptance

VWSCoronaDashboard25:
feature(sanity debug tool): show sanity debug tool on develop and acceptance environment

Closes: COR-1349

### DIFF
--- a/packages/app/src/locale/use-lokalize-text.tsx
+++ b/packages/app/src/locale/use-lokalize-text.tsx
@@ -20,9 +20,8 @@ const datasets = ['development', 'production', 'keys'] as const;
 export type Dataset = typeof datasets[number];
 
 const query = `*[_type == 'lokalizeText']`;
-const enableHotReload = process.env.NEXT_PUBLIC_PHASE === 'develop';
-
-export const IS_STAGING_ENV = typeof window !== 'undefined' && window.location.host === process.env.NEXT_HOST_URL_STG;
+const isDevelopBuild = process.env.NEXT_PUBLIC_PHASE === 'develop';
+const isAcceptanceBuild = process.env.NEXT_PUBLIC_PHASE === 'acceptance';
 
 /**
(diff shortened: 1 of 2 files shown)


VWSCoronaDashboard26 (approved):

packages/app/src/locale/use-lokalize-text.tsx
```diff
@@ -38,7 +37,7 @@ export function useLokalizeText(initialLocale: LanguageKey) {
   const [locale, setLocale] = useState(initialLocale);
   const [text, setText] = useState<SiteText>(languages[locale]);
   const lokalizeTextsRef = useRef<SanityDocument<LokalizeText>[]>([]);
-  const showSanityDebugToggle = enableHotReload || IS_STAGING_ENV;
+  const showSanityDebugToggle = isDevelopBuild || isAcceptanceBuild;
```
VWSCoronaDashboard28, line 40: Could/should we also add a condition so that it always shows on our local builds while developing?

VWSCoronaDashboard28 (approved):
Pre-approving as everything looks good. I've left a comment/question though, but it is non-blocking.

packages/app/src/locale/use-lokalize-text.tsx
VWSCoronaDashboard25: You can already if you like. When you run `yarn dev`, you'll see a log line:
```
info  - Loaded env from /.../nl-covid19-data-dashboard/packages/app/.env.local
```

So if you add `NEXT_PUBLIC_PHASE=develop` there, (or `NEXT_PUBLIC_PHASE=acceptance` after this PR), it will trigger this feature. Just tried it and it works but will lead to different problems.

Also note that setting `NEXT_PUBLIC_PHASE` will not result in a build similar to the specific pipeline build. You'd need to set all `.env.local` variables manually to mimic a specific build.

PR #4657, opened 2023-02-13
bugfix/magic-number-bugfixes

VWSCoronaDashboard26:
## Summary

* fixed faulty use of `asResponsiveArray` within Styled Components;

#### Before
image

#### After
image

### DIFF
--- a/packages/app/src/components/anchor-tile.tsx
+++ b/packages/app/src/components/anchor-tile.tsx
@@ -7,7 +7,7 @@ import { Box } from '~/components/base';
 import { asResponsiveArray } from '~/style/utils';
 import { Link } from '~/utils/link';
 import { ExternalLink } from './external-link';
-import { space } from '~/style/theme';
+import { mediaQueries, space } from '~/style/theme';
 
 interface AnchorTileProps {
   title: string;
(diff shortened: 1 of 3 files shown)


packages/app/src/components/anchor-tile.tsx
```diff
@@ -56,11 +56,18 @@ export const IconWrapper = styled.span(
 );
 
 const Container = styled.article`
+  border-top: 2px solid ${colors.gray2};
   display: flex;
-  padding-top: ${asResponsiveArray({ _: space[2], sm: space[3] })};
```
VWSCoronaDashboard28, line 60: 😬

VWSCoronaDashboard28 (dismissed):

VWSCoronaDashboard21 (dismissed):

VWSCoronaDashboard18 (commented):
Screenshot 2023-02-13 at 11 33 04

At least two more files have this issue as well. Would that also be part of this fix? Or should that be done separately?

VWSCoronaDashboard18 (changes requested):
See my other comment... (which didn't need resolving for some reason...)

VWSCoronaDashboard28 (approved):

VWSCoronaDashboard18 (approved):
Nice!

packages/app/src/components/anchor-tile.tsx
```diff
@@ -56,11 +56,18 @@ export const IconWrapper = styled.span(
...
-  padding-bottom: ${asResponsiveArray({ _: space[3], sm: space[4] })};
-  flex-direction: ${asResponsiveArray({ _: 'column', lg: 'row' })};
-  border-top: solid 2px ${colors.gray2};
+  flex-direction: column;
+  padding-block: ${space[2]} ${space[3]};
```
VWSCoronaDashboard25, line 62: Does `padding-block` stay `padding-top` and `padding-bottom` when the `flex-direction` changes?

VWSCoronaDashboard26: Good question! According to the [docs](https://developer.mozilla.org/en-US/docs/Web/CSS/padding-block#description), `padding-block` is only affected by values defined for `writing-mode`, `direction` and `text-orientation`. The same applies for `padding-inline` as can be read [here](https://developer.mozilla.org/en-US/docs/Web/CSS/padding-inline#description).

VWSCoronaDashboard25 (approved):

PR #4658, opened 2023-02-13
Feature/COR-1342-automate-sanity-summary-page-kpi-trend-icons

VWSCoronaDashboard28:
## Summary

This feature introduces new logic which enables the trend icons on the homepage KPI tiles to be rendered based on the value set for the KPI in sanity. For the logic to work, the trend icon fields must be left empty. If they are filled, the logic will be overwritten. 

There are no real before or after screenshots to show here as there are no visual changes on the dashboard.

One thing to note is that there is a new switch in the theme tile document in sanity. See below:
image


packages/app/src/domain/topical/components/topical-kpi-tile/topical-tile.tsx
```diff
@@ -0,0 +1,135 @@
+import { TrendIcon as TrendIconType, TrendIconColor, TrendIconDirection } from '@corona-dashboard/app/src/domain/topical/types';
```
VWSCoronaDashboard28, line 1: This file has been moved, therefore it looks like everything has changed. A lot has indeed changed in this file, so maybe review it as you would a new file.

packages/app/src/domain/topical/components/topical-kpi-tile/topical-tile.tsx
```diff
@@ -0,0 +1,135 @@
...
+    ...(hasManualTrendIconConfig && {
+      ...trendIcon,
+      color: mapStringToColors(trendIcon.color as TrendIconColor),
+      direction: TrendDirection[trendIcon.direction as TrendIconDirection],
+    }),
+  };
+
+  return (
+    <Tile as={cta.href ? 'a' : 'div'} href={cta.href ?? undefined}>
+      <>
```
VWSCoronaDashboard26, line 47: This is not part of your changes, but seeing this file again triggers me into wondering about two things.

1. I do not see any valid reasoning for keeping the Fragment inside this component, as we are not in a map or trying to render multiple components in our return statement.

2. This is completely out of scope but more so a thing to consider. Currently, the `Tile` is rendered as an `<a>` element when an `href` is supplied. However, it would be nice if we could use Next's `<Link>` component for this instead so that we can leverage functionality such as prefetching links in production. Is this something we would still want to pick up within the scope of this feature?

packages/app/src/domain/topical/components/topical-kpi-tile/topical-tile.tsx
```diff
@@ -0,0 +1,135 @@
...
+            padding={{ _: space[3], xs: space[4] }}
+            paddingTop={formattedKpiValue ? { _: space[2] } : undefined}
+          >
+            <RichContent blocks={description} elementAlignment="start" />
+          </Box>
+        </Box>
+
+        {sourceLabel && (
+          <Box padding={{ _: `0 ${space[3]} ${space[3]}`, xs: `0 ${space[4]} ${space[4]}` }}>
+            <InlineText color="gray7">{sourceLabel}</InlineText>
```
VWSCoronaDashboard26, line 85: Very nitpicky, but you could update this to `color={colors.gray7}` if you'd like, especially because all other colours are referred to as such within this file.

packages/app/src/domain/topical/components/topical-kpi-tile/topical-tile.tsx
VWSCoronaDashboard26: Is the motivation to move this file mainly the fact that this pull request also introduces a new `logic` folder for this component? In any case, I would personally be totally fine to put the `logic` folder on the same level as the current `components` folder and keep the folder structure as-is. This approach is also taken for other 'domain' components such as those found in the 'behavior' or 'situations' folders. Is there any particular reason why this setup was chosen over the other?

packages/app/src/domain/topical/index.ts
```diff
@@ -4,7 +4,7 @@ export { TopicalLinksList } from './components/topical-links-list';
 export { TopicalMeasureTile } from './components/topical-measure-tile';
 export { TopicalSectionHeader } from './components/topical-section-header';
 export { TopicalThemeHeader } from './components/topical-theme-header';
-export { TopicalTile } from './components/topical-tile';
+export { TopicalTile } from './components/topical-kpi-tile/topical-tile';
```
VWSCoronaDashboard26, line 7: One could argue that this is a barrel file: something we opt to avoid using. I propose to get rid of this file and update any imports where necessary. I guess that this is not too much of a change, as this is all scoped to the topical domain/page.

packages/app/src/domain/topical/components/topical-kpi-tile/topical-tile.tsx
VWSCoronaDashboard28: There is. While the component is already scoped to the topical domain, I personally think this new logic (which is very tightly coupled to the topical tile) should be scoped to the topical tile, hence the creation of a component specific folder. If this logic was to be used elsewhere within the topical domain, then I agree that what you suggest would be the way to go.

VWSCoronaDashboard28: Having said that, if the preference is to revert the structure back to what it was, and create a higher level logic folder, I'm not against doing that.

packages/app/src/domain/topical/components/topical-kpi-tile/topical-tile.tsx
VWSCoronaDashboard28: 1. Would rather convert the fragment into a `Box`? 
2. I'm not sure if the `Box`'s `as` attribute supports passing in components like `Link`. Would you then want to create some kind of logic to either render the `Link` or the `Tile`? And then the same for their closing tags? - Gives me major `isml` vibes 😅

packages/app/src/domain/topical/components/topical-kpi-tile/topical-tile.tsx
VWSCoronaDashboard28: Good point, I've done this.

packages/app/src/domain/topical/index.ts
VWSCoronaDashboard28: Good point, I've covered this in my latest commit.

packages/app/src/domain/topical/components/topical-kpi-tile/topical-tile.tsx
VWSCoronaDashboard26: 1. For starters, I don't even think we need the Fragment at all. It is already something that is not rendered in the DOM, so converting this to a `Box` would just clutter the DOM with one more extra `<div>` element per tile.
2. This is a bit more complex than using the `Box`'s `as` prop here. An example of something we _could_ look at it as below.

```typescript
const TopicalTileWrapper = ({ children }: { children: React.ReactNode }) =>
  cta.href ? (
    <Link href={cta.href} passHref>
      <Tile as="a">{children}</Tile>
    </Link>
  ) : (
    <Tile>{children}</Tile>
  );
```

It is a bit more complex and I am really not sure why we need the `<Tile as="a">` as child here, but I guess that has something to do with the way our Next.js application is set up or the version we are on.. 🤷🏻 The new `TopicalTileWrapper` would then in turn replace the top-level `Tile` component on line 46.

VWSCoronaDashboard21 (approved):
LGTM!

packages/app/src/domain/topical/components/topical-kpi-tile/topical-tile.tsx
VWSCoronaDashboard28: Appreciate the sharp eye and feedback. I've updated the code to include this.

VWSCoronaDashboard21 (approved):

packages/app/src/domain/topical/components/topical-kpi-tile/topical-tile.tsx
VWSCoronaDashboard26: Completely fair reasoning. I will resolve this conversation.

packages/app/src/domain/topical/components/topical-kpi-tile/topical-tile.tsx
VWSCoronaDashboard26: Great solution! 🙌🏻

VWSCoronaDashboard26 (approved):

PR #4660, opened 2023-02-15
bugfix/COR-1249-image-article-fix

VWSCoronaDashboard26:
## Summary

* temporary revert to basic img element to see if that fixes anything;

## Motivation

To be honest, @VWSCoronaDashboard21 and I are a bit lost on how to further approach the bug mentioned in COR-1249. Aside from the findings documented in the ticket, we have also Dockerized a production build which does not result in the same behaviour as on any of the hosted environments such as development, staging and production. This pull request is a means to see if a standard good old HTML `img` element without any `sizes` or `srcset` attributes resolves anything on the hosted environments. For obvious reasons, this solution should be considered as a temporary fix and is to be reverted when an outcome is clear.

### DIFF
--- a/packages/app/src/components/article-teaser.tsx
+++ b/packages/app/src/components/article-teaser.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable */
 import css from '@styled-system/css';
 import { ReactNode } from 'react';
 import styled from 'styled-components';
@@ -9,6 +10,7 @@ import { Link } from '~/utils/link';
 import { BackgroundImage } from './background-image';
 import { Box } from './base';
 import { Anchor, Heading, Text, BoldText } from './typography';
+import { getImageProps } from '~/lib/sanity';
 
 export type ArticleSummary = Pick<Article, 'title' | 'slug' | 'summary' | 'cover' | 'category' | 'categories'>;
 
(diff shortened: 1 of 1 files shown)


VWSCoronaDashboard28 (approved):

VWSCoronaDashboard18 (approved):
Good luck testing this!

packages/app/src/components/article-teaser.tsx
```diff
@@ -24,12 +26,15 @@ export function ArticleTeaser(props: ArticleTeaserProps) {
...
 
+  const { src } = getImageProps(cover, {});
+
   return (
     <Link passHref href={`/artikelen/${slug}`}>
       <StyledArticleTeaser>
         {cover.asset && (
           <ZoomContainer height="200px">
-            <BackgroundImage image={cover} height="200px" sizes={coverSizes} />
+            {/* <BackgroundImage image={cover} height="200px" sizes={coverSizes} /> */}
```
VWSCoronaDashboard18, line 36: Is this intentional to leave in there? We have this in our Git history. So maybe delete it for production? And put it back when necessary?

VWSCoronaDashboard18: I know before the article fix test it was already commented out. But I think it can be removed just in case.

VWSCoronaDashboard26: I am not entirely sure what to make of these comments on an already merged pull request?

VWSCoronaDashboard18: Ai, wrong PR... I took this one side-by-side with the revert for comparing...

PR #4662, opened 2023-02-15
Feature/COR-1392-text-gm-positive-tests-is-in-bold-should-be-normal-text

VWSCoronaDashboard28:
## Summary

Converted two `BoldText` components to `Markdown`. Specified that only the `infected_per_100k` variable should be bold - achieved by adding markdown to the lokalize key's value.

The ticket was not too clear for me, so if someone has a better understanding, could you please check this out and make sure its working as expected?

I ran into the issue with the lokalize keys where it would not let me update the existing value. So I had to create two new ones with `__renamed` in the name of the key.

Before:
image

After:
image

### DIFF
--- a/packages/app/src/pages/gemeente/[code]/positief-geteste-mensen.tsx
+++ b/packages/app/src/pages/gemeente/[code]/positief-geteste-mensen.tsx
@@ -1,36 +1,34 @@
 import { colors, TimeframeOption, TimeframeOptionsList } from '@corona-dashboard/common';
-import { useState } from 'react';
 import { GgdTesten } from '@corona-dashboard/icons';
 import { GetStaticPropsContext } from 'next';
+import { useState } from 'react';
 import { Box } from '~/components/base';
-import { Text } from '~/components/typography';
-import {
-  ChartTile,
-  DynamicChoropleth,
-  TwoKpiSection,
-  ChoroplethTile,
-  TimeSeriesChart,
-  TileList,
-  InView,
-  CollapsibleContent,
-  KpiTile,
-  KpiValue,
-  Markdown,
-  PageInformationBlock,
-} from '~/components';
+import { ChartTile } from '~/components/chart-tile';
+import { DynamicChoropleth } from '~/components/choropleth';
+import { ChoroplethTile } from '~/components/choropleth-tile';
 import { thresholds } from '~/components/choropleth/logic/thresholds';
-import { Layout, GmLayout } from '~/domain/layout';
+import { CollapsibleContent } from '~/components/collapsible/collapsible-content';
+import { InView } from '~/components/in-view';
+import { KpiTile } from '~/components/kpi-tile';
+import { KpiValue } from '~/components/kpi-value';
+import { Markdown } from '~/components/markdown';
+import { PageInformationBlock } from '~/components/page-information-block';
+import { TileList } from '~/components/tile-list';
+import { TimeSeriesChart } from '~/components/time-series-chart/time-series-chart';
+import { TwoKpiSection } from '~/components/two-kpi-section';
+import { Text } from '~/components/typography';
+import { GmLayout, Layout } from '~/domain/layout';
 import { useIntl } from '~/intl';
 import { Languages, SiteText } from '~/locale';
 import { ElementsQueryResult, getElementsQuery, getTimelineEvents } from '~/queries/get-elements-query';
 import { getArticleParts, getPagePartsQuery } from '~/queries/get-page-parts-query';
 import { createGetStaticProps, StaticProps } from '~/static-props/create-get-static-props';
-import { createGetChoroplethData, createGetContent, getLastGeneratedDate, selectGmData, getLokalizeTexts } from '~/static-props/get-data';
+import { createGetChoroplethData, createGetContent, getLastGeneratedDate, getLokalizeTexts, selectGmData } from '~/static-props/get-data';
 import { filterByRegionMunicipalities } from '~/static-props/utils/filter-by-region-municipalities';
 import { ArticleParts, PagePartQueryResult } from '~/types/cms';
-import { replaceComponentsInText, replaceVariablesInText, getVrForMunicipalityCode, useReverseRouter } from '~/utils';
-import { getLastInsertionDateOfPage } from '~/utils/get-last-insertion-date-of-page';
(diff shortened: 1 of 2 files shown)


packages/cms/src/lokalize/key-mutations.csv
```diff
@@ -9,3 +9,7 @@ timestamp,action,key,document_id,move_to
 2023-02-10T14:27:20.288Z,add,pages.positive_tests_page.shared.tooltip_labels.ggd_tested_total_moving_average__renamed,xjLKnlFrh2WM7l3TeFR9rD,__
 2023-02-10T14:27:20.295Z,delete,pages.positive_tests_page.shared.tooltip_labels.ggd_infected_percentage,jF33EuwumlGuwav2FD3Zd6,__
 2023-02-10T14:27:20.301Z,delete,pages.positive_tests_page.shared.tooltip_labels.ggd_tested_total_moving_average,6s2ymgdY1hbg7zMfb8Z5wD,__
+2023-02-15T15:31:58.572Z,add,pages.positive_tests_page.gm.map_last_value_text__renamed,cQD0A3OGksV3pY6pYGz5LJ,__
+2023-02-15T15:31:59.442Z,add,pages.positive_tests_page.gm.map_safety_region_last_value_text__renamed,xjLKnlFrh2WM7l3TehfRNX,__
+2023-02-15T15:47:31.649Z,delete,pages.positive_tests_page.gm.map_last_value_text,Hpcrbv9k01wdsU0EheZaiI,__
+2023-02-15T15:47:31.651Z,delete,pages.positive_tests_page.gm.map_safety_region_last_value_text,0ztyMU2xVVOl2tFNiaUuSE,__
```
VWSCoronaDashboard18, line 15: Why are new keys used? We can just use the old keys, right?

VWSCoronaDashboard28: Similar to the bugs which have been fixed this past Friday and 2 weeks ago, there are some issues which prevent us from using the same key. When I try to use the existing keys (so removing the ID's and changing the values so the key-mutations pick them up) it says that the keys already exist or already have a value so will not be created again. This issue is most likely caused by a change from camelCase to snake_case for lokalize keys.

VWSCoronaDashboard28: I've reverted this, please resolve is you think this is aright.

packages/cms/src/lokalize/key-mutations.csv
```diff
@@ -9,3 +9,5 @@ timestamp,action,key,document_id,move_to
 2023-02-10T14:27:20.288Z,add,pages.positive_tests_page.shared.tooltip_labels.ggd_tested_total_moving_average__renamed,xjLKnlFrh2WM7l3TeFR9rD,__
 2023-02-10T14:27:20.295Z,delete,pages.positive_tests_page.shared.tooltip_labels.ggd_infected_percentage,jF33EuwumlGuwav2FD3Zd6,__
 2023-02-10T14:27:20.301Z,delete,pages.positive_tests_page.shared.tooltip_labels.ggd_tested_total_moving_average,6s2ymgdY1hbg7zMfb8Z5wD,__
+2023-02-16T10:12:52.969Z,delete,pages.positive_tests_page.gm.map_last_value_text__renamed,cQD0A3OGksV3pY6pYGz5LJ,__
+2023-02-16T10:12:52.970Z,delete,pages.positive_tests_page.gm.map_safety_region_last_value_text__renamed,xjLKnlFrh2WM7l3TehfRNX,__
```
VWSCoronaDashboard18, line 13: This doesn't need removal. As they are not on production.

packages/cms/src/lokalize/key-mutations.csv
VWSCoronaDashboard28: I've removed the file

packages/app/src/pages/gemeente/[code]/positief-geteste-mensen.tsx
```diff
@@ -3,7 +3,7 @@ import { useState } from 'react';
 import { GgdTesten } from '@corona-dashboard/icons';
 import { GetStaticPropsContext } from 'next';
 import { Box } from '~/components/base';
-import { Text, InlineText, BoldText } from '~/components/typography';
+import { Text } from '~/components/typography';
 import {
```
Comment (VWSCoronaDashboard18) [inline packages/app/src/pages/gemeente/[code]/positief-geteste-mensen.tsx:7]: We opted not to use barrel files anymore. And since you are using the Markdown component. I think we should update this, according to our previous decisions.

VWSCoronaDashboard26 (dismissed):

packages/cms/src/lokalize/key-mutations.csv
```diff
@@ -9,3 +9,5 @@ timestamp,action,key,document_id,move_to
 2023-02-10T14:27:20.288Z,add,pages.positive_tests_page.shared.tooltip_labels.ggd_tested_total_moving_average__renamed,xjLKnlFrh2WM7l3TeFR9rD,__
 2023-02-10T14:27:20.295Z,delete,pages.positive_tests_page.shared.tooltip_labels.ggd_infected_percentage,jF33EuwumlGuwav2FD3Zd6,__
 2023-02-10T14:27:20.301Z,delete,pages.positive_tests_page.shared.tooltip_labels.ggd_tested_total_moving_average,6s2ymgdY1hbg7zMfb8Z5wD,__
+2023-02-17T09:24:14.946Z,delete,pages.positive_tests_page.gm.map_last_value_text__renamed,cQD0A3OGksV3pY6pYGz5LJ,__
+2023-02-17T09:24:14.948Z,delete,pages.positive_tests_page.gm.map_safety_region_last_value_text__renamed,xjLKnlFrh2WM7l3TehfRNX,__
```
VWSCoronaDashboard18, line 13: Please, do remove this, as this is not yet added to production and this will have the pipeline try deleting a non-existing key.

packages/app/src/pages/gemeente/[code]/positief-geteste-mensen.tsx
Comment (VWSCoronaDashboard18) [reply packages/app/src/pages/gemeente/[code]/positief-geteste-mensen.tsx]: Nice, it's solved!

packages/cms/src/lokalize/key-mutations.csv
VWSCoronaDashboard28: Done done done

VWSCoronaDashboard18: Nice!

VWSCoronaDashboard26 (approved):

PR #4664, opened 2023-02-20
Feature/cor 1336 add kpi value as variable to description

VWSCoronaDashboard18:
Added the possibility to add variables to the rich content block.
This is made possible by adding a new type of annotation. When the annotation is applied to the selected text that part of the text is used as the variable name.

**_Before_**
<details>
  <summary>Toggle before screenshot</summary> 
Screenshot 2023-02-21 at 15 19 47
</details>

**_After_**
<details>
  <summary>Toggle after screenshot</summary> 
Screenshot 2023-02-21 at 17 12 42
Screenshot 2023-02-21 at 17 16 59
</details>

### DIFF
--- a/packages/app/src/components/cms/rich-content.tsx
+++ b/packages/app/src/components/cms/rich-content.tsx
@@ -95,7 +95,6 @@ export function RichContent({ contentWrapper, blocks, imageSizes, elementAlignme
         return <ContentWrapper>{PortableText.defaultSerializers.types.block(props)}</ContentWrapper>;
       },
       image: (props: { node: ImageBlock | RichContentImageBlock }) => <ContentImage contentWrapper={contentWrapper} sizes={imageSizes} {...props} />,
-
       inlineCollapsible: (props: { node: InlineCollapsibleList }) => {
         if (!props.node.content.inlineBlockContent) return null;
 
@@ -227,18 +226,25 @@ export function RichContent({ contentWrapper, blocks, imageSizes, elementAlignme
     marks: {
       inlineAttachment: InlineAttachmentMark,
       link: InlineLinkMark,
-      richContentVariable: (props: { children: [] }) => {
+      richContentVariable: (props: { children: string[] }) => {
         const { children } = props;
         if (!children) {
           return <>{children}</>;
         }
-        return children.map((child: string, index: number) => {
-          if (child === '' || !variableValue) {
-            return <>{child}</>;
-          }
-          const replacedText = replaceVariablesInText(child as string, { kpiValue: variableValue as string });
-          return <InlineText key={index}>{replacedText}</InlineText>;
-        });
+        if (children[0] === '' || !variableValue) {
+          return <>{children[0]}</>;
+        }
+        return (
+          <InlineText>
+            {children.map((child: string) => {
+              if (child === '' || !variableValue) {
+                return <>{child}</>;
+              }
+              const replacedText = replaceVariablesInText(child as string, { kpiValue: variableValue as string });
+              return replacedText;
+            })}
+          </InlineText>
+        );
       },
     },
   };
--- a/packages/app/src/components/cms/rich-content.tsx
+++ b/packages/app/src/components/cms/rich-content.tsx
@@ -228,20 +228,16 @@ export function RichContent({ contentWrapper, blocks, imageSizes, elementAlignme
       link: InlineLinkMark,
       richContentVariable: (props: { children: string[] }) => {
         const { children } = props;
-        if (!children) {
+        if (!children || !variableValue) {
           return <>{children}</>;
-        }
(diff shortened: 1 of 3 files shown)


packages/app/src/components/cms/rich-content.tsx
```diff
@@ -225,6 +226,26 @@ export function RichContent({ contentWrapper, blocks, imageSizes, elementAlignme
...
+            {children.map((child: string) => {
+              if (child === '' || !variableValue) {
+                return <>{child}</>;
+              }
+              const replacedText = replaceVariablesInText(child as string, { kpiValue: variableValue as string });
+              return replacedText;
+            })}
+          </InlineText>
+        );
+      },
```
VWSCoronaDashboard26, line 248: I have some general questions about this code as I do not fully understand the approach taken and can not locally test the changes due to an error I run into in this branch.

1. Line 231 checks if there are no `children`, and then returns `children`. This seems contradictory? 
2. Line 234 checks if the first child of `children` is an empty string. What makes this check necessary (for only the first child)? 
3. Line 243 uses `kpiValue` as a `variableValue` to be replaced in the text, implying this functionality is tightly coupled to the _one_ component (`packages/app/src/domain/topical/components/topical-kpi-tile/topical-tile.tsx`) that is supposed to be using this change. Is this desirable?

packages/cms/src/schemas/objects/block-fields.ts
```diff
@@ -48,6 +48,19 @@ export const blockFields = [
...
+          title: 'Rich content variable',
+          fields: [
+            {
+              name: 'Variable naam',
+              type: 'string',
+              title: 'variableName',
+              hidden: true,
+            },
+          ],
+        },
```
VWSCoronaDashboard26, line 63: I personally have some questions about this new `richContentVariable` annotation for rich content. Basically, I wonder what the relative gain is of adding this functionality instead of 'simply' supporting the use of variables (following `{{variableName}}` format) inside rich content?

If I understood and tested everything correctly, content managers can use this new `richContentVariable` to select a text and 'convert' that to a variable, yet they still have to include the `{{` and `}}` manually, meaning it will not be a variable until the 'variable distinguishers' are also added to the text?

VWSCoronaDashboard26 (commented):
I have some comments/questions about this implementation. Aside from that, I am not able to test this locally.

image

packages/cms/src/schemas/objects/block-fields.ts
VWSCoronaDashboard18: For now, indeed this is only a solution for the KpiValue. And the '{{' & '}}' are included because of their similarity to the current way of using variables. This is more recognizable to the user.

There is a ticket COR-1338, which is an extended research ticket on this subject. I can elaborate more about that ticket. This would have an object lookup which gets used in Sanity and our frontend code. And where the (for now hidden) string gets replaced by a selection of the available 'in-text'-variables

VWSCoronaDashboard18: If this clears it up or the current solution has not te be changed, please resolve this comment.

packages/app/src/components/cms/rich-content.tsx
VWSCoronaDashboard18: For points 2 & 3: I've updated the logic.
There seemed to be some duplicate checks.

Regarding point 3.
This is for now a single request. It's built in a way that it can extend more easily to other variables as well.

VWSCoronaDashboard18: If you have any questions left, please reply. Otherwise could you resolve this comment?

packages/app/src/components/cms/rich-content.tsx
```diff
@@ -225,6 +226,26 @@ export function RichContent({ contentWrapper, blocks, imageSizes, elementAlignme
     marks: {
       inlineAttachment: InlineAttachmentMark,
       link: InlineLinkMark,
+      richContentVariable: (props: { children: string[] }) => {
+        const { children } = props;
+        if (!children) {
+          return <>{children}</>;
```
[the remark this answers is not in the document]
VWSCoronaDashboard18: As you mentioned in the pair-programming sessions this is solved now

packages/app/src/components/cms/rich-content.tsx
VWSCoronaDashboard18: As you mentioned in the pair-programming sessions this is solved now

packages/cms/src/schemas/objects/block-fields.ts
VWSCoronaDashboard18: As you mentioned in the pair-programming sessions this is solved now

VWSCoronaDashboard28 (approved):

PR #4665, opened 2023-02-22
bugfix/COR-1232-redirect-3-articles-to-all-articles-page

VWSCoronaDashboard28:
## Summary

Based on the discussion in the pre-release call yesterday, this solution aims to redirect the 3 articles mentioned in COR-1232. The initial idea was to create a more dynamic solution which would redirect any unpublished articles to the all articles page. The ticket for that is COR-1379. However, as that is still in progress, for the sake of the release, I have hardcoded redirects for the three articles in question in COR-1232. I have also moved COR-1232 back from DONE to REVIEW.

There are no screenshots to show for this change.

### DIFF
--- a/packages/app/src/next-config/redirects/redirects.js
+++ b/packages/app/src/next-config/redirects/redirects.js
@@ -111,6 +111,22 @@ async function redirects() {
       destination: `/landelijk/ziekenhuizen-en-zorg`,
       permanent: true,
     },
+    // Redirects for unpublished articles - COR-1232
+    {
+      source: '/artikelen/wat-betekent-de-britse-variant-voor-nederland',
+      destination: '/artikelen',
+      permanent: true,
+    },
+    {
+      source: '/artikelen/hoe-weten-we-hoeveel-besmettelijke-mensen-er-zijn',
+      destination: '/artikelen',
+      permanent: true,
+    },
+    {
+      source: '/artikelen/waarom-mogelijke-situaties-van-besmetting-niet-altijd-te-zien-zijn-op-het-dashboard',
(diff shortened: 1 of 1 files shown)


VWSCoronaDashboard21 (approved):

packages/app/src/next-config/redirects/redirects.js
```diff
@@ -111,6 +111,22 @@ async function redirects() {
...
+    {
+      source: '/artikelen/hoe-weten-we-hoeveel-besmettelijke-mensen-er-zijn',
+      destination: '/artikelen',
+      permanent: true,
+    },
+    {
+      source: '/artikelen/waarom-mogelijke-situaties-van-besmetting-niet-altijd-te-zien-zijn-op-het-dashboard',
+      destination: '/artikelen',
+      permanent: true,
+    },
```
VWSCoronaDashboard26, line 129: I see no mention of the article named 'Waarom worden de risiconiveaus niet meer getoond op het Coronadashboard?', which is mentioned in the ticket for this feature. Looking in the CMS, I get the feeling that the article was never published in the first place and therefore a redirect for this article would be unnecessary. However, this is just a feeling. Is this indeed the case?

VWSCoronaDashboard28: Good one! I used the list from the comments and this one was not mentioned there. Upon checking it out, I suspect your feeling may be correct. When inspecting the article in the production suite, it looks like it was never published.

Also, if you read the comment thread in the ticket (for which I should have updated the description) you'll see that I mentioned the 3 articles which can now be unpublished. And communications seems to be inline with those.

PR #4670, opened 2023-02-24
COR-1312 make link optional and remove right side

VWSCoronaDashboard18:
## Summary

* 
* 
* 

#### Before
<details>

Screenshot 2023-02-24 at 11 41 41

</details>

#### After
<details>

Screenshot 2023-02-24 at 11 41 29

</details>

### DIFF
--- a/packages/app/src/components/anchor-tile.tsx
+++ b/packages/app/src/components/anchor-tile.tsx
@@ -7,7 +7,7 @@ import { Box } from '~/components/base';
 import { asResponsiveArray } from '~/style/utils';
 import { Link } from '~/utils/link';
 import { ExternalLink } from './external-link';
-import { mediaQueries, space } from '~/style/theme';
+import theme, { mediaQueries, space } from '~/style/theme';
 
 interface AnchorTileProps {
   title: string;
@@ -76,6 +76,7 @@ const Content = styled.div(
   css({
     flexGrow: 1,
     flex: '1 1 70%',
+    maxWidth: theme.sizes.maxWidthText,
   })
 );
 
--- a/packages/app/src/components/anchor-tile.tsx
+++ b/packages/app/src/components/anchor-tile.tsx
@@ -26,24 +26,26 @@ export function AnchorTile({ title, href, label, children, external = false }: A
         {children}
       </Content>
 
-      <LinkContainer>
-        {external ? (
-          <ExternalLink href={href}>
-            <Box display="flex" alignItems="center">
-              <IconWrapper>
-                <ExternalLinkIcon />
-              </IconWrapper>
-              {label}
-            </Box>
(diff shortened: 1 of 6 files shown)


packages/app/src/components/anchor-tile.tsx
```diff
@@ -7,7 +7,7 @@ import { Box } from '~/components/base';
 import { asResponsiveArray } from '~/style/utils';
 import { Link } from '~/utils/link';
 import { ExternalLink } from './external-link';
-import { mediaQueries, space } from '~/style/theme';
+import theme, { mediaQueries, space } from '~/style/theme';
```
VWSCoronaDashboard28, line 10: We only need `sizes` from the theme right? Perhaps consider exporting `sizes` so we can import only what we need and not everything from `theme`.

packages/app/src/components/anchor-tile.tsx
```diff
@@ -74,6 +76,7 @@ const Content = styled.div(
   css({
```
VWSCoronaDashboard28, line 76: Do you want to convert this to match the component above? So using back ticks and not the CSS function? This comment applies to other instances of this as well. If you do this, you will have to refactor `asResponsiveArray` to `mediaQueries`. You can then get rid of two imports.

VWSCoronaDashboard18: Maybe wait until a decision is made on the backticks or not? Otherwise, we have to revert everything again...

Please resolve.

VWSCoronaDashboard28: There is a decision to be made?

packages/app/src/components/anchor-tile.tsx
VWSCoronaDashboard28: Not really, no. I exported `mediaQueries`  out of the `theme` instead of importing a bunch of stuff that is not needed. And even if it is out of scope, it's only going to create an improvement and has no negative impact.

packages/app/src/components/anchor-tile.tsx
VWSCoronaDashboard18: See the Frontend Chapter meetings.

packages/app/src/components/anchor-tile.tsx
VWSCoronaDashboard18: It's adjusted accordingly, please resolve it's ok now.

packages/app/src/components/anchor-tile.tsx
VWSCoronaDashboard18: It's adjusted accordingly, please resolve it's ok now.

VWSCoronaDashboard28 (approved):
🙌

PR #4671, opened 2023-02-24
Solve-sanity-custom-input-element

VWSCoronaDashboard18:


packages/cms/src/components/topical-tile-date.tsx
```diff
@@ -26,9 +27,7 @@ const ShowDate = (props: ShowDateProps) => {
   const formattedDate = !isConfigValid ? '' : getTopicalTileDateConfig({ config: dateConfig, inputDate: new Date(), language: type.title });
 
   // The following line makes it possible to do realtime updates on this field with the new date configurations
-  const initFormattedDateValue = set(formattedDate);
-  onChange(PatchEvent.from(initFormattedDateValue));
-  const [value, setValue] = useState('');
+  const [value, setValue] = useState(props.value);
```
VWSCoronaDashboard28, line 30: Good one, I assume now it will always initialise with the value stored in state?

VWSCoronaDashboard28 (approved):

packages/cms/src/components/topical-tile-date.tsx
VWSCoronaDashboard18: Yes! Correct! You can test it on the local dev instance as well...

Please resolve.

PR #4678, opened 2023-02-27
fix: PR feedback

VWSCoronaDashboard18:
changed `styled.a` to `styled(Anchor)`

Because of feedback on already merged PR.

### DIFF
--- a/packages/app/src/components/anchor-tile.tsx
+++ b/packages/app/src/components/anchor-tile.tsx
@@ -37,9 +37,7 @@ export function AnchorTile({ title, href, label, children, external = false }: A
             </ExternalLink>
           ) : (
             <Link href={href} passHref>
-              <Anchor display="flex">
-                <span>{label}</span>
-              </Anchor>
+              <Anchor>{label}</Anchor>
             </Link>
           )}
         </LinkContainer>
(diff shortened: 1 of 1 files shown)


packages/app/src/components/anchor-tile.tsx
```diff
@@ -80,7 +80,7 @@ const Content = styled.div`
   max-width: ${sizes.maxWidthText}px;
 `;
 
-const StyledAnchor = styled.a<AnchorProps>`
+const StyledAnchor = styled(Anchor)<AnchorProps>`
```
VWSCoronaDashboard26, line 83: You're not actively using `AnchorProps` in this file, so you can omit that.

Or, instead, have a look at the definition of `AnchorProps` and notice the `display` property already being on there. That means that you could even entirely omit this `StyledAnchor` component and use `<Anchor display="flex">` directly.

VWSCoronaDashboard18: You are totally correct! I've adjusted it!

Please resolve if it is to your liking

VWSCoronaDashboard26 (approved):

VWSCoronaDashboard26 (approved):

VWSCoronaDashboard28 (approved):

VWSCoronaDashboard27 (approved):

PR #4680, opened 2023-02-28
feature/COR-1435-hospitals-page-patient-influx-over-time-filter-null-values

VWSCoronaDashboard26:
## Summary

* added `countLeadingNullValues` utility and tests;
* added `trimLeadingNullValues` utility and tests;
* refactored `trimNullValues` to `trimTrailingNullValues`;
* implemented usage of new utilities on the 'Ziekenhuizen en de zorg' page;
* implemented small quality of life fix (persist state when toggling between the graphs on this page);

### Screenshots
The screenshot only show the  'Ziekenhuisinstroom' (or hospital scope) of the 'Instroom van covid-patiënten door de tijd heen' graph. However, it should be noted that this change is also applied for the 'Intensive care-instroom' (or ICU scope).
#### Before
<details>

COR-1435_before

</details>

#### After
<details>

COR-1435_after

</details>


packages/app/src/utils/count-leading-null-values.ts
```diff
@@ -0,0 +1,13 @@
+import { findIndex } from 'lodash';
```
VWSCoronaDashboard28, line 1: Why not use `findIndex` which is available in JS? Not needed to import it from `lodash` I would say

packages/app/src/utils/count-leading-null-values.ts
```diff
@@ -0,0 +1,13 @@
...
+import { isFilled, isPresent } from 'ts-is-present';
+
+/**
+ * Return the number of leading null values from the given value array.
+ * If a property name is specified it is used to check each value item's
+ * property for null.
+ */
+export const countLeadingNullValues = <T>(values: T[], property?: keyof T) => {
+  const predicate = isPresent(property) ? (d: T) => isFilled(d[property]) : (d: T) => isFilled(d);
```
VWSCoronaDashboard28, line 10: Why is this called predicate? I'm not really sure what to understand by that variable name.

I am also not too big a fan of `(d: T)` what is `d` supposed to be?

packages/app/src/utils/count-trailing-null-values.ts
```diff
@@ -6,10 +6,8 @@ import { isFilled, isPresent } from 'ts-is-present';
  * If a property name is specified it is used to check each value item's
  * property for null.
  */
-export function countTrailingNullValues<T>(values: T[], property?: keyof T) {
-  const predicate = isPresent(property)
-    ? (d: T) => isFilled(d[property])
-    : (d: T) => isFilled(d);
+export const countTrailingNullValues = <T>(values: T[], property?: keyof T) => {
```
VWSCoronaDashboard28, line 9: Same comments here as made for `count-leading-bull-values`

packages/app/src/utils/trim-leading-null-values.ts
```diff
@@ -0,0 +1,16 @@
...
+ * checked for null on each item.
+ *
+ * @param values The given array
+ * @param property Optional property to check
+ * @returns The existing array without the leading null values
+ */
+export const trimLeadingNullValues = <T>(values: T[], property?: keyof T) => {
+  const numberOfLeadingNulls = countLeadingNullValues(values, property);
+
+  return values.filter((value) => !(values.indexOf(value) < numberOfLeadingNulls));
```
VWSCoronaDashboard28, line 15: I'm a little confused here, so `countLeadingNullValues` returns the number of null values at the beginning of an array so for example if i returns 20, that means the first 20 items of the array contain null values? 

Wouldn't `values.slice(numberOfLeadingValues)` suffice?

packages/app/src/utils/trim-leading-null-values.ts
```diff
@@ -0,0 +1,16 @@
+import { countLeadingNullValues } from './count-leading-null-values';
+
+/**
+ * Trims all the leading null values from the given array.
+ * When the property parameter is defined, this property will be
+ * checked for null on each item.
+ *
+ * @param values The given array
```
VWSCoronaDashboard28, line 8: Could you perhaps change this @param values The given array to @param values - The given array? Same for the param below

packages/app/src/utils/trim-trailing-null-values.ts
```diff
@@ -1,16 +1,16 @@
 import { countTrailingNullValues } from './count-trailing-null-values';
 
 /**
- * Right trims all of the null values from the given array.
- * When the property parameter is defined this property will be
+ * Trims all the trailing null values from the given array.
+ * When the property parameter is defined, this property will be
  * checked for null on each item.
  *
  * @param values The given array
```
VWSCoronaDashboard28, line 8: Could you perhaps change this @param values The given array to @param values - The given array? Same for the param below

packages/app/src/utils/count-leading-null-values.ts
VWSCoronaDashboard26: It is called `predicate` because its value is used to affirm/assert if the passed in value meets what is expected.

I do agree on `d` being very cryptic, and seeing as it is an instance of `T`, one can conclude that a more logical name for it would be `value`. I have updated the scripts accordingly, so feel free to resolve this if you agree.

packages/app/src/utils/trim-leading-null-values.ts
VWSCoronaDashboard26: Exactly: `countLeadingNullValues` will find the first index of a filled value (a value that is asserted through the predicate). Anything before that does not meet the predicate and is therefore a `null` value.

I am 100% sure I had tried `values.slice(numberOfLeadingValues)` earlier, but that did not seem to work at the time. However, it works fine now so I have updated this script accordingly, so feel free to resolve this if you agree.

packages/app/src/utils/trim-leading-null-values.ts
VWSCoronaDashboard26: Good one! Applied this and added more docs to the other utils. Feel free to resolve this if you agree.

packages/app/src/utils/trim-trailing-null-values.ts
VWSCoronaDashboard26: Good one! Applied this and added more docs to the other utils. Feel free to resolve this if you agree.

VWSCoronaDashboard28 (approved):

packages/app/src/utils/count-trailing-null-values.ts
VWSCoronaDashboard26: See other comments. Feel free to resolve this if you agree.

packages/app/src/utils/trim-leading-null-values.ts
```diff
@@ -0,0 +1,16 @@
...
+ * @param property Optional property to check
+ * @returns The existing array without the leading null values
+ */
+export const trimLeadingNullValues = <T>(values: T[], property?: keyof T) => {
+  const numberOfLeadingNulls = countLeadingNullValues(values, property);
```
VWSCoronaDashboard18, line 13: Is there a reason you created two new functions? Instead of using the existing one with a double reverse for example?

VWSCoronaDashboard18:
Nice that you did this one too!
- implemented usage of new utilities on the 'Ziekenhuizen en de zorg' page;

VWSCoronaDashboard21 (dismissed):
Really like the persist state when toggling between the graphs on this page! :clap:

packages/app/src/utils/trim-leading-null-values.ts
VWSCoronaDashboard26: I personally think this approach allows for more granular 'plug-n-play' usage of either utility without having to extend the existing functionality or having to worry about figuring out how the functionality works when unaware of the 'directional' argument (either leading or trailing) you'd have to pass into the method. Basically, all functions are kept to a minimum/simpler version this way.

VWSCoronaDashboard18: Good argument! Thanks for elaborating. I thought it was weird to have both functions as the original one is not being used anywhere. Therefore I thought it would make sense to just extend that functionality or replace it.

packages/app/src/utils/count-leading-null-values.ts
```diff
@@ -0,0 +1,16 @@
...
+ * Return the number of leading null values from the given value array.
+ * If a property name is specified it is used to check each value item's
+ * property for null.
+ *
+ * @param values - The given array
+ * @param property - Optional property to check
+ * @returns The index of the last element in the array meeting the predicate
+ */
+export const countLeadingNullValues = <T>(values: T[], property?: keyof T) => {
+  const predicate = isPresent(property) ? (value: T) => isFilled(value[property]) : (value: T) => isFilled(value);
```
VWSCoronaDashboard18, line 13: You can maybe shorten this further by having one function and do the isPresent check inside that single function...

Because now you have two `(value: T) => {}` functions

VWSCoronaDashboard18: That would look something like this:

`const predicate = (value: T) => (isPresent(property) ? isFilled(value[property]) : isFilled(value));`

VWSCoronaDashboard18: Feel free to resolve it either way, depending on your perspective.

packages/app/src/utils/__tests__/trim-leading-null-values.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+
+TrimLeadingNullValues('should trim null values from an array of objects', () => {
+  const data = [{ test: null }, { test: null }, { test: 1 }, { test: 2 }];
+  assert.is(trimLeadingNullValues(data, 'test').length, 2);
+});
+
+TrimLeadingNullValues('should trim null values from an array of values', () => {
+  const data = [null, null, 1, 2];
+  assert.is(trimLeadingNullValues(data).length, 2);
+});
```
VWSCoronaDashboard18, line 15: Would it be nice to extend this test by checking if it only removes the leading values? And not deletes other null values?

Maybe also include a test with no null values at all?

Feel free to resolve this comment if it would take to much time or is out of scope.

packages/app/src/utils/__tests__/count-leading-null-values.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+
+CountLeadingNullValues('should count the amount of leading null values in an array of values', () => {
+  const data = [{ test: null }, { test: null }, { test: 1 }, { test: 2 }];
+  assert.is(countLeadingNullValues(data, 'test'), 2);
+});
+
+CountLeadingNullValues('should count the amount of leading null values in an array of objects', () => {
+  const data = [null, null, 1, 2];
+  assert.is(countLeadingNullValues(data), 2);
+});
```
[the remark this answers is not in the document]
VWSCoronaDashboard27: I think it is a good suggestion.

VWSCoronaDashboard27:
Looks good for me.

VWSCoronaDashboard27 (dismissed):

packages/app/src/utils/count-leading-null-values.ts
VWSCoronaDashboard26: Great idea! However, I've taken this one step further. Please see if you like the solution and resolve accordingly.

packages/app/src/utils/__tests__/count-leading-null-values.spec.ts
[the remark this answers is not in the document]
VWSCoronaDashboard26: I've updated the test cases for this file and the other utilities. I personally do think this covers some nice and sufficient scenarios, so feel free to resolve this conversation if you agree.

packages/app/src/utils/count-leading-null-values.ts
VWSCoronaDashboard18: Nice! Great imrpovement!

packages/app/src/utils/__tests__/count-leading-null-values.spec.ts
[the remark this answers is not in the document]
VWSCoronaDashboard18: Nice ones!

packages/app/src/utils/__tests__/trim-leading-null-values.spec.ts
VWSCoronaDashboard18: Nice ones!

VWSCoronaDashboard18 (approved):
Looks wonderfull!

VWSCoronaDashboard28 (approved):

PR #4682, opened 2023-03-08
feature/COR-1455-remove-topical-page-measure-theme-components

VWSCoronaDashboard26:
## Summary

* removed measure themes section from topical page;
* removed all measure theme component schemas from Sanity;
* updated topical page groq query and query types;
* updated theme schema to mark tiles and other properties as not-required, allowing for 'simple' themes without tiles or links;
* reverted removal of `sizes` export from theme, as this resulted in breaking changes;

### Motivation
I opted to completely remove the measure theme (component) schemas as alter the existing (base) theme schema instead. This means that a base theme can now contain tiles but these are no longer required. This essentially simplifies the workflow within the CMS as can be seen in the attached screenshots.

#### Before
<details>

image

COR-1455_before
</details>

#### After
<details>

image

COR-1455_after
</details>


VWSCoronaDashboard18:
Screenshot 2023-03-08 at 10 48 07

Would it be nice to remove the icon as well? Or is that out of scope!
Feel free to resolve this nontheless

VWSCoronaDashboard18 (approved):
Nice adjustment! And great use of the Sanity themes!

VWSCoronaDashboard26:
> Would it be nice to remove the icon as well? Or is that out of scope? Feel free to resolve this nonetheless.

I have thought about this as well, though I think it would be nice to keep this around. The 'tricky' thing about our icons is that they use a name that _implies_ their use case, but that does not necessarily mean that _should_ always be used in that specific use case. As per example, this specific `maatregelen` icon is still being used in the remaining component (see screenshots) and the sidebar.

VWSCoronaDashboard28 (approved):
Looks good!

PR #4693, opened 2023-03-21
feat(variants): added XBB (Omicron) variants 1.9, 1.5.1, 2.3, 1.9.1, 1.6 and XBF

VWSCoronaDashboard25:
Closes: COR-1523

### DIFF
--- a/packages/app/schema/nl/__named_difference.json
+++ b/packages/app/schema/nl/__named_difference.json
@@ -43,6 +43,12 @@
             "XBB",
             "XBB_1_5",
             "CH_1_1",
+            "XBB_1_9",
+            "XBB_1_5_1",
+            "XBB_2_3",
+            "XBB_1_9_1",
+            "XBB_1_6",
+            "XBF",
             "other_table",
             "other_graph"
           ]
--- a/packages/app/schema/nl/variants.json
+++ b/packages/app/schema/nl/variants.json
@@ -45,6 +45,12 @@
             "XBB",
             "XBB_1_5",
             "CH_1_1",
+            "XBB_1_9",
+            "XBB_1_5_1",
+            "XBB_2_3",
+            "XBB_1_9_1",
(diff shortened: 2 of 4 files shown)


packages/app/schema/nl/__named_difference.json
```diff
@@ -43,6 +43,12 @@
             "XBB",
             "XBB_1_5",
             "CH_1_1",
+            "XBB_1_9",
+            "XBB_1_5_1",
+            "XBB_2_3",
+            "XBB_1_9_1",
+            "XBB_1_6",
```
VWSCoronaDashboard26, line 50: Small typo: the variant to be added is `XBB.1.16`, not `XBB.1.6`. This comment also applies for the other files changed in this pull request.

VWSCoronaDashboard26 (commented):
Nitpicky comment and this can totally be ignored if you're not down for it, but I think it would be nice to group different (sub-)variant together (XBB all grouped together, for instance). 

Generally the code looks fine, but do be aware of the incorrect variant being added.

packages/app/schema/nl/__named_difference.json
VWSCoronaDashboard25: Aaah crêpe

VWSCoronaDashboard25:
> @VWSCoronaDashboard27 said:
> I think it would be nice to group different (sub-)variant together (XBB all grouped together, for instance).

I was thinking that, because I noticed it wasn't in any order. And looking in it further made me realise the order is the order of adding the variances, because it seems we're always adding them at the end. So I thought to respect that order and keep that information in the list.

packages/app/schema/nl/__named_difference.json
VWSCoronaDashboard25: Fixed!

VWSCoronaDashboard28 (approved):

PR #4695, opened 2023-03-21
Feature/COR-1469-add-source

VWSCoronaDashboard21:
## Summary

* Added the source textline on the Vaccination page 
* On NL and GM level.

#### Before
<details>
Scherm­afbeelding 2023-03-21 om 14 18 53

</details>

#### After
<details>

Scherm­afbeelding 2023-03-21 om 14 18 45

</details>


VWSCoronaDashboard18 (dismissed):
Nice job

packages/cms/src/lokalize/key-mutations.csv
VWSCoronaDashboard28, line 1: Are the keys used on NL and GM the same? As in do they have the same value? I ask because i see that you added three keys for GM but none for NL, so I assume NL already exists. If the values are the same, then maybe they can be part of the shared object rather than individually on the NL and GM level?

VWSCoronaDashboard26 (dismissed):

VWSCoronaDashboard28 (approved):

VWSCoronaDashboard26 (approved):

VWSCoronaDashboard28 (approved):

VWSCoronaDashboard26 (approved):

PR #4720, opened 2023-03-29
feature/COR-1510-non-clickable-choropleth

VWSCoronaDashboard21:
## Summary

* Choropleth on 'Gehandicaptenzorg' page is no longer clickable.

#### Before
<details>

Scherm­afbeelding 2023-03-29 om 14 17 31

</details>

#### After
<details>

Scherm­afbeelding 2023-03-29 om 14 16 57

Scherm­afbeelding 2023-03-29 om 14 17 15

</details>

### DIFF
--- a/packages/app/src/pages/landelijk/gehandicaptenzorg.tsx
+++ b/packages/app/src/pages/landelijk/gehandicaptenzorg.tsx
@@ -25,7 +25,6 @@ import { createGetStaticProps, StaticProps } from '~/static-props/create-get-sta
 import { createGetChoroplethData, createGetContent, getLastGeneratedDate, selectNlData, getLokalizeTexts } from '~/static-props/get-data';
 import { ArticleParts, PagePartQueryResult } from '~/types/cms';
 import { getBoundaryDateStartUnix } from '~/utils/get-boundary-date-start-unix';
-import { useReverseRouter } from '~/utils/use-reverse-router';
 import { getLastInsertionDateOfPage } from '~/utils/get-last-insertion-date-of-page';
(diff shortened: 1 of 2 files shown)


VWSCoronaDashboard26 (approved):

packages/app/src/pages/landelijk/gehandicaptenzorg.tsx
```diff
@@ -250,7 +248,7 @@ function DisabilityCare(props: StaticProps<typeof getStaticProps>) {
                 },
               }}
               dataOptions={{
-                getLink: reverseRouter.vr.gehandicaptenzorg,
+                isPercentage: true,
```
VWSCoronaDashboard26, line 251: I must admit that this change catches my off guard a little. I would originally argue that we'd rather not want this change, but then I noticed that the data displayed in this choropleth actually contains percentage values, meaning this does make sense. However, this property is used to correctly format the percentage number in the choropleth's tooltip, which is also done by `dataFormatters` in `dataConfig`. That means that we could consider that `dataFormatters` as redundant.

VWSCoronaDashboard26 (approved):

VWSCoronaDashboard18 (approved):
Nice one! (I would say remove the IsPercentage, but that is nonblocking, therefor you have my approval!)

packages/app/src/pages/landelijk/gehandicaptenzorg.tsx
VWSCoronaDashboard21: @VWSCoronaDashboard26 and @VWSCoronaDashboard25 as discussed I'll leave this for now like it is in this commit and will create a new ticket to do some investigation on `dataOptions`.

PR #4727, opened 2023-04-04
feature/COR-1516-404-page-improvements-multiple-pages-client-side-only

VWSCoronaDashboard28:
### Summary

Implements the new 404 pages using the existing 404 page. The branch name is lying to you, this is not client side ONLY (it was initially). All config is fetched at build time, and then the page decides which config to apply at run time.

- If there is no config found the fallback is the old version of the 404 pages.

NOTE: All feedback given on the PR for the SSR solution has already been applied in this branch as well. The main differences to look out for are in the `404.tsx` page. Ill post a link with a diff checker so you don't have to review the whole thing.

## Screenshots
**_Before_**
<details>

  <summary>Toggle before screenshot</summary> 

image

</details>

**_After_**
<details>

  <summary>Toggle After screenshot</summary> 

image

image

image

image

</details>


packages/app/src/pages/404.tsx
VWSCoronaDashboard28, line 1: See this link to see differences between this 404 and the 404 from the SSR branch: https://www.diffchecker.com/SxUgRAKb/

NOTE: The original text is the SSR (on the left) and the client side stuff is on the right

packages/app/src/pages/404.tsx
```diff
@@ -1,27 +1,227 @@
...
+      <StyledAnchor hasIcon={!!icon} href={linkUrl} isCTA={isCTA} underline={!isCTA}>
+        {linkLabel}
+      </StyledAnchor>
+
+      {hasChevron && <ChevronRight color={colors.blue8} height="10px" />}
+    </Box>
+  );
+};
+
+const NotFoundCTA = NotFoundLink; // Renaming for the sake of readability.
```
VWSCoronaDashboard18, line 183: Isn't this extra confusing, on the implementation on lines 112 and 125 you don't see it's that same element? It feels weird to have a 'CTA' component with a prop: 'isCTA' required for it to be a CTA
Could a description for both at the same time be a clearer option? It feels weird

For example: {{<NotFoundCTA>}} and use a hasBuyttonStyling={true} instead of a 'isCTA'
Or maybe it would be just as simple as: {{<NotFoundCtaLink>}}

This is non-blocking. And because your code works feel free to resolve this! With or without an answer. This comment is just {{check, check double-check}}.

VWSCoronaDashboard18 (approved):
Looks great! I like this PR. Here the approval!

VWSCoronaDashboard26 (approved):

VWSCoronaDashboard21 (approved):
LGTM!

packages/app/src/pages/404.tsx
VWSCoronaDashboard28: I see your point, however, I do prefer to have this renamed as `NotFoundCTA` simply because thats what is, i.e. it is supposed to be a call to action visually, and not a link. I wanted to do:
```
const NotFoundCTA = styled(NotFoundLink)`...`
```

But i can't do that.

PR #4742, opened 2023-04-17
feature/COR-1569-add-vaccination-page-keys

VWSCoronaDashboard21:


packages/app/schema/nl/vaccine_campaigns.json
```diff
@@ -3,14 +3,8 @@
...
       "type": "object",
-      "required": [
-        "vaccine_campaign_order",
-        "vaccine_campaign_name_nl",
-        "vaccine_campaign_name_en",
-        "vaccine_administered_total",
-        "vaccine_administered_last_week"
-      ],
-      "additionalProperties": false,
+      "required": ["vaccine_campaign_order", "vaccine_campaign_name_nl", "vaccine_campaign_name_en", "vaccine_administered_total", "vaccine_administered_last_timeframe"],
```
VWSCoronaDashboard28, line 6: This is a non blocking comment but I do feel its better for readability if these array items are on their own lines.

VWSCoronaDashboard28 (approved):
Looks good so Ill approve it already. Just one non-blocking comment.

packages/app/schema/nl/vaccine_campaigns.json
VWSCoronaDashboard21: I understand that this can be better for readability but this is auto formatted. As discussed I will  therefor resolve this conversation.

packages/app/schema/nl/vaccine_campaigns.json
```diff
@@ -3,14 +3,8 @@
...
+      "additionalProperties": true,
```
VWSCoronaDashboard25, line 7: I don't think this should be set to `true`. This allows basically anything.
If there are additional properties that are allowed, I think they should just be added in the schema, without adding them to the `required` array.

packages/common/src/types/data.ts
```diff
@@ -989,7 +989,9 @@ export interface NlVaccineCampaign {
   vaccine_campaign_name_nl: string;
   vaccine_campaign_name_en: string;
   vaccine_administered_total: number;
-  vaccine_administered_last_week: number;
+  vaccine_administered_last_week?: number;
+  vaccine_administered_last_timeframe: number;
+  [k: string]: unknown;
```
VWSCoronaDashboard25, line 994: This is a weird entry to have, but it's because `allowAdditionalProperties` is set to `true`.

packages/app/schema/nl/vaccine_campaigns.json
VWSCoronaDashboard18: As discussed with the backend and IA team the last_week is no longer optional but removed.

packages/common/src/types/data.ts
VWSCoronaDashboard18: This is updated because the archived section still refers to `_last_week`.

VWSCoronaDashboard18 (approved):
Looks good to me!

VWSCoronaDashboard26 (approved):

PR #4753, opened 2023-04-20
Task/COR-1493-sanity-vr-cleanup

VWSCoronaDashboard18:
## Summary

* 
* Removed all unused or unnecessary Vr keys in the lokalize keys.
* This PR does a final 'sanity-check' for remaining lokalize keys within Sanity, making sure all keys for ticket COR-1475 have been removed.
* By removing the removal of the VR pages and the corresponding keys, there are shared keys now which are only used on 1 level. In this PR those keys that are not shared anymore are moved to the scope of which they are used.
* 

It's in draft for now, I'll wait for the CSV to be updated after the sync to release script is run. And I will resolve the merge conflicts afterward.


packages/app/src/components/breadcrumbs/logic/use-breadcrumbs.tsx
```diff
@@ -114,22 +90,14 @@ export function useBreadcrumbs(): Breadcrumb[] {
...
-              }
-            );
+          if (str.includes('GM')) {
+            const pageTemplate = replaceVariablesInText(commonTexts.breadcrumbs.redirects['gemeente'], {
+              name: ctx[str],
+            });
+
+            return replaceVariablesInText(commonTexts.breadcrumbs.redirects.template, {
+              page: pageTemplate,
+            });
```
VWSCoronaDashboard18, line 100: A lot of formatting happening in this file. But only this part is changed

packages/app/src/pages/gemeente/[code]/positief-geteste-mensen.tsx
```diff
@@ -233,7 +233,7 @@ function PositivelyTestedPeople(props: StaticProps<typeof getStaticProps>) {
               }
               legend={{
                 thresholds: thresholds.gm.infected_per_100k,
-                title: textShared.chloropleth_legenda.titel,
+                title: textShared.chloropleth_legenda_titel,
```
Comment (VWSCoronaDashboard18) [inline packages/app/src/pages/gemeente/[code]/positief-geteste-mensen.tsx:236]: A single element was left. So I converted it from a object to a string

packages/app/src/pages/landelijk/besmettelijke-mensen.tsx
```diff
@@ -19,7 +19,6 @@ import { useDynamicLokalizeTexts } from '~/utils/cms/use-dynamic-lokalize-texts'
 const selectLokalizeTexts = (siteText: SiteText) => ({
   metadataTexts: siteText.pages.topical_page.nl.nationaal_metadata,
   textNl: siteText.pages.infectious_people_page.nl,
-  textShared: siteText.pages.infectious_people_page.shared,
```
VWSCoronaDashboard28, line 22: Just out of curiosity, why do I see the removal of shared text in multiple places? This is not exactly a VR reference if im not wrong, and stuff like `textShared.belangrijk_bericht` is used in multiple places.

packages/app/src/pages/landelijk/gedrag.tsx
```diff
@@ -59,7 +58,7 @@ export default function BehaviorPage(props: StaticProps<typeof getStaticProps>)
   const behaviorLastValue = data.behavior.last_value;
 
   const { formatNumber, formatDateFromSeconds, formatPercentage, locale } = useIntl();
-  const { caterogyTexts, metadataTexts, text, textNl, textShared } = useDynamicLokalizeTexts<LokalizeTexts>(pageText, selectLokalizeTexts);
```
VWSCoronaDashboard28, line 62: Same question here. Is the shared text object only a combination of NL and VR text? This comment also applies to other cases where `textShared` has been removed.

VWSCoronaDashboard28 (commented):
Please see my comments.

I'm also curious what your approach was when working on this. Do you think you could add a small note about that to the PR overview? Nothing elaborate just the general idea of the process you followed.

packages/app/src/pages/landelijk/besmettelijke-mensen.tsx
VWSCoronaDashboard18: I have updated the summary: see this point over there:

> By removing the removal of the VR pages and the corresponding keys, there are shared keys now which are only used on 1 level. In this PR those keys that are not shared anymore are moved to the scope of which they are used.

packages/app/src/pages/landelijk/gedrag.tsx
VWSCoronaDashboard18: I have updated the summary: see this point over there:

> By removing the removal of the VR pages and the corresponding keys, there are shared keys now which are only used on 1 level. In this PR those keys that are not shared anymore are moved to the scope of which they are used.

packages/cms/src/lokalize/key-mutations.csv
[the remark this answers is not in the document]
VWSCoronaDashboard26: Would you be able to elaborate on what issues happened earlier when using `move` rather than `delete` and `add`? Is that something we could somehow avoid? And how can we be certain that this will not introduce issues this time around?

VWSCoronaDashboard28 (dismissed):
Not sure how its going to go with the `moved` keys, but otherwise looks good

packages/cms/src/lokalize/key-mutations.csv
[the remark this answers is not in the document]
VWSCoronaDashboard18: The problems were related to the camel vs. snake cases page names. And previews were not updating when keys were changed. Now I've checked the nl_export multiple times. For this reason, I will exclude it from the upcoming release.

packages/app/src/pages/landelijk/gedrag.tsx
```diff
@@ -59,7 +58,7 @@ export default function BehaviorPage(props: StaticProps<typeof getStaticProps>)
...
+  const { caterogyTexts, metadataTexts, text, textNl } = useDynamicLokalizeTexts<LokalizeTexts>(pageText, selectLokalizeTexts);
```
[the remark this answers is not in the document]
VWSCoronaDashboard18: Thanks! Good catch! I've adjusted it and will resolve this comment.

VWSCoronaDashboard28 (approved):

VWSCoronaDashboard26 (approved):

PR #4755, opened 2023-04-24
feature/COR-1521-vulnerable-groups-page-optimizations-2

VWSCoronaDashboard21:
## Summary

* Deleted KPI
* Deleted PageInformationBlocks
* Deleted unused Sanity Keys
* Created `BorderedKpiSection`
* Replaced `VaccineCoverageTile`, `HospitalsTile` and InfectLocationsTile with `BorderedKpiSection`
* Deleted unused Icon ElderlyPeople
* Archived data for KPI
* Fixed a bug which caused the `PercentageBar` not showing

#### Before
<details>
<summary>Toggle before screenshots</summary>

Scherm­afbeelding 2023-04-18 om 17 04 47

Scherm­afbeelding 2023-04-18 om 17 04 54

**Percentagebar bug**
Scherm­afbeelding 2023-04-24 om 15 08 42

</details>

#### After
<details>
<summary>Toggle after screenshots</summary>

after-vulnerable-groups

**Percentagebar**
after_percentagebar

</details>

PR #4757, opened 2023-04-25
feature/COR-1521-vulnerable-groups-page-optimizations-typefix

VWSCoronaDashboard21:
## Summary

* Fix based on PR feedback which [can be found here](https://github.com/minvws/nl-covid19-data-dashboard/pull/4755#discussion_r1175472345)
* No visible changes and therefor no screenshots.

### DIFF
--- a/packages/app/src/components/kpi/types.ts
+++ b/packages/app/src/components/kpi/types.ts
@@ -1,6 +1,6 @@
 import { DifferenceInteger } from '@corona-dashboard/common';
 
-export type TileData = {
+type TileData = {
   description: string;
   title: string;
   value: number | null;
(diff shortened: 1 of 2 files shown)


packages/app/src/components/kpi/types.ts
```diff
@@ -1,6 +1,6 @@
 import { DifferenceInteger } from '@corona-dashboard/common';
 
-type TileData = {
```
VWSCoronaDashboard28, line 3: I don't think this needs to be exported right?

VWSCoronaDashboard21: You are right, I have adjusted this.

VWSCoronaDashboard26 (approved):
:shipit:

VWSCoronaDashboard28 (approved):

PR #4760, opened 2023-05-02
feature/COR-1529-advices-section-homepage

VWSCoronaDashboard26:
## Summary

* added advice component schemas for Sanity;
* updated Sanity desk structure;
* added Sanity query and types for fetching the topical page's new advice section;
* added Advice component;

#### Before
<details>

COR-1529_before_desktop

COR-1529_before_mobile
</details>

#### After
<details>

COR-1529_after_desktop

COR-1529_after_mobile
</details>


packages/cms/src/schemas/topical/advice-link.ts
```diff
@@ -0,0 +1,29 @@
+import { REQUIRED } from '../../validation';
+
+export const adviceLink = {
+  name: 'adviceLink',
+  title: "'Adviezen' sectie link",
```
VWSCoronaDashboard28, line 5: Why is `Adviezen` in single quotes? 🤔

VWSCoronaDashboard28 (approved):

packages/app/src/domain/topical/components/advice.tsx
```diff
@@ -0,0 +1,106 @@
...
+  image: ImageBlock;
+}
+
+export const Advice = ({ title, description, links, image }: AdviceProps) => {
+  return (
+    <Box width="100%" backgroundColor={colors.blue1} marginY={space[4]}>
+      <MaxWidth
+        alignItems={{ _: 'flex-start', sm: 'center' }}
+        display="flex"
+        flexDirection={{ _: 'column', sm: 'row' }}
```
VWSCoronaDashboard18, line 31: It looks like the image is not centered on the mobile viewport.
Which I would expect to happen.

Screenshot 2023-05-04 at 11 03 09

packages/cms/src/schemas/topical/advice.ts
```diff
@@ -0,0 +1,46 @@
...
+      description: "Configureer een beschrijving. Wordt onder de titel in de 'Adviezen' sectie getoond.",
+      type: 'localeRichContentBlock',
+      validation: REQUIRED,
+    },
+    {
+      name: 'links',
+      title: 'Links',
+      description: "Configureer een lijst van een of meer links. Wordt onder de beschrijving getoond in de 'Adviezen' sectie.",
+      type: 'array',
+      of: [{ type: 'reference', to: { type: 'adviceLink' } }],
```
VWSCoronaDashboard18, line 27: Is there a reason you created a new specific link document? There is a document of type link that also has proper validation.

packages/app/src/pages/index.tsx
```diff
@@ -222,7 +223,9 @@ const Home = (props: StaticProps<typeof getStaticProps>) => {
           </Box>
         </MaxWidth>
 
-        <Spacer marginBottom={space[5]} />
+        <Spacer marginBottom={{ _: space[5], md: space[6] }} />
```
VWSCoronaDashboard18, line 226: Was this change intentional? It doesn't appear in the designs and also not on other pages. The distance between other sections on the page is also 'just' space[5], unlike the distance between themes where it is [6].

Feel free to resolve this yourself, as I understand the inconsistency.

packages/app/src/domain/topical/components/advice.tsx
```diff
@@ -0,0 +1,106 @@
...
+  margin-block: ${space[4]};
+`;
+
+const ListItem = styled.li`
+  a {
+    align-items: center;
+    background-color: ${colors.white};
+    border-radius: ${radii[1]}px;
+    color: ${colors.blue8};
+    display: flex;
```
VWSCoronaDashboard18, line 91: Why add a display flex? What's wrong with inline?

packages/app/src/domain/topical/components/advice.tsx
```diff
@@ -0,0 +1,106 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  gap: ${space[3]};
+  list-style: none;
+  margin-block: ${space[4]};
+`;
+
+const ListItem = styled.li`
+  a {
+    align-items: center;
```
VWSCoronaDashboard18, line 87: Why align-items? There is no need for alignment here and should work out-of-the-box.

VWSCoronaDashboard18 (commented):
I've added some nonblocking comments, as everything works as intended. But there are things that could be improved.

There is something I would suggest would be blocking: The not-center alignment of the image on mobile.

Feel free tor resolve the others, or if it is solved.

packages/cms/src/schemas/topical/advice.ts
VWSCoronaDashboard25: @VWSCoronaDashboard18, for now and in the future, if you know there is another solution somewhere in the code, presumably better, please add a link to that with your comment.

You (knowing where to look) going through the effort once is better then making @VWSCoronaDashboard26  **and all interested reviewers** search for it one by one.

VWSCoronaDashboard25 (approved):
Still interested for the replies to some of the earlier comments by @VWSCoronaDashboard18.

But assuming a satisfying response to those comments and assuming that Design agrees with the current mobile view solution, I'll approve.

packages/app/src/domain/topical/components/advice.tsx
VWSCoronaDashboard18: @VWSCoronaDashboard25, that will not work. As the box doesn't have a 100% width. But line 31 as I suggested would also not work, as with would also center the Title, text, and buttons.

I would suggest removing the `<Box>` element on line 60.
Create a new styled component at the end if the file:

```
const AdviceImage = styled(SanityImage)`
  align-self: center;
`;
```

And replace the image with this styled component:

`<AdviceImage {...getImageProps(image, { defaultWidth: '300px' })} />`

packages/cms/src/schemas/topical/advice.ts
VWSCoronaDashboard18: Thanks #25 here is the file I was referring to:
`packages/cms/src/schemas/objects/link.ts`

packages/app/src/domain/topical/components/advice.tsx
VWSCoronaDashboard25: Ah yes, you are right.

Another solution can be `align-self: 'center'` to the line 60 box. https://css-tricks.com/snippets/css/a-guide-to-flexbox/#aa-align-self
I don't think it needs a display flex then.

```
<Box align-self={{ _: 'center', sm: 'flex-start' }}>
```

packages/app/src/pages/index.tsx
VWSCoronaDashboard26: I would personally argue that this _is_ in the design and it is also not something that can be considered inconsistent as the homepage is not like any other page. This essentially balances out the spacing on either side of the new component.

I will resolve this comment for now, but please do unresolve it if you disagree.

packages/app/src/queries/get-topical-structure-query.ts
```diff
@@ -92,6 +92,24 @@ export function getTopicalStructureQuery(locale: string) {
...
       },
+    },
+    'advice': *[
+      _type == 'advice' && !(_id in path('drafts.**'))
+    ][0]{
+      'title': title.${locale},
+      'description': description.${locale},
+      'links': links[]->{
+        'id': _id,
+        'url': url,
```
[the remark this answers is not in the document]
VWSCoronaDashboard26: Given how the `link` type works, which I will now use according to your [suggestion](https://github.com/minvws/nl-covid19-data-dashboard/pull/4760/files/f07b6f294819d989eda872033f656470c72ee9db#r1184748915), there is no way to 'localise' internal links as far as I am aware. However, I think it is always wise to use relative links for internal URLs and absolute URLs for external URLs. This logic is also handled in the component itself.

packages/cms/src/schemas/topical/advice-link.ts
VWSCoronaDashboard26: Adjusted this accordingly, though not in this file as it is removed due to this [comment](https://github.com/minvws/nl-covid19-data-dashboard/pull/4760/files/f07b6f294819d989eda872033f656470c72ee9db#r1184748915). Please feel free to resolve.

packages/cms/src/schemas/topical/advice.ts
VWSCoronaDashboard26: Good suggestion to opt with the `link` type. Feel free to resolve this conversation if you both agree on the result.

packages/app/src/domain/topical/components/advice.tsx
VWSCoronaDashboard26: This is a by-product of the previous usage of `display: flex;`, so this is removed and this comment can be resolved accordingly.

VWSCoronaDashboard27 (approved):

VWSCoronaDashboard25 (approved):
Looks good to me, speaking for myself, all comments can be resolved.

packages/app/src/domain/topical/components/advice.tsx
```diff
@@ -0,0 +1,103 @@
...
+import { ImageBlock, LinkProps } from '~/types/cms';
+import { isInternalUrl } from '~/utils/is-internal-url';
+import { Link } from '~/utils/link';
+
+interface AdviceProps {
+  title: string;
+  description: PortableTextEntry[];
+  links: LinkProps[];
+  image: ImageBlock;
+}
```
VWSCoronaDashboard18, line 22: This could be directly imported like this:


```import { Advice as AdviceProps } from '~/queries/query-types';```

**Review by VWSCoronaDashboard18 — COMMENTED** [review]
Nice! Great job!

File: packages/app/src/domain/topical/components/advice.tsx [hunk 9]
```diff
@@ -0,0 +1,98 @@
...
+      </Box>
+
+      <AdviceImage {...getImageProps(image, { defaultWidth: '300px' })} />
+    </AdviceContainer>
+  </Box>
+);
+
+const AdviceContainer = styled(MaxWidth)`
+  column-gap: ${space[4]};
+`;
```
VWSCoronaDashboard18, line 56: Nice! Good solution!

VWSCoronaDashboard18 (dismissed):
Great stuff!

VWSCoronaDashboard27 (approved):

packages/app/src/domain/topical/components/advice.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { External as ExternalIcon } from '@corona-dashboard/icons';
+import styled from 'styled-components';
+import { Box } from '~/components/base/box';
+import { RichContent } from '~/components/cms/rich-content';
+import { SanityImage } from '~/components/cms/sanity-image';
+import { ExternalLink } from '~/components/external-link';
+import { MaxWidth } from '~/components/max-width';
+import { Heading } from '~/components/typography';
+import { getImageProps } from '~/lib/sanity';
+import { Advice as AdviceProps } from '~/queries/query-types';
```
VWSCoronaDashboard25, line 11: Not too keen on using the query type as component props.

But if you really want this tight coupling, you should double down on it and do `<Advice {{...advice}} />` instead of `<Advice title={advice.title} description={advice.description} links={advice.links} image={advice.image} />`.

If that feels wrong, you should remove this tightly coupling.

VWSCoronaDashboard25 (changes requested):

packages/app/src/domain/topical/components/advice.tsx
VWSCoronaDashboard26: I must admit that I also had my doubts about using the type of the query for the component props definition. It just _happens_ to be the exact same structure and thus can we use it. I did however, when working on this ticket, also opt to go with `<Advice {...advice} />` earlier. I reverted this after some thought and a second opinion as

* spreading the `advice` props might not always be as clear
* spreading the `advice` props will not tell you what is there on hover, forcing one to click-through to see what it takes

I agree with that using the query type might not be desirable. I will revert that change.

VWSCoronaDashboard18 (approved):
Looks great! Nice esthetic.

VWSCoronaDashboard28 (approved):

packages/app/src/domain/topical/components/advice.tsx
VWSCoronaDashboard25: I'd argue that the first is just something we should learn to feel comfortable with because it's a common pattern to use in React, just not our project.
And the second, that's true. But you would just be piping the data through **without interacting with it**. So at the level where `<Advice ... />` is used, there is no real reason to know what it takes apart from personal interest.

VWSCoronaDashboard25 (approved):

PR #4762, opened 2023-05-09
COR-1588: use archived data from a different zip file for verdenkingen-huisartsen.tsx

VWSCoronaDashboard25:


packages/app/src/pages/landelijk/verdenkingen-huisartsen.tsx
```diff
@@ -27,12 +27,12 @@ type LokalizeTexts = ReturnType<typeof selectLokalizeTexts>;
...
   ({ locale }: { locale: keyof Languages }) => getLokalizeTexts(selectLokalizeTexts, locale),
   getLastGeneratedDate,
-  selectNlData('difference.doctor__covid_symptoms_per_100k', 'difference.doctor__covid_symptoms', 'doctor')
+  selectArchivedNlData('difference.doctor__covid_symptoms_per_100k_archived_20210903', 'difference.doctor__covid_symptoms_archived_20210903', 'doctor_archived_20210903')
 );
 
 const SuspectedPatients = (props: StaticProps<typeof getStaticProps>) => {
-  const { pageText, selectedNlData: data, lastGenerated } = props;
-  const lastValue = data.doctor.last_value;
+  const { pageText, selectedArchivedNlData: data, lastGenerated } = props;
```
VWSCoronaDashboard26, line 34: Would it make sense to define some kind of convention here? There will be times where we could be using both active and archived data on the same page, hence I would suggest to opt to go with `selectedArchivedNlData: archivedData` here. In the other case, for active data, we can opt to go with `selectedNlData: data`.

packages/common/src/types/data.ts
```diff
@@ -4,6 +4,44 @@
...
+  values: NlDoctorValue[];
+  last_value: NlDoctorValue;
+}
+export interface NlDoctorValue {
+  date_start_unix: number;
+  date_end_unix: number;
+  covid_symptoms_per_100k: number;
+  covid_symptoms: number;
+  date_of_insertion_unix: number;
+}
```
VWSCoronaDashboard26, line 43: Any particular reason why this is moved?

VWSCoronaDashboard26 (dismissed):
Pre-approved, but small non-blocking remark(s).

packages/app/src/pages/landelijk/verdenkingen-huisartsen.tsx
VWSCoronaDashboard25: Yes, that makes sense, I'll do that right now.

VWSCoronaDashboard26 (approved):

packages/app/src/utils/get-current-page-scope.ts
VWSCoronaDashboard18, line 1: Not sure how this relates to the PR.
Bot non-blocking, so feel free to resolve this

packages/app/infra/download.js
VWSCoronaDashboard18, line 1: I would have liked to see this setup dynamically, instead of this fixed-in-stone amount of two different download URLs.
To make this "amount-agnostic" wouldn't change the effort put into this, I think.

But it works perfectly, so this is definitely non-blocking!
Feel free to resolve this yourself!

packages/common/src/types/data-utils.ts
```diff
@@ -7,15 +7,19 @@ import { Gm, Nl } from '.';
 export type ScopedData = {
   gm: Gm;
   nl: Nl;
+  archived_nl: ArchivedNl;
 };
 
+export type ScopedRouterData = Omit<ScopedData, 'archived_nl'>;
+
 export type DataScopeKey = keyof ScopedData;
+export type RouterDataScopeKey = keyof ScopedRouterData;
```
VWSCoronaDashboard18, line 16: This ties in with the previous comment on the get-current-page-scope file.

I don't entirely understand why the "archived_nl" data type is removed from the router, but non-blocking.
So feel free to resolve this yourself!

VWSCoronaDashboard18 (approved):

PR #4763, opened 2023-05-10
Bugfix/cor 1617 restore bug from cor 1521

VWSCoronaDashboard18:
## Summary

* 
* Restore the percentage values not being used correctly and cleanup some props
* 

#### Before
<details>

Screenshot 2023-05-10 at 11 02 16
Screenshot 2023-05-10 at 11 02 27

</details>

#### After
<details>

Screenshot 2023-05-10 at 11 02 11
Screenshot 2023-05-10 at 11 02 39

</details>


packages/app/src/components/kpi/types.ts
VWSCoronaDashboard28, line 1: Mostly looks good, but I don't see the need to get rid of this file. I feel like it adds more stuff to this PR when the actual fix is just the `isPercentage` thing? Ill be honest, I like when the types for a particular component are organised in a dedicated `types.ts` file, so i'm curious why you removed it?

packages/app/src/components/kpi/components/kpi-content.tsx
```diff
@@ -7,40 +7,48 @@ import { parseBirthyearRange } from '~/domain/vaccine/logic/parse-birthyear-rang
...
+  birthyear?: string | null;
+  differenceValue?: DifferenceInteger;
+};
+
+export type BarType = {
+  value: number;
+  color: string;
+};
+
+export const KpiContent = ({ title, description, value, bar, birthyear, differenceValue, isPercentage = false }: TileData) => {
```
VWSCoronaDashboard26, line 12: I would prefer to keep `TileData` as `KpiContentProps`, as we are describing the property structure of the component because the name `TileData` could imply that it is used for different purposes.

This also ties in with the removal of `KpiContentProps` from `packages/app/src/components/kpi/types.ts`.

packages/app/src/components/kpi/types.ts
VWSCoronaDashboard28: I personally thing this is a nice way of organising code, and If i look in the project there are numerous files doing the same or similar thing (See the screenshot). So I'd say its not uncommon in the project. I also like the fact that the types or interfaces for all the components or functions follow the pattern of component name plus "Props", I think this makes it clear what your referring to and simple `cmd + click` will take you to the full definition. 

image

PR #4769, opened 2023-05-17
feature/COR-1606-dynamic-variants-v2

VWSCoronaDashboard18:
## Summary

* 
* Applying logic to form the variant data for the time series correctly.
* This with a dynamic label coming from the Backend
* There are no screenshots as this is something that has no visual output. (as it should stay exactly the same)
* During the development and research of this ticket there was a way to do it without refactoring the whole Time Series which would even be harder than setting up a whole new structure for a line chart and building it from scratch.
*


packages/app/src/domain/variants/static-props/get-variant-chart-data.ts
```diff
@@ -1,16 +1,13 @@
...
+export type VariantCode = string;
 
 export type VariantChartValue = {
   date_start_unix: number;
   date_end_unix: number;
   is_reliable: boolean;
-} & Partial<{
-  [key in `${VariantCode}_percentage`]: number;
-}>;
+} & Partial<{ [key: string]: number }>;
```
VWSCoronaDashboard26, line 10: I'm having a bit of a hard time wrapping my head around this. I can imagine we previously used a `Partial` utility type as we wanted to assign a value's percentage to only one of the variant codes inside the map (this file, lines 37 through 43). The partial allowed us to still have some type of code completion as an instance of `VariantChartValue` would be an object containing the three key-value pairs on lines 7 through 9 _and_ could contain all the different variant codes coming from Sanity as optionals. In practice, an instance of `VariantChartValue` only contained one of those variant codes. Given the current set-up, do we still need to use `Partial`?

VWSCoronaDashboard18: I've updated this to accept an undetermined amount of key/value pairs...

VWSCoronaDashboard26 (dismissed):

VWSCoronaDashboard26 (approved):

PR #4771, opened 2023-05-25
Log additional properties instead of breaking the code.

VWSCoronaDashboard18:
We often don't want the code/build/validation to break because of additional properties.
They are just extra loads. But we want to be notified if there are additional (meaning "unnecessary") properties.

In this PR I've implemented a solution to deal with this and convert a breaking situation to a logging situation.

Screenshot 2023-05-25 at 12 38 38
Screenshot 2023-05-25 at 12 38 58

### DIFF
--- a/packages/app/schema/archived_nl/__difference.json
+++ b/packages/app/schema/archived_nl/__difference.json
@@ -14,7 +14,7 @@
     "doctor__covid_symptoms_per_100k_archived_20210903",
     "doctor__covid_symptoms_archived_20210903"
   ],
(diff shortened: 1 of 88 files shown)


VWSCoronaDashboard18:
Al the 'additionalProperties" keys are converted and set to true (even if the original ones were different) this is because you always want a log, even if it used to not break on that schema.

VWSCoronaDashboard18:
No need for atm, and to much of a hassle to deal with right now. So to the 🧊 icebox, it is...

PR #4750, opened 2023-04-19
feature/COR-1504-VR-schema-files-removal

VWSCoronaDashboard26:
## Summary

* removed all VR schema files;
* removed all VR_collection schema files (exceptions mentioned below);
* updated schema references throughout the code;
* generated types;
* resolved any type issues based on newly generated types;
* cleaned up some associated docs;

### Notes
This pull request has been labeled as 'Awaiting Data' and 'On hold' as the removal of keys in VR_COLLECTION.json has to be processed by BE first in order to move this feature forward. Any VRXX.json files are ignored by the various validation scripts.

I approached this feature as specified in the associated ticket, which actually proved to be a fairly straight forward way of working. However, the ticket mentions maintaining a few less VR_collection schema files than this pull requests maintains. Below lists the metrics that are kept and the location they will be used once the safety regions have been removed from the dashboard.

| Metric                            | Page                                                                          | Location                                                          |
|-----------------------------------|-------------------------------------------------------------------------------|-------------------------------------------------------------------|
| disability_care_archived_20230126 | [/landelijk/gehandicaptenzorg](https://coronadashboard.government.nl/landelijk/gehandicaptenzorg)             | Disability care homes with positive tests                         |
| elderly_at_home_archived_20230126 | [/landelijk/thuiswonende-ouderen](https://coronadashboard.government.nl/landelijk/thuiswonende-ouderen)          | Positive tests among people aged over 70 living at home           |
| nursing_home_archived_20230126    | [/landelijk/kwetsbare-groepen-70-plussers](https://coronadashboard.government.nl/landelijk/kwetsbare-groepen-70-plussers) | Nursing and care homes with positive tests                        |

### Screenshots
This pull request does not include any screenshots as the changes do not have a visual impact on the dashboard.

PR #4772, opened 2023-05-25
bugfix/COR-1644-vulnerable-groups-metric-references

VWSCoronaDashboard26:
## Summary

* re-introduced vulnerable_nursing_home metric for VR_COLLECTION schemas;
* updated types;
* updated VulnerableGroups route component to point to the correct data metrics for the two KPIs, the choropleth and a time series chart;
  * KPIs found at 'Nursing and care homes'
  * Choropleth found at 'Nursing and care homes with positive tests'
  * Time series chart found at 'Number of locations with positive tests over time'

### Screenshots

There is no big visual difference as a result of this change, or at least not a change that is easily conveyed through screenshots. Essentially, the aforementioned KPIs and charts now show up-to-date data.

### DIFF
--- a/packages/app/schema/vr_collection/vulnerable_nursing_home.json
+++ b/packages/app/schema/vr_collection/vulnerable_nursing_home.json
@@ -1,12 +1,9 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
 
-  "title": "vr_collection_nursing_home_archived_20230126",
+  "title": "vr_collection_vulnerable_nursing_home",
   "type": "object",
   "properties": {
-    "newly_infected_people": {
-      "type": "integer"
-    },
     "newly_infected_locations": {
       "type": "integer"
     },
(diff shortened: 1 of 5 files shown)


packages/app/schema/vr_collection/vulnerable_nursing_home.json
VWSCoronaDashboard26, line 1: This change seems a bit more complex than it actually is, because GitHub marks the file as renamed rather than deleted and added.

In any case, this is what the original file looked like when previously deleted in this [pull request](https://github.com/minvws/nl-covid19-data-dashboard/pull/4750/files#diff-4c4d667a8fdaf290e22c2f073004f039fa28f46f53ee13cc1539d8d8aa8cb6f3).

VWSCoronaDashboard18 (approved):
Good that it is fixed now!

VWSCoronaDashboard25 (approved):
Looks good!

packages/app/schema/vr_collection/vulnerable_nursing_home.json
```diff
@@ -16,29 +13,24 @@
...
     "date_unix": {
       "type": "integer"
     },
-    "date_of_insertion_unix": {
-      "type": "integer"
-    },
     "vrcode": {
       "type": "string",
-      "pattern": "^VR[0-9]+$"
+      "equalsRootProperty": "code"
```
VWSCoronaDashboard28, line 24: Whats `"equalsRootProperty": "code"` doing exactly?

VWSCoronaDashboard28 (approved):

packages/app/schema/vr_collection/vulnerable_nursing_home.json
VWSCoronaDashboard26: This keyword checks whether a property of a sub property equals that of the value of the root object.
This is mainly used to check if 'vrcode' properties are equal to their root object. Preventing regional data from one region to be added to a different one.

PR #4775, opened 2023-05-30
Sanity-updates-for-COR-1618-1620-1622-1624

VWSCoronaDashboard18:
## Summary

* 
* The removal of the keys for branches: 
* 
* task/COR-1618-remove-KPI-from-Disability-care-homes
* task/COR-1920-remove-KPI-from-People-over-70-living-at-home
* task/COR-1622-remove-KPI-from-Coronamelder
* task/COR-1624-remove-KPI-from-GPs
* 
* This is done to prevent merge conflicts. So new approvals will not be necessary after one of them is merged
*

### DIFF
--- a/packages/cms/src/lokalize/key-mutations.csv
+++ b/packages/cms/src/lokalize/key-mutations.csv
@@ -5,3 +5,10 @@ timestamp,action,key,document_id,move_to
 2023-05-26T09:50:37.407Z,delete,common.verdenkingen_huisartsen.normalized_kpi_titel,jF33EuwumlGuwav2FD3X8m,__
 2023-05-26T09:50:37.407Z,delete,common.verdenkingen_huisartsen.normalized_kpi_toelichting,jF33EuwumlGuwav2FD3XAS,__
 2023-05-26T09:50:37.407Z,delete,common.verdenkingen_huisartsen.titel_kpi,jF33EuwumlGuwav2FD3XC8,__
+2023-05-30T05:55:00.820Z,delete,common.corona_melder_app.rapport.description,jF33EuwumlGuwav2FD4Hq0,__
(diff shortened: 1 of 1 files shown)


VWSCoronaDashboard28 (dismissed):

VWSCoronaDashboard26 (approved):
I totally understand the approach taken here and I comment you for the out-of-the-box thinking: this will make reviewing the individual PRs easier and they are ever so slightly decoupled.

However, the downside to this approach is that it might be harder/more cumbersome to map the deletion of keys to individual PRs or to check if all relevant keys are actually removed. In any way, you take some, you lose some.

:shipit:

VWSCoronaDashboard28 (approved):

VWSCoronaDashboard18:
> I totally understand the approach taken here and I comment you for the out-of-the-box thinking: this will make reviewing the individual PRs easier and they are ever so slightly decoupled.
> 
> However, the downside to this approach is that it might be harder/more cumbersome to map the deletion of keys to individual PRs or to check if all relevant keys are actually removed. In any way, you take some, you lose some.
> 
> :shipit:
> 
> Edit: Another question I have, though, is the following: I see that compilation is failing. Will that be resolved once the individual PRs are merged?

Thanks for your comment. I've considered this downside as well. And that's why I had the removal committed for every PR separately.

And about you 'another question'. I have this because in this branch all those files from the other PRs are not present. Now those files are still referencing those keys. But after those are merged, this is instantly solved.

VWSCoronaDashboard21 (approved):

VWSCoronaDashboard27 (approved):

VWSCoronaDashboard26 (approved):
LG2M - :shipit:

VWSCoronaDashboard28 (approved):

PR #4779, opened 2023-06-01
feature/COR-1602-FAQ-improved

VWSCoronaDashboard21:
## Summary

* Created `FaqSection`
* Divided the information in two columns.
* Added and changed styling.

#### Before
<details>

Scherm­afbeelding 2023-06-01 om 16 31 58

</details>

#### After
<details>

Scherm­afbeelding 2023-06-01 om 16 31 10

</details>

### DIFF
--- a/packages/app/src/components/collapsible/collapsible-section.tsx
+++ b/packages/app/src/components/collapsible/collapsible-section.tsx
@@ -12,24 +12,28 @@ interface CollapsibleSectionProps extends BoxProps {
   summary: string;
   children: ReactNode;
   id?: string;
+  hideBorder?: boolean;
   textColor?: string;
   borderColor?: string;
   border?: string;
   borderRadius?: string;
   marginBottom?: string;
   hasNormalFontWeight?: boolean;
+  hasSmallerFontSize?: boolean;
 }
 
 export const CollapsibleSection = ({
   summary,
   children,
   id,
+  hideBorder,
   textColor = colors.blue8,
   borderColor = colors.gray2,
   border,
   borderRadius,
   marginBottom,
   hasNormalFontWeight,
+  hasSmallerFontSize,
 }: CollapsibleSectionProps) => {
   const section = useRef<HTMLElement>(null);
 
@@ -64,8 +68,8 @@ export const CollapsibleSection = ({
   }, [toggle, id]);
 
   return (
-    <Box as="section" borderTop={`1px solid ${borderColor}`} id={id} ref={section} border={border} borderRadius={borderRadius} marginBottom={marginBottom}>
-      <StyledSummary textColor={textColor} onClick={() => collapsible.toggle()} hasNormalFontWeight={hasNormalFontWeight}>
+    <Box as="section" borderTop={hideBorder ? undefined : `1px solid ${borderColor}`} id={id} ref={section} border={border} borderRadius={borderRadius} marginBottom={marginBottom}>
+      <StyledSummary textColor={textColor} onClick={() => collapsible.toggle()} hasNormalFontWeight={hasNormalFontWeight} hasSmallerFontSize={hasSmallerFontSize}>
         <Box width="100%" position="relative">
           {summary}
           {id && (
@@ -100,6 +104,7 @@ const StyledAnchor = styled(Anchor)`
 interface SummaryProps {
   textColor: string;
   hasNormalFontWeight?: boolean;
+  hasSmallerFontSize?: boolean;
 }
 
 const StyledSummary = styled(Box)<SummaryProps>`
@@ -107,7 +112,7 @@ const StyledSummary = styled(Box)<SummaryProps>`
   color: ${({ textColor }) => textColor};
   cursor: pointer;
   display: flex;
-  font-size: ${fontSizes[5]};
+  font-size: ${({ hasSmallerFontSize }) => (hasSmallerFontSize ? `${fontSizes[2]}` : `${fontSizes[5]}`)};
   font-weight: ${({ hasNormalFontWeight }) => (hasNormalFontWeight ? `${fontWeights.normal}` : `${fontWeights.bold}`)};
   justify-content: space-between;
   padding: ${space[3]};
--- a/packages/app/src/components/collapsible/collapsible-section.tsx
+++ b/packages/app/src/components/collapsible/collapsible-section.tsx
@@ -12,7 +12,6 @@ interface CollapsibleSectionProps extends BoxProps {
   summary: string;
   children: ReactNode;
   id?: string;
-  hideBorder?: boolean;
   textColor?: string;
   borderColor?: string;
   border?: string;
@@ -24,7 +23,6 @@ export const CollapsibleSection = ({
   summary,
   children,
   id,
-  hideBorder,
   textColor = colors.blue8,
   borderColor = colors.gray2,
   border,
@@ -64,7 +62,7 @@ export const CollapsibleSection = ({
   }, [toggle, id]);
 
   return (
-    <Box as="section" borderTop={hideBorder ? undefined : `1px solid ${borderColor}`} id={id} ref={section} border={border} borderRadius={borderRadius} marginBottom={marginBottom}>
+    <Box as="section" borderTop={`1px solid ${borderColor}`} id={id} ref={section} border={border} borderRadius={borderRadius} marginBottom={marginBottom}>
       <StyledSummary textColor={textColor} onClick={() => collapsible.toggle()}>
         <Box width="100%" position="relative">
           {summary}
--- a/packages/app/src/components/collapsible/collapsible-section.tsx
+++ b/packages/app/src/components/collapsible/collapsible-section.tsx
@@ -17,6 +17,7 @@ interface CollapsibleSectionProps extends BoxProps {
   border?: string;
   borderRadius?: string;
   marginBottom?: string;
+  hasNormalFontWeight?: boolean;
 }
 
 export const CollapsibleSection = ({
@@ -28,6 +29,7 @@ export const CollapsibleSection = ({
   border,
   borderRadius,
   marginBottom,
+  hasNormalFontWeight,
 }: CollapsibleSectionProps) => {
   const section = useRef<HTMLElement>(null);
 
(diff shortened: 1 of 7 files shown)


packages/app/src/pages/veelgestelde-vragen.tsx
```diff
@@ -62,44 +61,42 @@ const Verantwoording = (props: StaticProps<typeof getStaticProps>) => {
   const { commonTexts } = useIntl();
 
   const groups = groupBy<FAQuestionAndAnswer>(content.questions, (x) => x.group);
+  const groupsArray = Object.entries(groups);
+  const groupsArrayDivider = Math.ceil(groupsArray.length / 2);
+  const firstHalf = groupsArray.slice(0, groupsArrayDivider);
+  const secondHalf = groupsArray.slice(groupsArrayDivider);
```
VWSCoronaDashboard18, line 67: So we ended up with just straight-up randomness on the two halves? There is no sorting or calculation going on?

VWSCoronaDashboard28: The order of groups and questions can still be controlled from Sanity. If COM wants, they can switch around the order of groups and questions to create whatever kind of sorting they want. 

There was a commit which was reverted, which introduced `groupLeft` and `groupRight`, which was also an option, however that solution is not backwards compatible out of the box (as far as I know), and it introduces a new functionality increasing the scope of what communication would need to do in Sanity.

Now, they can decide whether they wan't equal columns, or whether they wan't an alphabetic order. And if the amount of groups and questions is right, they can also have both. But to respect the design, there's no real way for us to ensure an alphabetic order AND exactly equal columns.

VWSCoronaDashboard28 (commented):
Please see my comments.

packages/app/src/pages/veelgestelde-vragen.tsx
VWSCoronaDashboard18: Yeah, I got that, but I find it hard to go for a solution where the separation of columns happens in one list in sanity where the top half is the left side and the other half is the left side. Especially with there being no explanation if it. 

This will very hard to get for a future colleague to get this info from looking at the page, or even sanity alone...

packages/app/src/components/collapsible/collapsible-section.tsx
```diff
@@ -51,7 +52,7 @@ export const CollapsibleSection = ({ summary, children, id, hideBorder, textColo
   }, [toggle, id]);
 
   return (
-    <Box as="section" borderTop={hideBorder ? undefined : `1px solid ${borderColor}`} id={id} ref={section}>
+    <Box as="section" borderTop={hideBorder ? undefined : `1px solid ${borderColor}`} id={id} ref={section} className={`${className} ${collapsible.isOpen ? 'open' : ''}`}>
```
VWSCoronaDashboard25, line 55: I understand this pattern as the ability to provide extra classnames to `CollapsibleSection`, but I don't think we use this anywhere and it seems not the be used in the PR. If this is true, maybe just do `${collapsible.isOpen ? 'open' : ''}` instead and don't add the extra `className` property.

But maybe there is a good reason to do this. I might be a bit out of the loop and `className` could be a property like `children` that is always there, even if you don't specify it. Then you need to do this to make sure you're not breaking default behaviour.

packages/app/src/pages/veelgestelde-vragen.tsx
```diff
@@ -62,44 +61,42 @@ const Verantwoording = (props: StaticProps<typeof getStaticProps>) => {
   const { commonTexts } = useIntl();
 
   const groups = groupBy<FAQuestionAndAnswer>(content.questions, (x) => x.group);
+  const groupsArray = Object.entries(groups);
+  const groupsArrayDivider = Math.ceil(groupsArray.length / 2);
```
VWSCoronaDashboard25, line 65: For naming, a divider is something you put in-between stuff to divide them, here you find the index of the last element that is still part of the first group.

It's okay to be quite literal with naming stuff if you cannot think of a satisfying name, because that's what most people will always understand. Here I think I'd go with `lastIndexForFirstHalf`. It's not pretty, but at least it's technical correct. :)

packages/app/src/components/collapsible/collapsible-section.tsx
VWSCoronaDashboard28: Your assumption is semi-correct. We are not required to pass the `className` prop, and we do not often do this. However, systems which work with React - in this case, `styled-components` - requires a `className` prop to be present so it can attach an obfuscated `className` to the newly created styled component so it may receive its styles. 

So in this case, for us to be able to "restyle" `CollapsibleSection` into `StyledCollapsibleSection`, this was required.

This is something new which we have just learnt while working on this ticket. Essentially, if you are trying to use `styled-system` to style an existing component, the `className` prop must be present on that component. When we just use `styled.div` this is not required because that component is created by `styled-system`, however when we do `styled(MyCustomComponent)`, it is required.

packages/app/src/pages/veelgestelde-vragen.tsx
VWSCoronaDashboard28: A future colleague as in a developer or someone working in the communications team? I believe @VWSCoronaDashboard21 spoke with someone from COM and got confirmation that its good to go like this.

VWSCoronaDashboard28 (dismissed):

packages/app/src/pages/veelgestelde-vragen.tsx
VWSCoronaDashboard21: @VWSCoronaDashboard18 I have added a comment to the code to clarify this more to future developers. Please resolve this comment if you agree.

packages/app/src/pages/veelgestelde-vragen.tsx
VWSCoronaDashboard21: @VWSCoronaDashboard25 I have changed the naming of `groupsArrayDivider` to `middleIndexOfGroups`. Please resolve this comment if you are okay with that.

VWSCoronaDashboard28 (approved):

packages/app/src/components/collapsible/collapsible-section.tsx
VWSCoronaDashboard25: Thanks @VWSCoronaDashboard28 for the explaination!

packages/app/src/pages/veelgestelde-vragen.tsx
VWSCoronaDashboard18: @VWSCoronaDashboard21, thanks that helps. I think it is way, way more important to have this explained in Sanity itself. Otherwise, we are called every now, and then, asking why a question is moved over to the other column.

For a developer it should be clear there is no sorting going on. A communication colleague filling this out in Sanity has no possible way to see this before publishing.

VWSCoronaDashboard28 (dismissed):

packages/app/src/pages/veelgestelde-vragen.tsx
VWSCoronaDashboard27: @VWSCoronaDashboard18 done

packages/cms/src/schemas/documents/pages/veelgestelde-vragen-page.ts
```diff
@@ -18,7 +18,7 @@ export const veelgesteldeVragen = {
       type: 'array',
       title: 'Vragen',
       description:
-        'Je kan veel gestelde vragen toevoegen, de volgorde veranderen, de tekst bijwerken of verwijderen',
+        'Je kan veelgestelde vragen toevoegen, de volgorde veranderen, de tekst bijwerken of verwijderen. Om de groepen en vragen opnieuw te ordenen, hoeft je slechts één van de vragen in een groep te verplaatsen. Waar de eerste vraag van een bepaalde groep staat, is waar de groep zal verschijnen op de Veelgestelde vragen pagina. Heb je bijvoorbeeld één vraag uit de groep Algemeen bovenaan en de andere vragen onderaan de lijst, dan wordt de groep Algemeen als eerste groep getoond.',
```
VWSCoronaDashboard18, line 21: Sorry, I feel a bit like a nag...
If I imagine myself being a user this is a hard-to-understand text and not really explaining how the left and right columns are determined. I would suggest to explain it more like this:

> De veelgestelde vragen worden in productie in twee kolommen weergegeven. Welke groep vragen gekoppeld is aan de kolom wordt bepaald door de groepen in tweeën te delen. De sortering van het onderstaande overzicht wordt aangehouden en er wordt gekeken naar welke groep als eerste een vraag heeft in deze lijst. Die groep is groep 1. Vervolgens wordt gekeken naar de volgende vraag van een nieuwe groep. En dat wordt groep 2. Totdat elke groep een plaats in de sortering heeft. Vervolgens wordt de eerste helft van deze volgorde links weergegeven op de FAQ pagina. En de tweede helft van deze volgorde rechts.
> Om een groep van van de linker naar de rechter kolom te wisselen moeten de vragen van deze groep lager staan dan de helft van het aantal groepen. Om een groep van van de rechter naar de linker kolom te wisselen moet de eerste vraag van deze groep hoger staan dan de helft van het aantal groepen.

VWSCoronaDashboard18 (dismissed):

packages/cms/src/schemas/documents/pages/veelgestelde-vragen-page.ts
VWSCoronaDashboard28: I don't really mind changing the description to what you think is right. I agree that its better to give more information than necessary to avoid any confusion. However, if we do implement this longer description from @VWSCoronaDashboard18, we should consider implementing the description the same way the description is implemented for `Lokalize` documents, so that this massive description does not clutter the viewport.

VWSCoronaDashboard18 (approved):
Great job done!

packages/app/src/pages/veelgestelde-vragen.tsx
VWSCoronaDashboard18: This is adjusted in a video call session

VWSCoronaDashboard27 (approved):

VWSCoronaDashboard28 (approved):

PR #4781, opened 2023-06-05
Feature/cor 1613 data explained redesign

VWSCoronaDashboard28:
## Summary

* Updated Cijferverantwoording design - Desktop and Mobile
* Created a new layout for data explanation
* Added a few new fields to sanity for icons and metadata(slugs)
* Fixed broken back button on GM level

#### Before
<details>

image

image
</details>

#### After
<details>

image

image

image

image
</details>


packages/app/src/components/layout/app-content.tsx
```diff
@@ -24,121 +22,109 @@ export function AppContent({ children, sidebarComponent, searchComponent, hideBa
   const reverseRouter = useReverseRouter();
   const { commonTexts } = useIntl();
 
-  const isMenuOpen = router.pathname == '/landelijk' || router.pathname == '/veiligheidsregio/[code]' || router.pathname == '/gemeente/[code]' || router.query.menu === '1';
+  const isMenuOpen = router.pathname == '/landelijk' || router.pathname == '/verantwoording' || router.pathname == '/gemeente/[code]' || router.query.menu === '1';
```
VWSCoronaDashboard18, line 25: Nice usage of VR leftovers!

packages/cms/src/data/data-structure.ts
```diff
@@ -3,6 +3,9 @@
  * This file is generated based on the JSON schema's by yarn generate-data-structures in the cli package.
  */
 export const dataStructure = {
+  archived_nl: {
+    doctor_archived_20210903: ['covid_symptoms_per_100k', 'covid_symptoms'],
+  },
```
VWSCoronaDashboard18, line 8: I think this is removed in an earlier PR, you sure this needs to be here?

packages/cms/src/data/data-structure.ts
```diff
@@ -301,205 +303,9 @@ export const dataStructure = {
...
-      'birthyear_range_18_plus',
-      'birthyear_range_60_plus',
-      'vaccinated_percentage_12_plus',
-      'vaccinated_percentage_12_plus_label',
-      'vaccinated_percentage_18_plus',
-      'vaccinated_percentage_18_plus_label',
-      'vaccinated_percentage_60_plus',
-      'vaccinated_percentage_60_plus_label',
-    ],
+    vulnerable_nursing_home: ['newly_infected_locations', 'infected_locations_total', 'infected_locations_percentage'],
```
VWSCoronaDashboard18, line 309: I thought these were removed in the KPI PRs as well..

VWSCoronaDashboard18 (dismissed):
Looking great. (some comments can be solved instantly, just wanted to share what I liked)

packages/cms/src/data/data-structure.ts
VWSCoronaDashboard28: I'm not sure now that you mention it, this was created by running `yarn cms` I believe 🤔. Either way, it is in no way connected to this ticket so I will remove it.

packages/cms/src/data/data-structure.ts
VWSCoronaDashboard28: Same thing as the previous comment, ill undo this stuff.

packages/cms/src/data/data-structure.ts
VWSCoronaDashboard28: Ill resolve this as the file is no longer part of this PR.

packages/app/src/components/layout/app-content.tsx
```diff
@@ -24,121 +22,109 @@ export function AppContent({ children, sidebarComponent, searchComponent, hideBa
...
+  };
 
-  /**
-   * @TODO Open the menu purely client side without loading a new page
-   */
-  const backButtonUrl = currentPageScope ? (isMenuOpen ? (isNational ? reverseRouter.topical.nl : undefined) : reverseRouter[currentPageScope].index(currentCode)) : undefined;
-
-  const backButtonText = currentPageScope ? (isMenuOpen ? (isNational ? commonTexts.nav.back_topical.nl : '') : commonTexts.nav.back_all_metrics[currentPageScope]) : '';
+  const backButtonUrl = currentPageScope ? (backButtonConfig.isMenuOpen ? backButtonConfig.url.truthyValue : backButtonConfig.url.falsyValue) : undefined;
+  const backButtonText = currentPageScope ? (backButtonConfig.isMenuOpen ? backButtonConfig.text.truthyValue : backButtonConfig.text.falsyValue) : '';
```
VWSCoronaDashboard25, line 44: This feels like it can use some additional thought. Basically you've setup an elaborate structure to avoid triple nested ternaries.
I like the goal, but not the means. From what I understand, the variables to determine the url or text are `typeof currentPageScope === 'undefined'`, `isMenuOpen`, `isNational` and `isGeneral`, but I still don't really understand by just reading the code how it is related.

I'd like to take a crack on this, to see if I can come up with something with the same result but easier to understand.

VWSCoronaDashboard28: Fair enough man. My goal was just to make the ternaries more readable as I think nesting to such a level in ternaries is quite unreadable at a glance (which I feel is the point of ternaries.

You are correct about the variables. Their relation to each other however, is not so strong. `isGeneral` is created from `currentPageScope` simply for some abstraction. Same with `isNational`. 

There's definitely better ways to go about this. A simple getter function with some regular if/else blocks would be a good way to go.

This solution tries to balance time and effort so that this feature definitely makes it into the release. Its apparently the most important feature in this release. Feel free to give this crack but also keep in mind that this is rather important for the release in case your version takes too long.

packages/cms/src/data/data-structure.ts
VWSCoronaDashboard18: The unfortunate thing of not rebasing the branches...

VWSCoronaDashboard18 (approved):
Looking great! Let's go with this!

Nice improvements on the get-back-button-logic!

packages/app/src/components/layout/logic/get-back-button-values.ts
```diff
@@ -0,0 +1,42 @@
...
+      return {
+        url: isMenuOpen ? reverseRouter.topical.nl : reverseRouter[currentPageScope].index(),
+        text: isMenuOpen ? commonTexts.nav.back_topical.nl : commonTexts.nav.back_all_metrics[currentPageScope],
+      };
+    case 'gm':
+      if (isMenuOpen) break;
+      return {
+        url: reverseRouter[currentPageScope].index(currentCode),
+        text: commonTexts.nav.back_all_metrics[currentPageScope],
+      };
```
VWSCoronaDashboard25, line 29: Inside the switch case the value for `currentPageScope` is known, so similar to line 33, you should be able to use the literal value instead of using `[currentPageScope]`, right?
I didn't test this:
```suggestion
      return {
        url: isMenuOpen ? reverseRouter.topical.nl : reverseRouter.nl.index(),
        text: isMenuOpen ? commonTexts.nav.back_topical.nl : commonTexts.nav.back_all_metrics.nl,
      };
    case 'gm':
      if (isMenuOpen) break;
      return {
        url: reverseRouter.gm.index(currentCode),
        text: commonTexts.nav.back_all_metrics.gm,
      };
```

packages/app/src/domain/layout/data-explained-layout.tsx
```diff
@@ -0,0 +1,76 @@
...
+  groups: DataExplainedGroups;
+}
+
+export const DataExplainedLayout = ({ children, groups }: DataExplainedLayoutProps) => {
+  const { commonTexts } = useIntl();
+
+  const items = Object.entries(groups).map(([name, items]) => ({
+    key: name,
+    title: name,
+    icon: <DynamicIcon name={getFilenameToIconName(items[0].groupIcon as IconName) as IconName} />,
```
VWSCoronaDashboard25, line 26: `getFilenameToIconName(items[0].groupIcon as IconName) as IconName`
I didn't catch this before, but I don't think casting to `IconName` twice is needed here.

For the first casting, `getFilenameToIconName` expects a string as input, and within this PR, the type of the `groupIcon` property of `DataExplainedGroup` is typed as a string, so that should not be necessary. If casting is really needed, it should be `as string` instead.

The second casting is on the return value of `getFilenameToIconName`. Looking at that method, it comes directly from an `Array.find()` method on an `IconName[]`, which means the result is typed `IconName | undefined`. So instead, handling the `undefined` scenario should be enough.
`icon` expects a `React.ReactElement`, so if the result of `getFilenameToIconName()` is `undefined`, it should be sufficient to use `null` for `icon`.

packages/app/src/domain/layout/data-explained-layout.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+            <Box paddingX={space[3]}>
+              <Heading level={2} variant={'h3'}>
+                {commonTexts.sidebar.nl.title}
+              </Heading>
+            </Box>
+
+            <Box paddingBottom={space[3]}>
+              <Menu spacing={2}>
+                <MenuRenderer items={items as ExpandedSidebarMap<'custom'>} />
```
VWSCoronaDashboard25, line 66: `items` is created on line 23 as an array of plain JS objects, which is why you need to cast them here to `ExpandedSidebarMap<'custom'>`, because that's what `MenuRenderer` is expecting. But casting should be avoided.

Instead, it's better to start line 23 with:
```const items: ExpandedSidebarMap<'custom'> = ...```
This way, you are telling your IDE that you are going to create just that. And then, if something is not correct, you will get red squiggly lines within the JS object on the faulty piece of code instead of here.

File: packages/app/src/pages/verantwoording/[slug].tsx [hunk 8]
```diff
@@ -0,0 +1,77 @@
...
+          <link key="dc-type" rel="dcterms:type" href="https://standaarden.overheid.nl/owms/terms/webpagina" />
+          <link key="dc-type-title" rel="dcterms:type" href="https://standaarden.overheid.nl/owms/terms/webpagina" title="webpagina" />
+        </Head>
+
+        <Box maxWidth={sizes.contentWidth} spacing={4} paddingTop={{ _: space[3], md: space[5] }} paddingLeft={{ _: space[3], sm: space[5] }} paddingRight={{ _: space[3], sm: 0 }}>
+          <PageInformationBlock
+            category={page.title || 'Cijferverantwoording'}
+            screenReaderCategory={page.title || 'Cijferverantwoording'}
+            title={item.title}
+            icon={item.icon ? <DynamicIcon name={getFilenameToIconName(item.icon as IconName) as IconName} /> : undefined}
```
Comment (VWSCoronaDashboard25) [inline packages/app/src/pages/verantwoording/[slug].tsx:67]: `getFilenameToIconName(item.icon as IconName) as IconName`
See [the other comment](https://github.com/minvws/nl-covid19-data-dashboard/pull/4781/files#r1220404858) about this casting.

packages/app/src/utils/get-current-page-scope.ts
```diff
@@ -4,6 +4,14 @@ import { NextRouter } from 'next/router';
...
+export function getCurrentPageScope(router: NextRouter): (RouterDataScopeKey | 'general') | undefined {
+  const { pathname } = router;
+
+  const pageScopeMap: Record<string, RouterDataScopeKey | 'general'> = {
+    landelijk: 'nl',
+    gemeente: 'gm',
+    verantwoording: 'general',
+  };
+
+  return pageScopeMap[Object.keys(pageScopeMap).find((key) => pathname.startsWith(`/${key}`)) as keyof typeof pageScopeMap] || undefined;
```
VWSCoronaDashboard25, line 16: `Object.keys(pageScopeMap).find(..) as keyof typeof pageScopeMap` is a very elaborate way of claiming this is probably not `undefined`.
And then putting it in `pageScopeMap[...] || undefined` is... peculiar.

The `keyof typeof pageScopeMap` is `string` according to line 10. Using `Array.find()` makes it `string | undefined`.

So extracting the key as separate value and then catching the `undefined` scenario should take care of that:
```suggestion
  const pageScopeMapKey = Object.keys(pageScopeMap).find((key) => pathname.startsWith(`/${key}`));
  return typeof pageScopeMapKey === 'undefined' ? undefined : pageScopeMap[pageScopeMapKey];
```

packages/app/src/components/layout/app-content.tsx
VWSCoronaDashboard25: Thanks, I like the current solution. Just a small remark (in a different comment).

VWSCoronaDashboard25 (dismissed):
I left a few comments, mostly on [type casting](https://www.w3schools.com/typescript/typescript_casting.php) and how to avoid it.
Due discussed time constraints and my absence, resolve at your own discretion, but please pick up the type casting soon after the release if not now.

That said, take my cursed approval if you really need it now.

VWSCoronaDashboard18 (approved):
Looking great!

VWSCoronaDashboard27 (approved):

PR #4788, opened 2023-06-16
feature/COR-1491-migrate-to-sanity-studio-V3

VWSCoronaDashboard26:
## Summary

At the moment of writing this description, various scripts do not work. This is due to some unexpected results based on environment variables and the ability to read them through either the Sanity studio or the CLI. This means that files found in the following paths are subject to change and do not have to be reviewed yet.

* `packages/cms-v3/src/lokalize/scripts`
* `packages/cms-v3/src/lokalize/utils`
* `packages/cms-v3/src/studio/scripts`

This will have to be optimised before merging.

The end goal of this pull request is to remove the `packages/cms/` package and have the `packages/cms-v3/` package replace the former.

Given the above challenges, we could theoretically release Sanity V3 as a Studio, but keep the 'legacy' (working) scripts for Sanity V2 around until we have solved any issues with the new scripts.

**Edit:** any of the above issues have now been resolved. These files can be reviewed accordingly.


packages/cms-v3/package.json
```diff
@@ -0,0 +1,64 @@
...
+  "private": true,
+  "version": "1.0.0",
+  "license": "EUROPEAN UNION PUBLIC LICENCE v. 1.2",
+  "type": "module",
+  "scripts": {
+    "start": "sanity start",
+    "build": "sanity build",
+    "clean": "npx rimraf dist .sanity-asset-cache .lokalize-reference",
+    "nuke": "yarn clean && npx rimraf node_modules",
+    "dev": "yarn workspace @corona-dashboard/common build && yarn workspace @corona-dashboard/icons build && yarn workspace @corona-dashboard/cli generate-data-structures && sanity dev",
```
VWSCoronaDashboard18, line 12: `sanity dev` gives an error
`Error: "dev" is not a sanity command. Did you mean "debug"? See "sanity --help"
    at _.runCommand (/usr/local/lib/node_modules/@sanity/cli/bin/sanity-cli.js:3628:1339)
    at t.exports (/usr/local/lib/node_modules/@sanity/cli/bin/sanity-cli.js:2006:2422)
❯ yarn dev`

I think it should be `sanity start`

VWSCoronaDashboard28: No its not supposed to be `sanity start`. `sanity start` now has its own functionality which is to build a production version of the studio. The new command is indeed `sanity dev`, which you can use for now from the `cms-v3` directory, using `yarn dev`. If this still gives you some issues, i would suggest running installing the packages in the new CMS again.

See this link for more info on `sanity dev`: https://www.sanity.io/docs/example-migrating-the-blog-template-from-studio-v2-to-v3#468bc06dfa37

VWSCoronaDashboard18 (approved):
🎉LG2M

VWSCoronaDashboard21 (commented):
I noticed there is a error message of max 0 articles but that does not look correct to me.
Scherm­afbeelding 2023-06-19 om 15 13 39

packages/cms-v3/src/schemas/documents/page-parts/articles.ts
```diff
@@ -29,15 +29,13 @@ export const articles = defineType({
...
       fieldset: 'articleConfiguration',
-      validation: (rule) => rule.min(0),
     }),
     defineField({
       title: 'Maximum aantal artikelen',
       name: 'maxNumber',
       type: 'number',
       hidden: ({ currentUser }) => !isAdmin(currentUser),
       fieldset: 'articleConfiguration',
-      validation: (rule) => rule.min(0),
```
VWSCoronaDashboard28: I don't really see how this works. The error was saying that `You can have max 0 articles` but the rule here was a `mini` rule and not `max`. Doesn't really make sense that this would cause that error right? Or maybe I don't see the full picture.

VWSCoronaDashboard21: Alright, thanks for looking into this. I'll resolve this conversation.

packages/cms-v3/src/studio/desk-structure/lokalize-structure-item.ts
```diff
@@ -0,0 +1,68 @@
...
+          .listenQuery(
+            `//groq
+            *[_type == $type]{ subject }
+          `,
+            { type },
+            {}
+          )
+          .pipe(
+            // @ts-expect-error - The below code works like a charm and creates list items accordingly.
+            // The old CMS would also face challenges here, but does not report this as such given the TS implementation.
```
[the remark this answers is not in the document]
VWSCoronaDashboard26: This comment, and similar comments, are meant to be kept. The code does work, but does not satisfy TypeScript. The same issue does theoretically occur in the old CMS, yet does not physically show up because the TypeScript implementation isn't correctly done.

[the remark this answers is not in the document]
VWSCoronaDashboard21: Alright, I'll resolve this comment.

VWSCoronaDashboard18 (approved):
Good updates!

VWSCoronaDashboard28 (approved):
Wow, I am speechless. Beauty is an understatement.

VWSCoronaDashboard21 (approved):

VWSCoronaDashboard18 (approved):
I'm not only speechless but it also cannot be described by any visual or auditive representation. This masterpiece is indescribable awesome!

PR #4789, opened 2023-06-21
COR-1649 update rwzi measurements

VWSCoronaDashboard18:
## Summary

* 
* Updating the schema to have a separate metric for the sewer installations
* ALSO THE GM_COLLECTION is updated to have a property removed for the sewer installation
* 
* 
* There are no screenshots as there is no visual change on the dashboard.
* 
*

### DIFF
--- a/packages/app/schema/gm/__index.json
+++ b/packages/app/schema/gm/__index.json
@@ -4,20 +4,22 @@
   "title": "gm",
   "additionalProperties": false,
   "required": [
-    "last_generated",
-    "proto_name",
     "name",
+    "proto_name",
     "code",
-    "tested_overall",
-    "hospital_nice",
-    "deceased_rivm_archived_20221231",
     "difference",
     "static_values",
+    "booster_coverage_archived_20220904",
+    "deceased_rivm_archived_20221231",
+    "hospital_nice",
+    "sewer_installation_measurement",
+    "sewer_per_installation",
     "sewer",
+    "tested_overall",
     "vaccine_coverage_per_age_group",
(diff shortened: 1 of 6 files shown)


packages/app/schema/gm/__index.json
```diff
@@ -4,20 +4,22 @@
   "title": "gm",
   "additionalProperties": false,
   "required": [
-    "last_generated",
```
VWSCoronaDashboard28, line 7: Im a little curious what happened in this file? It looks like everything (except for sewer_installation_measurement) is the same but rearranged?

packages/common/src/types/data.ts
```diff
@@ -340,7 +346,7 @@ export interface GmCollectionSewer {
   date_end_unix: number;
   gmcode: string;
   average: number;
-  total_installation_count: number;
+  total_installation_count?: number;
```
VWSCoronaDashboard28, line 349: This property has been removed right? Should it then not be removed completely instead of being made optional here? 🤔

VWSCoronaDashboard28 (approved):
Looks good! Just a couple of comments.

VWSCoronaDashboard21 (approved):

packages/app/schema/gm/__index.json
VWSCoronaDashboard18: It's rearranged indeed

packages/common/src/types/data.ts
VWSCoronaDashboard18: It's an auto-generated file, so this is not giving a problem

PR #4796, opened 2023-06-23
COR-1663: adding sewer data on NL, GM and GM_collection to the archived section

VWSCoronaDashboard25:


VWSCoronaDashboard26 (dismissed):
Pre-approved. However, I think it would be helpful if you also were to run either `yarn cms` from the root of the project or run `yarn generate-data-structures` from the CLI package. Either of these scripts will update the data structure for the CMS found at `packages/cms/src/studio/data/data-structure.ts`. I propose to introduce this as a practice whenever making changes to the data in our repository. What do you think?

packages/cli/src/schema/schema-info.ts
```diff
@@ -35,5 +35,14 @@ export function getSchemaInfo(jsonDirectory: string = defaultJsonDirectory): Sch
       files: ['NL.json'],
       basePath: path.join(jsonDirectory, 'archived'),
     },
+    archived_gm: {
+      files: getFileNames(fileList, /^GM[0-9]+.json$/),
```
VWSCoronaDashboard18, line 39: I know this should have the duplicate files as the Zips root folder (all the gm's)
But is it good to use the file list for the actual base path mentioned on the line below? (the archived folder)

This is not blocking though as it is a quality improvement an not a code-breaking comment

VWSCoronaDashboard18 (dismissed):
Good stuff
(Approved with a non-blocking comment)

VWSCoronaDashboard25:
> Pre-approved. However, I think it would be helpful if you also were to run either `yarn cms` from the root of the project or run `yarn generate-data-structures` from the CLI package. Either of these scripts will update the data structure for the CMS found at `packages/cms/src/studio/data/data-structure.ts`. I propose to introduce this as a practice whenever making changes to the data in our repository. What do you think?

Wow, my bad. I didn't realise we have **two** files generated based on **the same** schema's, but generated by **different** scripts.

Neglecting this like I did is perfectly fine to consider a blocking issue, in my opinion. As it is now, changing the schema's should both change the data.ts and the data-structure.ts, so if you see a schema change without both the data.ts and data-structure.ts modified, just throw in a change request.

VWSCoronaDashboard26 (approved):

VWSCoronaDashboard28 (approved):

packages/cli/src/schema/schema-info.ts
VWSCoronaDashboard25: Fixed!

VWSCoronaDashboard28 (approved):

VWSCoronaDashboard18 (approved):
Nice! Great PR

PR #4828, opened 2023-08-04
Task: Create the possibility to get the archived data from GM level

VWSCoronaDashboard18:
Because we now have the first GM proto's in the archived folder for GM, we also need to update the code to get the archived data from the GM.

Also, we updated the types for getStaticProps as they were missing the excellent structure to receive more static props than we now have. There were some inconsistencies in the typing where the props were not built the same.
After this fix we also tested the new typing and if the old stuff would break. And we were not able to find any breaking code.


VWSCoronaDashboard29 (approved):
Approved

PR #4829, opened 2023-08-07
Feature/cor 1667 move non suffix keys to archive

VWSCoronaDashboard29:
## Summary

Archive the following keys on pages:
**National**

_Coronamelder app_
| Old name of key| New name|
|--------|--------|
|corona_melder_app_warning|corona_melder_app_warning_archived_20220421|
|corona_melder_app_download|corona_melder_app_download_archived_20220421|

_Besmettelijke mensen_
| Old name of key| New name|
|--------|--------|
|infectious_people|infectious_people_archived_20210709|

_Vaccinaties_
| Old name of key| New name|
|--------|--------|
|repeating_shot_administered|repeating_shot_administered_20220713|
|vaccine_coverage|vaccine_coverage_archived_20220518|
|vaccine_delivery_per_supplier|vaccine_delivery_per_supplier_archived_20211101|
|vaccine_stock|vaccine_stock_archived_20211024|
|vaccine_vaccinated_or_support|vaccine_vaccinated_or_support_archived_20230411|

_Verpleeghuiszorg_
| Old name of key| New name|
|--------|--------|
|vulnerable_nursing_home|vulnerable_nursing_home_archived_20230711|

**Municipal**
| Old name of key| New name|Comments|
|--------|--------|-------|
|hospital_nice|-|Key isn't used anywhere (anymore) in the frontend, removed|


packages/app/schema/gm_collection/__index.json
```diff
@@ -17,20 +17,12 @@
...
     },
-    "hospital_nice": {
-      "type": "array",
-      "minItems": 342,
-      "maxItems": 342,
-      "items": {
-        "$ref": "hospital_nice.json"
-      }
-    },
     "hospital_nice_choropleth": {
```
VWSCoronaDashboard18, line 20: *Disclaimer: Heavily opinionated comment here, so feel free to resolve it.

In my honest opinion, I think the choropleth one should be the one to go. We should not define the type of graph in the metric name. Only what metric it is about. And the metric is hipsotal_nice regardless of where it is showing up in our Frontend.

packages/app/src/pages/gemeente/index.tsx
```diff
@@ -33,12 +33,12 @@ const Municipality = (props: StaticProps<typeof getStaticProps>) => {
   };
 
   const data = useMemo(() => {
-    return gmData.map<GmCollectionHospitalNice>(
+    return gmData.map<GmCollectionHospitalNiceChoropleth>(
```
VWSCoronaDashboard18, line 36: hmm, I'm in doubt. I think we should clean this hospital-type stuff here as well. And make it dependent only on the gmData array. But probably not for now. But you might want to create a ticket when you encounter stuff like this.

packages/app/src/pages/landelijk/vaccinaties.tsx
```diff
@@ -105,7 +107,7 @@ export const getStaticProps = createGetStaticProps(
     }>((context) => {
       return `{
         "parts": ${getPagePartsQuery('vaccinations_page')},
-        "elements": ${getElementsQuery('nl', ['vaccine_coverage', 'vaccine_administered'], context.locale)}
+        "elements": ${getElementsQuery('nl', ['vaccine_coverage_archived_20220518', 'vaccine_administered'], context.locale)}
```
VWSCoronaDashboard18, line 110: I think @VWSCoronaDashboard28  mentioned this before. This is a dangerous thing to do. As sanity is not auto-updated with the archived suffix...

VWSCoronaDashboard29: What are the implications of this? What will go wrong?

VWSCoronaDashboard18: @VWSCoronaDashboard30 is able to tell you! What the implications are and how he dealt with them.

VWSCoronaDashboard29: I have updated sanity with the appriopriate archived suffix. I have also run the `elements:add` script to add back timeline events to the newly archived timeseries chart on the vaccines page.

packages/app/src/pages/landelijk/vaccinaties.tsx
```diff
@@ -387,11 +389,11 @@ function VaccinationPage(props: StaticProps<typeof getStaticProps>) {
 
               <VaccinationsOverTimeTile
                 text={textNl}
-                coverageData={data.vaccine_coverage}
+                coverageData={archiveData.vaccine_coverage_archived_20220518}
                 administrationData={administrationData}
                 vaccineAdministeredPlannedLastValue={data.vaccine_administered_planned.last_value}
                 timelineEvents={{
-                  coverage: getTimelineEvents(content.elements.timeSeries, 'vaccine_coverage'),
+                  coverage: getTimelineEvents(content.elements.timeSeries, 'vaccine_coverage_archived_20220518'),
```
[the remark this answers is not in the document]
VWSCoronaDashboard29: Idem. Sanity and it's timeline events have been updated to use the newly archived metric instead.

packages/app/schema/gm_collection/__index.json
VWSCoronaDashboard18: Can you maybe elaborate on this a bit more @VWSCoronaDashboard29 ?

VWSCoronaDashboard29: I understand this way of thinking. The reason we chose to do the opposite is because we believed it would be clearer to have a metric that's only used for _choropleth_ charts, have _choropleth_ in the name itself.

VWSCoronaDashboard18 (approved):
Awesome! Nice aligned for the sprint demo!

packages/app/schema/gm_collection/__index.json
VWSCoronaDashboard29: @VWSCoronaDashboard18 would you agree to resolve the comment?

VWSCoronaDashboard18: I will resolve this for now. But with your explanation @VWSCoronaDashboard29, Are we extending this to other metrics as well? Almost all metrics are for a single type of graph... For example a table, a time-series, or a KPI tile. Following your logic, I think you can create tickets for those as well...

VWSCoronaDashboard18 (dismissed):
Looks awesome!

PR #4832, opened 2023-08-07
Add functionality to use archived vr_collection data

VWSCoronaDashboard29:
## Summary

Addition of a function to create choropleth with archived VR_COLLECTION data

### DIFF
--- a/packages/app/schema/archived_vr_collection/__index.json
+++ b/packages/app/schema/archived_vr_collection/__index.json
@@ -1,0 +1,32 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "title": "archived_vr_collection",
+  "additionalProperties": false,
+  "required": [
+    "last_generated",
+    "proto_name",
+    "name",
+    "code"
+  ],
+  "properties": {
+    "last_generated": {
+      "type": "string"
+    },
+    "proto_name": {
+      "$ref": "#/$defs/archived_vr_collection_id"
+    },
+    "name": {
+      "$ref": "#/$defs/archived_vr_collection_id"
+    },
+    "code": {
+      "$ref": "#/$defs/archived_vr_collection_id"
+    }
+  },
+  "$defs": {
+    "archived_vr_collection_id": {
+      "type": "string",
+      "enum": ["ARCHIVED_VR_COLLECTION"]
+    }
+  }
+}
--- a/packages/app/schema/archived_vr_collection/__index.json
+++ b/packages/app/schema/archived_vr_collection/__index.json
@@ -26,7 +26,7 @@
   "$defs": {
     "archived_vr_collection_id": {
       "type": "string",
-      "enum": ["ARCHIVED_VR_COLLECTION"]
+      "enum": ["VR_COLLECTION"]
     }
   }
 }
--- a/packages/app/src/components/choropleth/logic/types.ts
+++ b/packages/app/src/components/choropleth/logic/types.ts
@@ -65,16 +65,6 @@ export type GmDataItem = GmDataCollection[number];
 export type ArchivedVrDataCollection = null[] | null[];
 export type ArchivedVrDataItem = ArchivedVrDataCollection[number];
 
-/**
- * Here we map a MapType to a corresponding DataCollection type
- */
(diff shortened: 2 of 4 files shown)


packages/app/schema/archived_vr_collection/__index.json
```diff
@@ -0,0 +1,32 @@
...
+      "$ref": "#/$defs/archived_vr_collection_id"
+    }
+  },
+  "$defs": {
+    "archived_vr_collection_id": {
+      "type": "string",
+      "enum": ["ARCHIVED_VR_COLLECTION"]
+    }
+  }
+}
```
VWSCoronaDashboard18, line 32: This should just be VR_COLLECTION as it's the scope and not if it's archived or not. line 27 already takes care of that...
Schema validation would also break because it wouldn't match the current Backend agreement.

packages/app/src/components/choropleth/logic/types.ts
```diff
@@ -45,16 +46,25 @@ export const mapToCodeType: Record<MapType, CodeProp> = {
...
 
 export type GmDataCollection = GmCollectionHospitalNice[] | GmCollectionTestedOverall[] | GmCollectionSewer[] | GmCollectionVaccineCoveragePerAgeGroup[];
 export type GmDataItem = GmDataCollection[number];
 
+export type ArchivedVrDataCollection = null[] | null[];
+export type ArchivedVrDataItem = ArchivedVrDataCollection[number];
+
 /**
  * Here we map a MapType to a corresponding DataCollection type
  */
```
VWSCoronaDashboard18, line 70: The two functions below are redundant and can be removed.

packages/app/src/static-props/get-data.ts
```diff
@@ -288,6 +302,23 @@ export function createGetChoroplethData<T1, T2>(settings?: {
...
+export function createGetArchivedChoroplethData<T1, T2>(settings?: {
+  vr?: (collection: ArchivedVrCollection, context: GetStaticPropsContext) => T1;
+  gm?: (collection: ArchivedGmCollection, context: GetStaticPropsContext) => T2;
+}) {
+  return (context: GetStaticPropsContext) => {
+    const filterVr = settings?.vr ?? NOOP;
+    const filterGm = settings?.gm ?? NOOP;
+
+    return {
+      choropleth: {
```
VWSCoronaDashboard18, line 314: This name needs an update. Otherwise we would have trouble when we have archived an non archived data on the same page

VWSCoronaDashboard18 (changes requested):
Already looking good!!

packages/app/schema/archived_vr_collection/__index.json
VWSCoronaDashboard29: @VWSCoronaDashboard30 instructed me to use `ARCHIVED_VR_COLLECTION` here. 

Let's agree first on what is the correct way. @VWSCoronaDashboard30 please check this.

packages/app/src/static-props/get-data.ts
VWSCoronaDashboard29: renamed `filter*` function to `filterArchived*`

packages/app/src/components/choropleth/logic/types.ts
VWSCoronaDashboard29: Removed `MappedDataCollection` and `MappedDataItem`

packages/app/schema/archived_vr_collection/__index.json
VWSCoronaDashboard18: Nice! Good stuff!

packages/app/src/components/choropleth/logic/types.ts
VWSCoronaDashboard18: Nice! Good stuff!

packages/app/src/static-props/get-data.ts
VWSCoronaDashboard18: This is not exactly what I meant. The `filterGm` has no impact on the name of the data.
I referred to the key at the root of the object, not the filter function.

VWSCoronaDashboard29: I see, updated `return: {choropleth}` to `return: {archivedChoropleth}`.

Also refactored `types.ts` to prevent type check errors

PR #4833, opened 2023-08-07
feat(COR-1420): Added required redirects for NL and GM levels

VWSCoronaDashboard30:
- Added the new redirect and wildcards for `redirects.js` file.
- Updated the file names.
- Updated all reverse router references and variables.


packages/app/src/next-config/redirects/redirects.js
```diff
@@ -68,12 +68,40 @@ async function redirects() {
...
+    // Redirects for the archived pages (COR-1420)
+    // NL level
     {
       source: '/landelijk/intensive-care-opnames',
-      destination: `/landelijk/ziekenhuizen-en-zorg`,
+      destination: '/landelijk/ziekenhuizen-en-zorg',
+      permanent: true,
+    },
+    {
+      source: '/landelijk/ziekenhuizen-en-zorg',
```
VWSCoronaDashboard18, line 82: Now on lines 71 and 78, the destination is the source on this line. Update their destination to the same as over here so you won't redirect the user's al around the site...

VWSCoronaDashboard18 (dismissed):
Nice PR. My comment is non-blocking! But might still be worth the attention.

packages/app/src/next-config/redirects/redirects.js
VWSCoronaDashboard30: Fixed!

VWSCoronaDashboard18 (approved):
Looks good!

PR #4834, opened 2023-08-09
task: Update code on leftover comments on the PR for COR-1666

VWSCoronaDashboard18:
Since we archived the VR_collection we need to remove the original one that is not being used. We get updated data today and need to clean up the `vr_collection` references and instead use `archived_vr_collection`

small update: Schema validation is successful as we can handle obsolete files, which we don't reference in the schema's


VWSCoronaDashboard30 (approved):
Amazing job!!! Thank you for the effort!

VWSCoronaDashboard28 (approved):

VWSCoronaDashboard27 (approved):

PR #4840, opened 2023-08-14
Feature/cor 1760 spacing text and arrow components

VWSCoronaDashboard35:
## Summary

There's two parts to this feature:
1.  Spacing of the text and arrow in the button that is found in the pageinformation block
2. Spcacing between the text and arrow in the faq section, mainily the question text.

## Motivation

To maintain a clear distance from text and the icons because the text could first overflow onto/ really close to the icon.

## Changes

MicrosoftTeams-image (3)

Also made smilar changes to the faq part.
image


packages/app/src/components/collapsible/collapsible-section.tsx
```diff
@@ -64,7 +64,7 @@ export const CollapsibleSection = ({
   return (
     <Box as="section" borderTop={hideBorder ? undefined : `1px solid ${borderColor}`} id={id} ref={section} className={`${className} ${collapsible.isOpen ? 'open' : ''}`}>
       <Summary color={textColor} onClick={() => collapsible.toggle()} fontSize={fontSize}>
-        <Box width="100%" position="relative">
```
VWSCoronaDashboard18, line 67: Screenshot 2023-08-14 at 10 12 06

It doesn't work entirely correctly right now. (In the screenshot you can see the arrow overflowing into the padding)

VWSCoronaDashboard18 (changes requested):
See the comments

packages/app/src/components/collapsible/collapsible-section.tsx
VWSCoronaDashboard35: My appologies, I have edited my description. The changes in the collapsible-section affect this
image

packages/app/src/components/page-information-block/page-information-block.tsx
```diff
@@ -110,8 +110,10 @@ export function PageInformationBlock({
               <Box flex="1" display="flex" flexDirection="column" spacing={3}>
                 {pageInformationHeader.dataExplained && (
                   <PageInformationButton href={pageInformationHeader.dataExplained.link}>
-                    <BoldText>{pageInformationHeader.dataExplained.button.header}</BoldText>
-                    <RichContent blocks={pageInformationHeader.dataExplained.button.text} />
+                    <Box pr={5}>
```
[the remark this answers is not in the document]
VWSCoronaDashboard35: I'm not quite sure how else  I would go about this one. 
image

packages/app/src/components/collapsible/collapsible-section.tsx
VWSCoronaDashboard18: We stopped the use of `magic numbers`. You can check the documentation on this one.

And I think you mixed up the screenshot. As this screenshot is not part of this piece of code.

packages/app/src/components/page-information-block/page-information-block.tsx
```diff
@@ -110,8 +110,10 @@ export function PageInformationBlock({
               <Box flex="1" display="flex" flexDirection="column" spacing={3}>
                 {pageInformationHeader.dataExplained && (
                   <PageInformationButton href={pageInformationHeader.dataExplained.link}>
```
VWSCoronaDashboard18, line 112: It is also happening at the FAQ section
 
Screenshot 2023-08-15 at 10 13 00

VWSCoronaDashboard18 (changes requested):
Some comments left.

packages/app/src/components/page-information-block/page-information-block.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard28: You could also opt to convert this entire button into a flexbox or a grid.

packages/app/src/components/page-information-block/page-information-block.tsx
VWSCoronaDashboard35: I understood that as well, but I have changed stylings for the button so the changes will be propagated to it as well.

packages/app/src/components/page-information-block/page-information-block.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard35: I took the grid approach for this.

packages/app/src/components/page-information-block/page-information-block.tsx
```diff
@@ -113,7 +113,7 @@ export function PageInformationBlock({
                     <BoldText>{pageInformationHeader.dataExplained.button.header}</BoldText>
                     <RichContent blocks={pageInformationHeader.dataExplained.button.text} />
 
-                    <ChevronRight />
+                    <ChevronRight class="arrow" />
```
VWSCoronaDashboard28, line 116: This class is not being used and also `class` is a reserved keyword and cannot be used in React like this. What you want instead is `className`.

packages/app/src/components/page-information-block/page-information-block.tsx
```diff
@@ -187,17 +187,50 @@ const PageInformationButton = styled(Anchor)`
...
   text-align: left;
+  display: inline-grid;
+  grid-template-columns: 3fr 3fr 1.25fr;
+  grid-template-rows: 1fr;
+  column-gap: ${space[5]};
+
+  
+  @media ${mediaQueries.xs} {
+    grid-template-columns: 4fr 4fr 1fr;
+
```
VWSCoronaDashboard28, line 201: Each media query here has this extra empty line within the block.

packages/app/src/components/page-information-block/page-information-block.tsx
```diff
@@ -187,17 +187,50 @@ const PageInformationButton = styled(Anchor)`
...
-    position: absolute;
-    right: ${space[2]};
-    top: 50%;
-    transform: translate(-50%, -50%);
+    grid-column-start: 3;
+    grid-column-end: 3;
     width: 24px;
+    transform: translate(-10%, -50%);
+    display: block;
+    margin-top: auto:
```
VWSCoronaDashboard28, line 233: Type on this line: `:`

packages/app/src/components/page-information-block/page-information-block.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard28: Any particular reason for choosing one over the other?

packages/app/src/components/page-information-block/page-information-block.tsx
VWSCoronaDashboard35: I have removed this since I don't need it anymore.

packages/app/src/components/page-information-block/page-information-block.tsx
VWSCoronaDashboard35: Good catch!

packages/app/src/components/page-information-block/page-information-block.tsx
VWSCoronaDashboard35: It's consistent now.

packages/app/src/components/page-information-block/page-information-block.tsx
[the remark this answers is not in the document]
VWSCoronaDashboard35: Since the title and text are two separate components positioned vertically under each other and the svg is positioned horizontally I went for Grid.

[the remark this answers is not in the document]
VWSCoronaDashboard28: Please see my message in our direct channels.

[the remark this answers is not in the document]
VWSCoronaDashboard35: Thank you for your message. I have considered you changes and used the flexbox approach to maintain the consistency.

packages/app/src/components/page-information-block/page-information-block.tsx
```diff
@@ -188,16 +192,24 @@ const PageInformationButton = styled(Anchor)`
   color: ${colors.black};
   cursor: pointer;
   padding: ${space[3]} ${space[4]};
-  position: relative;
   text-align: left;
+  display: flex;
+  justify-content: space-between;
+
+  div {
+    margin-right: ${space[5]};
```
VWSCoronaDashboard28, line 200: Is this `margin-right` actually needed? It doesn't seem to be doing much on larger screens and on smaller screens it creates a very large gap. 

If it is really needed, I would suggest changing the `div` to a `Box` and adding the margin as a prop - `marginRight`.

VWSCoronaDashboard35: The box approach is something I already attemted. I think you can see it in the commit history.

packages/app/src/components/collapsible/collapsible-section.tsx
VWSCoronaDashboard35: I have removed the use of the magic number.

VWSCoronaDashboard28 (approved):

VWSCoronaDashboard28 (dismissed):

VWSCoronaDashboard28:
> On mobile, I still have problems with the arrow not having space to the left.
> 
> Screenshot 2023-08-22 at 10 38 37
> 
> And I don't see why for the same issue, you are creating two solutions. (on padding-right on the collapsible, and a flexbox on the page information block items) But this last comment is non-blocking.

This is a very specific edge case, one which could be deemed totally unrealistic because there are very low chances of COM using single letters to form the copy of their buttons. While I do agree it's good to take edge cases into consideration, this is quite an extreme edge case. The screenshot below is a little more realistic, in my opinion, and it looks just fine.

Additionally, your comment is non-blocking however you are still requesting changes. 

image

VWSCoronaDashboard18:
Let me say it like this: there needs to be a gap between the elements! Not a visual gap between the text and the SVG. There is a mayor difference in those two things.

VWSCoronaDashboard28 (approved):

VWSCoronaDashboard18 (approved):
Awesome!

PR #4845, opened 2023-08-21
feat(COR-1674): Added the new lokalize key

VWSCoronaDashboard30:
- Added the new lokalize key _population_count_connected_to_rwzis_


packages/cms/src/studio/data/data-structure.ts
VWSCoronaDashboard18, line 1: This file should not be part of this PR, I think...

VWSCoronaDashboard18 (changes requested):
See my comment, the other two files are great!

VWSCoronaDashboard28 (approved):

VWSCoronaDashboard18 (changes requested):
test for realtime github updates

VWSCoronaDashboard18 (approved):
Now it's OK to merge. Because the other PR's (where the data structure changes) are on the way.
When the data structure is deleted in this PR, a new commit needs to happen, and two new approvals are needed. So for time/effort constraints we do it already in this PR

PR #4847, opened 2023-08-22
task(key-mutations): Fix issue with cluttered keys

VWSCoronaDashboard30:
- Removed the 3 batches of keys that need to be replaced with the back up


VWSCoronaDashboard28 (dismissed):

VWSCoronaDashboard18 (changes requested):
I think we need to remove the tested items from here and put them in the branch for COR-1705. Now the validation fails, and emerging this would break the dev environment.

VWSCoronaDashboard28 (approved):

VWSCoronaDashboard18 (dismissed):
Good fixing!

*Normally the files have an empty line at the end of the file. But the file is still valid, so there is no issue to merge. But even with priority stuff and fixing try to stay constant and keep these things in mind.

VWSCoronaDashboard28 (approved):

PR #4852, opened 2023-08-30
feature/COR-1772_adjusted-KPIs-for-hospital-admissions-per-municipality

VWSCoronaDashboard30:


packages/app/src/components/choropleth-legenda.tsx
```diff
@@ -29,10 +29,10 @@ export function ChoroplethLegenda({ title, thresholds, valueAnnotation, pageType
       : replaceVariablesInText(commonTexts.common.value_and_higher, {
           value: formatNumber(x.threshold),
         });
-    if (pageType === 'sewer' && i === 0 && x.threshold === 0) {
+    if ((pageType === 'sewer' || pageType === 'patienten-in-beeld' || pageType === 'ziekenhuis-opnames') && i === 0 && x.threshold === 0) {
```
VWSCoronaDashboard30, line 32: Should I make this a more generic type instead of just making the list bigger and bigger? @VWSCoronaDashboard28 @VWSCoronaDashboard29

packages/app/schema/archived_gm_collection/__index.json
```diff
@@ -17,6 +17,14 @@
     "code": {
       "$ref": "#/$defs/archived_gm_collection_id"
     },
+    "hospital_nice_choropleth_archived_20230830": {
```
VWSCoronaDashboard28, line 20: This is no longer used therefore it is being archived, if I'm not wrong. However, the ticket does not mention archiving anything. Are we sure this needs to be archived and not removed?

The only mention I see in the ticket about archiving: `admissions_on_date_of_admission_per_100000 needs to be archived for patienten-in-beeld page.`

packages/app/schema/archived_gm/__difference.json
```diff
@@ -16,6 +16,9 @@
     },
     "deceased_rivm__covid_daily_archived_20221231": {
       "$ref": "#/definitions/diff_integer"
+    },
+    "hospital_nice__admissions_on_date_of_reporting_moving_average_archived_20230905": {
+      "$ref": "#/definitions/diff_integer"
```
VWSCoronaDashboard28, line 21: Why does this need to be archived? Is it still used in an archived graph somewhere? Additionally, why have you added the `_archived_20230905`? As far as I know, we do not use these suffixes any more.

Also, if it is supposed to be archived, why is it not a required property?

packages/app/schema/archived_gm_collection/hospital_nice.json
VWSCoronaDashboard28, line 1: Again, I'm missing some context or the ticket is not clear because I see a bunch of archiving happening and I don't really see where this is mentioned in the ticket.

packages/app/schema/archived_gm_collection/hospital_nice_choropleth.json
VWSCoronaDashboard28, line 1: Same comment as above.

packages/app/src/components/choropleth-legenda.tsx
VWSCoronaDashboard28: This could get out of hand if we keep adding to it like this. 

I'd opt for either some kind of object lookup with the appropriate config, or a switch statement.

packages/app/src/components/choropleth-legenda.tsx
```diff
@@ -45,7 +45,7 @@ export function ChoroplethLegenda({ title, thresholds, valueAnnotation, pageType
     } as LegendItem;
   });
 
-  if (pageType === 'sewer') {
+  if (pageType === 'sewer' || pageType === 'patienten-in-beeld' || pageType === 'ziekenhuis-opnames') {
```
VWSCoronaDashboard28, line 48: I also do not see anything about adding the outdated data label to the legend in the ticket.

packages/app/src/components/choropleth/logic/thresholds.ts
```diff
@@ -183,6 +183,41 @@ const hospitalAdmissionsPer100000Thresholds: ChoroplethThresholdsValue[] = [
   },
 ];
 
+const admissions_in_the_last_7_days_per_100000: ChoroplethThresholdsValue[] = [
```
VWSCoronaDashboard28, line 186: Please look at the naming convention used for other arrays in this file of the same type.

packages/app/src/components/choropleth-legenda.tsx
VWSCoronaDashboard29: I think turning it into a generic type would increase readability, so yes.

packages/app/src/components/kpi/bordered-kpi-section.tsx
```diff
@@ -9,9 +9,9 @@ import { Text } from '../typography';
 import { KpiContent } from './components/kpi-content';
 import { BorderedKpiSectionProps } from './types';
 
-export const BorderedKpiSection = ({ title, description, source, dateUnix, tilesData }: BorderedKpiSectionProps) => {
+export const BorderedKpiSection = ({ title, description, source, dateOrRange, tilesData }: BorderedKpiSectionProps) => {
   const metadata: MetadataProps = {
-    date: dateUnix,
+    date: dateOrRange,
```
VWSCoronaDashboard28, line 14: This is already on development, why's it showing up here? 🤔

packages/app/src/components/metadata.tsx
```diff
@@ -29,13 +35,13 @@ export function Metadata({ date, source, obtainedAt, isTileFooter, datumsText, m
...
           date: formatDateFromSeconds(date, 'weekday-long'),
         })
-      : Array.isArray(date)
-      ? replaceVariablesInText(commonTexts.common.metadata.date_from_to, {
-          startDate: formatDateFromSeconds(date[0], 'weekday-long'),
-          endDate: formatDateFromSeconds(date[1], 'weekday-long'),
-        })
       : typeof date === 'string'
       ? date
+      : date && date.start && date.end
```
VWSCoronaDashboard28, line 40: Are these changes coming from another branch? I think these were the changes made in COR-1765.

packages/app/src/pages/gemeente/[code]/ziekenhuis-opnames.tsx
```diff
@@ -49,6 +51,11 @@ export const getStaticProps = createGetStaticProps(
       hospital_nice_choropleth: filterByRegionMunicipalities(hospital_nice_choropleth, context),
     }),
   }),
+  createGetArchivedChoroplethData({
```
Comment (VWSCoronaDashboard28) [inline packages/app/src/pages/gemeente/[code]/ziekenhuis-opnames.tsx:54]: The ticket does not mention the archiving of a choropleth, as far as I can see.

packages/app/src/components/choropleth/logic/thresholds.ts
```diff
@@ -319,6 +354,7 @@ export const thresholds: Thresholds = {
   gm: {
     infected_per_100k: positiveTestedThresholds,
     admissions_on_date_of_admission_per_100000: hospitalAdmissionsPer100000Thresholds,
```
VWSCoronaDashboard29, line 356: The ticket specifies that we will stop using the metric `admissions_on_date_of_admissions_per_100000` and start using the new metric `admissions_in_the_last_7_days_per_100000`.

Shouldn't the threshold definition for the old metric be removed if it's not used anywhere else?

packages/app/src/pages/landelijk/patienten-in-beeld.tsx
```diff
@@ -44,6 +46,9 @@ export const getStaticProps = createGetStaticProps(
   createGetChoroplethData({
     gm: ({ hospital_nice_choropleth }) => ({ hospital_nice_choropleth }),
   }),
+  createGetArchivedChoroplethData({
```
VWSCoronaDashboard28, line 49: Same comment as before, Ticket does not mention anything about archiving graphs

packages/app/src/components/kpi/types.ts
```diff
@@ -10,8 +10,13 @@ export type TileData = {
...
 };
 
+interface DateRange {
+  start: number;
+  end: number;
+}
+
 export interface BorderedKpiSectionProps {
-  dateUnix: number;
+  dateOrRange: number | DateRange;
```
VWSCoronaDashboard29, line 19: Is the new `DateRange` type also using unix timestamps? 

If yes - my suggestion would be to rename the variable `dateOrRange` to `dateOrRangeUnix`.

If no - I would advise to reconsider combining two different types of date annotations (unix and non-unix) in one type.

Please let me know your thoughts.

VWSCoronaDashboard29 (changes requested):
I have made some suggestions. Please let me know your thoughts on them.

packages/app/src/components/kpi/types.ts
VWSCoronaDashboard30: I don't we need to add the suffix since all the dates we are using are in Unix format

packages/app/src/pages/gemeente/index.tsx
```diff
@@ -1,4 +1,4 @@
-import { colors, GmCollectionHospitalNiceChoropleth, gmData } from '@corona-dashboard/common';
+import { colors, ArchivedGmCollectionHospitalNice, gmData } from '@corona-dashboard/common';
```
VWSCoronaDashboard35, line 1:
```suggestion
import { colors, ArchivedGmCollectionHospitalNice, gmData } from '@corona-dashboard/common';
``` 
would need to be changed to  ArchivedGmCollectionHospitalNiceChoropleth which is what I see in ```data.ts```

VWSCoronaDashboard30: Good catch!

packages/app/schema/archived_gm/__difference.json
VWSCoronaDashboard30: I missed the part of it being in the required, it is fixed now. I don't recall not using the suffix anymore. We did change the approach of naming conventions for the schema data types, but not about the suffixes. Please correct me if I missed anything!

packages/app/src/components/choropleth/logic/thresholds.ts
VWSCoronaDashboard30: Good catch, it is corrected now!

packages/app/src/components/choropleth/logic/thresholds.ts
VWSCoronaDashboard30: It still uses it as the choropleth was archived. It's the same for national and municipality levels!

packages/app/src/components/kpi/types.ts
VWSCoronaDashboard29: I would argue that it's better for consistency across the codebase to add the suffix. However I don't want to further block the release so I will mark this as resolved.

VWSCoronaDashboard29 (dismissed):

VWSCoronaDashboard29 (approved):

VWSCoronaDashboard28 (approved):

packages/icons/src/svg/calendar_gear.svg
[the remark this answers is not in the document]
VWSCoronaDashboard30: Will do!

PR #4854, opened 2023-09-01
Feature/cor 1705 correct timeseries agegroups

VWSCoronaDashboard29:
Remove underreported annotation from agegroups timeseries


VWSCoronaDashboard28 (commented):
Something seems off in this PR. I'm seeing over 250 commits. What's going on?

packages/cms-v2/src/schemas/topical/theme-tile.ts
VWSCoronaDashboard30, line 1: @VWSCoronaDashboard29 Why are we changing changing stuff to V2?

VWSCoronaDashboard30 (commented):
Why are there 500 commits to the branch?

VWSCoronaDashboard28 (changes requested):
Something is off given the number of commits, please recreate this PR with only the changes you have introduced.

PR #4855, opened 2023-09-04
feat(COR-1772): Fix no patient numbers shown on zkh opnames GM level

VWSCoronaDashboard29:
Extra property to fix inaccurate data for in get-data.ts

### DIFF
--- a/packages/app/src/static-props/get-data.ts
+++ b/packages/app/src/static-props/get-data.ts
@@ -312,13 +312,15 @@ export function createGetArchivedChoroplethData<T1, T2>(settings?: {
 function replaceInaccurateLastValue(data: any) {
   const metricsInaccurateItems = ['intensive_care_nice', 'hospital_nice'];
 
-  const inaccurateMetricProperty = 'admissions_on_date_of_admission_moving_average_rounded';
+  const inaccurateMetricProperties = ['admissions_on_date_of_admission_moving_average_rounded', 'admissions_in_the_last_7_days'];
 
   const metricsWithInaccurateData = metricsInaccurateItems.filter((m) => m in data) as (keyof typeof data & keyof typeof metricsInaccurateItems)[];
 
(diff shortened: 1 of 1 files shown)


packages/app/src/static-props/get-data.ts
```diff
@@ -312,13 +312,15 @@ export function createGetArchivedChoroplethData<T1, T2>(settings?: {
 function replaceInaccurateLastValue(data: any) {
   const metricsInaccurateItems = ['intensive_care_nice', 'hospital_nice'];
 
-  const inaccurateMetricProperty = 'admissions_on_date_of_admission_moving_average_rounded';
+  const inaccurateMetricProperties = ['admissions_on_date_of_admission_moving_average_rounded', 'admissions_in_the_last_7_days'];
 
   const metricsWithInaccurateData = metricsInaccurateItems.filter((m) => m in data) as (keyof typeof data & keyof typeof metricsInaccurateItems)[];
```
VWSCoronaDashboard28, line 317: Shorthands for arguments like `x, `i`, `m` etc. reduce readability. I recommend refactoring a little bit to get rid of the shorthand.

VWSCoronaDashboard28 (commented):
Please see my comment, looks good otherwise.

packages/app/src/static-props/get-data.ts
VWSCoronaDashboard30: You mean having `metric` instead of `m`?

VWSCoronaDashboard28: Yes

VWSCoronaDashboard30 (approved):

packages/app/src/static-props/get-data.ts
VWSCoronaDashboard29: As the `m` notation already existed before we made this change I think we'l leave it for now. Let's take note of this in the future.

PR #4870, opened 2023-09-18
Bugfix/cor 1769 variants table and graph 2

VWSCoronaDashboard29:
Bugfix for COR-1769

### DIFF
--- a/packages/app/src/domain/variants/static-props/get-variant-chart-data.ts
+++ b/packages/app/src/domain/variants/static-props/get-variant-chart-data.ts
@@ -18,31 +18,32 @@ const EMPTY_VALUES = {
   },
 } as const;
 
+/**
+ * Returns values for variant timeseries chart
+ * @param variants
+ */
 export function getVariantChartData(variants: NlVariants | undefined) {
   if (!isDefined(variants) || !isDefined(variants.values)) {
     return EMPTY_VALUES;
   }
 
-  const variantsOfConcern = variants.values
-    .filter((variant) => variant.last_value.is_variant_of_concern || variant.last_value.has_historical_significance)
-    .filter((variant) => variant.variant_code !== 'other_graph' && variant.variant_code !== 'other_table')
-    .sort((a, b) => b.last_value.order - a.last_value.order);
+  const sortedVariants = variants.values.sort((a, b) => b.last_value.order - a.last_value.order);
 
-  const firstVariant = variantsOfConcern.shift();
+  const firstVariantInList = sortedVariants.shift();
 
-  if (!isDefined(firstVariant)) {
+  if (!isDefined(firstVariantInList)) {
     return EMPTY_VALUES;
   }
 
-  const values = firstVariant.values.map<VariantChartValue>((value, index) => {
+  const values = firstVariantInList.values.map<VariantChartValue>((value, index) => {
     const item = {
       is_reliable: true,
       date_start_unix: value.date_start_unix,
       date_end_unix: value.date_end_unix,
-      [`${firstVariant.variant_code}_percentage`]: value.percentage,
+      [`${firstVariantInList.variant_code}_percentage`]: value.percentage,
     } as VariantChartValue;
 
-    variantsOfConcern.forEach((variant) => {
+    sortedVariants.forEach((variant) => {
       (item as unknown as Record<string, number>)[`${variant.variant_code}_percentage`] = variant.values[index].percentage;
     });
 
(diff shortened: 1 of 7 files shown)


packages/app/src/domain/variants/static-props/get-variant-chart-data.ts
```diff
@@ -18,31 +18,32 @@ const EMPTY_VALUES = {
...
   }
 
-  const variantsOfConcern = variants.values
-    .filter((variant) => variant.last_value.is_variant_of_concern || variant.last_value.has_historical_significance)
-    .filter((variant) => variant.variant_code !== 'other_graph' && variant.variant_code !== 'other_table')
-    .sort((a, b) => b.last_value.order - a.last_value.order);
+  const sortedVariants = variants.values.sort((a, b) => b.last_value.order - a.last_value.order);
 
-  const firstVariant = variantsOfConcern.shift();
+  const firstVariantInList = sortedVariants.shift();
```
VWSCoronaDashboard30, line 32: If we are changing the logic, is this still needed?

@VWSCoronaDashboard29 Have you tested on what happens if we remove this line?

packages/app/src/domain/variants/static-props/get-variant-table-data.ts
```diff
@@ -31,14 +37,14 @@ export function getVariantTableData(variants: NlVariants | undefined, namedDiffe
...
-        return null;
-      }
-
-      return difference;
+      return difference ?? null;
     }
   }
 
+  function mapVariantToNamedDifference(namedDifferenceVariantCode: string) {
+    return variants?.values.find((x) => x.variant_code === namedDifferenceVariantCode) ?? null;
```
VWSCoronaDashboard30, line 45: If you're using the nullable symbol, won't this return null by default if nothing is found?

packages/app/src/domain/variants/variants-stacked-area-tile/variants-stacked-area-tile.tsx
```diff
@@ -92,10 +92,7 @@ export const VariantsStackedAreaTile = ({ text, values, variantColors, metadata
 
           const reorderContext = {
             ...context,
-            config: [
-              ...context.config.filter((value) => !hasMetricProperty(value) || filteredValues[value.metricProperty] || hasSelectedMetrics),
-              context.config.find((value) => hasMetricProperty(value) && value.metricProperty === 'other_graph_percentage'),
-            ].filter(isDefined),
+            config: context.config.filter((value) => !hasMetricProperty(value) || filteredValues[value.metricProperty] || hasSelectedMetrics).filter(isDefined),
```
VWSCoronaDashboard30, line 95: I think we need to keep the property spread notation (...) here

packages/app/src/domain/variants/variants-stacked-area-tile/variants-stacked-area-tile.tsx
```diff
@@ -35,9 +35,9 @@ export const VariantsStackedAreaTile = ({ text, values, variantColors, metadata
 
   const { list, toggle, clear } = useList<keyof VariantChartValue>(alwaysEnabled);
 
-  const [seriesConfig, otherConfig, selectOptions] = useSeriesConfig(text, values, variantColors);
```
VWSCoronaDashboard30, line 38: Do you know what otherConfig does?

packages/app/src/domain/variants/variants-stacked-area-tile/variants-stacked-area-tile.tsx
VWSCoronaDashboard29: Why is that?

packages/app/src/domain/variants/variants-stacked-area-tile/variants-stacked-area-tile.tsx
VWSCoronaDashboard29: Yes, it supplies a different visual style for the `other variants` metric on the variants graph. We've agreed with IA, BE and Design that `other variants` are to be treated as all the rest of the variants. Thus, no special configuration is necessary anymore.

packages/app/src/domain/variants/static-props/get-variant-table-data.ts
VWSCoronaDashboard29: It wouldn't. It would return `undefined` instead of `null`

VWSCoronaDashboard30: Fair enough

packages/app/src/domain/variants/static-props/get-variant-chart-data.ts
VWSCoronaDashboard30: I got the idea now, thanks!

packages/app/src/domain/variants/variants-stacked-area-tile/variants-stacked-area-tile.tsx
VWSCoronaDashboard30: You're not interacting with the object directly, but with a clone. This way you exclude the possibility of introducing inconsistencies

VWSCoronaDashboard29: Updated the code

VWSCoronaDashboard30 (approved):

PR #4871, opened 2023-09-18
feature/COR-1801_vaccinations-modifications-municipalities

VWSCoronaDashboard30:
- Split the choropleth into basic series and fall series
- Removed useful links from the header.
- Archived Vaccination coverage repeat vaccination against COVID-19 section.
- Archived several metrics.
- Added new lokalize keys for the choropleths.


packages/app/schema/archived_gm/vaccinaties/vaccine_campaign_coverage_per_age_group.json
VWSCoronaDashboard29, line 1: Why have you kept the name `vaccine_coverage_per_age_group.json` for this file unchanged? Now we have 2 different `vaccine_coverage_per_age_group.json` files in the archived_gm folder that have different contents but use the same name

packages/app/schema/archived_gm/vaccinaties/vaccine_campaign_coverage_per_age_group.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "title": "gm_vaccine_coverage_per_age_group",
+  "title": "gm_vaccine_campaign_coverage_per_age_group",
```
VWSCoronaDashboard29, line 3: What is the reasoning behind renaming from `gm_vaccine_coverage_per_age_group` to `gm_vaccine_campaign_coverage_per_age_group`?

packages/app/src/domain/vaccine/vaccine-coverage-choropleth.tsx
```diff
@@ -15,37 +15,29 @@ import { BoldText } from '~/components/typography';
...
     title: string;
     description: string;
     vaccinationKindLabel?: string;
     ageGroupLabel?: string;
   };
+  isPrimarySeries?: boolean;
 }
 
-export const VaccineCoverageChoropleth = ({ data, dataOptions, text }: VaccineCoverageChoroplethProps) => {
+export const VaccineCoverageChoropleth = ({ data, dataOptions, text, isPrimarySeries }: VaccineCoverageChoroplethProps) => {
```
VWSCoronaDashboard29, line 33: I'm not really a fan of passing conditions to a component like is done here. I believe this goes against the mantra that 'components should be as dumb as possible' like @VWSCoronaDashboard18 explained.

Is there a reason why we can't create a new component that has just the age group selector instead of conditionally showing the coverage kind selector?

VWSCoronaDashboard29: Perhaps you can create the type `vaccination_coverage_type: 'primary_series' | 'autumn_2022'` and use that as input for the selectedCoverageKind?

That way we can move a way from a construction wherein a vaccination coverage choropleth is either of the primary series, or it is not.

packages/common/src/types/data.ts
```diff
@@ -15,6 +15,7 @@ export interface ArchivedGm {
   code: ArchivedGmCode;
   vaccine_coverage_per_age_group_archived_20220622: GmVaccineCoveragePerAgeGroupArchived_20220622;
   vaccine_coverage_per_age_group_archived_20220908: GmVaccineCoveragePerAgeGroupWithBoosterShot;
+  vaccine_coverage_per_age_group_archived_202310xx: GmVaccineCampaignCoveragePerAgeGroup;
```
VWSCoronaDashboard29, line 18: For 3 different archived instances of `vaccine_coverage_per_age_group` we are using 3 different datatypes with wildly varying names.

I would argue that we should align the names of those datatypes with the values they are representing. Thus moving towards a scheme along the lines of `Archived{xxxxxxxx}GmVaccineCoveragePerAgeGroup` for every type.

VWSCoronaDashboard29 (commented):
Please look at my comments.

packages/app/schema/archived_gm/vaccinaties/vaccine_campaign_coverage_per_age_group.json
VWSCoronaDashboard30: They do have the same name, but they are on different levels. There's no clash in the import because of this

packages/app/src/domain/vaccine/vaccine-coverage-choropleth.tsx
VWSCoronaDashboard30: This component is only used twice. Once on the national level and once on the municipality level. Both of them are getting archived. The logic is already bound to only two types: `primary_series`  or `autumn_2022`. The component is still dumb since the configuration is coming from the outside. The parent component defines what choropleth is being rendered, and the `vaccine_coverage_choropleth` is just rendering everything.

The recommendation in the second comment is happening on line 36. We use this to render the components. I don't think a type is required since we are not going to enhance the choropleth anymore.

packages/common/src/types/data.ts
VWSCoronaDashboard30: That might require extra refactoring

packages/app/src/domain/vaccine/vaccine-coverage-choropleth.tsx
VWSCoronaDashboard29: While I don't entirely feel comfortable with the current approach, I recognise that taking a step back and rebuilding the component without any future updates in sight seems like wasted work. I'll approve it.

packages/common/src/types/data.ts
VWSCoronaDashboard29: How much work would you estimate roughly?

VWSCoronaDashboard30: Absolutely no clue, I need to check in the code

VWSCoronaDashboard29: Also, I'm wondering how much refactoring would need to take place. As we're only changing the name of the datatype. Not the name of the variable, nor the datatype itself.

VWSCoronaDashboard30: You're still importing the interfaces, and I don't exactly how many times they are being used

packages/app/schema/archived_gm/vaccinaties/vaccine_campaign_coverage_per_age_group.json
VWSCoronaDashboard29: Good that the import still works. I think my other two comments would apply here too.

packages/app/schema/archived_gm/vaccinaties/vaccine_campaign_coverage_per_age_group.json
VWSCoronaDashboard30: I can change it, no issue in this. Might be more suggestive as well

packages/common/src/types/data.ts
VWSCoronaDashboard29: I can see the following with regards to usage:
`vaccine_coverage_per_age_group_archived_20220622`: used a total of 7 times in the vaccines page on both municipal and national level
`vaccine_coverage_per_age_group_archived_20220908`: also used a total of 7 times
`vaccine_coverage_per_age_group_archived_202310xx`: used 3 times on the vaccines municipal page

I also don't see the interface names being used anywhere outside of `data.ts`

packages/app/schema/archived_gm_collection/__index.json
```diff
@@ -3,7 +3,16 @@
...
-  "required": ["last_generated", "proto_name", "name", "code", "hospital_nice_choropleth_archived_20230830", "sewer_archived_20230623", "tested_overall_archived_20230331"],
+  "required": [
+    "last_generated",
+    "proto_name",
+    "name",
+    "code",
+    "hospital_nice_choropleth_archived_20230830",
+    "sewer_archived_20230623",
+    "tested_overall_archived_20230331",
+    "vaccine_coverage_per_age_group_choropleth_archived_202310xx"
```
VWSCoronaDashboard28, line 14: What's with the `xx`?

packages/app/src/pages/gemeente/[code]/vaccinaties.tsx
```diff
@@ -201,14 +177,15 @@ export const VaccinationsGmPage = (props: StaticProps<typeof getStaticProps>) =>
...
-            data={choropleth.gm.vaccine_coverage_per_age_group}
+            data={archivedChoropleth.gm.vaccine_coverage_per_age_group_choropleth_archived_202310xx}
             dataOptions={{ getLink: reverseRouter.gm.vaccinaties, selectedCode: currentData.code, isPercentage: true }}
             text={{
               title: replaceVariablesInText(commonTexts.choropleth.choropleth_vaccination_coverage.gm.title, { municipalityName: municipalityName }),
               description: replaceVariablesInText(commonTexts.choropleth.choropleth_vaccination_coverage.gm.description, { municipalityName: municipalityName }),
               vaccinationKindLabel: commonTexts.choropleth.vaccination_coverage.shared.dropdown_label_vaccination_coverage_kind_select,
               ageGroupLabel: commonTexts.choropleth.vaccination_coverage.shared.dropdown_label_age_group_select,
             }}
+            isPrimarySeries
```
Comment (VWSCoronaDashboard28) [inline packages/app/src/pages/gemeente/[code]/vaccinaties.tsx:188]: So on the municipality level `isPrimarySeries` will always be true? 

If that is indeed the case, then perhaps consider using `getCurrentPageScope()`.

VWSCoronaDashboard28 (commented):
Please see my comments

packages/app/schema/archived_gm_collection/__index.json
VWSCoronaDashboard30: we still did not receive the release date for this metric

packages/app/src/pages/gemeente/[code]/vaccinaties.tsx
Comment (VWSCoronaDashboard28) [reply packages/app/src/pages/gemeente/[code]/vaccinaties.tsx:188]: Right, so this is a way to distinguish between the archived and non-archived versions of the same Choropleth, right?

VWSCoronaDashboard28 (approved):

packages/app/src/pages/gemeente/[code]/vaccinaties.tsx
Comment (VWSCoronaDashboard30) [reply packages/app/src/pages/gemeente/[code]/vaccinaties.tsx:188]: exactly

PR #4872, opened 2023-09-18
feature/COR-1788_refactor_archived_schemas_titles

VWSCoronaDashboard30:
- Renamed and refactored all the required archived schemas


packages/app/schema/archived_gm/vaccine_coverage_per_age_group_with_booster_shot.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "title": "gm_vaccine_coverage_per_age_group_with_booster_shot",
+  "title": "archived_gm_vaccine_coverage_per_age_group_with_booster_shot",
```
VWSCoronaDashboard29, line 3: See the comments made on #4871. I would argue that this name should reflect the data property it's forming a schema for. In this case I'd suggest `archived_20220908_vaccine_coverage_per_age_group`. This applies to other archived keys as well.

VWSCoronaDashboard29 (commented):
With which version of the protos did you test this? Can you supply me with that information so I can run a test locally too?

packages/app/schema/archived_gm/vaccine_coverage_per_age_group_with_booster_shot.json
VWSCoronaDashboard30: @VWSCoronaDashboard29 I don't agree with the naming. I've removed all the numbers from the naming for the schemas. The archived date is covered per value in the index folder and the name of the schema is used to generate the interfaces.

VWSCoronaDashboard27 (approved):

VWSCoronaDashboard28 (approved):

packages/app/schema/archived_gm/vaccine_coverage_per_age_group_with_booster_shot.json
VWSCoronaDashboard29: Fair enough

PR #4873, opened 2023-09-18
feature/COR-1597-contact-page-improved-design

VWSCoronaDashboard28:
## Summary

* Redesigned contact page
* Introduced `<ContentLayout>` which is now used on content pages like: About, FAQ, and Contact.
* Added a new icon
* Created new schemas for Sanity
* Extended the typography components to also accept `MarginProps` to avoid having to constantly wrap those components in boxes to space them out.

The feature works by dividing the contact page into groups and items. Items belong to groups, and the groups belong to the Contact page document in Sanity. Each group has a list of items which are connected to it by the way the schemas are built in Sanity (using references), no additional logic is required to link items and groups. The order of appearance of groups can be controlled in Sanity as well. 

#### Before
<details>

image

</details>

#### After
<details>

image

</details>


packages/app/src/pages/contact.tsx
```diff
@@ -1,86 +1,82 @@
...
-        {content.title && <Heading level={1}>{content.title}</Heading>}
-        {content.description && (
-          <RichContent
-            blocks={content.description}
-            contentWrapper={RichContentWrapper}
-          />
-        )}
-      </Content>
+      <VisuallyHidden>
+        <Heading level={1}>Contact</Heading>
```
VWSCoronaDashboard28, line 62: This is a visually hidden `H1`. The page as designed does not have an H1. Tagging the first title in the page as an H1 does not make sense as that title does not describe the page. I did some research and saw that adding a visually hidden one like this retains the benefits of an H1.

VWSCoronaDashboard29 (approved):

VWSCoronaDashboard27 (approved):

PR #4880, opened 2023-09-21
feature/COR-1470_separate_source_from_value_row_on_kpi_tiles

VWSCoronaDashboard27:
## Summary
The date and source were splitted by creating an extra sanity field. Now they are on separate lines and alined.

#### Before
<details>

Screenshot 2023-09-21 at 13 39 40

</details>

#### After
<details>

Screenshot 2023-09-21 at 13 40 10

</details>

### DIFF
--- a/packages/app/src/domain/topical/components/topical-kpi-tile/topical-tile.tsx
+++ b/packages/app/src/domain/topical/components/topical-kpi-tile/topical-tile.tsx
@@ -24,12 +24,13 @@ interface TopicalTileProps {
   hideTrendIcon: boolean;
   kpiValue: string | null;
   sourceLabel: string | null;
+  dateLabel: string | null;
   tileIcon: TopicalIcon;
   title: string;
   trendIcon: TrendIconType;
 }
 
-export const TopicalTile = ({ title, tileIcon, trendIcon, description, kpiValue, cta, sourceLabel, hideTrendIcon }: TopicalTileProps) => {
+export const TopicalTile = ({ title, tileIcon, trendIcon, description, kpiValue, cta, sourceLabel, dateLabel, hideTrendIcon }: TopicalTileProps) => {
   const { formatNumber } = useIntl();
 
   const formattedKpiValue = typeof kpiValue === 'number' ? formatNumber(kpiValue) : typeof kpiValue === 'string' ? kpiValue : false;
@@ -45,7 +46,7 @@ export const TopicalTile = ({ title, tileIcon, trendIcon, description, kpiValue,
 
   return (
     <TopicalTileWrapper cta={cta}>
-      <Box display="flex" flexDirection="column" justifyContent="start">
+      <Box display="flex" flexDirection="column" justifyContent="start" flexGrow={1}>
         <Box display="flex" justifyContent="space-between">
           <Box fontSize={{ _: fontSizes[6], xs: fontSizes[7] }} paddingLeft={{ _: space[3], xs: space[4] }} paddingTop={{ _: space[3], xs: space[4] }}>
             <StyledHeading level={3} color={colors.blue8}>
@@ -79,6 +80,12 @@ export const TopicalTile = ({ title, tileIcon, trendIcon, description, kpiValue,
           <RichContent blocks={description} elementAlignment="start" variableValue={formattedKpiValue} />
         </Box>
       </Box>
+
+      {dateLabel && (
+        <Box padding={{ _: `0 ${space[3]} ${space[3]}`, xs: `0 ${space[4]} ${space[1]}` }}>
+          <InlineText color={colors.gray7}>{dateLabel}</InlineText>
+        </Box>
+      )}
 
       {sourceLabel && (
         <Box padding={{ _: `0 ${space[3]} ${space[3]}`, xs: `0 ${space[4]} ${space[4]}` }}>
--- a/packages/app/src/domain/topical/components/topical-kpi-tile/topical-tile.tsx
+++ b/packages/app/src/domain/topical/components/topical-kpi-tile/topical-tile.tsx
@@ -82,7 +82,7 @@ export const TopicalTile = ({ title, tileIcon, trendIcon, description, kpiValue,
       </Box>
 
       {dateLabel && (
-        <Box padding={{ _: `0 ${space[3]} ${space[3]}`, xs: `0 ${space[4]} ${space[1]}` }}>
+        <Box padding={{ _: `0 ${space[3]} ${space[1]}`, xs: `0 ${space[4]} ${space[1]}` }}>
           <InlineText color={colors.gray7}>{dateLabel}</InlineText>
         </Box>
       )}
--- a/packages/app/src/domain/topical/components/topical-kpi-tile/topical-tile.tsx
+++ b/packages/app/src/domain/topical/components/topical-kpi-tile/topical-tile.tsx
@@ -24,7 +24,7 @@ interface TopicalTileProps {
   hideTrendIcon: boolean;
   kpiValue: string | null;
   sourceLabel: string | null;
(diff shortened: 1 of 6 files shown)


packages/cms-v2/src/schemas/topical/theme-tile.ts
VWSCoronaDashboard28, line 1: No need to add it in `cms-v2`. Could you get rid of this change, please?

packages/cms/src/schemas/documents/pages/homepage/theme-tile.ts
```diff
@@ -64,6 +64,12 @@ export const themeTile = defineType({
     defineField({
       title: 'Metadata label',
       description: 'Bij {{date}} wordt de tekst geplaatst van het tegeldatumveld. Deze kan handmatig overschreven worden.',
```
VWSCoronaDashboard28, line 66: The description for the `dateLabel` and the `sourceLabel` fields are now the same. Are you sure that's how it should be? 

Personally, I would imagine the source label would get a different description.

packages/cms/src/schemas/documents/pages/homepage/theme-tile.ts
```diff
@@ -64,6 +64,12 @@ export const themeTile = defineType({
...
+      name: 'dateLabel',
```
VWSCoronaDashboard28, line 67: The `dateLabel` field is not a required field, right? If that's the case, then I would make it optional in `packages/app/src/queries/query-types.ts` as well.

packages/app/src/queries/query-types.ts
```diff
@@ -83,6 +83,7 @@ export interface BaseTile {
 
 interface TopicalTile extends BaseTile {
   title: string;
+  dateLabel: string | null;
```
VWSCoronaDashboard28, line 86: If the field in Sanity is optional, then I think it should be marked as optional over here as well.

packages/app/src/domain/topical/components/topical-kpi-tile/topical-tile.tsx
```diff
@@ -24,12 +24,13 @@ interface TopicalTileProps {
   hideTrendIcon: boolean;
   kpiValue: string | null;
   sourceLabel: string | null;
+  dateLabel: string | null;
```
VWSCoronaDashboard28, line 27: Here too, if its an optional field in Sanity, that means its not neccessarily filled, so it should probably be `dateLabel?:` here.

VWSCoronaDashboard28 (commented):
Good stuff! Please see my comments.

packages/cms-v2/src/schemas/topical/theme-tile.ts
VWSCoronaDashboard27: Done

packages/cms/src/schemas/documents/pages/homepage/theme-tile.ts
VWSCoronaDashboard27: Done

packages/cms/src/schemas/documents/pages/homepage/theme-tile.ts
VWSCoronaDashboard27: Done

packages/app/src/queries/query-types.ts
VWSCoronaDashboard27: Done

packages/cms/src/schemas/documents/pages/homepage/theme-tile.ts
```diff
@@ -64,6 +64,12 @@ export const themeTile = defineType({
...
+      type: 'localeString',
+    }),
+    defineField({
+      title: 'Source label',
+      description: 'Deze velden krijgen hun input van de bron.   Deze kan handmatig overschreven worden.',
```
VWSCoronaDashboard28, line 72: Is this not a manual entry field? The source is not being automatically filled right? With the description here, to me, it seems like this field gets filled automatically somehow, and if they want, they can use this field to overwrite the automatic value.

packages/app/src/queries/query-types.ts
```diff
@@ -83,6 +83,7 @@ export interface BaseTile {
 
 interface TopicalTile extends BaseTile {
   title: string;
+  dateLabel?: string | null;
```
VWSCoronaDashboard28, line 86: I'm not sure in what scenarios this will be null. I think if its not filled in it'll be `undefined`? If that is the case, then I think `dateLabel?: string` is good enough.

packages/app/src/queries/query-types.ts
```diff
@@ -83,6 +83,7 @@ export interface BaseTile {
...
   sourceLabel: string | null;
```
VWSCoronaDashboard28, line 87: Now that I think about it, this is also an optional field, so should be made optional as well.

packages/app/src/domain/topical/components/topical-kpi-tile/topical-tile.tsx
```diff
@@ -24,12 +24,13 @@ interface TopicalTileProps {
   hideTrendIcon: boolean;
   kpiValue: string | null;
   sourceLabel: string | null;
+  dateLabel?: string | null;
```
VWSCoronaDashboard28, line 27: Same comment here, I don't think we need to explicitly type this as `null`.

packages/app/src/queries/query-types.ts
VWSCoronaDashboard28: The comments about `null` vs `undefined` apply here too.

packages/cms/src/schemas/documents/pages/homepage/theme-tile.ts
VWSCoronaDashboard27: Done

packages/app/src/queries/query-types.ts
VWSCoronaDashboard27: Done

packages/app/src/queries/query-types.ts
VWSCoronaDashboard27: Done

packages/app/src/domain/topical/components/topical-kpi-tile/topical-tile.tsx
VWSCoronaDashboard27: Done

VWSCoronaDashboard28 (approved):

VWSCoronaDashboard18 (approved):
It looks awesome. Nice code adjustments!
I have left one small question, but that's is definitely non blocking. So feel free to resolve it, and definitely feel free to have another opinion on the comment.

packages/app/src/domain/topical/components/topical-kpi-tile/topical-tile.tsx
```diff
@@ -80,6 +81,12 @@ export const TopicalTile = ({ title, tileIcon, trendIcon, description, kpiValue,
         </Box>
       </Box>
 
+      {dateLabel && (
+        <Box padding={{ _: `0 ${space[3]} ${space[3]}`, xs: `0 ${space[4]} ${space[1]}` }}>
```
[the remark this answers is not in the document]
VWSCoronaDashboard18: Do note this comment is an opinion and I'm not able to verify any mobile design at the moment.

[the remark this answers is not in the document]
VWSCoronaDashboard27: Good catch! Done

VWSCoronaDashboard28 (approved):

VWSCoronaDashboard18 (approved):
Awesome! Nice PR!

PR #4883, opened 2023-09-22
feature/COR-1784_vaccinations-national-2

VWSCoronaDashboard29:
## Summary

- Update schemas to make compatible with protos from COR-1803
- Split vaccine coverage per age group tables up by coverage type
- Split vaccine coverage choropleths up by coverage type
- Apply changes to vaccines page national to align with COR-1784


packages/app/src/components/tables/logic/use-get-single-coverage-percentage-data.ts
VWSCoronaDashboard28, line 1: Seems like this is more or less the same as `packages/app/src/components/tables/logic/use-get-percentage-data.ts`. Any particular reason why you did not choose to refactor that for single coverage as well?

packages/app/src/domain/vaccine/autumn-2022-shot-coverage-per-age-group/autumn-2022-shot-coverage-per-age-group.tsx
```diff
@@ -36,20 +35,19 @@ export const Autumn2022ShotCoveragePerAgeGroup = ({ title, description, metadata
...
+  const percentageTitles = text.headers.autumn_2022_shot;
+  const percentageColors = COLOR_AUTUMN_2022_SHOT;
+  const percentageFormattingRules = { shouldFormat: true };
+  const percentageData = useGetSingleCoveragePercentageData(sortedData, percentageTitles, percentageColors, percentageFormattingRules);
 
   return (
     <ChartTile title={title} description={description} metadata={metadata}>
       {breakpoints.lg ? (
-        <WideTable
+        <WideTable<SingleCoverageTableData>
```
VWSCoronaDashboard28, line 46: Interesting, don't think I've seen this pattern before. Could you explain why `<WideTable<SingleCoverageTableData>` is necessary?

packages/app/src/domain/vaccine/campaign-banner/campaign-banner.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+interface CampaignBannerProps {
+  title: string;
+  description: string;
+}
+
+export const CampaignBanner = ({ title, description }: CampaignBannerProps) => (
+  <Tile>
+    <Box backgroundColor={colors.blue1} marginY={space[4]}>
+      <BannerContainer
```
VWSCoronaDashboard28, line 18: If this is a style component anyway, consider placing all your styles in one place. This applies to other styled components in this file.

packages/app/src/domain/vaccine/campaign-banner/campaign-banner.tsx
```diff
@@ -0,0 +1,76 @@
...
+          paddingY={space[4]}
+        >
+          <Heading level={2} variant="h2">
+            {title}
+          </Heading>
+
+          <Markdown content={description} />
+        </CampaignDescription>
+        <Box paddingX={{ sm: space[5], md: space[2] }} alignSelf={{ _: 'center', sm: 'left' }}>
+          <CampaignIllustration src={'/images/Najaarsronde-coronaprik-hero.svg'} extension="svg" />
```
VWSCoronaDashboard28, line 41: Based on the other banner, I would argue that this image should come from Sanity.

packages/app/src/domain/vaccine/campaign-banner/index.ts
VWSCoronaDashboard28, line 1: Barrel files are typically discouraged. If that's something you guys would like to introduce once again, that's up to you, but just so you know.

packages/app/src/domain/vaccine/primary-series-coverage-per-age-group/index.ts
VWSCoronaDashboard28, line 1: Barrel files are typically discouraged. If that's something you guys would like to introduce once again, that's up to you, but just so you know.

packages/app/src/domain/vaccine/primary-series-coverage-per-age-group/primary-series-coverage-per-age-group.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const sortedData = requiredData.sort((a, b) => getSortingOrder(a.ageGroupRange, sortingOrder, componentName) - getSortingOrder(b.ageGroupRange, sortingOrder, componentName));
+  const percentageTitles = text.headers.fully_vaccinated;
+  const percentageColors = COLOR_FULLY_VACCINATED;
+  const percentageFormattingRules = { shouldFormat: true };
+  const percentageData = useGetSingleCoveragePercentageData(sortedData, percentageTitles, percentageColors, percentageFormattingRules);
+
+  return (
+    <ChartTile title={title} description={description} metadata={metadata}>
+      {breakpoints.lg ? (
+        <WideTable<SingleCoverageTableData>
```
VWSCoronaDashboard28, line 46: Same comment here as in a previous file about `<WideTable<SingleCoverageTableData>`

packages/app/src/domain/vaccine/vaccine-campaigns-tile/components/narrow-vaccine-campaign-table.tsx
```diff
@@ -73,11 +82,15 @@ const VaccineCampaignRow = ({ campaign, campaignDescriptions, headers, isFirst }
               </StyledCell>
             </tr>
 
-            <tr>
-              <StyledCell paddingY="0" isMobile>
-                {headers.total}: {isOpen ? <BoldText>{formatNumber(campaign.vaccine_administered_total)}</BoldText> : formatNumber(campaign.vaccine_administered_total)}
-              </StyledCell>
-            </tr>
+            {hideTotals ? (
```
VWSCoronaDashboard28, line 85: A ternary is not necessary here. While this will work fine, consider flipping the boolean and using the logical AND operator instead. 

`!booleanVariable && JSX/component`

packages/app/src/domain/vaccine/vaccine-campaigns-tile/components/wide-vaccine-campaign-table.tsx
```diff
@@ -12,22 +12,23 @@ interface WideVaccineCampaignTableProps {
...
-export const WideVaccineCampaignTable = ({ campaigns, campaignDescriptions, headers }: WideVaccineCampaignTableProps) => {
+export const WideVaccineCampaignTable = ({ campaigns, campaignDescriptions, headers, hideTotals }: WideVaccineCampaignTableProps) => {
   return (
     <StyledTable>
       <thead>
         <tr>
           <StyledHeaderCell>{headers.vaccine}</StyledHeaderCell>
           <StyledHeaderCell>{headers.last_week}</StyledHeaderCell>
-          <StyledHeaderCell>{headers.total}</StyledHeaderCell>
+          {hideTotals ? <></> : <StyledHeaderCell>{headers.total}</StyledHeaderCell>}
```
VWSCoronaDashboard28, line 25: A ternary is not necessary here. While this will work fine, consider flipping the boolean and using the logical AND operator instead.

`!booleanVariable && JSX/component`

packages/app/src/domain/vaccine/vaccine-campaigns-tile/vaccine-campaigns-tile.tsx
```diff
@@ -15,20 +15,25 @@ interface VaccineCampaignsTileProps {
...
-export const VaccineCampaignsTile = ({ title, headers, campaigns, campaignDescriptions, description, descriptionFooter, metadata }: VaccineCampaignsTileProps) => {
+export const VaccineCampaignsTile = ({ title, headers, campaigns, campaignDescriptions, description, descriptionFooter, metadata, campaignOptions }: VaccineCampaignsTileProps) => {
   const breakpoints = useBreakpoints();
 
-  const sortedCampaigns = campaigns.sort((campaignA, campaignB) => campaignA.vaccine_campaign_order - campaignB.vaccine_campaign_order);
+  const sortedCampaigns = campaigns
+    .filter((c) => !campaignOptions?.hide_campaigns?.includes(c.vaccine_campaign_order))
+    .sort((campaignA, campaignB) => campaignA.vaccine_campaign_order - campaignB.vaccine_campaign_order);
+
+  const containsEmptyTotals = sortedCampaigns.some((camp) => (camp.vaccine_administered_total !== null ? false : true));
```
VWSCoronaDashboard28, line 28: While this will work just fine, you don't need to explicitly return a boolean for `Array.some` to work. You can simply pass it a condition like `camp.vaccine_administered_total !== null` and some will receive whatever the expression evaluates to.

packages/app/src/domain/vaccine/vaccine-campaigns-tile/vaccine-campaigns-tile.tsx
```diff
@@ -15,20 +15,25 @@ interface VaccineCampaignsTileProps {
...
+  campaignOptions?: VaccineCampaignOptions;
 }
 
-export const VaccineCampaignsTile = ({ title, headers, campaigns, campaignDescriptions, description, descriptionFooter, metadata }: VaccineCampaignsTileProps) => {
+export const VaccineCampaignsTile = ({ title, headers, campaigns, campaignDescriptions, description, descriptionFooter, metadata, campaignOptions }: VaccineCampaignsTileProps) => {
   const breakpoints = useBreakpoints();
 
-  const sortedCampaigns = campaigns.sort((campaignA, campaignB) => campaignA.vaccine_campaign_order - campaignB.vaccine_campaign_order);
+  const sortedCampaigns = campaigns
+    .filter((c) => !campaignOptions?.hide_campaigns?.includes(c.vaccine_campaign_order))
```
VWSCoronaDashboard28, line 25: While `c` might seem readable now, I would advise replacing it with its full form for the sake readability in the future.

packages/cms-v2/src/elements/schemas/shared/index.ts
VWSCoronaDashboard28, line 1: You could omit these changes if you'd like, its the old CMS.

VWSCoronaDashboard28 (commented):
Looking good, I've left some comments.

packages/app/public/images/Najaarsronde-coronaprik-hero.svg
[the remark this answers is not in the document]
VWSCoronaDashboard29: What query could we have used to fetch it then? Also, I've taken inspiration from how the `IllustrationTile` component on the Reproductiegetal page is built.

packages/app/schema/archived_nl/vaccine_campaigns_archived_202310xx.json
```diff
@@ -0,0 +1,52 @@
...
+      "type": "integer"
+    },
+    "date_end_unix": {
+      "type": "integer"
+    },
+    "date_of_insertion_unix": {
+      "type": "integer"
+    }
+  }
+}
```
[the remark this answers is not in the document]
VWSCoronaDashboard29: Why is that required?

packages/app/src/components/tables/logic/use-get-single-coverage-percentage-data.ts
VWSCoronaDashboard29: As the new function takes a different input than `use-get-percentage-data.ts` and also returns a different output I thought it would be best to build a new hook.

packages/app/src/domain/vaccine/autumn-2022-shot-coverage-per-age-group/autumn-2022-shot-coverage-per-age-group.tsx
VWSCoronaDashboard29: Yes. I have made the component types for `WideTable` and `NarrowTable` generic. This way we can determine what props the component takes at build time and we don't have to build a separate component when we want to display `WideTable` or `NarrowTable` with just one column instead of two.

packages/app/src/domain/vaccine/campaign-banner/campaign-banner.tsx
VWSCoronaDashboard29: Could you elaborate please?

packages/app/src/domain/vaccine/campaign-banner/campaign-banner.tsx
VWSCoronaDashboard29: What query would we use for it then? The banner image on the topical page comes from the `topicalPageQuery`. There is no query for retrieving sanity images that we can use on the vaccination page. Please elaborate on how we would retrieve the image from Sanity?

packages/app/src/domain/vaccine/campaign-banner/index.ts
VWSCoronaDashboard29: What is the alternative?

packages/app/src/domain/vaccine/primary-series-coverage-per-age-group/index.ts
VWSCoronaDashboard29: Removed

packages/app/public/images/Najaarsronde-coronaprik-hero.svg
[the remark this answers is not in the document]
VWSCoronaDashboard28: The query would look something like:

```
'image': {
    ...image,
    'altText': image.altText.${locale},
    'asset': {
      ...image.asset->,
    }
}
```

Up to you where you would like to store the asset. The advice banner on the homepage, which is similar to this one works in that way. The query above is extracted from the query for the advice component.

packages/app/schema/archived_nl/vaccine_campaigns_archived_202310xx.json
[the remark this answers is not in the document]
VWSCoronaDashboard28: It's not required, but a good practice. Hence the big red circle at the end of the file.

packages/app/src/domain/vaccine/autumn-2022-shot-coverage-per-age-group/autumn-2022-shot-coverage-per-age-group.tsx
VWSCoronaDashboard28: Alright but when I omit `SingleCoverageTableData` in `<WideTable<SingleCoverageTableData>` nothing changes? I don't get any kind of error indicating something is going wrong, so I don't see what it's doing ehre.

packages/app/src/domain/vaccine/campaign-banner/campaign-banner.tsx
VWSCoronaDashboard28: Sure.

There are numerous styles applied as props/attributes on the `BannerContainer` component. Further down in this file, you add `column-gap: ${space[4]};` to the style of this component as well. From a readability perspective, it's a lot easier and intuitive to have all the styles in one place, rather than having to look in two places for one component.

packages/app/public/images/Najaarsronde-coronaprik-hero.svg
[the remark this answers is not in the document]
VWSCoronaDashboard29: That I understand. But do I make a fully seperate sanity query, that'll exist next to the existing `get-elements-query.ts`, `get-page-parts-query.ts`, etc.? Or can I make the above snippet part of an existing query? If so, which one?

packages/app/src/domain/vaccine/vaccine-campaigns-tile/components/wide-vaccine-campaign-table.tsx
VWSCoronaDashboard29: Good catch! Corrected.

packages/app/src/domain/vaccine/vaccine-campaigns-tile/vaccine-campaigns-tile.tsx
VWSCoronaDashboard29: This will be updated regardless because of the new design as well as the weekly updated primary series vaccine coverage not being delivered anymore.

Feedback noted.

packages/app/src/domain/vaccine/vaccine-campaigns-tile/vaccine-campaigns-tile.tsx
VWSCoronaDashboard29: This will be updated regardless because of the new design as well as the weekly updated primary series vaccine coverage not being delivered anymore.

Feedback noted.

packages/cms-v2/src/elements/schemas/shared/index.ts
VWSCoronaDashboard29: Are they still being taken into account in any typecheck or validation?

packages/app/public/images/Najaarsronde-coronaprik-hero.svg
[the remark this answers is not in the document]
VWSCoronaDashboard28: I did a little investigation and it seems rather complicated to query a stand-alone image which is not part of a schema of some sort. I don't think its worth the effort so I will resolve this comment and the other one relating to this too.

packages/app/src/domain/vaccine/campaign-banner/campaign-banner.tsx
VWSCoronaDashboard28: Resolving based on the other comment.

packages/app/src/domain/vaccine/campaign-banner/index.ts
VWSCoronaDashboard28: There is no alternative, if you don't use the barrel file, then you will simply import directly from `campaign-banner.tsx`. This is adding an additional layer.

packages/app/src/domain/vaccine/vaccine-campaigns-tile/vaccine-campaigns-tile.tsx
VWSCoronaDashboard28: I'm unsure whether I should resolve these threads or not. Given that this is code quality-related, and it's up to you if you want to go with my suggestion, feel free to resolve this.

packages/cms-v2/src/elements/schemas/shared/index.ts
VWSCoronaDashboard28: They should not be.

packages/app/schema/archived_nl/vaccine_campaigns_archived_202310xx.json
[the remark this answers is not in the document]
VWSCoronaDashboard29: Added newline.

packages/app/src/domain/vaccine/autumn-2022-shot-coverage-per-age-group/autumn-2022-shot-coverage-per-age-group.tsx
VWSCoronaDashboard29: You're right, it seems `<SingleCoverageTableData>` is not necessary for this to work.

Though I would argue that adding it does increase readability for the code, granted that the type indication is also added to all other instances where `<WideTable>` is used, e.g. the behaviour page.

packages/app/src/domain/vaccine/vaccine-campaigns-tile/components/wide-vaccine-campaign-table.tsx
```diff
@@ -22,7 +22,7 @@ export const WideVaccineCampaignTable = ({ campaigns, campaignDescriptions, head
         <tr>
           <StyledHeaderCell>{headers.vaccine}</StyledHeaderCell>
           <StyledHeaderCell>{headers.last_week}</StyledHeaderCell>
-          {hideTotals ? <></> : <StyledHeaderCell>{headers.total}</StyledHeaderCell>}
+          {!hideTotals ?? <StyledHeaderCell>{headers.total}</StyledHeaderCell>}
```
VWSCoronaDashboard28, line 25: I'm not sure I get what's going on here. `HideTotals` is a boolean, right? And here you're saying if it's null or undefined then display the right-hand side? When does it become undefined?

packages/app/src/domain/vaccine/campaign-banner/campaign-banner.tsx
VWSCoronaDashboard29: Updated component

packages/app/src/domain/vaccine/vaccine-campaigns-tile/vaccine-campaigns-tile.tsx
VWSCoronaDashboard29: I'll take the suggestion into account when rebuilding this component according to the new design. As for now I shall leave the conversation open as a reminder.

packages/app/src/domain/vaccine/vaccine-campaigns-tile/components/wide-vaccine-campaign-table.tsx
VWSCoronaDashboard29: In the file `vaccine-campaigns-tile.tsx` we have the following code at line 28 that determines wether or not the `hideTotals` boolean is false or true.

`const containsEmptyTotals = sortedCampaigns.some((camp) => (camp.vaccine_administered_total !== null ? false : true));`

Does this answer your question?

VWSCoronaDashboard28: `??` returns the right-hand side when the left is `null` or `undefined`. With the example above, `hideTotals` will always be either `true` or `false`, but never `undefined` or `null`. Am I missing something?

VWSCoronaDashboard29: I see your point. The original reason I updated this was based on your feedback in [this comment](https://github.com/minvws/nl-covid19-data-dashboard/pull/4883#discussion_r1335979977). We use true/false instead of defined/undefined because elsewhere we need true/false in a ternary operator to determine whether or not to show the dropdown button 

See line 64-99 in `wide-vaccine-campaign-table.tsx`

VWSCoronaDashboard29: I have updated the code

AP2610: Fair enough, I understand that. So, why not do the same here? 

`!hideTotals && <StyledHeaderCell>{headers.total}</StyledHeaderCell>`. I'm just curious

packages/app/src/domain/vaccine/vaccine-campaigns-tile/vaccine-campaigns-tile.tsx
VWSCoronaDashboard29: I have switched out `c` for `vaccineCampaign` now.

VWSCoronaDashboard29:
follow up https://github.com/minvws/nl-covid19-data-dashboard/pull/4887

PR #4936, opened 2023-11-20
feature/COR-1758_shorten-sidebar

VWSCoronaDashboard30:
Implemented the functionality for the new collapsible category menu

### DIFF
--- a/packages/app/src/components/aside/menu.tsx
+++ b/packages/app/src/components/aside/menu.tsx
@@ -14,6 +14,7 @@ import { asResponsiveArray } from '~/style/utils';
 import { Link } from '~/utils/link';
 import { useBreakpoints } from '~/utils/use-breakpoints';
 import { AsideTitle } from './title';
+import { useCollapsible } from '~/utils/use-collapsible';
 
 type Url = UrlObject | string;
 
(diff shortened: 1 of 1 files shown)


packages/app/src/components/aside/menu.tsx
```diff
@@ -99,6 +150,30 @@ function isActivePath(router: NextRouter, href: Url) {
   return currentPath === hrefPath;
 }
 
+function openArchivedCategoryMenu(router: NextRouter) {
+  const archivedNlPaths: ArchivedPath[] = [
```
VWSCoronaDashboard29, line 154: As discussed. Please see if you can find a way to have the determination of whether the URL is an archived path based on the `archived_metrics` key of the useSideBar map

VWSCoronaDashboard29 (changes requested):

packages/app/src/components/aside/menu.tsx
VWSCoronaDashboard29, line 1: Please look into adding a horizontal divider on the sidebar between the non-archived and the archived pages

packages/app/src/components/aside/menu.tsx
VWSCoronaDashboard30: Second PR for comparison is ready #4937 .

PR #4937, opened 2023-11-22
feature/COR-1758_shorten-sidebar-with-hooks

VWSCoronaDashboard30:
Refactored the sidebar to make use of hooks

### DIFF
--- a/packages/app/src/components/aside/menu.tsx
+++ b/packages/app/src/components/aside/menu.tsx
@@ -60,7 +60,7 @@ export function CollapsibleCategoryMenu({ title, children, icon }: { children: R
   const archivedPaths = useArchivedPaths();
   const collapsible = useCollapsible({ isOpen: isCategoryMenuOpen(router, archivedPaths) });
   return (
-    <Box as="li" spacing={2}>
+    <Box as="li" spacing={2} borderTop={`1px solid ${colors.gray2}`}>
       {title && icon && (
         <>
           <Box display="flex" flexDirection="row" flexWrap="nowrap" alignItems="center">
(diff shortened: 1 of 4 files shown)


packages/common/src/data/reverse-router.ts
```diff
@@ -49,3 +49,32 @@ export function getReverseRouter(isMobile: boolean) {
...
   return reverseRouter;
 }
+
+export function getArchivedRoutes() {
+  /**
+   * Add the same keys that are part of the archived_metrics array
+   * in the nl-layout and gm-layout files
+   */
+
+  const archivedPaths = {
```
VWSCoronaDashboard29, line 59: I think this is an OK way to solve the issue of archived paths for now. I'd suggest that in the future we can look into a way of splitting the NL paths block on line 17 into archived and non-archived paths and work from there.

VWSCoronaDashboard29 (changes requested):
Looks quite allright to me but what about the cijferverantwoording?

packages/app/src/components/aside/menu.tsx
```diff
@@ -99,6 +135,21 @@ function isActivePath(router: NextRouter, href: Url) {
   return currentPath === hrefPath;
 }
 
+function isCategoryMenuOpen(router: NextRouter, archivedPaths: ArchivedPath) {
```
VWSCoronaDashboard29, line 138: I would suggest the following name.

```suggestion
function isCurrentRouteArchivedPage(router: NextRouter, archivedPaths: ArchivedPath) {
```

VWSCoronaDashboard29 (dismissed):

packages/app/src/components/aside/menu.tsx
VWSCoronaDashboard30: The function name was changed!

packages/common/src/data/reverse-router.ts
VWSCoronaDashboard30: True, we should look into this, as we also have the third sidebar

VWSCoronaDashboard29 (approved):

PR #4945, opened 2023-12-11
fix: set OS and NodeJS version to use in GitHub workflow

VWSCoronaDashboard29:
- Set specific version number for NodeJS version used in GH workflow to 18.18.4
- Set specific OS version for Ubuntu in GH workflow to 22.04


yorickdevries (approved):
Looks good; fixes the problem with newe node versions

PR #4952, opened 2023-12-13
feature/fix ts node esm+upgrade sharp

yorickdevries:
## Summary

We saw that the latest version of node added a breaking change for running ts-node commands.
I applied this fix to mitigate this; https://github.com/TypeStrong/ts-node/issues/1997
In addition, the sharp package needed to be upgraded as well to be able to build.
As a reasult we can use the latest version of node 18 again.


packages/cms-v2/package.json
```diff
@@ -13,16 +13,16 @@
     "test:ci": "echo 'Sanity check is deprecated \n@TODO add sanity test ci script'",
```
VWSCoronaDashboard30, line 13: These changes are not needed since we are not using this version anymore

VWSCoronaDashboard30 (approved):

packages/cms-v2/package.json
yorickdevries: Then I propose to still apply the changes in case we ever need to run it again

PR #4987, opened 2024-02-26
feature(COR-1939): Added tooltip year for date range values

VWSCoronaDashboard30:
## Summary
 - Added year for date range tooltip


yorickdevries (approved):
approved

